PR #3, opened 2019-05-14
fix broken link to github project

vitaliis:
Maybe there is a better place to refer to, then please adjust.
For now some generic place for the link to not be broken anymore.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -16,7 +16,7 @@ description = "Easily collect telemetry like metrics and distributed traces from
 description = "Vendor-neutral APIs and instrumentation for distributed tracing"
 
 [[menu.social]]
-url = "https://github.com/open-telemetry/open-telemetry"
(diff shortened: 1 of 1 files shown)


bogdandrutu:
Please sign the CLA.

vitaliis:
It is crazy how many things you have to go through to sign a CLA. Create a Linux Foundation ID, confirm email, link github account, go to some hellosign, fill the same there. It must be simpler!!!

Also... Having some fields that you not necessarily want to fill, like phone number, which can lead to people simply cancelling!

SergeyKanzhelev:
@vitaliis it's the price to pay to get the code in the vendor-neutral organization. Good thing you only have to do it once. I bet (and hope) if something can be done easier - CNCF already thought about it.

Thank you for patience and willingness to go the extra mile to help!

PR #18, opened 2019-07-30
Url for CNCF main blog was incorrect.

Chad-Ballay:
Originally the URL was:
https://blog.cncf.io/

Changing to:
https://www.cncf.io/category/blog/

The original URL does not resolve/exist.  Looks like a straightforward mistype.


Chad-Ballay:
Deleting to set missing github user and resubmit.

PR #25, opened 2019-09-10
Add a status board to the website.

austinlparker:


tedsuo:
After some discussion, I suggest we remove "Spec" from this list. Otherwise LGTM.

SergeyKanzhelev:
Can you pls post a screenshot? If it's the same date for all languages - is there a value having it as a table with a separate rows?

austinlparker:
We’ve got an image version that I’m going to replace this with, will update
the PR when that’s ready.

austinlparker:
Here's what it looks like with an image rather than a table. Keep in mind this is really just a placeholder for something that will be replaced in a couple weeks, so unless there's some really stringent objections I think we should go ahead and go with this.

image

PR #27, opened 2019-09-11
Removed timezone enforcement from calendar link

GreyCat:
Google's calendar web view can detect one's timezone from browser / OS settings, and this is actually the preferred way to use it to be friendlier to people from non-Pacific timezones.


GreyCat:
I signed it

PR #30, opened 2019-09-22
Redesign for alpha release

austinlparker:
Adding the initial work I've done to retool the site for alpha. This isn't ready for merge yet, but I wanted to get a deploy preview and have something for people to look at.


austinlparker:
This is pretty much good to go in terms of layout/organization, just needs content. IMO it'd be better to go ahead and merge so we can lay out issues for bringing in content individually from each SIG/author/etc.

mtwo:
Just got back from vacation - apologies for the delay. Is there a way to see the preview site with the changes that you've made? No worries if not

austinlparker:
@mtwo Yeah, it's at https://deploy-preview-30--opentelemetry.netlify.com/

layouts/partials/home/hero.html
```diff
@@ -20,10 +20,13 @@
           <p class="is-size-3-desktop is-size-4-tablet is-size-5-mobile has-text-weight-light">
             {{ $subtagline }}
           </p>
+        
+          <br/><br/>
+          <p class="is-size-4-desktop is-size-5-tablet is-size-6-mobile has-text-weight-bold">
```
mtwo: I suggest that we add a description of what OpenTelemetry provides (libraries, agents, etc.) before previewing the latest blog post. Here's a modified version of our FAQ text that I propose we use:

OpenTelemetry provides a single set of APIs, libraries, agents, and collector services that capture distributed traces and metrics from your application and send them to backends like Prometheus and Jaeger for processing.

austinlparker: Done -- also pushed the blog content down a row. We can add more content to this area later, but the hero was getting full.

layouts/partials/home/progress_chart.html
```diff
@@ -0,0 +1,38 @@
+{{ $projects := site.Data.progress }}
+{{ $yOffset := 140 }}
+
+<svg class="progressChart" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 1500 1000">
+<title>Progress Chart</title>
+<g id="chartTop">
+  <text class="yLabel" x="400" y="75">Alpha</text>
+  <text class="yLabel" x="700" y="75">Beta</text>
+  <text class="yLabel" x="1000" y="75">RC</text>
+  <text class="yLabel" x="1300" y="75">v1.0</text>
```
jparsana: v.1.0 cuts off on render. ( https://deploy-preview-30--opentelemetry.netlify.com/project-status/ )

austinlparker: fixed

yurishkuro:
I'd recommend to call the top menu item "Project", instead of "More"

The doc pages have weird behavior where they first display without the stylesheets, and then redraw. I am getting it consistently in the preview. Also, occasionally Chrome offers to translate the page from Norwegian 🤷‍♂

austinlparker:
@yurishkuro I'm not sure what's up with the redraw on the doc pages. I'll look into it a bit more.

My best guess as to why Chrome is trying to translate them after some quick googling is that there's not enough english text on the page, which will get fixed as the content fills out. 

I did update the 'More' header to say 'Project', though.

tedsuo:
Thanks @austinlparker! One thing I noticed: at most page widths, the sidebar looks like it has an excessive amount of padding on its left.

PR #49, opened 2019-10-02
Add Ruby to SIG progress list

fbogsany:


fbogsany:
@austinlparker I'm not sure what's up with CI for this. I *assume* I broke something, but I couldn't find a CI run for the previous PR #30

austinlparker:
Yeah, looks like I've got a bug somewhere. Looking into it, stand by.

austinlparker:
Bug is fixed in #50, but it'll need to merge first. You could probably take that change into your branch and push it then I can approve this?

fbogsany:
@austinlparker thanks, that seems to have fixed it.

austinlparker:
That fixed CI, but it looks like there's some bugs left in the chart logic, so let me take a stab at those.

fbogsany:
I picked up that change as well.

austinlparker:
Great! Thanks, this lgtm

PR #50, opened 2019-10-02
Fixup milestone script to handle various edge cases

austinlparker:
If a milestone existed but had 0 total issues, we'd write a string that the template would try to render but would fail due to dividing by zero


austinlparker:
Made some more improvements to the data script -- should always pull the lowest open version first. Also did some misc code cleanup for readability.

austinlparker:
Closed by #49

PR #53, opened 2019-10-02
updated learn more links

mangotron:


austinlparker:
Checks failed due to a fix that was merged in #49, should be g2g now.

PR #67, opened 2019-10-31
Reduce whitespace on docs pages with sidebar

### LINKED ISSUE #62: Too much padding on Sidebar
Just feels a little pushed in.
For example, here with sidebar: https://opentelemetry.io/docs/js/metrics/
Versus here with no sidebar: https://opentelemetry.io/get-involved/

* Sidebar feels to far from the left edge
* Content is too far from the sidebar.
* Content registration is different depending on whether the sidebar is present. It "jumps" to the left when the sidebar appears.

austinlparker:
Modifies the navbar to take up a single column.

### DIFF
--- a/layouts/docs/list.html
+++ b/layouts/docs/list.html
@@ -5,7 +5,7 @@
 <section class="section">
   <div class="container">
     <div class="columns">
-      <div class="column is-one-quarter">
+      <div class="column is-1">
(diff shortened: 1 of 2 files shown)


austinlparker:
Could I get a review on this? we've got some PR's coming in for docs and the formatting looks weird without this change. @mtwo

PR #77, opened 2019-11-19
add erlang/elixir source to data

tsloughter:


tsloughter:
Well that isn't right. I Checked the preview and it has the bar full... Is that because there isn't a release yet?

austinlparker:
Lemme take a look, maybe it’s a bug in how the data gets generated

tsloughter:
I created a release so it might work now if it gets rerun/rebuilt.

tsloughter:
Argh, I created a release yesterday https://github.com/open-telemetry/opentelemetry-erlang-api/releases and then this morning force pushed to get a new build and it is still just a solid black bar...

austinlparker:
oh! ok, I think I know what's going on here - there's probably no style information in the css for this.

data/progress_source.yaml
```diff
@@ -74,6 +74,15 @@ data:
     releaseName: "n/a"
     releaseDescription: "n/a"
     releaseUrl: ""
+  - label: Erlang/Elixir
```
austinlparker: If you change this value to 'Erlang' it'll work - the slash in the label breaks the css class.

tsloughter: Ah, *facepalm*, ok, I've pushed a change.

tsloughter:
The full black bar is fixed but all of them are now showing no releases. Maybe that is just a preview thing?

austinlparker:
Yeah, I don’t think we run the script to pull milestones on PRs

PR #94, opened 2020-01-10
Add OpenTelemetry Registry

### LINKED ISSUE #59: Migrate the OpenTracing Registry to OpenTelemetry
We should move the registry from OpenTracing to this site, and enhance it if needed.

austinlparker:
This PR adds a registry for plugins, instrumentation, tracers, and other items to the OpenTelemetry site. I've pre-populated it with items in the opentelemetry organization.

This implementation is mostly ported directly from opentracing.io. We should consider what other features we might like to add either in this PR or in a future one.

Fixes #59


austinlparker:
Notes:
- Add badge/tag to distinguish between projects that live in Otel org vs. non-Otel org
- Headers should either be shorter/smaller/just the language name?

PR #98, opened 2020-02-04
Fix link to add a project to the Registry

austinlparker:


austinlparker:
cc @mtwo if I could get a sign-off on this today it'd be good, fixes a broken link

austinlparker:
this also fixes the issue where only one result would get returned from a search

layouts/registry/list.html
```diff
@@ -11,7 +11,7 @@ <h1 class="title">OpenTelemetry Registry</h1>
       <p>This section of the site will allow you to discover tracers, instrumentation libraries, interfaces,
```
bhs: Misc copy suggestion:

`The OpenTelemetry Registry allows you to search for instrumentation libraries, tracer implementations, utilities, and other useful projects in the OpenTelemetry ecosystem.`

austinlparker: I like it!

PR #105, opened 2020-02-05
fix the ref for the GitHub badge while on the root page

zenmoto:
The Hugo script for the GitHub badge was returning an empty page
when on the root of the site, which then got concatenated into
a URL that didn't exist on GitHub. This commit checks to see if
that page portion is empty, and if so sets the URL for the badge
to the OpenTelemetry organization page


zenmoto:
I signed the CLA.

austinlparker:
Thanks! This is probably a better option than how I had it before.

PR #139, opened 2020-03-16
Fix extra space between bracket and paren

ymotongpoo:
Nit fix for the markdown to be rendered properly.


ejoreo:
This was fixed in #121 -- thanks though!

PR #140, opened 2020-03-17
Add initial collector docs

flands:


content/docs/collector/about.md
```diff
@@ -0,0 +1,106 @@
...
+
+1. An agent running with the application or on the same host as the application
+(e.g. binary, sidecar, or daemonset)
+2. A collector running as a standalone service (e.g. container or deployment)
+
+### Agent
+
+It is recommended to deploy the Agent on every host within an environment. In
+doing so, the Agent is capable of adding metadata to observability data it
+receives or collects. In addition, the Agent can offload responsibilities that
```
tigrannajaryan: >capable of adding metadata

This is not limited to Agent deployment. Perhaps mention elsewhere.

content/docs/collector/about.md
```diff
@@ -0,0 +1,106 @@
...
+## Getting Started
+
+### Demo
+
+Deploys a load generator, agent, and collector as well as Jaeger, Zipkin and
+Prometheus back-ends. More information can be found on the demo
+[README.md](https://github.com/open-telemetry/opentelemetry-collector/tree/master/examples/demo)
+
+```bash
+$ svn export https://github.com/open-telemetry/opentelemetry-collector/trunk/examples/demo; \
```
tigrannajaryan: What does `svn export` do? Can SVN connect to github? Why not `git clone`?

content/docs/collector/about.md
```diff
@@ -0,0 +1,106 @@
...
+Builds the latest version of the collector based on the local operating system,
+runs the binary with all receivers enabled and exports all the data it receivers
+locally to a file. Data is sent to the container and the container scrapes its own
+Prometheus metrics.
+
+```bash
+$ git clone git@github.com:open-telemetry/opentelemetry-collector.git; \
+    cd opentelemetry-collector; make otelcol; \
+    export OTEL_AGENT_ENDPOINT=0.0.0.0:55678; \
+    ./bin/$($GOOS)/otelcol --config ./examples/otel-local-config.yaml & pid1="$!"; \
```
tigrannajaryan: Is $GOOS assumed to be defined?

content/docs/collector/about.md
flands: It downloads a specific file instead of the whole repository -- I am not aware of a better way to do this but open to suggestions. Cloning everything to get a file is not ideal (but neither is having a dependency on svn...)

content/docs/collector/about.md
flands: Hmm, I thought I fixed this -- I will update

content/docs/collector/about.md
mtwo: Agreed with Tigran - git commands are preferable to svn, even if it means downloading extra data. In my experience most demos require cloning an entire repository.

austinlparker:
Is this documentation also available in the collector repository?

mtwo:
@flands same question as Austin - are the docs also in the Collector repository?

content/docs/collector/about.md
flands: Switched

flands:
@austinlparker @mtwo -- my plan is to remove the duplicate documentation from the collector repository and add a link to this documentation (see https://github.com/open-telemetry/opentelemetry-collector/pull/581). Having/maintaining the documentation in two places is a recipe for issues down the line.

PR #146, opened 2020-03-20
tabs shortcode with erlang/elixir as example

tsloughter:
It took me a lot of fiddling but I think I've got a usable bulma based tab shortcode added.

Example:

```
{{<tabs Erlang Elixir>}}

{{<tabsection>}}
{{< highlight erlang >}}
{deps, [{opentelemetry_api, "~> 0.3"}]}
{{< / highlight >}}
{{</tabsection>}}

{{<tabsection>}}
{{< highlight elixir >}}
defp deps() do
  [
    {:opentelemetry_api, "~> 0.3"}
  ]
end
{{< / highlight >}}
{{</tabsection>}}
{{</tabs>}}
```

Screenshot_2020-03-20 OpenTelemetry


tsloughter:
@austinlparker can you tell what is going wrong with the checks? I took a look and it is something about a hash..

layouts/shortcodes/tabs.html
```diff
@@ -0,0 +1,12 @@
+<div id="tabs-with-content">
+  <div class="tabs is-left is-toggle">
+    <ul>
+      {{ range $index, $value := .Params }}
+      <li><a href="#{{$index}}">{{$value}}</a></li>
```
austinlparker: I think this is what the build is complaining about -- whatever anchor this is pointing to doesn't exist.

austinlparker:
ok luc's change should fix this, rebase and you'll be g2g

PR #166, opened 2020-04-02
Add homepage update about logs working group

mtwo:
Fits under the beta box, though I'd prefer if this and the "Discover Integrations" box could sit side by side. @austinlparker is that possible?


austinlparker:
Yeah, it's possible. Do you want to merge this and have me make another PR to fix it, or do you want to try to and change it yourself? You'd need to edit the beta column as well to make it not extend across the window.

mtwo:
Let's merge and then followup with another PR if that's ok

austinlparker:
kk

PR #171, opened 2020-04-08
Update demo Docker image to core

brianashby-sfx:
Demo Docker image changed from distrib to core version.

CC @flands @pjanotti


brianashby-sfx:
I signed it.

PR #191, opened 2020-06-10
add kafkajs plugin to the registry

blumamir:


austinlparker:
Hi, there's a slight change to the registry format, we've added a new `language` tag - check out the updated files in master to see how it works. If you could add that, it'd be great!

blumamir:
Thanks @austinlparker 
Changed the format :)

PR #196, opened 2020-06-11
Change build to not fail if token unset

austinlparker:


austinlparker:
Changed netlify settings to pass token to branch builds

PR #199, opened 2020-06-15
Optimize page load for main page by converting images to Hugo assets

austinlparker:
Big speedup by reducing file sizes, addresses #172


austinlparker:
Lighthouse report from live: 
image

Lighthouse report from preview:
image

'bout 9s speedup on load

PR #205, opened 2020-06-18
Add OpenTelemetry-Web to site

austinlparker:
This adds opentelemetry-web to the site with the DocumentLoad() plugin to analyze performance.


austinlparker:
it's working, woo

PR #211, opened 2020-06-30
Adding pre-existing go instrumentation plugins

evantorrie:
Adding
   * Macaron
   * Labstack
   * Mongodb


austinlparker:
Could you add a language field, as in the README? Thanks!

evantorrie:
Language field added to the plugins I added, but also to the preexisting plugins owned by the Otel-go repo (which were also missing).

Issue #212, opened 2020-06-30
Remove jQuery dependency from registry search


connorlindsey:
The registry search currently uses jQuery which could be removed by refactoring the search.js script. Lighthouse lists removing jQuery as a good opportunity to increase our performance scores. I plan to replace any jQuery used with standard JS functions.

I would like to pick up this issue as well if the maintainers could please assign me to it. Thanks!

austinlparker:
Closed by #218, thanks!

PR #218, opened 2020-07-16
Removed jQuery dependency from registry search

connorlindsey:
I changed all jQuery calls to vanilla JS and removed jQuery from the registry html files. 

This closes issue #212


austinlparker:
Thanks! I'll take a look at this tomorrow probably.

PR #230, opened 2020-08-12
Updated google opentelemetry python exporter registry

aabmass:
- The exporters are a single package now, so combined them into one
- Updated repo link, authors, and description
- Added a registry entry for the opentelemetry-tools-google-cloud package.
  * creates a new category "tools"


content/registry/tools-python-google-cloud.md
```diff
@@ -0,0 +1,19 @@
+---
+title: Google Cloud Tools
+registryType: tools
```
aabmass, line 3: There might be a better category name for this than "tools". The package has a trace propagator and a resource detector. Thoughts?

assets/sass/registry.sass
```diff
@@ -5,6 +5,7 @@ $dark-blue: #39a9da;
 $core: #2C3ED3
 $instrumentation: #D32C3E
 $exporter: #3ED32C
+$tools: #2CD3B4
```
aabmass, line 8: I just generated this color with some tool online. Is there a better color available?

content/registry/tools-python-google-cloud.md
austinlparker: 'extensions', maybe?

assets/sass/registry.sass
austinlparker: it's probably fine for now

content/registry/tools-python-google-cloud.md
aabmass: I noticed the registry page says (https://opentelemetry.io/registry/)
>The OpenTelemetry Registry allows you to search for instrumentation libraries, tracer implementations, utilities, and other useful projects in the OpenTelemetry ecosystem.

Maybe "utilities" then?

austinlparker: yeah, let's go with that, it's probably the most accurate for now.

Issue #203, opened 2020-06-17
Documentation Tracking Issue


austinlparker:
For GA, we need to have docs for each language in GA on the site. This issue will track discovery and implementation.

**Please tag any docs issue or PR with the docs label**
Kanban: https://github.com/open-telemetry/opentelemetry.io/projects/1

**Need To Have**
- Discuss/edit this page link
- Search
- i18n
- Accessible

**To Do**
- [x] Evaluate existing CNCF docs sites
- [x] Create requirements for SIG docs
- [x] Supporting documentation and examples for creating docs in new format
- [ ] Write documentation!

austinlparker:
Adding some notes here for myself/anyone else following at home - 

I've started looking into the feasibility of porting the site over to the Docsy theme (used by k8s doc site, most notably). To its credit, it checks a lot of boxes for us out of the box. Here's a preview of what the docs section could look like (most options in the theme are on defaults, so colors/fonts/etc are technically WIP) - 
![image](https://user-images.githubusercontent.com/4140740/92012933-17f42e80-ed1b-11ea-8e32-4a1a05e5cb19.png)

We get a lot of stuff for free with this - deep linking back into GH for issues/edits to pages, automatic revision history slugs, i18n, and significantly better accessibility audits (77 on current custom theme, 91 with docsy, mostly due to better ARIA compliance). 

The biggest lift is probably going to be migrating over the registry, but I don't think it'd be _that_ bad all things considered. We'll also need to rework the status page, but that was getting long in the tooth anyway.

austinlparker:
Pardon the auto-translation, but here's an example of how the multi-language works - you get a dropdown in the corner (these languages need to be defined in `config.toml`) - 
![image](https://user-images.githubusercontent.com/4140740/92015049-33ad0400-ed1e-11ea-9568-180abb093614.png)

In the repo itself, you just need to ensure that your content is defined per-language - 
![image](https://user-images.githubusercontent.com/4140740/92015126-5212ff80-ed1e-11ea-96f2-3599fe18ec6a.png)

One thing to note, the language dropdown appears to be global, but there's no fallback to the english text for pages that aren't defined (they're just blank, that's why there's no sections in top nav in the above screenshot, I didn't create the subtrees for other stuff). It's not really a huge problem, I reckon.

austinlparker:
working on some ideas for the landing page: 
[...]

PR #247, opened 2020-09-03
WIP: Site Refresh

austinlparker:
For #203


ffe4:
Should we pick up here from #203 to reduce crosstalk (the issue is called Docs Section Rework after all)? I'm not sure if you want to get the whole site done in this PR, so it might be better to move the discussion into separate issues?

One thing I immediately noticed while scrolling was that the fixed navbar does not really work as well when it's transparent.

Screen Shot 2020-09-17 at 20 35 16

On 1080p, Chrome with bookmarks bar, I also have to scroll to reach the scroll button.

As I mentioned in the issue, I think it would be good to define the content we want to have first and then build around that, so I'll try to share some of my thoughts. For now, just on the landing page.

Regarding content I think we _must_ have on the landing page:
- blurb explaining OpenTelemetry
- link to getting started guide
- mention of library and language support

I think all of these should be included in the first impression on desktop. I think library and language support deserves an extra section, however, so I would make that the first scroll target, and tease it either via text near the scroll, or by adjusting the layout so that the section heading is visible without scrolling on most desktop browsers.

We might also want to have:
- link to GitHub
- exporters
- some talk about metrics and tracing (and logs?)

The GitHub link can go in the header. Not sure whether exporters are an important selling point or not, but they represent the next step after instrumentation, so it makes sense to have them be the section after library support. However, I am not sure what to do about the metric/tracing topic. It might kind of fit in with exporters, but maybe they should be mentioned in the blurb, and then be part of a detailed description of the problem domain and OpenTelemetry's role in it on another page.

Anything you think we should or should not have?

jkowall:
Before I submit a PR, I wanted to suggest some new text:

**Current:**
Robust and portable metrics, logs, and traces for cloud-native software.
OpenTelemetry provides a single set of APIs, libraries, agents, and collector services to capture distributed traces and metrics from your application. You can analyze them using Prometheus, Jaeger, and other observability tools.

**Proposed 1 :**
Metrics, logs, and traces for the cloud-native engineer
OpenTelemetry is a set of multi-language open source APIs, libraries, and agents to capture distributed logs, traces, and metrics from your application and infrastructure. The collector delivers telemetry to the observability system of your choosing, including popular CNCF projects like Prometheus and Jaeger.

**Proposed 2 :**
Metrics, logs, and traces for the cloud-native engineer
OpenTelemetry is a set of open source APIs, libraries, and agents to capture distributed logs, traces, and metrics and deliver telemetry to the observability system of your choosing.

I am also working on design elements for the technology list and an overall high-level diagram with icons for the components. One of our designers did this for me, which is already more usable.

image

austinlparker:
ok, there might be some nitpicks but I think I've got enough of the site moved over that we can think about pushing it out and getting started on content dev

cc @mtwo for review

mtwo (approved):

PR #268, opened 2020-10-16
add erlang/elixir to list of languages with support

tsloughter:


austinlparker (approved):

austinlparker:
out of curiosity should I make all references to 'erlang' ref 'erlang/elixir'?

tsloughter:
@austinlparker probably. Some newer Elixir people or people considering Elixir might be more likely to check it out if it is clearly stated that there is an Elixir API included.

austinlparker:
mmk, lemme make a note. thanks!

PR #282, opened 2020-10-23
Add templates for documentation

austinlparker:
This adds template markdown files that can be used to bootstrap docs writers

PR #300, opened 2020-11-05
Add documentation templates

austinlparker:
This is #282 but cleaned up because I made git and myself very unhappy.


flands (approved):
LGTM

Issue #284, opened 2020-10-23
Documentation - Go


austinlparker:
Tracking issue for the Go documentation section. See `./templates/language-docs` for markdown templates.

Content should live at `./content/en/docs/go/`

- [x] Index
- [x] Getting Started
- [x] Instrumentation
- [x] Export and Config

PR #311, opened 2020-11-09
Initial OTel Go Documentation

austinlparker:
This adds initial documentation for Go per #284


content/en/docs/go/exporting_data.md
```diff
@@ -0,0 +1,62 @@
+---
+title: "Processing and Exporting Data"
+weight: 4
+---
+
+Once you've instrumented your code, you need to get the data out in order to do much useful with it. This page will cover the basics of the process and export pipeline.
```
flands, line 6:
```suggestion
Once you've instrumented your code, you need to get the data out in order to do anything useful with it. This page will cover the basics of the process and export pipeline.
```

content/en/docs/go/getting-started.md
```diff
@@ -0,0 +1,127 @@
...
+When using OpenTelemetry, it's a good practice to set a global tracer provider. Doing so will make it easier for libraries and other dependencies that use the OpenTelemetry API will be able to easily discover the SDK, and emit telemetry data. In addition, you'll want to configure context propagation options. Context propagation allows for OpenTelemetry to share values across multiple services - this includes trace identifiers, which ensure that all spans for a single request are part of the same trace, as well as baggage, which are arbitrary key/value pairs that you can use to pass observability data between services (for example, sharing a customer ID from one service to the next).
+
+Setting up global options uses the `global` package - add these options to your `main.go` file as shown -
+
+```
+	global.SetTracerProvider(tp)
+	global.SetTextMapPropagator(propagators.Baggage{})
+```
+
+It's important to note that if you do not set a propagator, the default is to use the NoOp option, which means that context will not be shared between multiple services.
```
flands, line 89:
```suggestion
It's important to note that if you do not set a propagator, the default is to use the `NoOp` option, which means that context will not be shared between multiple services.
```

content/en/docs/go/getting-started.md
```diff
@@ -0,0 +1,127 @@
...
+
+# Quick Start
+
+Let's put the concepts we've just covered together, and create a trace in a single process. In our main function, after the initialization code, add the following -
```
flands, line 93:
```suggestion
Let's put the concepts we've just covered together, and create a trace in a single process. In our main function, after the initialization code, add the following:
```

flands (approved):
LGTM, minor comments

PR #327, opened 2020-11-18
make concepts come first in the docs page sidebar

tsloughter:
Each language entry has weight 10 so that they are sorted in
alphabetical order. Concepts is weight 1 so it comes first and
other entries to come after each language, like collector and
workshops get a multiple of 10 weight.

Before this change `.Net` was the first entry on the docs page sidebar, now it is `Concepts`.


austinlparker (approved):

austinlparker:
lgtm, thanks

PR #331, opened 2020-12-02
Add AWS SDK Extension package to registry

NathanielRN:
Adds a reference to the python package which allows OpenTelemetry Python users to configure the generation of traces to be compatible for the AWS X-Ray service.

### DIFF
--- a/content/en/registry/sdk-extension-python-aws.md
+++ b/content/en/registry/sdk-extension-python-aws.md
@@ -1,0 +1,14 @@
+---
+title: AWS SDK Extension
+registryType: utilities
+isThirdParty: true
+language: python
+tags:
+  - python
+  - sdk-extension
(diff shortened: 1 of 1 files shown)


content/en/registry/sdk-extension-python-aws.md
```diff
@@ -0,0 +1,14 @@
+---
+title: AWS SDK Extension
+registryType: sdk-extension
```
austinlparker, line 3: Could this be changed to 'utilities'? We don't have a mapping for sdk-extension so it breaks the filters/labels.

austinlparker (changes requested):

content/en/registry/sdk-extension-python-aws.md
NathanielRN: I've changed it to be `utilities`! I think that's a good enough label of what it is 👍

austinlparker (approved):

austinlparker:
lgtm thanks!

PR #348, opened 2020-12-27
use css `scroll-margin-top` to offsetting an html anchor to adjust for fixed header 

WPH95:
1. found the problem.
image

reproduce this problem by the following URL:
- https://opentelemetry.io/about/#what-is-an-observability-framework
- https://opentelemetry.io/community/#opentelemetryio

2. found CSS code has handled this problem, btw not working.
3. fount the `h2` parent `section` used `display: flex`, make old logic not working.
4. use CSS `scroll-margin-top: {header-heigh}` to fix it :)

### DIFF
--- a/assets/scss/main.scss
+++ b/assets/scss/main.scss
@@ -50,13 +50,10 @@ footer {
         visibility: hidden;
     }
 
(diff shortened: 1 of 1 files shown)


austinlparker (approved):
lgtm, thanks!

PR #349, opened 2020-12-27
registry page: persistent value from filter language/component filter

WPH95:
## why need it?
for example, a freshman needs to search a JAVA AWS OTel SDK, maybe follow this step:
1. select language to `java`.
2. input search query: `aws` 
3. bomb. return a list with `aws`, not include `java` (because we don't persistent selected language/component)

## how this PR to fix it.
add two hide input, when form submit. will take language/component value.
when language/component value changed, update input value and urlParams


austinlparker (approved):
lgtm thanks!

PR #372, opened 2021-02-04
add index and getting started guide for Erlang along with tabs shortcode

tsloughter:
The index and getting started pages are basically copies of the same Go pages. I have 2 other pages that I'm working on to be added later. This PR is a draft so I can share it for comments but version 0.6.0 has not be released yet so it is not ready to be published.

The tabs shortcode was added to allow both Erlang and Elixir examples to be listed together.

There is an issue I need help with, assuming the tabs shortcode is acceptable aside from this, there is a gap between the tabs and the content that I can't explain (image below). I tried inspecting the element in firefox but couldn't find what was causing it to space like this on every set of tabs.

Screenshot_2021-02-04 Getting Started


austinlparker:
The gap is due to the `highlight` class that's applied to preformatted blocks. It doesn't show on the parent div for weird css reasons, presumably, but if you drill down a bit you should see it. I don't think that it's a huge problem tbh.

tsloughter:
Aah, yea, now I see the spacing coming from the div with highlight class.

tsloughter:
Ok, so I have some css to fix it, but not sure where to actually include it. I shoved it in `main.scss`, is that right?

```
.td-content .highlight {
	margin: 2rem 0;
	padding: 0;
}
```

austinlparker:
yeah, or just make a new css file and import it to main

tsloughter:
Ok, cool. I just went by what seemed to be the convention (a file with an underscore, `_tabs.scss`, and import the name without an underscore, `@import "tabs";`) and it seems to have worked :)

mtwo (approved):
Just skimmed this, looks great

austinlparker (approved):

PR #414, opened 2021-02-17
improve link css 

### LINKED ISSUE #384: CSS for links needs to be fixed when primary color is used
If the background for a section is set to the primary color (blue) then links are hard to see on hover.

Screen Shot 2021-02-14 at 11 31 53 PM

Screen Shot 2021-02-14 at 11 31 58 PM

In addition, even without hover the "white" color is not apparent enough and should be fixed.

WPH95:
resolved #384 
1. all a/link use the same color
2. remove hover color change, used underline

before

image

now
image

### DIFF
--- a/assets/scss/main.scss
+++ b/assets/scss/main.scss
@@ -69,3 +69,14 @@ section#announcement {
     justify-content: center;
     padding: 6px;
 }
+
+
+.td-box--dark p > a {
+    color: #3176d9;
+}
+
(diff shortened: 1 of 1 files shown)


austinlparker (approved):

flands:
Nice! A nit, which can be handle as a separate issue:

Screen Shot 2021-02-17 at 8 47 16 AM

PR #415, opened 2021-02-17
registry update plugins to instrumentation

nirsky:
In our (Aspecto) extensions [repo](https://github.com/aspecto-io/opentelemetry-ext-js) we updated our plugins to the new [instrumentation API](https://github.com/open-telemetry/opentelemetry-js/tree/main/packages/opentelemetry-instrumentation).  
As part of this change, the term "plugin" changes, this PR is updating the term where needed.  

We also created a new mongoose instrumentation, that contains bug fixes to the original plugin, as well follows the semantic conventions (as opposed to the previous plugin) and uses the new API, so that's added in this PR as well.


flands (approved):
LGTM, thanks!

PR #417, opened 2021-02-18
Update Sumo Logic OTLP support status

pmm-sumo:
Sumo Logic now accepts OTLP/HTTP traces and I believe the status for Native OTLP support can be now updated


austinlparker (approved):

austinlparker:
thanks!

PR #456, opened 2021-02-24
Added core collector receivers to registry

gramidt:
Added core collector receivers to registry

### DIFF
--- a/content/en/registry/collector-receiver-fluent-forward.md
+++ b/content/en/registry/collector-receiver-fluent-forward.md
@@ -1,0 +1,15 @@
+---
+title: Fluent Forward Collector Receiver
+registryType: receiver
+isThirdParty: false
+language: collector
+tags:
+  - go
+  - receiver
+  - collector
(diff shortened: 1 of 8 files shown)


austinlparker (approved):

austinlparker:
Oh, I'm actually not sure if that 'receiver' type is going to work with the type dropdown.

austinlparker:
Yeah, looks like it doesn't show up. Maybe 'core'? You could also add a new type (look at how they're implemented in `layouts/registry/list.html`, you need a new lookup there and some css)

gramidt:
Great catch, @austinlparker! Didn't serve the page this morning when starting. Dope...Need more coffee. Adding now.

gramidt:
@austinlparker - Do you have a recommendation on the background color for types "receiver" and "processor"? I'll submit this via a different PR.

austinlparker:
I don't really have any strong opinions. The text is white, so something that's readable with white text on it.

PR #487, opened 2021-03-10
added additional details about slack sign-up and channels

lakamsani:


lakamsani:
will submit another PR with a cleaner merge

PR #501, opened 2021-03-27
Improvements while going through the Elixir guide

GregMefford:
While working through the guide, I found a few things to correct grammatically, and also tried to clarify a few things that I think may not be obvious to new-comers.


GregMefford:
It looks like `CODEOWNERS` is set up to be global to the whole repo, so just a heads up to @open-telemetry/erlang-approvers about this as well.

ferd (approved):

flands (approved):

PR #532, opened 2021-04-13
Remove duplicated file

vmihailenco:
`instrumentation-js-fetch.md` content is exactly the same


shelbyspees:
Good catch. Just for clarification, the reason you chose to keep `fetch` over `xml-http-request` is because it's the newer API? ([TIL](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch))

vmihailenco:
I did not give it much thought, but the plugin is called `opentelemetry-plugin-fetch` so I guess markdown should have resembling name...

shelbyspees (approved):

austinlparker:
Superseded by #534

PR #533, opened 2021-04-13
erlang/elixir: update getting started to use 1.0.0-rc1

tsloughter:
Forgot to update this when the RC was released. Once 1.0 is released it can just be updated to use like `~> 1.0` so that it doesn't have to be continually updated when new releases come out.

### DIFF
--- a/content/en/docs/erlang/getting-started.md
+++ b/content/en/docs/erlang/getting-started.md
@@ -54,15 +54,15 @@ run as a Release and export spans from.
 {{< tabs Erlang Elixir >}}
 
 {{< tab >}}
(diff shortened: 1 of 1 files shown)


mtwo (approved):
Nice!

shelbyspees (changes requested):
Heads up: this docs page now lives in the Erlang/Elixir SDK repo: https://github.com/open-telemetry/opentelemetry-erlang/blob/main/website_docs/getting-started.md

Please make the corresponding change over there and it'll get pulled into this repo.

tsloughter:
oh, right, forgot about that. Closing this and opening on the other repo. Thanks!

PR #571, opened 2021-05-07
.gitignore: don't ignore themes, and other tweaks

chalin:


.gitignore
```diff
@@ -1,14 +1,11 @@
 # Hugo-generated assets
-public/
+/public
 resources/
 
 # npm assets
 node_modules/
 
-# ignore themes
-themes/
```
chalin, line 9: Unless I'm misunderstanding the intent here, we don't want to ignore the top-level `themes` directory -- it was committed and is being tracked already. I doubt that the intent was to ignore `themes` subdirectories.

.gitignore
```diff
@@ -1,14 +1,11 @@
 # Hugo-generated assets
-public/
+/public
```
chalin, line 2: This change allows us to ignore the top-level `public` directory (which contains the generated site files), while allowing `/public` to be a symlink (e.g., to another git tracked repo -- a trick that is very useful for examining diffs of generated site files).

austinlparker (approved):

PR #587, opened 2021-05-20
Update vendors page

arminru:
The Dynatrace OneAgent serves as distro by picking up OpenTelemetry traces without the need for explicit configuration.


austinlparker:
Could you explain this a bit more? Is the Dynatrace agent a wrapper around upstream OpenTelemetry SDKs/tools, or is it just integrating OTLP export from your existing agent?

arminru:
@austinlparker Sure. It acts as a wrapper around the upstream SDK which is independent from the native OTLP ingest we offer in addition to that.

austinlparker:
Ah, cool, thanks for the clarification!

austinlparker (approved):

PR #626, opened 2021-07-05
Add missing JS instrumentations to registry

jtmalinowski:


MetinSeylan (approved):

Rauno56 (approved):
There are a few missing from [contrib](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/plugins/node) still. At least `restify` and `router`, if you don't mind adding those as well. Didn't do an exhaustive check.

jtmalinowski:
All Node.js contrib instrumentations are now in. Thanks @Rauno56

flands (approved):

Issue #636, opened 2021-07-14
Upgrade Hugo


chalin:
The website is currently using [Hugo 0.65.3](https://gohugo.io/news/0.65.3-relnotes/) from 2020/02/23. There have been a number of improvements (including performance) since then. As of the time of writing, the current version of [Hugo is 0.85.0](https://gohugo.io/news/0.85.0-relnotes/).

PR #659, opened 2021-08-11
Generate production CSS and JS in production environments only

chalin:
A simple override of a couple of Docsy partials (replacing `.Site.IsServer` by `ne (getenv "HUGO_ENV") "production"`), so that it is easier to locally build a non-production version of the site, which in turn makes it easier to `git diff` the generated site files. Being able to diff generated site files provides a means to validate refactors and sanity-check most infrastructure updates.


chalin:
Once we've updated Hugo to a more recent version (#636), we'll be able to use `hugo.IsProduction` -- cf. https://github.com/google/docsy/pull/413.

mtwo (approved):

PR #695, opened 2021-08-27
Publish OpenTelemetry Schema File 1.6.1

carlosalberto:
Also retire the broken 1.6.0 schema.

### DIFF
--- a/static/schemas/1.6.0
+++ b/static/schemas/1.6.0
@@ -1,6 +1,0 @@
-file_format: 1.0.0
-schema_url: https://opentelemetry.io/schemas/1.6.0
-versions:
-  1.6.0:
-  1.5.0:
-  1.4.0:
--- a/static/schemas/1.6.1
+++ b/static/schemas/1.6.1
@@ -1,0 +1,6 @@
+file_format: 1.0.0
+schema_url: https://opentelemetry.io/schemas/1.6.1
+versions:
+  1.6.1:
+  1.5.0:
+  1.4.0:


carlosalberto:
@open-telemetry/docs-approvers please review this.

austinlparker (approved):
lgtm

static/schemas/1.6.0
```diff
@@ -1,6 +0,0 @@
-file_format: 1.0.0
```
chalin, line 1: @austinlparker @carlosalberto - just curious: is there an issue here for folks attempting to access the 1.6.0 and getting a 404?

chalin: My guess is that we either want to preserve all version files "forever", or introduce appropriate redirect rules. But I could be wrong. @tigrannajaryan WDYT?

chalin: For my own reference, here's a link to the schema doc I found: https://github.com/open-telemetry/oteps/blob/main/text/0152-telemetry-schemas.md. (I don't know if this OTEP is superseded by some other doc once the proposal is approved.)

tigrannajaryan: This is `file_format` which is different from and unrelated to schema versions, which are listed in the `versions` section. `file_format` MUST be 1.0.0 for now. `file_format` may change in the future only if  we amend the schema file format itself via a new OTEP which introduces new capabilities in the schema files.

Unlike `file_format`, a new schema version is introduced every time we make a specification release. We are currently on 1.6.1 of specification and of the schema. 1.6.0 was a broken release and because of that it is not listed in the schema files _as if 1.6.0 never happened_.

chalin: I wasn't commenting on the `file_format` line, I was just attaching the comment to the deleted file `static/schemas/1.6.0` -- sorry if that wasn't clear.

> 1.6.0 was a broken release and because of that it is not listed in the schema files _as if 1.6.0 never happened_.

That answers my question, thanks.

chalin: Actually, I have a followup question: is there a problem if the content of these **two directories drift apart**?

- https://github.com/open-telemetry/opentelemetry.io/tree/main/static/schemas
- https://github.com/open-telemetry/opentelemetry-specification/tree/main/schemas

Should 1.6.0 be deleted from `opentelemetry-specification/schemas` too, "as if it never happened"?

tigrannajaryan: >Should 1.6.0 be deleted from opentelemetry-specification/schemas too, "as if it never happened"?

Good catch. Yes, it should be deleted.

The content of the directories should be in sync and I plan to add an action that syncs the content automatically.

chalin: The ideal situation is to avoid having the same files in two repos. I've opened #709 to track this issue and discussion.

PR #702, opened 2021-08-31
Link to Community Marketing Guidelines

### LINKED ISSUE #700: Add marketing guidelines to website
**What needs to be changed?**

Currently our marketing guidelines are not very discoverable to non-technical contributors, such as marketing people and managers working at contributing organizations. They should be added to our website to make them more visible.

I propose adding a new page - openTelemetry.io/guidelines - and link to it from opentelemetry.io/community.

Possibly it could be added to the bottom of the documentation sidebar as well?

**Additional context**
The current marketing guidelines can be found here:
https://github.com/open-telemetry/community/blob/main/marketing-guidelines.md

We could set it up to have the guidelines be pulled in automatically via website_docs to create a single source of truth, but I don't want to overcomplicate things.

austinlparker:
fixes #700

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -184,6 +184,11 @@ name ="Privacy Policy"
 url ="https://www.linuxfoundation.org/en/privacy/#:~:text=The%20Linux%20Foundation%27s%20core%20purpose,%2C%20a%20%E2%80%9CProject%E2%80%9D).&text=The%20Privacy%20Policy%20does%20not,employees%20or%20other%20TLF%20personnel." 
 icon =" fa fa-book"
 desc ="CNCF works in close association with Linux Foundation. Read the privacy policy here"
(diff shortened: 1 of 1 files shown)


chalin (approved):
LGTM

chalin:
Rereading #700, I see that this PR doesn't introduce a new guidelines page (as @tedsuo suggested), but it a step in the direction of making the guidelines more easily accessible. 

I agree with @tedsuo that we should aim to keep visitors on the website rather than direct them to a GH file, but we can handle that in a followup PR. If y'all agree, then we could say that this PR contributes to but doesn't yet close #700.

Thoughts?

austinlparker:
Yeah, we can take an item to translate those guidelines into a specific page but I'd like to get something out there as a stopgap.

PR #723, opened 2021-09-09
Add missing instruction for running locally

jessitron:
The 'hugo run serve' gave an error like this:

```
Error: Error building site: failed to render pages: render of "section" failed: "/workspaces/opentelemetry.io/layouts/registry/list.html:3:23": execute of template failed: template: registry/list.html:3:23: executing "main" at <resources.Get>: error calling Get: stat /workspaces/opentelemetry.io/themes/docsy/assets/js/registrySearch.js: no such file or directory
```

until I ran `npm run webpack`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -7,6 +7,7 @@
   * Please note that you need to install the "extended" version of Hugo (with built-in support) to run the site locally.
 * Install [npm](https://npmjs.com)
 * Run `npm install`
+* Run `npm run webpack`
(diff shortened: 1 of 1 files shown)


chalin:
Thanks for reporting the issue! I've created #724 with your opening comment to track the problem.

chalin:
I'm going to solve this slightly differently, so that users don't need to run another script.

chalin:
Closing in favor of #725.

jessitron:
Thank you!

PR #740, opened 2021-09-17
Add `Baggage` Signal to "Concepts"

### LINKED ISSUE #711: Data Sources should include baggage signals
Noticed that [the data sources page](https://opentelemetry.io/docs/concepts/data-sources/) doesn't yet include notes about [Baggage Signals](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/overview.md#baggage-signal). Should be trivial to translate the gist of the content -- I might come back and write it if nobody takes the issue :)

GaryPWhite:
Resolves #711

---

Preview: https://deploy-preview-740--opentelemetry.netlify.app/

Closes #711

### DIFF
--- a/content/en/docs/concepts/data-sources.md
+++ b/content/en/docs/concepts/data-sources.md
@@ -126,3 +126,17 @@ For more information, see the [logs
 specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/overview.md#log-signal),
 which covers topics including: log, defined fields, trace context fields and
 severity fields.
+
+## Baggage
+
+In addition to trace propagation, OpenTelemetry provides a simple mechanism for propagating name/value pairs, called `Baggage`. `Baggage` is intended for indexing observability events in one service with attributes provided by a prior service in the same transaction. This helps to establish a causal relationship between these events.
+
+While `Baggage` can be used to prototype other cross-cutting concerns, this mechanism is primarily intended to convey values for the OpenTelemetry observability systems.
+
+These values can be consumed from `Baggage` and used as additional dimensions for metrics, or additional context for logs and traces. Some examples:
+
+- A web service can benefit from including context around what service has sent the request
+- A SaaS provider can include context about the API user or token that is responsible for that request
+- Determining that a particular browser version is associated with a failure in an image processing service
+
+For backward compatibility with OpenTracing, `Baggage` is propagated as `Baggage` when using the OpenTracing bridge. New concerns with different criteria should consider creating a new cross-cutting concern to cover their use-case; they may benefit from the W3C encoding format but use a new HTTP header to convey data throughout a distributed trace.
--- a/content/en/docs/concepts/data-sources.md
+++ b/content/en/docs/concepts/data-sources.md
@@ -130,7 +130,7 @@ severity fields.
 ## Baggage
 
 In addition to trace propagation, OpenTelemetry provides a simple mechanism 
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/data-sources.md
```diff
@@ -126,3 +126,17 @@ For more information, see the [logs
 specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/overview.md#log-signal),
 which covers topics including: log, defined fields, trace context fields and
 severity fields.
+
+## Baggage
+
+In addition to trace propagation, OpenTelemetry provides a simple mechanism for propagating name/value pairs, called `Baggage`. `Baggage` is intended for indexing observability events in one service with attributes provided by a prior service in the same transaction. This helps to establish a causal relationship between these events.
```
chalin, line 132: Why use code font in this new section? E.g., this paragraph could read:

> ... called **baggage**. Baggage is ...

We can continue in the manner in the next paragraph:

> While baggage can ...

content/en/docs/concepts/data-sources.md
```diff
@@ -126,3 +126,17 @@ For more information, see the [logs
...
+
+While `Baggage` can be used to prototype other cross-cutting concerns, this mechanism is primarily intended to convey values for the OpenTelemetry observability systems.
+
+These values can be consumed from `Baggage` and used as additional dimensions for metrics, or additional context for logs and traces. Some examples:
+
+- A web service can benefit from including context around what service has sent the request
+- A SaaS provider can include context about the API user or token that is responsible for that request
+- Determining that a particular browser version is associated with a failure in an image processing service
+
+For backward compatibility with OpenTracing, `Baggage` is propagated as `Baggage` when using the OpenTracing bridge. New concerns with different criteria should consider creating a new cross-cutting concern to cover their use-case; they may benefit from the W3C encoding format but use a new HTTP header to convey data throughout a distributed trace.
```
chalin, line 142: "... `Baggage` is propagated as `Baggage` ..." this doesn't make sense to me w/o more context.

If OT has a concept of baggage, and OT baggage is propagated as OTel baggage through the bridge, why is it worth pointing this out?

chalin (changes requested):
Thanks for this @GaryPWhite.
I've inlined a few initial comments.

Could you line wrap all of the text to 80 columns? This makes it easer to: review, create diffs, and eventually, support internationalization efforts.

content/en/docs/concepts/data-sources.md
GaryPWhite: I've used the formatting I saw int he original snippet this is from. Happy to edit it!

content/en/docs/concepts/data-sources.md
GaryPWhite: from the examples I can find in "best practices" for opentracing, "baggage" exists in relation to a "SpanContext" -- or a trace. My understanding of `Baggage` for OTel is that it can be [attached to a trace, metric, or log](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/baggage/api.md#overview) -- the distinction here being that it may or may not be attached to a `span`. i'm also not married to including this -- this is something from the specification, not from my own brain.

chalin: Come to think of it, backwards compatibility topics probably don't belong in (what should be) a simple introduction to the concept. WDYT?

GaryPWhite: yeah, let's leave that for the in-depth docs. I'll take it out.

Issue #746, opened 2021-09-21
Refactor and cleanup assets/scss


chalin:
Contributes to #744 by refactoring and cleaning up SCSS files:

- Move project-specific SCSS definitions and overrides into `assets/scss/_styles_project.scss`
- Delete duplicate SCSS files that unnecessarily shadow docsy files of the same name
- Drop obsolete SCSS

PR #377, opened 2021-02-13
Add announcement so the index can display an announcemen

WPH95:
resolved #375 
image

PR #747, opened 2021-09-21
assests/scss: cleanup buttons & nav, a first pass

chalin:
- Contributes to #746
- No effective change to the website.


assets/scss/_buttons.scss
```diff
@@ -1,12 +0,0 @@
...
-  box-shadow: none;
-
-  &-primary {
-    background: $primary;
-  }
-
-  &-secondary {
-    background: $secondary;
-  }
-}
```
chalin, line 12: All of these definitions correspond to Bootstrap and/or Docsy defaults.

assets/scss/_nav.scss
```diff
@@ -24,10 +24,6 @@
     opacity: inherit;
 }
 
-.navbar-announcement-onscroll {
-    position: absolute !important;
-}
-
```
chalin, line 30: This was introduced by #377, which was improved by #413. `.navbar-announcement-onscroll` no longer seems to be used.

assets/scss/_nav.scss
```diff
@@ -99,41 +95,4 @@
...
-            text-rendering: auto;
-            -webkit-font-smoothing: antialiased;
-            font-family: "Font Awesome 5 Free";
-            font-weight: 900; 
-            content: "\f0d9";
-            padding-left: 0.5em;
-            padding-right: 0.5em;
-        }
-    }  
-    }
```
chalin, line 139: Moved into `_styles_project.scss`.

chalin: After this move, `_nav.scss` matches the Docsy version of the file, so we'll be removing it in a followup PR. (I didn't do it here to make it more obvious which part is being moved into `_styles_project.scss`.

PR #758, opened 2021-09-22
Fix banner on mobile

chalin:
- Closes #757 
- This moves the banner to be under the topnav, which IMHO makes it much more visibile.
- I still have to strip out the code mods from `src/js/base.js`, but I'll do that in a followup PR.

Preview: https://deploy-preview-758--opentelemetry.netlify.app/

---

### Screenshots (see #757 for the "before" screenshot):

Mobile:

Screen Shot 2021-09-22 at 6 15 00 AM

Desktop:

Screen Shot 2021-09-22 at 6 15 24 AM

Issue #757, opened 2021-09-22
Homepage doesn't display properly on mobile


chalin:
This is what the homepage looks on mobile (iOS, for example):

> <img width="400" alt="Screen Shot 2021-09-22" src="https://user-images.githubusercontent.com/4140793/134317725-d1e244de-7a15-49b3-a81d-f6f8286db784.png">

Initial experiments show that the "announcement" feature might be the problem.

PR #760, opened 2021-09-22
Homepage top nav: always use std bg color

chalin:
- Followup to #758, and so also contributes to #757
- This PR effectively undoes the changes introduced by:
  - #377
  - #413
- The "Bootstrap Fixed Header" code in `base.js` has been reverted to match Docsy

With the CSS changes in this PR, the homepage top nav has the same appearance and behaves the same as for any other site pages:

> image

### DIFF
--- a/assets/scss/_styles_project.scss
+++ b/assets/scss/_styles_project.scss
@@ -4,8 +4,8 @@
 @import "registry";
 @import "tabs";
 
-.navbar-position-absolute {
-  position: absolute !important;
+.td-navbar {
+  background: $primary !important;
 }
 
 // Icons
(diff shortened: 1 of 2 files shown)


assets/scss/_styles_project.scss
```diff
@@ -4,8 +4,8 @@
 @import "registry";
 @import "tabs";
 
-.navbar-position-absolute {
-  position: absolute !important;
+.td-navbar {
+    background: $primary !important;
```
chalin, line 8: IMHO, it never really worked to have a transparent top nav given the current homepage design. By default Docsy will set the top-nav background to the `$primary` color but only after the user has scrolled past the `blocks/cover`. This change sets the background to always be the primary color.

austinlparker (approved):
lgtm

PR #761, opened 2021-09-22
Add line-terminating whitespace to Metrics > SDK

chalin:
Followup to #752. /cc @reyang 

Preview: https://deploy-preview-761--opentelemetry.netlify.app/status/


chalin:
We _really_ should change the formatting of this file to use explicit `<br>` elements (or some other formatting entirely) rather than relying on trailing whitespaces.

WDYT @austinlparker?

chalin:
> WDYT @austinlparker?

In case you agreed, I went ahead and created the following issue to track the needed update: #764

PR #768, opened 2021-09-23
Setup use of a ConsoleSpanExporter during development

chalin:


chalin:
Approved by @austinlparker out of band.

PR #790, opened 2021-09-29
Upgrade to Hugo 0.88.1 via NPM

chalin:
I'm having some strange site-generation issues with the current version of Hugo. The latest version seems to be fine.

This PR switches to using the very convenient [hugo-extended](https://www.npmjs.com/package/hugo-extended) NPM package to fetch the proper version of Hugo. This makes is easier for contributors -- who will no longer need to separately install Hugo.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -3,13 +3,11 @@
 ## Dev Setup
 
 * Fork and clone the repository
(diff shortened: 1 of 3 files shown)


chalin:
The only change in the generated site files (modulo the Hugo version embedded in files), is the addition of a `tabindex="0"` on `<pre>` elements.

austinlparker (approved):

PR #763, opened 2021-09-22
Drop top-nav entry icons, and other tweaks

chalin:
Closes #753

PR #550, opened 2021-04-18
Addition of navbar icons

harshita19244:
Provides an extra feature modification in accordance with https://github.com/open-telemetry/opentelemetry.io/issues/511

PR #811, opened 2021-10-06
Drop obsolete navbar style customization

chalin:
- Final cleanup step following #810 and #763, completely reverting #550 (since, as I mentioned elsewhere, this is a standard feature in more recent versions of docsy.
- There is no change in generated site files and no resulting change in appearance (none that I can see when doing a side-by-side compare).

### DIFF
--- a/assets/scss/_styles_project.scss
+++ b/assets/scss/_styles_project.scss
@@ -52,39 +52,6 @@
   background: $primary !important;
 }
 
-// Icons
-#main_navbar {
-  li i {
-  padding-right: 0.5em;
-
-      &:before {
-          display: inline-block;
-          text-align: center;
-          min-width: 1em;
-      }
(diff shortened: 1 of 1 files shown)


assets/scss/_styles_project.scss
```diff
@@ -52,39 +52,6 @@
...
-  li i {
-  padding-right: 0.5em;
-
-      &:before {
-          display: inline-block;
-          text-align: center;
-          min-width: 1em;
-      }
-  }
-.alert {
```
chalin, line 66: Btw, this nested definition / override for `.alert` under `#main_navbar`, doesn't make any sense to me -- I can't see a case where a standard Bootstrap alert would be used inside the navbar. (Interestingly, the website doesn't have any alerts at all.)

PR #820, opened 2021-10-12
Add 3rd party instrumentation for Go database/sql from @XSAM

MrAlias:


bhs:
I think this may have caused opentelemetry.io to become unavailable due to a formatting issue in this PR. I'm going to try rolling it back.

PR #822, opened 2021-10-13
Fix opentelemetry.io

### LINKED ISSUE #821: the home page shows 404
**What happened?**
the home page of opentelemetry now shows page not found

bhs:
#820 introduced a syntax issue which broken the website build (or so I think??). This seems to fix the syntax, though admittedly I can't get the website to build locally with or without #820 so I'm in breakfix mode.

That said, I can verify that this PR fixes the metadata syntax which I believe is likely to be the underlying issue here.

Hopefully closes #821.


mtwo:
Merged!

mtwo:
I can load the home page!

austinlparker:
Thanks for fixing this y'all - not sure why that PR passed deployment checks then broke.

PR #832, opened 2021-10-15
Add examples to registry

### LINKED ISSUE #814: Add 'examples' as a category into the registry
As discussed during the SIG adding examples/tutorials to the registry seems like a good experiment: this would allow new-comers to OTel to explore different ways into getting started, also people creating those examples/tutorials can make them a little bit more visibile.

A requirement would be that those examples are vendor-agnostic.

svrnm:
Fix #814, this adds 'example' as a category to the registry and 2 examples that are currently contained in the docs from @mnadeem, @pragmaticivan, @niko-achilles as a starting point. 

I create this as a draft, because I think there are a few points to sort out: 
* What are the requirements for those examples? Should they be only OTel, or is it acceptable that they contain other components as well, especially a (non-commercial) backend? What is ok for a non-commercial backend, can it be optional, mentioned in the docs?
* Does the example need to be standalone in a repository or can it point to a folder within another project?
* Should examples of the individual language specific projects also be included?
* What is a good naming & description schema for those examples, right now I just copied the names from the repositories, but maybe there's a way to standardise on.


niko-achilles:
@svrnm 
if examples with great demonstration impact are needed for the community, i can participate and create examples .

Does the community need new examples or are improvements to be made ?
In case of new examples , is the content of the new examples discussed and is known ?  

I remember we had a discuss around this topic on opentelemetry js contrib. , but it is unknown to me what can be an improvement to what we already have ?!

svrnm:
Thank you all for jumping on the conversation, as said in my initial comment, I put this into draft state because the questions I want to get answered are still open:

What are the requirements for those examples? Should they be only OTel, or is it acceptable that they contain other components as well, especially a (non-commercial) backend? What is ok for a non-commercial backend, can it be optional, mentioned in the docs?
Does the example need to be standalone in a repository or can it point to a folder within another project?
Should examples of the individual language specific projects also be included?
What is a good naming & description schema for those examples, right now I just copied the names from the repositories, but maybe there's a way to standardise on.

@chalin thanks for the comments, I copied those descriptions from the repo, but I can fix that *if* those stay in the list of examples

@jkowall those repositories are not mine, I picked them from the existing docs (https://opentelemetry.io/docs/js/instrumentation_examples/ and https://opentelemetry.io/docs/java/instrumentation_examples/) -- I agree that the examples we link should satisfy a certain level of quality, that's one of my core questions

@niko-achilles sorry, you said it to my twice already that I should not include it, I just was digging up all the examples I had at hand, I will remove it. 

> if examples with great demonstration impact are needed for the community, i can participate and create examples .
> Does the community need new examples or are improvements to be made ?
> In case of new examples , is the content of the new examples discussed and is known ?
> I remember we had a discuss around this topic on opentelemetry js contrib. , but it is unknown to me what can be an improvement to what we already have ?!

The intend of this PR was to enable external repositories to be added to the registry, so people can explore all the things out there. From my understanding it is still too early with all the specs and implementations being evolving to have something "official"

Issue #852, opened 2021-10-21
Publish the specification via opentelemetry.io


chalin:
The spec is a core OTel resource, and as such it should be published through this repo. (#833)

- [x] Add spec pages to website
- [x] Ensure that repository links (like **Edit this page**) point to the canonical source (cf. #542)

bhs:
@chalin Per your note elsewhere, I will raise this with the GC today.

austinlparker:
As a first step, we should add a 'Specification' category to the left nav in `/docs` that links to the main GitHub repo for specs.

reyang:
Couple things need to be clarified:
1. Do we only publish Stable version of specs, or we also publish Experimental specs? What to do if a spec status is Mixed?
2. Do we only publish the latest-stable version, or we publish ALL the stable versions?

chalin:
Good questions.

I'd suggest _starting with_ the latest-stable version (no experimental).

austinlparker:
Perhaps it should be under 'concepts'?

chalin:
> Looking at the layout, I also suggest that the spec should not be listed so prominently in our documentation

Ok. Under the new IA (once it is defined via #770), I would imagine that it would make sense to have a **Reference** section, with the spec under that.

Until we have a new IA I can I can think of a these options:

1. Not show the Specification section in the sidenav.
2. Show the Specification in the sidenav but move it towards the end of the list.
3. Add a Referrence section now (even before we have the new IA) and place the spec there from the start.

Thoughts?

austinlparker:
I like #3

tedsuo:
Option 3 sounds good

chalin:
@tedsuo @austinlparker et al - Done:

- I've implemented option 3, and
- Replaced the "Edit page" button by a "View page source"

PTAL https://deploy-preview-866--opentelemetry.netlify.app/docs/reference/specification/

austinlparker:
Since we've addressed the issues raised here outside of automation (which would take a bit to solve in a generalized fashion), we'll commit to a less than 24 hour turnaround for updates when new spec releases are cut from notification. 

With that in mind, unless there are other blockers we'll be merging #866 next Tuesday.

chalin:
Closed by #866

PR #866, opened 2021-10-27
Add the specification to the docs

chalin:
Here's a **preview** of the spec published via the website: 

- https://deploy-preview-866--opentelemetry.netlify.app/docs/reference/specification/

This requires no extra effort on the part of the spec maintainers other than to let us know when the spec is updated so that we can update it here too.

Contributes to #852. (There will be some followup & cleanup items, but they can be addressed in a separate PR.)

@bhs @tedsuo @open-telemetry/specs-approvers WDYT?

/cc @celestehorgan @nate-double-u


chalin:
As mentioned in https://github.com/open-telemetry/opentelemetry.io/issues/852#issuecomment-961148591:

> * I've implemented option 3 [put the spec under /docs/reference] , and
> * Replaced the "Edit page" button by a "View page source"

PTAL

austinlparker (approved):

austinlparker:
Hold merge until after 11/9/21

chalin:
> With that in mind, unless there are other blockers we'll be merging #866 next Tuesday.

No objections raised. Merging.

PR #869, opened 2021-10-28
Compress PNG Images

mic-max:
Tool used: `PNGGauntlet`

During my compress job I noticed that some of the images that were compressed were included via a Git submodule, so I will make some follow up changes to those in the following repositories:
- [content-modules/opentelemetry-java](https://github.com/open-telemetry/opentelemetry-java)
- [content-modules/opentelemetry-specification](https://github.com/open-telemetry/opentelemetry-specification)
- [themes/docsy](https://github.com/cncf/docsy)


austinlparker:
@chalin Thoughts on this? Is this something that could be added at build time?

chalin:
@austinlparker - yes, this is something I've been wanting to setup, but this can probably go in as is for now.

chalin (approved):

mic-max:
Making this part of the PR checks would be cool.
One way to implement it would be something like:
1. The PR either modifies or creates a PNG
2. Run a compression program on all the modified and created PNGs
3. If all the compressed images have the same file sizes in the PR, succeed. Otherwise fail.

Then you'd likely want to add some guidance that before uploading they run the **same** compression program locally to a README

PR #885, opened 2021-11-04
fix otel local config file path.

woyaodeyangguang:
the otel local config   file path is incorrect.


woyaodeyangguang:
cc @chalin

woyaodeyangguang:
cc @Austin Parker

austinlparker (approved):

PR #923, opened 2021-11-15
Add Span Links to glossary

cartermp:
This has come up a few times (at my workplace), so this seems like a good place to start.


chalin (commented):
Thanks for the PR @cartermp!

A few thoughts:

- I think that it would be better to have an entry for **Span link** rather than the unqualified term "link"
- Use a definition like:
  > **Span Link:** A span link is a link between causally-related spans. For details see [Links between spans][] and [Specifying links][]

  [Links between spans]: /docs/reference/specification/overview/#links-between-spans
  [Specifying links]: /docs/reference/specification/trace/api#specifying-links

_Maybe_ such a definition **should be added to the Specification glossary**?
WDYT @open-telemetry/specs-approvers 

/cc @austinlparker

cartermp:
Yeah, I actually prefer calling it `Span Link` - I've noticed that it's called `Link` in go/.NET/etc. at the API level so I started here, but I think most people won't be confused if it's called `Span Link` here.

austinlparker (approved):

austinlparker (approved):

austinlparker:
It's actually 'Link' rather than 'Span Link' because links can exist between any telemetry points, so you can link a metric to a span or a log to a span or vice-versa.

cartermp:
Should be good to go now

PR #937, opened 2021-11-17
CI scripts: support GET and DEPTH parameters

chalin:
- Support GET/no-get and DEPTH parameters for submodules.
- Also switch to conventional naming for submodule scripts.
- Dev doc updates will follow.


package.json
```diff
@@ -20,22 +21,22 @@
...
     "prebuild:production": "npm run _prebuild",
     "prebuild": "npm run _prebuild",
     "precheck-links": "npm run build",
-    "preinstall": "npm run submodule:get",
+    "preinstall": "npm run get:submodule",
     "preserve:hugo": "npm run _prebuild",
     "preserve": "npm run _prebuild",
     "s": "run-s",
-    "schemas:update": "npm run submodule:update content-modules/opentelemetry-specification",
+    "schemas:update": "npm run update:submodule content-modules/opentelemetry-specification",
```
chalin, line 35: Leaving this name as is since I don't want it to match `update:*`.

Issue #703, opened 2021-09-01
Update JS Exporter docs with collector OTLP export instructions


wong2:
https://opentelemetry.io/docs/js/exporters/ 
At the bottom of this guide, it links to how to setup a OpenTelemetry Collector, but not how to use it with JavaScript sdk.

PR #956, opened 2021-11-20
Provide JS instructions to send traces to a collector

cartermp:
fixes https://github.com/open-telemetry/opentelemetry.io/issues/703

I suppose this doesn't tell you how to send metrics data to a collector, so I'm happy to figure that out and do it too if it's needed, but since metrics are still in development I wonder if it's better to just hold off until they stabilize in the SDK.

### DIFF
--- a/content/en/docs/instrumentation/js/exporters.md
+++ b/content/en/docs/instrumentation/js/exporters.md
@@ -107,5 +107,20 @@ const meter = new MeterProvider({
 
 ## OpenTelemetry Collector
 
-If you are looking for a vendor-agnostic way to receive, process and export your
-telemetry data follow the instructions to setup a [collector](/docs/collector/)
+To send trace data to a Collector you'll want to use the `exporter-collector` package:
+
+```shell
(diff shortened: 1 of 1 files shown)


content/en/docs/js/exporters.md
```diff
@@ -107,5 +107,20 @@ const meter = new MeterProvider({
...
+To send trace data to a Collector you'll want to use the `exporter-collector` package:
+
+```shell
+npm install --save @opentelemetry/exporter-collector
+```
+
+And configure the exporter to point at your Collector endpoint:
+
+```javascript
+import { CollectorTraceExporter } from '@opentelemetry/exporter-collector';
```
cartermp, line 119: hmm, I guess this should instead be one of

`@opentelemetry/exporter-trace-otlp-[http/grpc/proto]`

austinlparker (approved):

austinlparker:
Yeah, I'd wait until metrics stabilize and then we can get metrics export instructions going.

PR #978, opened 2021-12-09
Update to new java docs repo

trask:
I'm not sure if you prefer renaming the directory to match also `content-modules/opentelemetry-java` --> `content-modules/opentelemetry-java-docs?`

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -8,8 +8,8 @@
 [submodule "content-modules/opentelemetry-specification"]
 	path = content-modules/opentelemetry-specification
 	url = https://github.com/open-telemetry/opentelemetry-specification.git
-[submodule "content-modules/opentelemetry-java"]
-	path = content-modules/opentelemetry-java
+[submodule "content-modules/opentelemetry-java-docs"]
+	path = content-modules/opentelemetry-java-docs
 	url = https://github.com/open-telemetry/opentelemetry-java-docs
 [submodule "content-modules/opentelemetry-ruby"]
 	path = content-modules/opentelemetry-ruby
--- a/.gitmodules
+++ b/.gitmodules
@@ -10,7 +10,7 @@
 	url = https://github.com/open-telemetry/opentelemetry-specification.git
 [submodule "content-modules/opentelemetry-java"]
 	path = content-modules/opentelemetry-java
-	url = https://github.com/open-telemetry/opentelemetry-java
+	url = https://github.com/open-telemetry/opentelemetry-java-docs
 [submodule "content-modules/opentelemetry-ruby"]
 	path = content-modules/opentelemetry-ruby
 	url = https://github.com/open-telemetry/opentelemetry-ruby
--- a/content/en/docs/instrumentation/java
+++ b/content/en/docs/instrumentation/java
@@ -1,1 +1,1 @@
-../../../../content-modules/opentelemetry-java/website_docs
+../../../../content-modules/opentelemetry-java-docs/website_docs


trask:
the build doesn't like it, probably due to changing the submodule? the build log is not very descriptive, any ideas? thx

```
1:25:17 PM: Error fetching branch: https://github.com/open-telemetry/opentelemetry.dev pull/978/head
```

trask:
@open-telemetry/docs-approvers wondering if this is the issue: https://answers.netlify.com/t/error-fetching-branch-after-trying-to-update-submodule/5370. can you try clearing the netlify cache?

chalin (commented):
I'd like to instead propose that we move the content of the Java `website_docs` into **this** repo, rather than moving them out to yet another Java OTel repo. (That way your new repo can be for OTel Java examples.)

Thoughts @trask @austinlparker @tedsuo?

chalin:
[To follow through with my comments from earlier] Months ago, language SIGs were offered the following two choices:

(1) Their docs would reside here in the docs/website repo (this is simplest and preferred choice), or 
(2) As a _convenience_ to language SIG contributors, the docs would be in the main code repo with a submodule used to bring in the docs into the website.

@austinlparker et al., I don't think that we want to offer yet a third option, esp. since the convenience of having the docs + code in the same repo is lost once the docs are being moved elsewhere, so (again), I'd strongly vote for the docs to be moved here.

trask:
@austinlparker proposed similar: https://github.com/open-telemetry/opentelemetry-java/pull/3974#issuecomment-989411778

I believe the main concern to this is summarized by @jkwatson's comment:

> We also want to include non-website docs, with runnable code samples and gradle build scripts for them. I don't think that belongs in the website repo, does it?

chalin:
Thanks for offering some context @trask. I'll see if @austinlparker and I can meet tomorrow to discuss this.

austinlparker:
Sorry, this thread slipped my attention.

I agree that having code samples in this repository doesn't make a ton of sense, but my understanding that the rationale from maintainers to keep docs + samples in the main repo was to ensure they were updated alongside the SDK releases. If that isn't a concern any more, and they're going to be divorced from it anyway, then I'd strongly urge for _all_ documentation to be migrated directly into this repository, with a separate code sample repository being maintained. 

Our eventual goal would be to migrate all docs to the website at the end of the day (aside from things that don't make sense, like auto-generated API docs or things that require language-specific tooling; javadoc, et. al.) so that we can enable versioning of the documentation and promote the website as a "one stop shop" for OTel docs.

austinlparker:
I'd really like to understand what Java (and other) SIGs need to feel comfortable with moving their docs here. Is it approve/merge permissions? We can do that with codeowners and whatnot.

trask:
you will need to discuss this with @jkwatson and @anuraaga

I'm going to close this PR for now until it's resolved

chalin:
> you will need to discuss this with @jkwatson and @anuraaga

FYI, I've created #989 so that we can hold our discussion there.

cartermp:
In the interim @chalin @austinlparker can we go through with this PR? It solves a _now_ problem - getting existing enhancements to docs (e.g., how to configure automatic instrumentation) to be live on the site - which is important too.

How it gets organized is clearly important, but now that it's deferred to an issue for discussion I'd like to have the _now_ problem resolved 🙂

chalin:
Thanks for the ping. Let me chat with @austinlparker about this later today, once I'm back at the office.

austinlparker:
I suppose, yeah, if we're getting blocked on updates.

cartermp:
Not sure why, but same build error - something funky needed to "refresh" the submodule or something?

chalin:
> I suppose, yeah, if we're getting blocked on updates.

Ok, will proceed for now.

> Not sure why, but same build error - something funky needed to "refresh" the submodule or something?

It's actually a bit tricky to update -- which is another reason to avoid submodules when we can. I have another PR ready that changes the Java submodule, but I can't submit and merge until the broken links are fixed: https://github.com/open-telemetry/opentelemetry-java-docs/pull/7

chalin:
Closing, as this PR won't work. I'll submit another soon.

chalin:
Superseded by #995

Issue #1001, opened 2021-12-21
Rework .NET Docs - proposed mini-IA


cartermp:
.NET is a little special. So I propose a mini-IA for the .NET docs:

Under instrumentation/.NET:

* Getting started
* Terminology and APIs
* Automatic
* Manual
* Exporters
* .NET Framework

Most .NET developers doing modern services work use modern .NET (i.e., not .NET Framework, which is a legacy Windows-based technology).

So, I propose that _everything except for a page entitled ".NET Framework"_ (or something like it) uses modern .NET APIs - ASP.NET Core, Microsoft.Extensions.Hosting, etc. And in that one doc, the things you need to do to configure things for .NET Framework-based apps is covered instead, rather than putting it as an option for every single concept in the docs.

Additionally, since .NET recommends using the `System.Diagnostics.DiagnosticSource` API over the OpenTelemetry API, there should be a dedicated doc that explains the difference and maps concepts from one to the other in a table.

Thoughts?

chalin:
A few questions:

- Could **Terminology and APIs** (either or both the terminology or the API references) be a part of the index page?
- The main pages/topics for modern .NET are **Getting started**, **Automatic**, **Manual** and **Exporters**. Which of these topics would be covered under **.NET Framework**? Would it make sense to instead have a **.NET Framework** section at the bottom of each page?

cartermp:
> Could Terminology and APIs (either or both the terminology or the API references) be a part of the index page?

I think so, yes. At least certainly a terminology overview. The API reference itself is actually just here: https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics?view=net-6.0

...there's nothing in there about OTel (in fact, the core `Activity` class says it's for logging...), but, err, it's...there? Heh

> Which of these topics would be covered under .NET Framework? 

For .NET Framework, the main differences are in setup (where to "hook" in otel), and automatic instrumentation for ASP.NET (rather than ASP.NET Core, which doesn't run on .NET Framework). Everything else is the same OTel APIs, and exporters are even the same too (as far as I can tell).

So I think it could just be a separate doc.

cartermp:
I think I will close this one out. After sleeping on it a few times, I don't think there needs to be a dedicated terminology&APIs section.

Issue #974, opened 2021-12-02
Document creating a nested/child span


cartermp:
Tracking issue for the "Nested Spans" row here: https://docs.google.com/spreadsheets/d/1EO0HkjraZSN_7LlFpwFPoazR9xbdCCgQi-IFc0YA5rU/edit#gid=0

Emoji guide:

✔️ - this is basically good
➖ - this could be improved
❌ - this needs to exist or be rewritten

Currently documented with an example:

- [x] ✔️ Java
- [x] ✔️ C++
- [x] ➖ Go (probably needs explicit section for the sample)
- [x] ➖ JavaScript (probably needs explicit section for the code sample)
- [ ] ➖ Erlang/Elixir (exists but no explicit section)
- [x] ➖ Python (it's in the quickstart but that's it, no section, no explaining the relationship)

Has no section at all:

- [ ] ❌ Ruby
- [x] ❌ .NET
- [ ] ❌ Rust
- [ ] ❌ Swift
- [ ] ❌ PHP

svrnm:
@cartermp, Can this be superseded by #3229?

cartermp:
Yep, let's close in favor. Whew, this one was pretty old

Issue #973, opened 2021-12-02
Document getting the active/current span for each language SDK


cartermp:
Tracking issue for the "Getting Current Span" row here: https://docs.google.com/spreadsheets/d/1EO0HkjraZSN_7LlFpwFPoazR9xbdCCgQi-IFc0YA5rU/edit#gid=0

Emoji guide:

✔️ - this is basically good
➖ - this could be improved
❌ - this needs to exist or be rewritten

Currently documented with an example:

- [x] ✔️ Python ([link](https://opentelemetry-python.readthedocs.io/en/stable/faq-and-cookbook.html#getting-and-modifying-a-span))
- [x] ✔️ JavaScript
- [x] ✔️ Java
- [x] ✔️ Ruby
- [x] ✔️ Go

Has a section saying it exists, but needs an example of using at least one:

- [ ] ➖ C++ (doesn't have get, but has set, so it's pretty easy to figure it out from there)

Has no section at all:

- [x] ❌ .NET
- [ ] ❌ Rust
- [ ] ❌ Erlang/Elixir
- [ ] ❌ Swift
- [ ] ❌ PHP

chalin:
@cartermp - I added checkboxes to the opening comment for subtasks. Please mark items as done when they are completed, thanks.

svrnm:
Can this be superseded by #3229?

cartermp:
Yep

PR #1014, opened 2021-12-22
Add manual instrumentation doc for .NET

cartermp:
Contributes to https://github.com/open-telemetry/opentelemetry.io/issues/1001 (but really contributes to the implicit ".NET has no docs" issue).

Also contributes to:

https://github.com/open-telemetry/opentelemetry.io/issues/974
https://github.com/open-telemetry/opentelemetry.io/issues/973

### DIFF
--- a/content/en/docs/instrumentation/net/manual.md
+++ b/content/en/docs/instrumentation/net/manual.md
@@ -207,8 +207,8 @@ which is not likely to be desired.
 
 ## Add tags to an Activity
 
-Tags let you attach key/value pairs to an `Activity` so it carries more information about the current operation
-that it's tracking.
+Tags (the equivalent of Attributes in OpenTelemetry) let you attach key/value pairs to an `Activity`
+so it carries more information about the current operation that it's tracking.
 
 ```csharp
 using var myActivity = MyActivitySource.StartActivity("SayHello")
--- a/content/en/docs/instrumentation/net/manual.md
+++ b/content/en/docs/instrumentation/net/manual.md
@@ -80,7 +80,8 @@ using OpenTelemetry;
 using OpenTelemetry.Resources;
 using OpenTelemetry.Trace;
 
-// Define some important constants and the activity source
+// Define some important constants and the activity source.
+// These can come from a config file, constants file, etc.
 var serviceName = "MyCompany.MyProduct.MyService";
 var serviceVersion = "1.0.0";
 
@@ -119,7 +120,8 @@ public static class Telemetry
 {
     //...
 
-    // 'TelemetryConstants.ServiceName' comes from somewhere in your code
+    // Name it after the service name for your app.
+    // It can come from a config file, constants file, etc.
     public static readonly ActivitySource MyActivitySource = new(TelemetryConstants.ServiceName)
 
     //...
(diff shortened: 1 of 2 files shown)


chalin (commented):
LGTM after a quick glance. I'll let @austinlparker et all do a detailed read. If this is still open when I get back, I can make a careful pass too.

(Thanks for naming the file `manual.md` right from the start!)

content/en/docs/instrumentation/net/manual.md
```diff
@@ -0,0 +1,253 @@
...
+the same as the Service Name.
+
+```csharp
+using System.Diagnostics;
+
+public static class Telemetry
+{
+    //...
+
+    // 'TelemetryConstants.ServiceName' comes from somewhere in your code
```
austinlparker, line 122: This applies to this and the earlier reference, but is this idiomatic .NET? Should the service name be set as a runtime constant or be loaded from config?

content/en/docs/instrumentation/net/manual.md
```diff
@@ -0,0 +1,253 @@
...
+
+Sometimes it's helpful to access whatever the current `Activity` is at a point in time so you can enrich
+it with more information.
+
+```csharp
+var activity = Activity.Current;
+// may be null if there is none
+```
+
+Note that `using` is not used in the following example. Doing so will end current `Activity`,
```
austinlparker, line 179: Prior, not following, example? It might be more clear to use the brace syntax for `using` rather than the brace-less syntax in these examples.

content/en/docs/instrumentation/net/manual.md
```diff
@@ -0,0 +1,253 @@
...
+which is not likely to be desired.
+
+## Add tags to an Activity
```
austinlparker, line 182: I'd go with 'attributes' since that's the OTel term, then clarify that .NET Activity Tags === OTel Attributes.

austinlparker (approved):
lgtm outside of comments.

content/en/docs/instrumentation/net/manual.md
cartermp: I went with tags since every other subsection uses the System.Diagnostics terminology. The other idea I had was to use OTel terminology, but in each section describe the System.Diagnostics-equivalent.

I want to also create a "manual instrumentation but with the shim" doc too (named better) that would do this

content/en/docs/instrumentation/net/manual.md
cartermp: I'll leave a common saying that it can also come from config. There's a ton of options available and it really just depends on what people prefer. Having a TelemetryConstants fie is likely to be an approach some folks take. But there's also others (e.g., appsettings.json)

content/en/docs/instrumentation/net/manual.md
cartermp: Added a new section preceding this (good catch!) that talked about scoping with `using` and child operations.

content/en/docs/instrumentation/net/manual.md
cartermp: Happy to change btw, just wanted to explain the rationale

content/en/docs/instrumentation/net/manual.md
austinlparker: Sounds fine.

content/en/docs/instrumentation/net/manual.md
austinlparker: I think it'd be good to at least have a callout for 'hey, this is the same thing as an attribute' so we can be consistent across languages.

cartermp: Doneski

PR #1067, opened 2022-01-21
Fix NodeJS metrics example

MrMYHuang:
The new api.Counter doesn't support bind method anymore. Thus, using the new metrics package sdk-metrics-base, the NodeJS metrics example is also needed to update.


dyladan:
The new metrics SDK and API with this feature removed are under development and have not been released. AFAIK the released version still contains this feature.

chalin (commented):
Based on @dyladan's feedback, I'd suggest that we close this PR for now. You can consider resubmitting once the features are released. WDYT? /cc @austinlparker

austinlparker:
Closing until feature released.

PR #1079, opened 2022-01-26
Update javascript web getting started example.

jonasem:
Parcel will not bundle with esm-style imports unless it's script tag has the module type.


dyladan (approved):

chalin (approved):

chalin:
Thanks for your review and approval @dyladan.

PR #1081, opened 2022-01-27
Document setting span status and recording exceptions for .NET

cartermp:
This brings the .NET docs up to par with JS and Java.

Note that `RecordException` does not have an equivalent method on `Activity` in .NET, at it at least doesn't appear to be available based on API usage and docs I read. You could technically do it yourself in a helper extension method I guess?

### DIFF
--- a/content/en/docs/instrumentation/net/shim.md
+++ b/content/en/docs/instrumentation/net/shim.md
@@ -323,6 +323,49 @@ using var span = tracer.StartActiveSpan("another-span", links: links);
 // do some work
 ```
 
+## Set span status
+
+A status can be set on a span, typically used to specify that a span has not
+completed successfully - `Status.Error`. In rare scenarios, you could override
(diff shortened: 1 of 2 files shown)


content/en/docs/instrumentation/net/shim.md
```diff
@@ -323,6 +323,49 @@ using var span = tracer.StartActiveSpan("another-span", links: links);
...
+    span.SetStatus(Status.Error, "Something bad happened!");
+}
+```
+
+## Record exceptions in spans
+
+It can be a good idea to record exceptions when they happen. It's recommended to
+do this in conjunction with setting [span status](#set-span-status).
+
+```java
```
chalin, line 353: java -> csharp

chalin (changes requested):
See one inline comment.

content/en/docs/instrumentation/net/shim.md
cartermp: whoopsie!

Issue #833, opened 2021-10-15
Ensure non-API docs are hosted on opentelemetry.io


chalin:
As a part of **improving user experience** and **SEO**, all OTel docs should be hosted on [opentelemetry.io] possibly aside from low-level technical notes and autogenerated docs (like API references) that require language-specific tools.

The initial notes below come from the items marked as "(0)" in #730:

- [x] #852
- [ ] **Collector**: consider migrating relevant parts of the Collector [docs][collector-docs] folder
- [ ] **Community** pages:
  - [x] #2167
  - [x] #1289
- [ ] **C++**: has a [docs][cpp-docs] folder containing `.md` files and some RST documentation (under `public`) published at [opentelemetry-cpp.readthedocs.io].
- [ ] **Go**: #1935
- [x] **Java**:
  - [x] #989
  - [ ] **Java instrumentation**:
    - [x] https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/4523
      - [x] #1229
    - [x] https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/4524
      - [x] #1178
      - [x] https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/5516
    - [x] https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1580
- [ ] **.Net**/**C#** - has a [docs][dotnet-docs] folder containing `.md` files.
  - [x] https://github.com/open-telemetry/opentelemetry-dotnet/issues/2876
- [ ] **PHP**:
  - [x] #1154
  - [x] https://github.com/open-telemetry/opentelemetry-php/issues/423#issuecomment-1042389408
  - [x] #1157
  - [x] https://github.com/open-telemetry/opentelemetry-php/issues/405
- [ ] **Python**: the [Python docs] (from [docs directory][python-docs-dir] published via Read the Docs) seem extensive. **Migration is in progress**:
  - [x] https://github.com/open-telemetry/opentelemetry.io/issues/1109
    - [x] https://github.com/open-telemetry/opentelemetry-python/issues/2172 
  - [x] https://github.com/open-telemetry/opentelemetry-python/issues/2319
- [x] **Ruby**: #1942
- [ ] Rust: see [Rust docs]

[opentelemetry.io]: https://opentelemetry.io/
[collector-docs]: https://github.com/open-telemetry/opentelemetry-collector/tree/main/docs
[...]

PR #1085, opened 2022-01-28
Rough example layout of potentially-imported python docs

cartermp:
Contributes to:
- #833
- https://github.com/open-telemetry/opentelemetry-python/issues/2172

@chalin @austinlparker as promised:

**Preview**: https://deploy-preview-1085--opentelemetry.netlify.app/docs/instrumentation/python2/

This is rough since I'm not gonna spike hard on actually porting docs over, but it stays fairly consistent with the rest of the site without dropping things that were proposed here: https://github.com/open-telemetry/opentelemetry-python/issues/2319#issue-1076319100

Relative ordering of things, subsections and sub-subsections, etc. are all up for discussion of course 🙂 

As mentioned in our call, there's quite a lot of existing content, so getting this up to par with JS/Java/.NET wouldn't be too heavy of a lift.


chalin (commented):
I like this proposal!

austinlparker:
yeah this looks good. lets shop it

Issue #845, opened 2021-10-19
Host blog posts on website


chalin:
Steps to transitioning to the website-hosted blog:

- [x] Create Blog section - #1057
- [x] [@open-telemetry/blog-approvers]:
  - [x] Call for [@open-telemetry/blog-approvers] - https://lists.cncf.io/g/cncf-opentelemetry-comms/message/283
  - [x] https://github.com/open-telemetry/community/issues/953
  - [x] Add to README Roles section - #1064
  - [x] Add to CODEOWNERS - #1064
- [x] Create a blog [Hugo archetype file] - #1062
- [x] Add instructions for _Adding a blog post_, and create supporting archetypes:
  - #1062
  - #1139
- [x] Migrate references / abstracts linking to old posts:
  - [x] #1091
  - [x] #1136 & #1138
  - [x] #1137
  - [x] #1140
  - [x]  #1141
  - [x] #1142
  - Consider removing the `medium` archetype --> #1267
- Announcement that new blog posts will be hosted through the website:
  - Listserv - https://lists.cncf.io/g/cncf-opentelemetry-comms/message/305
  - Medium post to announce the move?

[@open-telemetry/blog-approvers]: https://github.com/orgs/open-telemetry/teams/blog-approvers
[Hugo archetype file]: https://gohugo.io/content-management/archetypes

---

Original opening comment:

Someone expressed the wish that blog posts be hosted directly on this site, but I can't recall who. Maybe @lizthegrey @tedsuo @austinlparker @mangotron? They are currently hosted on Medium at https://medium.com/opentelemetry.

Anyhow, we can use this issue to discuss the topic -- either before or after the next proposed post: https://lists.cncf.io/g/cncf-opentelemetry-comms/message/270 (cc @lalitb).

austinlparker:
I did! We should do this in Q1.

austinlparker:
AFAICT here's a rough plan/to-dos:

- [ ] Migrate existing Medium blog posts to /content/en/blog
- [ ] Figure out redirect strategy
- [ ] Tweak blog template if needed
- [ ] Create blog contribution guide
- [ ] Announce timeline to listserv

Probably some stuff I'm missing, but this should be a start.

chalin:
@austinlparker do you want to make the announcement re. the blog's move (see the opening comment for the subtask). /cc @tedsuo

chalin:
[...]

PR #1091, opened 2022-02-02
Blog entry for v1.0-trio announcement

chalin:
- @tedsuo @austinlparker : here's an example of the kind of post I have in mind for pages that have already been published on Medium. Given the nature of such a post, I don't even think that we need to bother with setting the canonical link. If you feel otherwise, I can look into it.
- Contributes to #845
- Closes #1092

Preview: https://deploy-preview-1091--opentelemetry.netlify.app/blog/2022/v1.0-trio/

### DIFF
--- a/content/en/blog/2022/v1.0-trio.md
+++ b/content/en/blog/2022/v1.0-trio.md
@@ -1,0 +1,12 @@
+---
+title: OpenTelemetry Erlang/Elixir, Javascript, and Ruby v1.0
+linkTitle: Erlang/Elixir, JS, and Ruby v1.0
+date: 2022-02-01
+---
+
+We are kicking off the new year with a bang! In the last couple months, three
+new languages (Ruby, Javascript, and Erlang/Elixir) have had their first 1.0
+releases, joining the existing GA releases from C++, Go, Java, .Net, Python and
+Swift. Read all the details from the [announcement][].
+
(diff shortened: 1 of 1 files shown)


content/en/blog/2022/v1.0-trio.md
```diff
@@ -0,0 +1,12 @@
+---
+title: OpenTelemetry Erlang/Elixir, Javascript, and Ruby v1.0
+linkTitle: Erlang/Elixir, JS, and Ruby v1.0
+date: 2022-02-01
+---
```
chalin, line 5: @tedsuo @austinlparker : note that I chose to exclude an `author` field because this isn't really a post but rather a reference to a post. WDYT?

If you prefer, I can add the authors like this:

```yaml
spelling: cSpell:ignore Dyla Laurin Sloughter
author: Tristan Sloughter, Daniel Dyla, and Robert Laurin
```

austinlparker: Yeah, this makes sense.

austinlparker (approved):

tedsuo:
Yeah I mean, I guess I'd prefer it just pop up the medium post in a new tab, rather than make the user click through twice. But if that's annoying to implement and we're doing it rough and fast, this is fine.

tedsuo (approved):
Based on our experiment, with direct linking, I'm now fine with this approach. Thanks @chalin!

PR #1092, opened 2022-02-02
Blog: redirect old posts to its Medium page directly

chalin:
- This is an alternative to #1091 
- Contributes to #845

Preview:

- Blog: https://deploy-preview-1092--opentelemetry.netlify.app/blog
  Notice that you see the v1.0-trio entry, but if you click on any link to that page, it'll send you directly to the Medium page.
- You can test the link here too: https://deploy-preview-1092--opentelemetry.netlify.app/blog/2022/v1.0-trio-redirect

### DIFF
--- a/.htmltest.yml
+++ b/.htmltest.yml
@@ -8,3 +8,5 @@ IgnoreInternalURLs:
   - HAHAHUGOSHORTCODE-s2-HBHB
   - HAHAHUGOSHORTCODE-s5-HBHB
   - HAHAHUGOSHORTCODE-s7-HBHB
+  # Blog entry redirects
+  - /blog/2022/v1.0-trio-redirect/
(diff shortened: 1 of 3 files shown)


chalin:
@tedsuo @austinlparker : PTAL. I think that this works as @tedsuo wanted. 

There is one "glitch", which we may or may not want to live with. Follow these steps:

- Visit https://deploy-preview-1092--opentelemetry.netlify.app/blog/2022/welcome/
- Click on the **Next** button at the bottom of the page.
- You'll be directed to the Medium post. This isn't ideal, but might be acceptable.

IMHO, #1091 is the simplest solution that requires no magic. It does require that folks click twice to get to the actual article. @tedsuo @austinlparker - it's your call. Let me know which PR you prefer.

austinlparker:
The 'click next' behavior feels kinda awkward to me. I think I prefer #1901 because yeah, its an extra click, but it doesn't break you out of your 'place' in the page.

chalin:
I agree with @austinlparker. I'd rather keep the UX consistent even at the cost of an extra click. Let's keep in mind that we'll only need this for about 10 posts whose importance will presumably diminish over the lifetime of the project. Thoughts @tedsuo?

tedsuo:
Yeah that's kind of a weird side effect. Ok fine, let's go with #1091

chalin:
Thanks for the feedback @tedsuo!

PR #1146, opened 2022-02-17
add c++ logo to instrumentation page

svrnm:
Minor thing, the c++ logo is missing at https://opentelemetry.io/docs/instrumentation/cpp/


chalin (approved):
As I mentioned to @austinlparker already, those icons are going to have to move out of the description because the are causing problems, ... but in the meantime, I'm ok with us being consistent.

PR #1147, opened 2022-02-17
add link to 'instrumenting' concept and to k8s operator injection to …

svrnm:
Two changes for https://opentelemetry.io/docs/instrumentation/:
* I linked back from that page to the concept of [instrumenting](https://opentelemetry.io/docs/concepts/instrumenting/). Idea is that someone landing on that page might not know about the concept and can go back from there
* I added a short paragraph that highlights that auto instrumentation libraries for java, node.js & python can be injected via the kubernetes operator. For someone running apps on k8s, who "just" wants to get services instrumented, this is a live saver. Let me know what you think, maybe this could also have it's own section within the docs and not pointing out?

Preview: https://deploy-preview-1147--opentelemetry.netlify.app/docs/instrumentation/

### DIFF
--- a/content/en/docs/instrumentation/_index.md
+++ b/content/en/docs/instrumentation/_index.md
@@ -3,7 +3,7 @@ title: Instrumentation
 weight: 2
 ---
 
-OpenTelemetry code instrumentation is supported for the languages listed below.
+OpenTelemetry code [instrumentation](/docs/concepts/instrumenting/) is supported for the languages listed below.
 Depending on the language, topics covered will include some or all of the
 following:
 
@@ -11,3 +11,4 @@ following:
 - Manual instrumentation
 - Exporting data
 
+If you are using Kubernetes, you can use the [OpenTelemetry Operator for Kubernetes](https://github.com/open-telemetry/opentelemetry-operator) to [inject auto-instrumentation libraries](https://github.com/open-telemetry/opentelemetry-operator#opentelemetry-auto-instrumentation-injection) for Java, NodeJS and Python into your application.


content/en/docs/instrumentation/_index.md
```diff
@@ -3,11 +3,12 @@ title: Instrumentation
 weight: 2
 ---
 
-OpenTelemetry code instrumentation is supported for the languages listed below.
+OpenTelemetry code [instrumentation](/docs/concepts/instrumenting/) is supported for the languages listed below.
```
chalin, line 6: 👍

chalin (commented):
Good idea to link to the concepts section.

As for the added paragraph, I'm thinking that a version might be better added under each SIGs landing page.

Thoughts @svrnm @cartermp @austinlparker?

cartermp:
I think adding it to each SIG page would be decent. It makes me wonder if there should be a SIG page-equivalent for kubernetes alone?

chalin:
I'd suggest that we merge this as is. I can then followup with a PR that distributes this comment to the Java, NodeJS and Python pages. How does that sound?

svrnm:
> I'd suggest that we merge this as is. I can then followup with a PR that distributes this comment to the Java, NodeJS and Python pages. How does that sound?

sounds good to me

> It makes me wonder if there should be a SIG page-equivalent for kubernetes alone?

I think that would be a great addition, and might at some point replace the pointers out to the operator.

chalin (approved):
Thanks @svrnm!

PR #1158, opened 2022-02-18
Registry: add new and update JS entries

jscherer92:
Add in new js instrumentation and update old packages.

- https://github.com/open-telemetry/opentelemetry-js-contrib/issues/803
- Related: #1010


content/en/registry/instrumentation-js-fastify.md
```diff
@@ -6,10 +6,9 @@ language: js
 tags:
   - Node.js
   - fastify
-  - fastify-plugin
-repo: https://github.com/autotelic/fastify-opentelemetry
```
dyladan, line 10: Why are we removing this instrumentation? Can't we list both?

content/en/registry/instrumentation-js-nestjs.md
```diff
@@ -1,14 +1,14 @@
...
 title: NestJS OpenTelemetry
 registryType: instrumentation
-isThirdParty: true
+isThirdParty: false
 language: js
 tags:
   - Node.js
   - instrumentation
   - nestjs
-repo: https://github.com/pragmaticivan/nestjs-otel
```
dyladan, line 10: Why is this removed?

content/en/registry/instrumentation-js-fastify.md
jscherer92: For both of we want to list both how do we want to go about official OpenTelemetry support vs. third party?

From what I gathered from the ticket created we want to change these over to our official repos.

chalin:
Thanks for the comments @dyladan.
What should be done with this PR now?

dyladan:
@chalin I don't think we should remove the external competing solutions. The whole point of the registry is for them to be able to be listed. Both of those repositories have recent activity. The otel contrib solutions should be added alongside them. IMO this PR should not be merged as it is.

chalin:
Thanks for the feedback @dyladan.

@jscherer92 - can you address the concerns that have been raised?

jscherer92:
No problem! How do we want to add back in third party libraries is my question. I don't see an example of multiple instrumentation repos being availble

dyladan:
You can just have two files in the directory. One for the contrib module and leave the existing one. I would just add `contrib` to the filename of the new file.

jscherer92:
Okay so have the third party be contrib and have the official left as is. Like it! I'll get this in the next day.

dyladan:
I don't think there is any distinction between "official" and "third party". The file names are just convention because most packages only have a single instrumentation and not competing ones. @chalin lmk if that's wrong

chalin:
Waiting on feedback from @austinlparker and maybe @tedsuo, here. (The setup of the registry was before my time, so I'll let Austin answer your question.)

jscherer92:
This has been fixed up in the latest changes. Review again please! Thanks!

content/en/registry/instrumentation-js-contrib-fastify.md
```diff
@@ -0,0 +1,15 @@
+---
+title: Fastify OpenTelemetry
+registryType: instrumentation
+isThirdParty: true
+language: js
+tags:
+  - Node.js
+  - fastify
+	- fastify-plugin
```
chalin, line 9: Could you replace the tab at the start of line 9 with two spaces?

> image

content/en/registry/instrumentation-js-contrib-nestjs.md
```diff
@@ -0,0 +1,13 @@
+---
+title: NestJS OpenTelemetry
+registryType: instrumentation
+isThirdParty: true
+language: js
+tags:
+  - Node.js
+	- instrumentation
+	- nestjs
```
chalin, line 9: Replace the tabs by spaces in these two lines too.

chalin (commented):
Nit: see inline comments pointing out tab characters (that should be replaced by spaces).

chalin:
Thanks for the updates @jscherer92. Btw, the build is failing because of the tab characters in two of the files, e.g.:

> Error: Error building site: "/opt/build/repo/content/en/registry/instrumentation-js-contrib-fastify.md:9:1": failed to unmarshal YAML: yaml: line 8: found a tab character that violates indentation

Once you fix those the build should pass.

content/en/registry/instrumentation-js-contrib-nestjs.md
jscherer92: should be fix

content/en/registry/instrumentation-js-contrib-fastify.md
jscherer92: should be fixed

dyladan:
The file js-contrib-fastify contains a fastify instrumentation that isn't in the contrib module which seems confusing. I would just rename the existing one to something like autotelic-fastify

jscherer92:
updated namings

chalin:
@dyladan - how's this looking now?

dyladan (approved):

dyladan:
LGTM

chalin (approved):
Thanks all!

PR #1162, opened 2022-02-21
Ensure configuration examples are illustrative but still secure

JonZeolla:


chalin (commented):
Can one of the @open-telemetry/collector-approvers comment on this?

content/en/docs/collector/configuration.md
```diff
@@ -303,7 +303,7 @@ exporters:
   jaeger:
     endpoint: "jaeger-all-in-one:14250"
     tls:
-      insecure: true
+      insecure: false
```
Aneurysm9, line 306: Isn't `false` the default value?  If we're going to set this it should probably be removed instead.  That said, does this configuration actually work without the user taking steps to ensure a valid TLS cert on the other end?  Are those steps documented here?

JonZeolla: I was just keeping it to illustrate that it's an option that can be configured, while avoiding an insecure example

jpkrohling: The example as it is will fail, as it's not specifying which CA to use when doing validation of the server's certificate.

chalin: So is the vote to close this PR @jpkrohling @Aneurysm9? Or request further changes?

Aneurysm9: We should do one of those two things.  I'm not sure that I care overmuch which it is.

JonZeolla: I'm happy to add to the docs and specify an example CA bundle

JonZeolla: It looks like specifying a CA bundle is optional, but a cert and key is required for mTLS according to [this](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/jaegerexporter#getting-started) and [this](https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/configtls/README.md#tls--mtls-configuration).

jpkrohling: Specifying a CA bundle is optional, as the target server might have a cert signed by a known root CA. This is not the case for self-signed certs. I know this would make the guide longer, but I do think we should instruct people how to generate the certs for dev/testing purposes (hinting that they might want to find an appropriate way to generate/provision the production certs), and use those certs in the examples.

Otherwise, people will just copy/paste the insecure config into their production's config.

jpkrohling: If you need, use the steps I outlined in this blog post to generate the certs: https://medium.com/opentelemetry/securing-your-opentelemetry-collector-1a4f9fa5bd6f

JonZeolla:
@austinlparker yes I will, hopefully this weekend or early next week

content/en/docs/collector/configuration.md
JonZeolla: @jpkrohling what do you think now?  I should have aligned it to your blog post, with some minor tweaks.

content/en/docs/collector/configuration.md
```diff
@@ -543,3 +547,43 @@ service:
       exporters:
         - otlp/auth
 ```
+
+
+### Setting up certificates
+
+Certain exporters require that x.509 certificates are used for mTLS security.
```
jpkrohling, line 554: How about something like this?

```suggestion
For a production setup, we strongly recommend using TLS certificates, either for secure communication or mTLS for mutual authentication.
```

content/en/docs/collector/configuration.md
```diff
@@ -543,3 +547,43 @@ service:
...
+See the below steps to generate the certificates used in this example.
```
jpkrohling, line 555:
```suggestion
See the below steps to generate self-signed certificates used in this example. You might want to use your current cert provisioning procedures to procure a certificate for production usage.
```

jpkrohling (commented):
I have a small suggestion for the cert provisioning part, but LGTM.

content/en/docs/collector/configuration.md
chalin: I'd vote like @Aneurysm9 and remove `insecure: false` if it is the default.

content/en/docs/collector/configuration.md
chalin: I like your suggested change @jpkrohling.

content/en/docs/collector/configuration.md
chalin: I like your suggested change @jpkrohling.

JonZeolla:
@jpkrohling I agree with your suggestions and merged them

chalin:
@JonZeolla: can you remove the two occurrences of `insecure: false` -- https://github.com/open-telemetry/opentelemetry.io/pull/1162#discussion_r828524264.

JonZeolla:
Sure

austinlparker:
@JonZeolla is this ready to go?

JonZeolla:
@austinlparker yup should be good!  Just pending a final approval.

I did look for the automerge option once I was comfortable but it doesn't seem to be enabled

jpkrohling (approved):

chalin (approved):
LGTM, thanks!

PR #1164, opened 2022-02-22
Use docsy with deps via NPM not git submodules

chalin:


chalin:
The site builds cleanly even after resetting the cache. For details, see the [Netlify log](https://app.netlify.com/sites/opentelemetry/deploys/6214f7f74484ecf88101d7b3).

PR #1177, opened 2022-02-25
add redirect from docs.opentelemetry.io -> /docs route

### LINKED ISSUE #976: Also support `docs.opentelemetry.io`  domain
I'm unopinionated about what the main URL should be, but I think both URLs:

* opentelemetry.io/docs
* docs.opentelemetry.io

should point to the same thing. I'm not sure if this improves SEO, but I wouldn't be shocked if several folks try to go to `docs.opentelemetry.io` and get surprised at the 404.

austinlparker:
Fixes #976 

cc @chalin, please check my work on the redirect syntax here. I've added `docs.opentelemetry.io` to the netlify DNS.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -4,3 +4,8 @@ command = "npm run build:preview"
 
 [context.production]
 command = "npm run build:production"
+
+[[redirects]]
+from = "https://docs.opentelemetry.io/*"
+to = "https://opentelemetry.io/docs/:splat"
+status = 200


chalin:
Thanks. I don't know if we can mix'n match redirect rule sources. I'll take a look and test things out as soon as I can, but it might not be before Tuesday.

netlify.toml
```diff
@@ -4,3 +4,8 @@ command = "npm run build:preview"
 
 [context.production]
 command = "npm run build:production"
+
+[[redirects]]
+from = "https://docs.opentelemetry.io/*"
```
chalin, line 9: The [`from` field is meant to be a path][from], not a URL. Have you read somewhere that it's possible to redirect subdomains in this way?

[from]: https://docs.netlify.com/routing/redirects/#syntax-for-the-netlify-configuration-file

netlify.toml
```diff
@@ -4,3 +4,8 @@ command = "npm run build:preview"
...
+to = "https://opentelemetry.io/docs/:splat"
+status = 200
```
chalin, line 11: I don't think that you want a status field here. Using the default 301 seems appropriate to me.

Nit: ensure that there's a newline at the end of the file.

chalin (commented):
See inline comments, thanks.

netlify.toml
chalin: Oh! I see that using 200 keeps the URL as is. Interesting.

chalin:
Ok, I think that this looks good, except possibly for dropping that `status` field.
Do you know if/how we can test this before merging? I haven't yet found a way.

netlify.toml
chalin: I stand corrected, apparently we _can_ redirect subdomains in this way.

netlify.toml
```diff
@@ -4,3 +4,8 @@ command = "npm run build:preview"
 
 [context.production]
 command = "npm run build:production"
+
+[[redirects]]
+from = "https://docs.opentelemetry.io/*"
+to = "https://opentelemetry.io/docs/:splat"
```
chalin, line 10: Hmm, I wondering if we can just use a path here so that Netlify redirects to the default domain? I.e., just use `/docs/:splat`.

chalin:
> Thanks. I don't know if we can mix'n match redirect rule sources. I'll take a look and test things out as soon as I can, but it might not be before Tuesday.

My comment about mix'n match was because we already use both the Hugo `aliases` feature and https://github.com/open-telemetry/opentelemetry.io/blob/main/layouts/index.redirects to support redirects. Maybe we can use these features rather than modify the Netlify config file? WDYT?

chalin:
Apparently Netlify supports the addition of rules in both places, as documented in [Rule processing order](https://docs.netlify.com/routing/redirects/#rule-processing-order).

chalin (approved):
After further investigation and thinking about this a bit more, I think that we can merge this as is and adjust later as needed.

austinlparker:
Let's go for it as-is and we'll tweak as needed? Only thing I'm not really sure about is 200 vs. 301 and how that affects crawlers (does googlebot dedupe this stuff for you automagically?)

PR #1183, opened 2022-03-01
Add github.com/MrAlias/flow to registry

MrAlias:
This uses a new `registryType` (`span-processor`). Please let me know if this is incorrect and something else should be used instead.


austinlparker (changes requested):
This should be 'processor' rather than 'span-processor', as we already have a 'processor' value for type.

MrAlias:
Updated.

PR #1178, opened 2022-02-27
Added @WithSpan documentation for Java

Gaurang-Patel:
Added documentation for @WithSpan as per the discussion  here: https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/4524

Resolves #1033 

---

Preview: https://deploy-preview-1178--opentelemetry.netlify.app/docs/instrumentation/java

PR #1193, opened 2022-03-07
Java docs: refer to v1.12.0, and some copyedits

chalin:
- Followup to #1178

/cc @open-telemetry/java-approvers

### DIFF
--- a/content/en/docs/instrumentation/java/_index.md
+++ b/content/en/docs/instrumentation/java/_index.md
@@ -55,7 +55,7 @@ using our BOM to keep the versions of the various components in sync.
       <dependency>
         <groupId>io.opentelemetry</groupId>
         <artifactId>opentelemetry-bom</artifactId>
-        <version>1.10.1</version>
+        <version>1.12.0</version>
         <type>pom</type>
         <scope>import</scope>
       </dependency>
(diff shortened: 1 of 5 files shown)


content/en/docs/instrumentation/java/_index.md
```diff
@@ -74,8 +74,8 @@ using our BOM to keep the versions of the various components in sync.
 
 ```kotlin
 dependencies {
-  implementation(platform("io.opentelemetry:opentelemetry-bom:1.10.1"))
-  implementation("io.opentelemetry:opentelemetry-api")
+  implementation(platform("io.opentelemetry:opentelemetry-bom:1.12.0"))
+  implementation("io.opentelemetry:opentelemetry-api:1.12.0")
```
chalin, line 78: Note the update here. Is this ok @open-telemetry/java-approvers?

jkwatson (approved):

cartermp (approved):

content/en/docs/instrumentation/java/_index.md
anuraaga: Since we have the bom we don't need a version here, and I think it's idiomatic to leave it out as previous. Otherwise it gets harder to reason about what's controlling what version (e.g., if the version of api was 1.13.0 I'm not really sure if it would actually be 1.13.0 or the bom would take precedence)

chalin: > Since we have the bom we don't need a version here, and I think it's idiomatic to leave it out as previous.

Thanks for the feedback @anuraaga. Done: I reverted the addition to the api dep.

austinlparker (approved):

PR #1217, opened 2022-03-17
Use standard license files

chalin:
This will make the Clomonitor happy :) - https://clomonitor.io/projects/open-telemetry/open-telemetry.

/cc @austinlparker


chalin:
@austinlparker - I'm going to merge this now so that the Clomonitor can rescan by tomorrow. We can always adjust later if you have comments.

PR #1220, opened 2022-03-19
Settle on Node.js name, and use it consistently

chalin:
Let's settle on the official name used on https://nodejs.org, which is **Node.js**.

Preview, e.g.: https://deploy-preview-1220--opentelemetry.netlify.app/docs/instrumentation/


cartermp (approved):

chalin:
I'm going to assume that this is non controversial, and merge now. We can always revert/adapt if necessary.

PR #1248, opened 2022-03-29
Document WithSpan future promise behavior

trask:
cc @open-telemetry/java-instrumentation-approvers

In particular, does anyone have better idea for the term "future promise"?

### DIFF
--- a/content/en/docs/instrumentation/java/automatic/annotations.md
+++ b/content/en/docs/instrumentation/java/automatic/annotations.md
@@ -55,21 +55,22 @@ denotes its duration and provides any thrown exceptions. By default, the span
 name will be `<className>.<methodName>`, unless a name is provided as an
 argument to the annotation.
 
-If the return type of the method annotated by `@WithSpan` is one of the below "future promise" types,
-then the span will not be ended until the "future promise" completes.
+If the return type of the method annotated by `@WithSpan` is one of the
+[Future](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/Future.html)-like
+types listed below, then the span will not be ended until the future completes.
 
-* java.util.concurrent.CompletableFuture
-* java.util.concurrent.CompletionStage
-* com.google.common.util.concurrent.ListenableFuture
-* org.reactivestreams.Publisher
-* reactor.core.publisher.Mono
-* reactor.core.publisher.Flux
-* io.reactivex.Completable
-* io.reactivex.Maybe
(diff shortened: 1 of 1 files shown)


content/en/docs/instrumentation/java/automatic/annotations.md
```diff
@@ -55,6 +55,22 @@ denotes its duration and provides any thrown exceptions. By default, the span
 name will be `<className>.<methodName>`, unless a name is provided as an
 argument to the annotation.
 
+If the return type of the method annotated by `@WithSpan` is one of the below "future promise" types,
```
chalin, line 58: In terms of an alternative formulation for "future promise" (and a slight reorg of the phrase), how about this:

> ... is one of the [Future]-like types listed below, then ... until the future completes:

[Future]: https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Future.html

(I don't know if you'll want to link to https://docs.oracle.com/en/java/javase/18/docs/api/java.base/java/util/concurrent/Future.html or https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Future.html or something else.)

Future-like is used here to mean: a type that represents an asynchronous computation and possible future result.

content/en/docs/instrumentation/java/automatic/annotations.md
```diff
@@ -55,6 +55,22 @@ denotes its duration and provides any thrown exceptions. By default, the span
...
+* com.google.common.util.concurrent.ListenableFuture
+* org.reactivestreams.Publisher
+* reactor.core.publisher.Mono
+* reactor.core.publisher.Flux
+* io.reactivex.Completable
+* io.reactivex.Maybe
+* io.reactivex.Single
+* io.reactivex.Observable
+* io.reactivex.Flowable
+* io.reactivex.parallel.ParallelFlowable
```
chalin, line 72: It might be nice if these were links to API docs :) ... though that can be added later.

chalin (commented):
See inline comments.

content/en/docs/instrumentation/java/automatic/annotations.md
trask: I like that, especially since we don't actually support `java.util.concurrent.Future` 😄

anuraaga (approved):

mateuszrzeszutek (approved):

chalin (approved):

PR #1250, opened 2022-03-30
Contrib java example: automatic instrumentation with multiple exporters

gwenblum:
Cookbook link and snippet


austinlparker:
Unfortunately, this is not appropriate content for the documentation as it explicitly references a single commercial vendor.

gwenblum:
> Unfortunately, this is not appropriate content for the documentation as it explicitly references a single commercial vendor.

Got it, i just removed vendor name and kept Jaeger and OpenTelemetry project.
It seems the snippet can be valuable for OpenTelemetry beginners, i can resubmit a new PR if you agree

PR #1256, opened 2022-03-31
Update dependencies for Prometheus exporter

ChrisChinchilla:
I'm a little unsure, but all other examples show using `@opentelemetry/sdk-metrics-base` and everything works using it, so I think this is a valid change, but maybe there's a reason for using the other package I am not clear on…


austinlparker:
I can't even find an `@opentelemetry/metrics` package on npm so this change looks correct.

austinlparker (approved):

PR #1262, opened 2022-04-01
Update cookbook for manually setting span context.

markfickett:
* Include an example of injecting the trace context into a text carrier dict.
* Add more explanatory text about what the different steps do, and which examples are different ways of doing the same thing.
* Add an example of context.attach as well as the pre-existing context=ctx passed to spans.

Followup to discussion in https://cloud-native.slack.com/archives/C01PD4HUVBL/p1648217884184149 .

### DIFF
--- a/content/en/docs/instrumentation/python/cookbook.md
+++ b/content/en/docs/instrumentation/python/cookbook.md
@@ -114,9 +114,13 @@ with tracer.start_as_current_span('child', context=ctx) as span:
     span.set_attribute('primes', [2, 3, 5, 7])
 
 # Or you can make it the current context, and then the next span will pick it up.
-context.attach(ctx)
-with tracer.start_as_current_span('child') as span:
-    span.set_attribute('evens', [2, 4, 6, 8])
+# The returned token lets you restore the previous context.
+token = context.attach(ctx)
+try:
+    with tracer.start_as_current_span('child') as span:
+        span.set_attribute('evens', [2, 4, 6, 8])
+finally:
+    context.detach(token)
 ```
 
 ## Using multiple tracer providers with different Resource
--- a/content/en/docs/instrumentation/python/cookbook.md
+++ b/content/en/docs/instrumentation/python/cookbook.md
@@ -66,26 +66,39 @@ print(baggage.get_baggage("context", child_ctx))
 
 ## Manually setting span context
 
+Usually your application or serving framework will take care of propagating your trace context for you. But in some cases, you may need to save your trace context (with `.inject`) and restore it elsewhere (with `.extract`) yourself.
+
 ```python
-from opentelemetry import trace
+from opentelemetry import trace, context
 from opentelemetry.trace import NonRecordingSpan, SpanContext, TraceFlags
 from opentelemetry.sdk.trace import TracerProvider
 from opentelemetry.sdk.trace.export import ConsoleSpanExporter, BatchSpanProcessor
 from opentelemetry.trace.propagation.tracecontext import TraceContextTextMapPropagator
 
+# Set up a simple processor to write spans out to the console so we can see what's happening.
 trace.set_tracer_provider(TracerProvider())
 trace.get_tracer_provider().add_span_processor(BatchSpanProcessor(ConsoleSpanExporter()))
 
 tracer = trace.get_tracer(__name__)
 
-# Extracting from carrier header
+# A TextMapPropagator can use a regular Python dict as its Carrier.
+carrier = {}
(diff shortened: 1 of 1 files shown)


content/en/docs/instrumentation/python/cookbook.md
```diff
@@ -66,26 +66,39 @@ print(baggage.get_baggage("context", child_ctx))
...
 from opentelemetry.trace.propagation.tracecontext import TraceContextTextMapPropagator
 
+# Set up a simple processor to write spans out to the console so we can see what's happening.
 trace.set_tracer_provider(TracerProvider())
 trace.get_tracer_provider().add_span_processor(BatchSpanProcessor(ConsoleSpanExporter()))
 
 tracer = trace.get_tracer(__name__)
 
-# Extracting from carrier header
+# A TextMapPropagator can use a regular Python dict as its Carrier.
```
aabmass, line 84:
```suggestion
# A TextMapPropagator works with any dict-like object as its Carrier by default. You can also implement custom getters and setters.
```

content/en/docs/instrumentation/python/cookbook.md
```diff
@@ -66,26 +66,39 @@ print(baggage.get_baggage("context", child_ctx))
...
+carrier = {}
+with tracer.start_as_current_span('first-trace'):
+    # Write the current context into the carrier.
+    TraceContextTextMapPropagator().extract(carrier)
```
aabmass, line 88: I think you meant `inject()` here

```suggestion
with tracer.start_as_current_span('first-trace'):
    carrier = {}
    # Write the current context into the carrier.
    TraceContextTextMapPropagator().inject(carrier)
```

content/en/docs/instrumentation/python/cookbook.md
```diff
@@ -66,26 +66,39 @@ print(baggage.get_baggage("context", child_ctx))
...
+
+# The below might be in a different thread, on a different machine, etc.
+# As a typical example, it would be on a different microservice and the carrier would
+# have been forwarded via HTTP headers.
+
+# Extract the trace context from the carrier.
+# Here's what a typical carrier might look like, as it would have been extracted above.
```
aabmass, line 95:
```suggestion
# Here's what a typical carrier might look like, as it would have been injected above.
```

content/en/docs/instrumentation/python/cookbook.md
```diff
@@ -66,26 +66,39 @@ print(baggage.get_baggage("context", child_ctx))
...
 carrier = {'traceparent': '00-a9c3b99a95cc045e573e163c3ac80a77-d99d251a8caecd06-01'}
+# Then we use a propagator to get a trace context from it.
```
aabmass, line 97: nit span context is a subset of the overall context

```suggestion
# Then we use a propagator to get a context from it.
```

content/en/docs/instrumentation/python/cookbook.md
```diff
@@ -94,7 +107,15 @@ span_context = SpanContext(
...
 
-with tracer.start_as_current_span("child", context=ctx) as span:
+# Now there are a few ways to make use of the trace context.
+
+# You can pass the context object when starting a span.
+with tracer.start_as_current_span('child', context=ctx) as span:
+    span.set_attribute('primes', [2, 3, 5, 7])
+
+# Or you can make it the current context, and then the next span will pick it up.
+context.attach(ctx)
```
aabmass, line 117: The best practice here would be wrapping this in a `try` and detaching in the `finally`. I believe it would look like

```py
try:
  token = context.attach(ctx)
  ...
finally:
  context.detach(token)
```

otherwise the context can leak

markfickett: Thanks, updated. I put the `attach` before the `try` since if that fails you won't have a `token` to detach with.

markfickett:
Ready for another look, thanks for reviewing!

lzchen (approved):

aabmass (approved):
Thanks for the PR!

codeboten (approved):

markfickett:
Hi all, thanks for the reviews. GitHub is saying "At least 1 approving review is required to merge this pull request." I see the docs approvers are @open-telemetry/docs-approvers but the link on the README 404s for me, so please let me know if I should be adding anyone. But, if everything's proceeding appropriately I'm in no rush, just want to make sure I'm not holding anything up!

lzchen:
PR to update codeowners: https://github.com/open-telemetry/opentelemetry.io/pull/1272

austinlparker (approved):

Issue #1200, opened 2022-03-10
Distributions - creating a distribution - language specific best practices.


svrnm:
**Desired feature or idea.**

When going to the [distributions page](https://opentelemetry.io/docs/concepts/distributions/) checking for some guidance on "creating my own distribution" I can only find details for the OpenTelemetry collector. 

It would be great to also have some details for the language specific SDKs/auto-instrumentations:

* Java provides an [extension mechanism](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/examples/extension/README.md). From my understanding the right approach for creating a distribution of that is building extension
* Python provides a [opentelemetry-distro package](https://opentelemetry.io/docs/instrumentation/python/distro/). I assume that building/extending this package is much more efficient than cobbling everything together yourself
* JavaScript provides the [SDK for Node.js](https://github.com/open-telemetry/opentelemetry-js/tree/main/experimental/packages/opentelemetry-sdk-node). Again, I assume it could be a best practice to build on top of it. 
* ...

While I don't want to prescribe vendors (and others) how to build their own distribution of opentelemetry, it would be great to provide some starting points & guidance. I see the following advantages:

* Having a "common structure" makes it easier for distributions to make use of existing installation/injection mechanisms like the K8s operator, (https://github.com/open-telemetry/opentelemetry-operator/issues/752 is one of the reasons why I opened this issue)
* Building on top of existing "base distributions" is also valuable for the OTel community: vendors might reconsider building their own stuff and prefer adding features to the existing packages and therefor help the community to make their implementations better.

**Next Steps**

I first of all wanted to check if this is a good/bad idea, and if it is reasonable, I think the next step would be to work with the individual language-specific SIGs to get an idea from them what they think a good practice for their language would look like. Maybe a first step is just linking from that page to existing work as mentioned before (Java Extension Mechanism, Python Base Distro, SDK for Node.js,...)

austinlparker:
I like the idea, but I'm not sure if 'distribution' is really the right concept we should be capturing. Maybe each language should have an 'extensibility' section that we can file docs like this under? Or, to reverse it, there should be an 'Extensibility' topic that has items like, 'Building a custom OTel Collector', 'Building a custom Java Agent', 'Extending the Node.JS SDK', etc. etc.

svrnm:
> I like the idea, but I'm not sure if 'distribution' is really the right concept we should be capturing.

[...]

Issue #1229, opened 2022-03-23
Migrate Java Agent doc pages to this repo, starting with Agent config


chalin:
@trask @jack-berg et al., as part of the work on

- https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/4523

What do you think about migrating selected Java-Agent pages from [opentelemetry-java-instrumentation/docs] to this repo under [Docs > Instrumentation > Java > Automatic]. Good candidate page to start off with might be:

- [x] [docs/agent-config.md] 
  - [x] #1233
  - [x] https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/5683
- [x] [docs/suppressing-instrumentation.md]
  - [x] #1237
  - [x] Factor out part to be left in code repo: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/5693
  - [x] https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/5694
  - [x] https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/5712
- [x] [docs/config/common.md]
  - [x] #1246
  - [x] https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/5707
  - Followup: #1247
- [x] https://github.com/open-telemetry/opentelemetry.io/issues/1247
- [ ] ...

/cc @austinlparker @cartermp @svrnm 

[Docs > Instrumentation > Java > Automatic]: https://opentelemetry.io/docs/instrumentation/java/automatic/
[opentelemetry-java-instrumentation/docs]: https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/docs
[docs/agent-config.md]: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/docs/agent-config.md
[docs/suppressing-instrumentation.md]: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/3ab1b8516feb5a0065d5b1e0d6a242049f2c7aea/docs/suppressing-instrumentation.md
[docs/config/common.md]: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/4ef6d165a9de883b2cebd727d7794c9cc530be97/docs/config/common.md

trask:
I support this

cc @open-telemetry/java-instrumentation-approvers

jack-berg:
[...]

PR #1263, opened 2022-04-01
Add "contribution vs distribution" & language specifics to Distributions page

svrnm:
Related to #1200:

* I thought it would be good to call out on the Distribution page that people should consider contributing back to the community before deviating through their distribution. 
* I added Java as an example to the language specific ones, I was not sure about the other languages, but we could check with them if there are some recommendations. (Note I copied the text for the extensibility from https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/examples/extension/README.md)


austinlparker (approved):

austinlparker:
This PR is causing the link checker to fail, but the content looks fine. Could you take care of the checks?

svrnm:
good catch @austinlparker, this comes from the fact that I just copied the content from the java repo. 

@chalin, this is related to #1229, would it be ok if I migrate https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/examples/extension/README.md to the docs via this PR? If so, I proceed and fix that links.

content/en/docs/instrumentation/java/extensibility.md
```diff
@@ -0,0 +1,69 @@
...
+
+Extensions add new features and capabilities to the agent without having to create a separate distribution (for examples and ideas, see [Use cases for extensions](#sample-use-cases)).
+
+The contents in this folder demonstrate how to create an extension for the OpenTelemetry Java instrumentation agent, with examples for every extension point. 
+
+> Read both the source code and the Gradle build script, as they contain documentation that explains the purpose of all the major components.
+
+## Build and add extensions
+
+To build this extension project, run `./gradlew build`. You can find the resulting jar file in `build/libs/`.
```
chalin, line 16: This line says "run `./gradlew build`" without any context. What steps do I need to following _before_ I can run this command? I assume that there are sources to fetch first.

content/en/docs/instrumentation/java/extensibility.md
```diff
@@ -0,0 +1,69 @@
...
+For more information, see the `extendedAgent` task in [build.gradle](build.gradle).
+## Extensions examples
+
+* Custom `IdGenerator`: [DemoIdGenerator](src/main/java/com/example/javaagent/DemoIdGenerator.java)
+* Custom `TextMapPropagator`: [DemoPropagator](src/main/java/com/example/javaagent/DemoPropagator.java)
+* New default configuration: [DemoPropertySource](src/main/java/com/example/javaagent/DemoPropertySource.java)
+* Custom `Sampler`: [DemoSampler](src/main/java/com/example/javaagent/DemoSampler.java)
+* Custom `SpanProcessor`: [DemoSpanProcessor](src/main/java/com/example/javaagent/DemoSpanProcessor.java)
+* Custom `SpanExporter`: [DemoSpanExporter](src/main/java/com/example/javaagent/DemoSpanExporter.java)
+* Additional instrumentation: [DemoServlet3InstrumentationModule](src/main/java/com/example/javaagent/instrumentation/DemoServlet3InstrumentationModule.java)
```
chalin, line 41: All of the local links here are invalid -- i.e., those files don't exist here. Do these files exist somewhere else? If so, link to their origin.

chalin (changes requested):
See inline comments.

chalin:
Just did a very quick pass over this PR so that I could get you some feedback before heading out. (See my inline comments from early.)

> good catch @austinlparker, this comes from the fact that I just copied the content from the java repo.
> 
> @chalin, this is related to #1229, would it be ok if I migrate https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/examples/extension/README.md to the docs via this PR? If so, I proceed and fix that links.

Oh, now I see where those files are coming from! So this PR _is_ effectively a migration of that README, right? Yes, it's ok for this PR to contribute to #1229, if that is what your question is. So yes, pls fix the links.

Note that you'll need to do a followup PR to remove the content of the original README and direct readers to the new OTel website page.

content/en/docs/instrumentation/java/extensibility.md
```diff
@@ -0,0 +1,69 @@
...
+     ```
+## Embed extensions in the OpenTelemetry Agent
+
+To simplify deployment, you can embed extensions into the OpenTelemetry Java Agent to produce a single jar file. With an integrated extension, you no longer need the `-Dotel.javaagent.extensions` command line option.
+
+For more information, see the `extendedAgent` task in [build.gradle](build.gradle).
+## Extensions examples
+
+* Custom `IdGenerator`: [DemoIdGenerator](src/main/java/com/example/javaagent/DemoIdGenerator.java)
+* Custom `TextMapPropagator`: [DemoPropagator](src/main/java/com/example/javaagent/DemoPropagator.java)
```
trask, line 36: what do you think about keeping the specifics of the examples in the instrumentation repo (alongside the code), and focusing this page on general documentation about javaagent extensions?

content/en/docs/concepts/distributions.md
```diff
@@ -62,6 +74,12 @@ If you are building your own distribution, the [OpenTelemetry Collector
 Builder](https://github.com/open-telemetry/opentelemetry-collector/tree/main/cmd/builder)
 might be a good starting point.
 
+### Language Specific Instrumentation libraries
+
+There are language specific extensibility mechanisms to customize the instrumentation libraries:
+
+- [Java](../../instrumentation/java/extensibility)
```
trask, line 81: this extensibility is specific to the Javaagent, there are different extensibility mechanisms for (what we call) "library" instrumentation, I'm not sure the best way to call this out though

content/en/docs/instrumentation/java/extensibility.md
svrnm: Yes, I would point to the source code in the repo here!

content/en/docs/concepts/distributions.md
svrnm: Understood, so if someone is using the SDK without the Agent, they will need to find another way to extensibility? Does something light this exist?

To add a little bit where I am coming from: As a vendor having a clear contract to extend on top of is highly valuable, because it takes away some uncertainty if things stay that way or change. Also there are multiple ways per language on how to do things, so having a "best practice" is great.  

Finally, _having a way to  add new features and capabilities to the agent without having to create a separate distribution_ [[1](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/examples/extension/README.md)] is adding value for the end user: instead of replacing distributions they only need to swap out extensions (or MAYBE can use extensions from multiple vendors).

trask: When not using the agent, I would say the "best practice" is to use the [autoconfigure module](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk-extensions/autoconfigure/README.md#opentelemetry-sdk-autoconfigure).

There are further extensibility points for (what we call) library instrumentation (but which still fall under otel "auto-instrumentation" definition). These allow you to customize individual instrumentation (as opposed to the autoconfigure module which allows you to customize the SDK). These library instrumentation extensibility points are not well documented yet though.

svrnm: @trask, thanks. I can update `Java` here with `Java Agent`. Or what is the best name for the agent?

content/en/docs/instrumentation/java/extensibility.md
svrnm: I removed that section, will remain in the repo.

svrnm:
Updated the extensibility document to only contain documentation, the instructions for the extension examples will remain in the main repo.

content/en/docs/concepts/distributions.md
trask: > what is the best name for the agent?

we have been using `Javaagent` (one word), to try to differentiate from common usage of `Agent` meaning out-of-proc agent

svrnm: thx, put Javaagent.

trask (approved):
thx!

svrnm:
@chalin: I closed all the requested changes

content/en/docs/concepts/distributions.md
```diff
@@ -51,6 +51,18 @@ distribution if they wish to use components in the
...
+Before you read on and learn how you can create your own distribution, ask yourself if your
+additions on top of an OpenTelemetry component would be beneficial for everyone and therefore
+should be included in the reference implementations:
+
+- Can your scripts for "ease of use" be generalized?
+- Can your changes to default settings be the better option for everyone?
+- Are your additional packaging options really specific?
+- Might your test, performance & security coverage work with the reference implementation as well?
+- Have you checked with the community if your additional capabilities could be part of the standard?
+
```
rquedas, line 65: Is the goal here to make a difference between a contribution and distribution? The distinction feels implicit in the content, I would make it a little more explicit.

rquedas (changes requested):
I would just address the "contribution vs distribution" piece, the rest seems ok to me

content/en/docs/concepts/distributions.md
svrnm: The goal is that people reconsider if all the features they create for their distribution are unique to them or if they should give them back to the community.

How would you make it more explicit?

svrnm:
> I think there's a change requested from @chalin? otherwise lgtm

I don't see them when I click on "view changes"🤔

content/en/docs/instrumentation/java/extensibility.md
```diff
@@ -0,0 +1,20 @@
+---
+title: Extensibility
+description: Extensions add new features and capabilities to the agent without having to create a separate distribution or to fork this repository. For example, you can create custom samplers or span exporters, set new defaults, and embed it all in the agent to obtain a single jar file.
+---
+
+## Introduction
+
+Extensions are designed to override or customize the instrumentation provided by the upstream agent without having to create a new OpenTelemetry distribution or alter the agent code in any way.
```
chalin, line 8: Nit: line-wrap paragraphs to 80 chars in this page.

content/en/docs/instrumentation/java/extensibility.md
```diff
@@ -0,0 +1,20 @@
+---
+title: Extensibility
+description: Extensions add new features and capabilities to the agent without having to create a separate distribution or to fork this repository. For example, you can create custom samplers or span exporters, set new defaults, and embed it all in the agent to obtain a single jar file.
```
chalin, line 3: This page is about extensions, so why not use that as a title?

> title: Extensions

Also, the description should be short, for example:

> What is an extensions and when should I create one

chalin (approved):
Haven't re-reviewed in detail, but overall LGTM.
See inline comments for suggested title change.

content/en/docs/instrumentation/java/extensibility.md
chalin: If you change the title, also rename the file.

content/en/docs/instrumentation/java/extensibility.md
svrnm: done

content/en/docs/instrumentation/java/extensibility.md
svrnm: done

svrnm:
@chalin, thanks, updated according to your review.

mateuszrzeszutek (approved):

svrnm:
@chalin , @austinlparker : anything missing from your side?

austinlparker:
nope, lgtm

PR #1275, opened 2022-04-05
fixing 'Building a custom authenticator' broken link

rquedas:
The reference in the config page is broken and generating a 404


chalin:
Got it! Ok, closing.

PR #1294, opened 2022-04-11
Add troubleshooting to node.js

svrnm:
Small change for the examples with node.js: it took me longer than I am happy to admit to find out how I can add debug output, so I thought it would be good to let people know during the getting started (inspired by @blumamir's [blog post](https://opentelemetry.io/blog/2022/troubleshooting-nodejs/), which has a bunch of great tips that should become part of the official docs:) )

### DIFF
--- a/content/en/docs/instrumentation/js/getting-started/nodejs.md
+++ b/content/en/docs/instrumentation/js/getting-started/nodejs.md
@@ -90,6 +90,10 @@ Create a file with a name like `tracing.js` which will contain your tracing setu
 // Require dependencies
 const opentelemetry = require("@opentelemetry/sdk-node");
 const { getNodeAutoInstrumentations } = require("@opentelemetry/auto-instrumentations-node");
+const { diag, DiagConsoleLogger, DiagLogLevel } = require('@opentelemetry/api');
+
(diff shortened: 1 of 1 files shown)


cartermp:
Hmmm, I feel like this might be better in an explicit troubleshooting section (could also perhaps be its own doc, but for now a section is likely fine)

svrnm:
> Hmmm, I feel like this might be better in an explicit troubleshooting section (could also perhaps be its own doc, but for now a section is likely fine)

Actually, I was wondering if we could take the [blog post about node.js troubleshooting](https://opentelemetry.io/blog/2022/troubleshooting-nodejs/) and turn it into that kind of section

However, I thought it would be a good place to call out already how people can simply turn on more or less logging while going through things

cartermp:
I think it'd be probably best in a `## Troubleshooting` section in this document for now, rather than in the first code sample. I don't necessarily think that it's harmful, just that it's a little uncommon to have troubleshooting lines of code in the happy path when getting started.

austinlparker (approved):
Eh, explicit vs. implicit is a good place to start for this kind of info.

PR #1312, opened 2022-04-22
Fix OpenTelemetry .NET status

reyang:
Metrics Stable version was shipped 7 days ago.
The integration with ILogger is also stable.

---

Preview: https://deploy-preview-1312--opentelemetry.netlify.app/docs/instrumentation/net/

### DIFF
--- a/content/en/docs/instrumentation/net/_index.md
+++ b/content/en/docs/instrumentation/net/_index.md
@@ -23,7 +23,14 @@ language:
 
 | Tracing | Metrics | Logging |
 | ------- | ------- | ------- |
-| Stable  | Stable  | Stable  |
+| Stable  | Stable  | Stable* |
+
+\* Most components are providing Stable support for
+[ILogger](https://docs.microsoft.com/dotnet/api/microsoft.extensions.logging.ilogger)
+integration. OTLP Log Exporter is still Experimental. Please refer to the
+official [OpenTelemetry
+.NET](https://github.com/open-telemetry/opentelemetry-dotnet) project to get the
+latest status of individual components.
 
 {{% latest_release "dotnet" /%}}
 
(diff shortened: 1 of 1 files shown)


cartermp (approved):

content/en/docs/instrumentation/net/_index.md
```diff
@@ -23,7 +23,7 @@ language:
 
 | Tracing | Metrics | Logging |
 | ------- | ------- | ------- |
-| 1.0    | Alpha   | Beta    |
+| Stable  | Stable  | Stable  |
```
cijothomas, line 26: Though  ILogger integration is stable, I'd put an "*" here, and indicate that the OTLP Log Exporter is not stable, to avoid any confusion (and incorrectly assume that OTLP Exporter for Logs is also stable)

cijothomas (approved):

content/en/docs/instrumentation/net/_index.md
reyang: Updated.

chalin (changes requested):
I'd vote against using an informal status of "stable*". Instead, I think that we should provide two entries with clearly distinguished statuses (**stable**, and **experimental**, respectively).

I also think that we should take this opportunity to convert the status table into the layout used in the [Python page](https://opentelemetry.io/docs/instrumentation/python/#status-and-releases). I can submit something along these lines tomorrow or Wednesday, if you'd like.

Thoughts @austinlparker @cartermp et al.?

reyang:
> I also think that we should take this opportunity to convert the status table into the layout used in the [Python page](https://opentelemetry.io/docs/instrumentation/python/#status-and-releases). I can submit something along these lines tomorrow or Wednesday, if you'd like.

Thanks @chalin. Yes please go ahead, I'll close this PR.

Issue #964, opened 2021-11-29
Getting started: provide a uniform experience for Collector and all language SIGs


chalin:
I think that each language SIG's **Getting started** page should use the equivalent of a "console exporter". Exporting to a backend should be covered elsewhere.

Thoughts @austinlparker @cartermp @theletterf

chalin:
☝️

cartermp:
Yep, agreed! I would also argue that every language section should have a dedicated page (or exporter section) that shows how to export to a backend, since it is important, just not for getting started.

The question in my mind is, what should the thing being instrumented cover? I can see two options:

1. An http server or something (think "hello world with express") that uses automatic instrumentation to track the inbound request, then creates at least 1 child span of that request's span, with the child/children tracking some distinct operation before a result is sent back as a response. Also demonstrates setting an attribute on a span.
2. A pure console app with no other dependencies that tracks some unit of work with a parent and child span. Also demonstrates setting an attribute on a span.

I lean a _little_ more towards (1) since I believe incorporate automatic instrumentation is critically important and we should encourage using it as a first step so that you can get some easy wins by using OTel. However, it makes setup more complex, any sample app is a little harder to run, and in some cases (e.g., Java) requires running an agent in addition to the app.

svrnm:
@chalin when you say "console collector" do you mean a console exporter or do you mean that there should be a collector receiving data via OTLP and the collector is then dumping the data to the console?

From an end-user experience I would argue that both would be great:
* Using the exporter is a quick win and it's easy to debug if anything goes wrong
* At the end, exporting with OTLP to a collector is recommended, so having a quick route to that as well is the next thing people will look for.

I think about the following flow

1. Do you have an application in <language> already? 
 a. Yes? great, you can use that one!
 b. No? ok, please use that "hello http server" as @cartermp suggested
2. Auto-instrument that service and use a console exporter to dump everything to the console. There are other exporters, go __`here`__ to learn more
3. Exporters are good, collectors are great! Do you have a collector already?
  a. Yes? great, proceed
  b. No?, ok, go here and learn how to install one that dumps everything to the console
4. Reconfigure your service to use OTLP and point your data to the collector. Collector is now dumping everything to the console. There are other exporters for the collector, go __`here`__ to learn more. 

I am happy to help to make this happen / find people to help with that, please let me know if this make sense :)

chalin:
> @chalin when you say "console collector" do you mean a console exporter

[...]

PR #1322, opened 2022-04-28
First draft of python get started rewrite

cartermp:
This is in service towards making the "get started" guides for each languages focus on autoinstrumentation first.

So, uhh, the CNCF posted a blog post recently that's basically what we'd like to have (more or less): https://www.cncf.io/blog/2022/04/22/opentelemetry-and-python-a-complete-instrumentation-guide/

It is not a complete instrumentation guide, but it is a _damn good guide on how to get started_. Since copying something is the greatest form of flattery, I decided to do that and put up a PR to see what people think. More than happy to change things up, and I expect we'd want to ask the original author's permission since so much is similar. If we'd prefer to take a different route, I'm happy to do that and adjust.

### DIFF
--- a/content/en/docs/instrumentation/python/getting-started.md
+++ b/content/en/docs/instrumentation/python/getting-started.md
@@ -1,285 +1,268 @@
 ---
 title: Getting Started
 weight: 1
-code_block_from__path_base: content-modules/opentelemetry-python/docs/getting_started
 ---
 
-This guide walks you through instrumenting a Python application with
-`opentelemetry-python`.
+In this page, you'll learn how to set up and get tracing telemetry from an HTTP
+server with Flask. If you're not using Flask, that's fine - this guide will also
+work with Django, FastAPI, [and
+more](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/instrumentation).
 
 For more elaborate examples, see
 [examples](https://github.com/open-telemetry/opentelemetry-python/tree/main/docs/examples/).
 
-## Hello world: emit a trace to your console
+## Installation
 
-To get started, install both the opentelemetry API and SDK:
+To begin, set up an environment and install dependencies:
 
-```sh
-pip install opentelemetry-api
-pip install opentelemetry-sdk
+```console
+$ mkdir otel-getting-started
+$ cd otel-getting-started
+$ python3 -m venv .
+$ source ./bin/activate
 ```
 
-The API package provides the interfaces required by the application owner, as
-well as some helper logic to load implementations.
+Now install the HTTP server:
 
-The SDK provides an implementation of those interfaces. The implementation is
-designed to be generic and extensible enough that in many situations, the SDK is
-sufficient.
+```console
+$ pip install flask
+```
 
-Once installed, you can use the packages to emit spans from your application. A
-span represents an action within your application that you want to instrument,
-such as an HTTP request or a database call. Once instrumented, you can extract
-helpful information such as how long the action took. You can also add arbitrary
-attributes to the span that provide more insight for debugging.
+And finally, install OpenTelemetry:
 
-The following example script emits a trace containing three named spans: “foo”,
-“bar”, and “baz”:
+```console
+$ pip install opentelemetry-distro
+```
 
-{{% code_block_from file="tracing_example.py" from="15" %}}
+The `opentelemetry-distro` package installs the API, SDK, and the
+`opentelemetry-bootstrap` and `opentelemetry-instrument` tools that you'll use
+soon.
 
-When you run the script you can see the traces printed to your console:
+## Add automatic instrumentation
 
-```sh
-$ python tracing_example.py
+Automatic instrumentation will generate telemetry data on your behalf. There are
+several options you can take, covered in more detail in [Automatic
+Instrumentation]({{< relref "automatic" >}}). Here we'll use the
+`opentelemetry-instrument` agent.
+
+First, create a file `app.py`:
+
+```python
+from random import randint
+from flask import Flask, request
+
+app = Flask(__name__)
+
+@app.route("/roll")
+def roll():
+    sides = int(request.args.get('sides'))
+    rolls = int(request.args.get('rolls'))
+    return roll_sum(sides,rolls)
+
+
+def roll_sum(sides, rolls):
+    sum = 0
+    for r in range(0,rolls):
+        result = randint(1,sides)
+        sum += result
+    return str(sum)
+```
+
+Next, install automatic instrumentation:
+
+```console
+$ opentelemetry-bootstrap -a install
+```
+
+This will install Flask instrumentation.
+
+## Run the instrumented app
+
+You can now run your instrumented app with `opentelemetry-instrument` and have
+it print to the console for now:
+
+```console
+$ opentelemetry-instrument --traces_exporter console flask run
+```
+
+When you access the server, you'll get a result in a trace printed to the
+console, such as the following:
+
+```console
 {
-    "name": "baz",
+    "name": "/roll",
     "context": {
-        "trace_id": "0xb51058883c02f880111c959f3aa786a2",
-        "span_id": "0xb2fa4c39f5f35e13",
-        "trace_state": "{}"
+        "trace_id": "0xdcd253b9501348b63369d83219da0b14",
+        "span_id": "0x886c05bc23d2250e",
+        "trace_state": "[]"
     },
-    "kind": "SpanKind.INTERNAL",
-    "parent_id": "0x77e577e6a8813bf4",
-    "start_time": "2020-05-07T14:39:52.906272Z",
-    "end_time": "2020-05-07T14:39:52.906343Z",
-    "status": {
-        "status_code": "OK"
-    },
-    "attributes": {},
-    "events": [],
-    "links": []
-}
-{
-    "name": "bar",
-    "context": {
-        "trace_id": "0xb51058883c02f880111c959f3aa786a2",
-        "span_id": "0x77e577e6a8813bf4",
-        "trace_state": "{}"
-    },
-    "kind": "SpanKind.INTERNAL",
-    "parent_id": "0x3791d950cc5140c5",
-    "start_time": "2020-05-07T14:39:52.906230Z",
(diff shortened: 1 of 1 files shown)


content/en/docs/instrumentation/python/getting-started.md
```diff
@@ -1,285 +1,268 @@
...
+more](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/instrumentation).
 
 For more elaborate examples, see
 [examples](https://github.com/open-telemetry/opentelemetry-python/tree/main/docs/examples/).
 
-## Hello world: emit a trace to your console
+## Installation
 
-To get started, install both the opentelemetry API and SDK:
+To begin, set up an environment and install dependencies:
```
svrnm, line 16: Suggestion: Add "in an empty directory"

content/en/docs/instrumentation/python/getting-started.md
```diff
@@ -1,285 +1,268 @@
...
+```console
+$ mkdir otel-getting-started
+$ cd otel-getting-started
+$ python3 -m venv .
+$ source ./bin/activate
 ```
 
-The API package provides the interfaces required by the application owner, as
-well as some helper logic to load implementations.
+Now install the HTTP server:
```
svrnm, line 25: is "HTTP server" really the correct description for Flask?

content/en/docs/instrumentation/python/getting-started.md
```diff
@@ -1,285 +1,268 @@
...
-        traces:
-            receivers: [otlp]
-            exporters: [logging]
-            processors: [batch]
+```console
+$ opentelemetry-instrument --traces_exporter console flask run
 ```
 
-Then start the Docker container:
+You'll now see two spans in the trace emitted to the console, and the one called
```
svrnm, line 182: Add that you need to send a web request to the server, then you will see the spans.

content/en/docs/instrumentation/python/getting-started.md
cartermp: I'll clean this up, I think at some point in the future we'd want to also cover Django, at which point wording would have to change up a bit. But that's a problem for another day

cartermp:
Thanks @svrnm, updated based on feedback!

cartermp:
Also, I suppose this should have a section for sending to a collector to be in service towards https://github.com/open-telemetry/opentelemetry.io/issues/964

cartermp:
Okay, this is ready for final review now that I added a "send to a local collector" section.

We may want to link to more detailed collector docs, but for now I think it is fine. The collector docs could use some love, particularly in listing out more configuration options and a grow-up story from a super-basic local configuration to a fully-fledged production configuration.

content/en/docs/instrumentation/python/getting-started.md
```diff
@@ -1,249 +1,265 @@
...
 * A single telemetry sink shared by multiple services, to reduce overhead of
-  switching exporters.
-
+  switching exporters
+* Aggregating traces across multiple services, running on multiple hosts
+* A central place to process traces prior to exporting them to a backend
 
-* Aggregating traces across multiple services, running on multiple hosts.
+Unless you have just a single service or are experimenting, you'll want to use a
+collector in production code.
```
jharley, line 258:
```suggestion
collector in production deployments.
```

content/en/docs/instrumentation/python/getting-started.md
```diff
@@ -265,21 +281,122 @@ service:
...
 
-```sh
+To do this, install the OTLP exporter package:
+
+```
 pip install opentelemetry-exporter-otlp
 ```
 
-Finally, execute the following script:
+Then configure the exporter. By default, it will send to `locahost:4317`, which is what the collector listens on.
```
jharley, line 305: I think worth mentioning how to set the OTLP endpoint via environment variables here as well? Or linking to a place that explains it

jharley (commented):
I really like this update. Added some suggestions.

content/en/docs/instrumentation/python/getting-started.md
```diff
@@ -1,249 +1,265 @@
...
 
-{{% code_block_from file="flask_example.py" from="15" %}}
-
-Now run the script, hit the root URL
-([http://localhost:5000/](http://localhost:5000/)) a few times, and watch your
-spans be emitted!
-
-```sh
-python flask_example.py
+```console
```
svrnm, line 83: replace with `json`?

content/en/docs/instrumentation/python/getting-started.md
```diff
@@ -1,249 +1,265 @@
...
 
-* Aggregating traces across multiple services, running on multiple hosts.
+Unless you have just a single service or are experimenting, you'll want to use a
+collector in production code.
 
-To enable a broad range of aggregation strategies, OpenTelemetry provides the
-[opentelemetry-collector](https://github.com/open-telemetry/opentelemetry-collector).
-The Collector is a flexible application that can consume trace data and export
-to multiple other backends, including to another instance of the Collector.
+### Configure and run a local collector
```
svrnm, line 260: I see this replicated between your .NET PR and the python PR here, would be good to externalise (not sure if this should be part of the "Getting Started" with the collector or if this could be an excerpt or something else that could be reused), with that one could use it for all the other languages as well.

cartermp: Yeah, I was thinking about using subpages here. @chalin any thoughts? I'm fine with duplicating for now and then consolidating later, but whatever is easiest

svrnm: I was just working on https://github.com/open-telemetry/opentelemetry.io/pull/1332 and I am wondering if this section you wrote (and especially that "output to console" collector config) should live somewhere on that getting started page for the collector (could be after those PRs are merged)

cartermp: Yeah, it does make for a good "Get started" experience for just local development before you stand anything up for real. And it's easy as heck to do! Tentatively I'd say including it in #1332 is better than not including it

content/en/docs/instrumentation/python/getting-started.md
cartermp: I think I'll leave this as-is for now, we don't really have docs for this yet, and I think two things would need to happen as well:

1. Get this in the Exporters docs for each langauge
2. Decide on which convention we document as the primary way to do things (configure in code or as env vars), then adjust all docs to handle that preference and then also document the equally-valid alternative

content/en/docs/instrumentation/python/getting-started.md
```diff
@@ -1,249 +1,265 @@
...
+# These are the necessary import declarations
+from opentelemetry import trace
+from opentelemetry.sdk.trace import TracerProvider
+
+from random import randint
+from flask import Flask, request
+
+# This is a way to initialize tracing
+provider = TracerProvider()
+trace.set_tracer_provider(provider)
```
srikanthccv, line 148: distro already sets up a tracer provider so this step is not effective.

content/en/docs/instrumentation/python/getting-started.md
```diff
@@ -265,21 +281,122 @@ service:
...
+)
+from opentelemetry.exporter.otlp.proto.http.trace_exporter import (
+    OTLPSpanExporter,
+)
+
+from random import randint
+from flask import Flask, request
+
+# Initialize tracing, but with OTLP
+provider = TracerProvider(resource=resource)
```
srikanthccv, line 322: Where is resource defined?

content/en/docs/instrumentation/python/getting-started.md
```diff
@@ -265,21 +281,122 @@ service:
...
+processor = BatchSpanProcessor(OTLPSpanExporter())
+provider.add_span_processor(processor)
+trace.set_tracer_provider(provider)
```
srikanthccv, line 325: Same as above. distro sets up the trace pipeline already so this part is not doing anything.

cartermp:
Thanks @srikanthccv - I should have all feedback addressed now.

One thing I noticed with the `opentelemetry-instrument` agent is that I still needed to wire up an `OTLPSpanExporter` in the application to export traces. It seems like once a trace crosses the boundary into the application, you can't just rely on the agent's default OTLP export? Perhaps I'm holding it wrong, though.

In any case, the final sample does work in my testing now.

srikanthccv:
> One thing I noticed with the opentelemetry-instrument agent is that I still needed to wire up an OTLPSpanExporter in the application to export traces. It seems like once a trace crosses the boundary into the application, you can't just rely on the agent's default OTLP export?

@cartermp you would use the `otlp_proto_grpc` or `otlp_proto_http` in place of `console` for --traces_exporter and it should work. Did it not send any traces? If so, then that's bug. Your code works since you can have the multiple provider and exporter pipelines but there can only be one global provider. So when people use `trace.get_tracer(...)` it's the global provider which creates new tracer and might be surprising for audience if they realise it's not from provider instantiated manually, what do you think?

austinlparker:
Is this ready for review/merge?

chalin (commented):
LGTM (I'd make a copyediting pass afterwards).

content/en/docs/instrumentation/python/getting-started.md
```diff
@@ -1,249 +1,271 @@
...
-visualized together.
+## Add automatic instrumentation
 
-This concept of aggregating span and trace information is known as distributed
-tracing. One such distributed tracing backend is known as Jaeger. The Jaeger
-project provides an all-in-one Docker container with a UI, database, and
-consumer.
+Automatic instrumentation will generate telemetry data on your behalf. There are
+several options you can take, covered in more detail in [Automatic
+Instrumentation]({{< relref "automatic" >}}). Here we'll use the
```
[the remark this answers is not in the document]
cartermp: It is on the site, yeah. I think we will be moving away from `relref`s though. I am likely going to go through and update these to "normal" relative links, but it's everywhere in the repo and so that might be done orthogonally

srikanthccv (approved):

cartermp:
@open-telemetry/python-approvers anything else you'd like to see here?

ocelotl (approved):

PR #1333, opened 2022-05-03
Add install collector script

svrnm:
Creating this as a draft because I wanted to pitch the idea first. While updating the collector getting started, I played around with the different ways of getting a collector up and running. Pulling and running it through a docker container is definitely the most convenient way but it comes with plenty of overhead and sometimes I just want to have a collector running on my laptop. This requires a few clicks right now (and you need to read closely because arm64 and amd64 are hard to distinguish), so I thought, why not have a similar install experience as `brew` or `nvm`:

```console
/bin/bash -c "$(curl -s http://opentelemetry.io/install/otelcol.sh)" 
./otelcol --config config.yml
```

or in a video:

https://user-images.githubusercontent.com/1519757/166491524-e1b2d65b-ebbc-44ec-a62f-cef605c43eca.mp4

what do you think, something OTel needs? :-)

### DIFF
--- a/content/en/install-collector.sh
+++ b/content/en/install-collector.sh
@@ -1,0 +1,44 @@
+#!/bin/bash
+
+echo -n "Installing the OpenTelemetry Collector into your current directory: $(pwd) in 3"
+#sleep 1
+#echo -n " 2"
+#sleep 1
+#echo -n " 1"
+#sleep 1
+echo ""
+
+LATEST_URL=$(curl -Ls -o /dev/null -w %{url_effective} https://github.com/open-telemetry/opentelemetry-collector-releases/releases/latest)
+
+VERSION=${LATEST_URL##*/v}
+
+case "$(uname -s)" in
+   Darwin)
+     SUFFIX="darwin_amd64.tar.gz"
+     # SUFFIX="darwin_arm64.tar.gz"
+     ;;
+   Linux)
+     SUFFIX="linux_amd64.tar.gz"
+     # SUFFIX="linux_arm64.tar.gz"
+     ;;
+   CYGWIN*|MINGW32*|MSYS*|MINGW*)
+     SUFFIX="linux_amd64.tar.gz"
+     ;;
+   # Add here more strings to compare
+   # See correspondence table at the bottom of this answer
+   *)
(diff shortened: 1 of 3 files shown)


austinlparker:
I like this idea! I kinda wonder if we should devote some energy towards getting a homebrew package or whatever for the collector.

cartermp:
Yeah, this is a great idea. Not only because it's a succinct way to get started, but it can be used to model other scenarios for acquiring a collector and running it.

svrnm:
Thank you both:)

> I kinda wonder if we should devote some energy towards getting a homebrew package or whatever for the collector.

I was thinking about homebrew as well but didn't get a hook into it so I started with this ;-) I see that https://github.com/open-telemetry/opentelemetry-collector-releases/issues mentions `brew` in their README.md but I don't see if there's a package already, maybe worth opening an issue

> Not only because it's a succinct way to get started, but it can be used to model other scenarios for acquiring a collector and running it.

True, there could also be a detection for deb/rpm systems and then auto installing those. I was also thinking about a way to pack some of the "reference configs" with it making it even easier to get started (e.g. the simple one that just outputs all data to the console you have in your PRs @cartermp)

svrnm:
> Would it not make more sense to have these scripts reside under a code (or release) repo, such as under one of the following (in order of decreasing preference, IMHO):

I put this on the website because my goal was to have a simple URL for people to get the installer, similar to other solutions doing that like sdkman: `curl -s "https://get.sdkman.io" | bash`

Altough I just checked homebrew, and nvm.sh:

`/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"`

and

`curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.1/install.sh | bash`

So maybe we can achieve something similar through https://github.com/open-telemetry/opentelemetry-collector-releases/tree/main/scripts, e.g.:

`/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/open-telemetry/opentelemetry-collector-releases/HEAD/scripts/install.sh)"`

Let me raise that issue there as well :-)

PR #1343, opened 2022-05-12
Fix a hyperlink in dotnet-instrumentation-first-beta.md

pellared:
Add a missing`[`

PR #1347, opened 2022-05-16
Fix issue in .NET automatic instrumentation announcement

zacharycmontoya:
The markdown for the beta release link is not written correctly, this fixes the issue.


svrnm:
Duplicate of #1343

PR #1356, opened 2022-05-17
Add Instana to the list of vendors supporting OpenTelemetry

cedricziel:


chalin (approved):
LGTM

Note that "The OpenTelemetry support in Instana is in **technology preview**", emphasis mine. (Is this ok with you @austinlparker and @tedsuo? I'll merge anyway, but welcome your feedback on this point.)

cedricziel:
We're graduating it to the next phase right now and it's not going away :)

Issue #1320, opened 2022-04-26
.NET has a local Getting-started page and an external reference


chalin:
_Originally posted by @chalin in https://github.com/open-telemetry/opentelemetry.io/pull/1319#discussion_r859032031_

> All: I have a question unrelated to the changes in this PR, but relevant to this page. Given that this section of the website (i.e., .NET instrumentation) has a [Getting started page](https://opentelemetry.io/docs/instrumentation/net/getting-started/), can we drop the **Getting Started** entry below (lines 35-36)?

- [x] #1364

chalin:
@reyang wrote:

> I can see the problem here - the [getting started page](https://opentelemetry.io/docs/instrumentation/net/getting-started/) is only talking about traces, it does not cover logs and metrics. And the doc wasn't reviewed by any .NET approver/maintainer.

So I believe that you're suggesting that the (site-local) **Getting started** page be updated? If so, @cartermp is this something that you have time for?

reyang:
> So I believe that you're suggesting that the (site-local) Getting started page be updated? If so, @cartermp is this something that you have time for?

Yes. And moving forward we might need some systematic way of keeping these docs up-to-date.

Couple things to consider:
* Do we require language maintainers/approvers to review/approve the change for each language?
* Do we expect the maintainers to keep the content fresh here once they release new versions of packages?
* Do we want to have contents in both places (language repo and here) or it should be one single place?

cartermp:
Happy to add sections for metrics and logs - it's pretty simple to add - Some additional thoughts:

We don't have a precedent for how to recommend which signal to start with. I believe traces are the best signal to use to diagnose a system, and new users should start there. This is the _de facto_ guidance anyways for the most part, since metrics and logs are not as mature as tracing across all of OpenTelemetry. But it's something we're going to want to think about outside the scope of .NET docs, perhaps with .NET docs leading the charge on how to actually organize this content.

As for these questions, here's my opinion:

> Do we require language maintainers/approvers to review/approve the change for each language?

We should definitely set up .NET approvers as a code owner for the .NET docs here. We do this for other languages too, so it's easy to do on this repo's side.

The biggest question then is if any changes should _require_ a .NET approver to actually merge. This hasn't been a problem in other languages so far, but I anticipate at some point that PRs could linger if we do require this, and if required approvers don't have the time to look over docs changes as well, we could hold things up unnecessarily.

[...]

PR #1364, opened 2022-05-18
Add .NET approvers to CODEOWNERS

chalin:
- Contributes to #1320

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -17,6 +17,7 @@
 
 # content owners
 content/en/docs/instrumentation/cpp/     @open-telemetry/docs-approvers @open-telemetry/cpp-maintainers
+content/en/docs/instrumentation/net/     @open-telemetry/docs-approvers @open-telemetry/dotnet-approvers @open-telemetry/dotnet-instrumentation-approvers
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -17,6 +17,7 @@
 
 # content owners
 content/en/docs/instrumentation/cpp/     @open-telemetry/docs-approvers @open-telemetry/cpp-maintainers
+content/en/docs/instrumentation/net/     @open-telemetry/docs-approvers @open-telemetry/dotnet-approvers @open-telemetry/dotnet-instrumentation-approvers
```
chalin, line 20: @austinlparker @reyang: are you ok with the addition of these teams, or would you prefer that I instead add the `-maintainers` teams? 

If you agree with the use of these teams, I'll address the access errors reported by GitHub.

reyang: I'm okay +1 👍

reyang (approved):
LGTM, thanks @chalin!

.github/CODEOWNERS
reyang, line 20: minor: consider putting `net` after `java` (alphabetic sorting)

chalin: @austinlparker +1'd out-of-band

chalin: Done

chalin:
Updated permissions of the two newly added teams. GH checks are green:

> image

Issue #799, opened 2021-10-01
Drop the banner announcing the 1.0 tracing specification?


chalin:
Given that the spec is now at 1.7 (#798), it seems reasonable to drop the 1.0 announcement banner?
WDYT? @tigrannajaryan @austinlparker @tedsuo?

tigrannajaryan:
It may be worth keeping for a while longer, since 1.0 Tracing GA was an important milestone. Perhaps we can replace it by 1.0 Metrics GA when it happens, hopefully soon.

@open-telemetry/governance-committee what do you think?

yurishkuro:
I don't mind the banner, but perhaps it should say GA rather than 1.0.

Below the fold there is a more out of place note that should be changed:

> OpenTelemetry is in beta across several languages and is suitable for use. We anticipate general availability soon.

lizthegrey:
agree, the latter thing is the more important thing to change

chalin:
> Below the fold there is a more out of place note that should be changed:
> 
> > OpenTelemetry is in beta across several languages and is suitable for use. We anticipate general availability soon.

@yurishkuro @lizthegrey: what would you like it changed to, or should those two sentences simply be dropped?

lizthegrey:
> > Below the fold there is a more out of place note that should be changed:
> > > OpenTelemetry is in beta across several languages and is suitable for use. We anticipate general availability soon.
> 
> @yurishkuro @lizthegrey: what would you like it changed to, or should those two sentences simply be dropped?

I'd say "OpenTelemetry is widely deployed, including at [...]" (pull list from ADOPTERS.MD)

yurishkuro:
> pull list from ADOPTERS.MD

We should be careful about this, it creates unnecessary friction of who gets to be on the homepage. Just linking to ADOPTERS is ok, but I often see ADOPTERS list quickly growing stale.

lizthegrey:
> > pull list from ADOPTERS.MD
> 
> We should be careful about this, it creates unnecessary friction of who gets to be on the homepage. Just linking to ADOPTERS is ok, but I often see ADOPTERS list quickly growing stale.

sure thing, how about saying "widely adopted" and link to ADOPTERS without naming specific names?

chalin:
- FYI, proposed banner change for metrics RCs: #1376

chalin:
Closed by #1376

PR #1376, opened 2022-05-20
update banner for metrics rc

austinlparker:
update announce banner to point to metrics RC blog


chalin:
- Closes #799

layouts/partials/banner.md
```diff
@@ -1,9 +1,9 @@
 {{ if .Params.show_banner -}}
 <div class="o-banner">
 
-The OpenTelemetry tracing specification is now 1.0!
+OpenTelemetry Metrics Release Candidates are now available!
 [Learn more][]
 
-[Learn more]: https://medium.com/opentelemetry/opentelemetry-specification-v1-0-0-tracing-edition-72dd08936978
+[Learn more]: https://opentelemetry.io/blog/2022/metrics-announcement/
```
chalin, line 7: Should be a local path -- we want to avoid "external" links into the website itself.

layouts/partials/banner.md
```diff
@@ -1,9 +1,9 @@
 {{ if .Params.show_banner -}}
 <div class="o-banner">
 
-The OpenTelemetry tracing specification is now 1.0!
+OpenTelemetry Metrics Release Candidates are now available!
```
chalin, line 4: I suggest sentence case here:

> OpenTelemetry metrics release candidates are now available

chalin (commented):
See inline comments (let me know if you'd like for me to make those updates). Otherwise, LGTM.

layouts/partials/banner.md
austinlparker: Hm, I tried using the {{< relref >}} directive and the build failed. What's the right way to do relative links from this component?

chalin: No need for relref (and I've been thinking of removing relref usage from the site pages, I can explain why at the next Comms meeting), just use the path `/blog/2022/metrics-announcement/`.

layouts/partials/banner.md
chalin: Maybe "metrics release candidates" can be in bold?

layouts/partials/banner.md
svrnm: It has been like that before, so only as optional change: There are 2 "Learn More" on that page, although web accessibility suggest to [ensure that links with the same accessible name serve a similar purpose](https://dequeuniversity.com/rules/axe/4.2/identical-links-same-purpose)

chalin:
@austinlparker - I think that you are working from an old version of the repo. You'll need to rebase first if you want build checks to pass.

austinlparker:
should be g2g now

chalin (approved):
LGTM, thanks!

Issue #1373, opened 2022-05-19
Update status of Logs OTLP and Logs Data model to stable


tigrannajaryan:
- OTLP Logs are now Stable: https://github.com/open-telemetry/opentelemetry-proto#maturity-level
- Log Data model is also Stable: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md

---
Edit(@chalin):

- [x] Update Logs note "The data model is experimental and released as part of the OTLP protocol" - #1379
- [x] Wait for https://github.com/open-telemetry/opentelemetry-specification/pull/2565
- [x] After the next official release of the spec (>v1.11.0), update this repo's spec git-submodule to a commit including the fix of https://github.com/open-telemetry/opentelemetry-specification/pull/2565
- Optional items: see below --> moved to #1492

Also, while we're making updates to the **Status** page:

- Add link into the spec's baggage section from the Baggage header (as is done for Metrics) once the spec is updated (since the spec's Baggage section has a landing page now). --> see #1492
- Make similar updates to the other sections (e.g., Logging) if those spec sections have landing pages too. --> see #1492

chalin:
1. The OTLP-logs status is automatically extracted from the spec (currently at v1.11.0). But even in the spec `@HEAD`, the status is still showing as `beta`, see https://github.com/open-telemetry/opentelemetry-specification/blame/main/specification/protocol/otlp.md#L57:

    > * Logs: **Beta**

    So an issue would need to be open over the spec. Would you like for me to create one?

2. Where does the "log data model" status show up in the status page?

tigrannajaryan:
> 1. So an issue would need to be open over the spec. Would you like for me to create one?

Created: https://github.com/open-telemetry/opentelemetry-specification/pull/2565

tigrannajaryan:
> 2\. Where does the "log data model" status show up in the status page?

See "The data model is experimental and released as part of the OTLP protocol.":

![image](https://user-images.githubusercontent.com/4194920/169578563-8829a4ab-89bd-4505-97bd-c2c6f463aff8.png)

chalin:
@tigrannajaryan - how about if I update that bullet to read the following instead:

> - The **data model** is released as part of the OTLP protocol

tigrannajaryan:
@chalin sounds good, thanks.

chalin:
Closed by #1464
Optional items moved to #1492

PR #1379, opened 2022-05-20
Status: add logs data-model status

chalin:
- Contributes to #1373

/cc @tedsuo @austinlparker

### DIFF
--- a/content/en/status.md
+++ b/content/en/status.md
@@ -108,9 +108,10 @@ same as the **Protocol** status.
 - **API:** draft
 - **SDK:** draft
 - {{% spec_status "Protocol" "protocol/otlp" "Logs" %}}
+  - {{% spec_status "Data model" "logs/data-model" %}}
 - Notes:
   - OpenTelemetry Logging is currently under active development.
-  - The data model is experimental and released as part of the OTLP protocol.
+  - The [logs data model][] is released as part of the OTLP protocol.
   - Log processing for many data formats has been added to the Collector, thanks
     to the donation of Stanza to the the OpenTelemetry project.
   - Log appenders are currently under develop in many languages. Log appenders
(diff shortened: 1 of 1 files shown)


content/en/status.md
```diff
@@ -108,9 +108,10 @@ same as the **Protocol** status.
 - **API:** draft
 - **SDK:** draft
 - {{% spec_status "Protocol" "protocol/otlp" "Logs" %}}
+  - {{% spec_status "Data model" "logs/data-model" %}}
```
chalin, line 111: Two notes here:

- By using `spec_status` the data-model status will get automatically extracted from the data-model spec page.
- I've put the **Data model** status entry as a subentry of **Protocol** since, as the note below explains, it "is released as part of the OTLP protocol". Does that work for you @tigrannajaryan? If not, I can put it at the top level. I'm open to other suggestions too.

content/en/status.md
```diff
@@ -108,9 +108,10 @@ same as the **Protocol** status.
...
 - Notes:
   - OpenTelemetry Logging is currently under active development.
-  - The data model is experimental and released as part of the OTLP protocol.
+  - The [logs data model][] is released as part of the OTLP protocol.
```
chalin, line 114: @tigrannajaryan - hmm, maybe we should drop "protocol" like this?

> - The logs ... part of the OTLP.

content/en/status.md
tigrannajaryan: We don't have this for other signals, although metrics has a data model page. Perhaps remove for logs as well?

chalin: It's your call. I can remove it. This was just a suggestion.

tigrannajaryan: The text is fine either way. My concern is that it is inconsistent: we talk about logs data model, but not about data models of traces and metrics. We should either do it for all 3 signals or delete this one.
I have a slight preference to delete, since I don't think it adds much value.

chalin: +1 to delete then

chalin:
Changes done. PTAL @tigrannajaryan

tigrannajaryan (approved):

cartermp (approved):

chalin:
Thanks for the reviews

PR #1381, opened 2022-05-21
Add SigNoz in the vendor list

pranay01:
I am not very clear on what does `Distribution` field mean. I have kept it as `Yes` as product similar to ours has this as `Yes`. 
Please advise on what does it mean and if our entry should be changed


cartermp:
I've interpreted "distribution" to mean having built SDK distributions or a collector distribution (but readily acknowledge some entries on the list could be inaccurate with respect to this).

pranay01:
Got it. We have added functionalities on top of opentelemetry, like exporters and processors - https://github.com/SigNoz/opentelemetry-collector-contrib

So, I think `Yes` for "distribution" should be accurate.

cartermp:
Sounds good!

cartermp (approved):

svrnm (approved):
lgtm

austinlparker (approved):

PR #1389, opened 2022-05-25
Update tracing content acknowledgement

cartermp:
The content in the gist was written by Alexandria Pigram as a part of work with Honeycomb. The gist was a way to "bootstrap" getting the content somewhere so we could utilize it here, but now we should do the proper acknowledgement.


cartermp:
something about my own branches being protected and can't commit back into them, closing

PR #1401, opened 2022-05-27
Suggestions for observability primer

svrnm:
Building on top of @avillela's addition of the "observability primer", two suggestions:

# 1 otel concepts => data sources
I found it confusing to have navigation item "OTel concepts" below "concepts" on the OpenTelemetry website. From the content it looks like an highly improved version of what "data sources" contained, so I suggest merging those two pages.   

I also did a few cosmetic changes on the same page (add some bold, add some uppercase).

A general question: should words like Span, Span Status, Span Link etc always be title case or not?

# 2 remove "unknown unknowns"
Sorry for being philosophical on that one, but I have my issue with the "unknown unknowns": I don't see how OpenTelemetry can help with catching issues of that category. If I add spans,logs,metrics at a place where I assume that something _might_ go wrong those are "known unknowns" (which are still novel problems). Unknown unknowns are issues I didn't even expect to have, so I can not plan for them. I *might* capture them or get a good approximation with high chance.

Similar issue with "properly instrumented": how should I know when my code is properly instrumented without assuming all the issues I might want to troubleshoot in the future? Maybe I get it right, but someone in an upstream service sends me some unexpected input and now all my assumptions are wrong.

So, I suggested to remove these words from the document.


cartermp:
Re: unknown unknowns

Philosophically the purpose of instrumenting things and capturing rich data about your application (and not just timings or errors) is so that you can react to novel problems (unknown unknowns) without having to instrument and redeploy your application. Things like "these user IDs and service names and service versions and [...] are most correlated with this new spike in latency" is the philosophical realm of unknown-unknowns. Moreover it is terminology that's been adopted (and not just by my employer!) that I think is valuable to keep.

cartermp:
> A general question: should words like Span, Span Status, Span Link etc always be title case or not?

Probably, yeah

svrnm:
re: unknown unknowns

@cartermp, thanks for going with me into that philosophical discussion :-)

I think it's important to think this through: OpenTelemetry is a great solution, that solves a lot of problems and for newcomers it should be clear what it can do and what it can not do. 

I also acknowledge that many vendors have adopted that terminology, but (a) do I not know if anyone has validated correctness and (b) is opentelemetry.io going to be the source of truth for this statement, so I prefer having it validated, before we have a closed loop of self-affirmation.

Using one of many definitions from the [wikipedia page on that topic](https://en.wikipedia.org/wiki/There_are_known_knowns) (emphasis are mine):

> 1. known unknowns (expected or foreseeable conditions), which can be **reasonably anticipated** but not quantified based on past experience as exemplified by case histories (in Appendix A) and 2. Unknown unknowns (unexpected or unforeseeable conditions), which pose a potentially greater risk simply because they **cannot be anticipated** based on past experience or investigation. Known unknowns result from recognized but poorly understood phenomena. On the other hand, unknown unknowns are phenomena which cannot be expected because there has been no prior experience or theoretical basis for expecting the phenomena.

If I take your example of a spike in latency that correlates with user IDs, service names & versions, this sounds like a condition that can be **reasonably anticipated but not quantified** beforehand (the correlation happens after the fact), which is by that definition a "known unknown".

cartermp:
I'd say this is just a matter of where you're looking.

From the perspective of "what am I looking at to tell me that a problem exists", yes, you instrument from _some_ perspective so that you can see things like if latency is spiking or errors have gone up.

From the perspective of "what is causing the problem I am seeing", then it's all about unknown-unknowns. The whole observability vs. APM spiel is an realization that you cannot rely on past measures or patterns to solve novel problems (a known-unknown).

svrnm:
> From the perspective of "what is causing the problem I am seeing", then it's all about unknown-unknowns. The whole observability vs. APM spiel is an realization that you cannot rely on past measures or patterns to solve novel problems (a known-unknown).

I start to understand that there's a need to distinguish between the things you could solve in the past and the things you now can solve additionally with observability (and/or OpenTelemetry). For me personally calling this "known unknowns" vs "unknown unknown" does obviously not match with my understanding of those terms, but since I can not come up with something better I will not continue complaining about it ;-) 

I think my concern narrows down to the fact, that the _handle novel problems (i.e. “unknown unknowns”)_ is not going into details what does novel problems are, and how existing solutions might not be sufficient. 

So an extension (or probably an additional page) on that would be helpful. I know there's lot of writing on this topic already from different vendors, that we can use as a starting point, but I think it would be good to have something comprehensive in the OpenTelemetry doc as well.

PR #1435, opened 2022-06-10
Simplify and make js manual doc more correct

cartermp:
Preview: https://deploy-preview-1435--opentelemetry.netlify.app/docs/instrumentation/js/instrumentation/

This change does a few things:

* Change `SimpleSpanProcessor` to `BatchSpanProcessor`. Got some feedback that this code is more likely to be copied into a production app and people probably shouldn't be using `SimpleSpanProcessor` for that
* Changes some of the code samples to be a little more clear (I think) in how they're laid out, with simpler code comments
* Changes other code samples to just have less code - some were combining multiple concepts like creating a child span _and_ showing how to set an attribute, and it's confusing to show both
* Formats the document to match other doc formatting here

### DIFF
--- a/content/en/docs/instrumentation/js/instrumentation.md
+++ b/content/en/docs/instrumentation/js/instrumentation.md
@@ -3,7 +3,9 @@ title: Instrumentation
 weight: 3
 ---
 
-This guide will cover creating and annotating spans, creating and annotating metrics, how to pass context, and a guide to automatic instrumentation for JavaScript. This simple example works in the browser as well as with Node.js
+This guide will cover creating and annotating spans, creating and annotating
+metrics, how to pass context, and a guide to automatic instrumentation for
+JavaScript. This simple example works in the browser as well as with Node.js
 
 ## Example Application
 
@@ -25,29 +27,32 @@ function doWork() {
 
 ## Initializing a Tracer
 
-As you have learned in the previous [Getting Started](../getting-started) guide you
-need a TracerProvider and an Exporter. Install the dependencies and add them to
-the head of your application code to get started:
+As you have learned in the previous [Getting Started](../getting-started) guide
+you need a TracerProvider and an Exporter. Install the dependencies and add them
+to the head of your application code to get started:
 
 ```shell
 npm install @opentelemetry/api
 npm install @opentelemetry/sdk-trace-base
 ```
 
-Next, initialize a tracer, preferably in a separate file (e.g., `instrumentation-setup.js`):
+Next, initialize a tracer, preferably in a separate file (e.g.,
+`instrumentation-setup.js`):
 
 ```javascript
 const {
   BasicTracerProvider,
   ConsoleSpanExporter,
-  SimpleSpanProcessor,
+  BatchSpanProcessor,
 } = require('@opentelemetry/sdk-trace-base');
 const opentelemetry = require('@opentelemetry/api');
 
 const provider = new BasicTracerProvider();
 
 // Configure span processor to send spans to the exporter
-provider.addSpanProcessor(new SimpleSpanProcessor(new ConsoleSpanExporter()));
+const exporter = new ConsoleSpanExporter()
+const processor = new BatchSpanProcessor(exporter)
+provider.addSpanProcessor(processor);
 provider.register();
 
 // This is what we'll access in all instrumentation code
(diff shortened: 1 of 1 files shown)


cartermp:
@open-telemetry/javascript-approvers it looks like you're not listed as code owners for these docs - would you like to be added? I intend on making more docs changes here soon and it'd be good to get your eyes/opinions.

content/en/docs/instrumentation/js/instrumentation.md
```diff
@@ -25,29 +27,32 @@ function doWork() {
 
 ## Initializing a Tracer
 
-As you have learned in the previous [Getting Started](../getting-started) guide you
-need a TracerProvider and an Exporter. Install the dependencies and add them to
-the head of your application code to get started:
+As you have learned in the previous [Getting Started](../getting-started) guide
```
dyladan, line 30: User may not have read the previous guide.

content/en/docs/instrumentation/js/instrumentation.md
```diff
@@ -25,29 +27,32 @@ function doWork() {
...
-Next, initialize a tracer, preferably in a separate file (e.g., `instrumentation-setup.js`):
+Next, initialize a tracer, preferably in a separate file (e.g.,
+`instrumentation-setup.js`):
 
 ```javascript
 const {
   BasicTracerProvider,
   ConsoleSpanExporter,
-  SimpleSpanProcessor,
+  BatchSpanProcessor,
```
dyladan, line 46: Batch span processor makes sense in some cases (typical server app) but not in others (serverless). It might make sense to add a comment about this because a serverless container may be killed before a batch is exported.

content/en/docs/instrumentation/js/instrumentation.md
```diff
@@ -56,9 +61,11 @@ export const tracer = opentelemetry.trace.getTracer(
 );
 ```
 
-This registers a tracer provider with the OpenTelemetry API as the global tracer provider, and exports a tracer instance that you can use to create spans.
+This registers a tracer provider with the OpenTelemetry API as the global tracer
+provider, and exports a tracer instance that you can use to create spans.
 
-If you do not register a global tracer provider, any instrumentation calls will be a no-op, so this is important to do!
+If you do not register a global tracer provider, any instrumentation calls will
+be a no-op, so this is important to do!
```
dyladan, line 68: We want to encourage using the API and not using SDK objects directly.

```suggestion
This registers a global `TracerProvider` with the OpenTelemetry API, which may be used to create spans.
If no global `TracerProvider` is registered, instrumentations which use the API will use a no-op tracer and no spans will be created or exported.
```

content/en/docs/instrumentation/js/instrumentation.md
```diff
@@ -93,25 +100,30 @@ Run your application and you will see traces being exported to the console:
...
-/* ... */
+  // Be sure to end the parent span!
+  parentSpan.end();
+}
 
-function doWork(parent) {
-  // Start another span. In this example, the main function already started a
-  // span, so that'll be the parent span, and this will be a child span.
+function doWork(parent, i) {
+  // To set a child span, we need to mark the parent span as the active span
```
dyladan, line 120: Not obvious what "set a child span" means for those not already very familiar with tracing.

content/en/docs/instrumentation/js/instrumentation.md
```diff
@@ -188,27 +192,26 @@ const { SemanticAttributes } = require('@opentelemetry/semantic-conventions');
...
+  // Make sure to end the span if you're done with it!
   span.end();
 }
 ```
 
 ## Span events
 
-An event is a human-readable message attached to a span that represents "something happening" during its lifetime. You can think of it like a primitive log.
+An event is a human-readable message attached to a span that represents
+"something happening" during its lifetime. You can think of it like a primitive
```
dyladan, line 213: IMO "something happening" is too general. I prefer to think of events as an activity that is tracked as a single point in time rather than a start time and and end time. For example, a page load may take time and be represented by a span, but the moment a page becomes interactive is a singular point in time and may be represented as a span event.

content/en/docs/instrumentation/js/instrumentation.md
```diff
@@ -252,37 +255,31 @@ function someFunction(spanToLinkFrom) {
...
   }
-
-  span.end();
 }
+
+span.end();
 ```
 
+By default, the status for all spans is `Unset`. In rare cases, you may also
+wish to set the status to `Ok`. This should generally not be necessary, though.
```
dyladan, line 277: Might be a good idea to explain why this is not necessary?

cartermp:
Superseded by https://github.com/open-telemetry/opentelemetry.io/pull/1445 - will pull feedback over

PR #1444, opened 2022-06-14
Add python manual instrumentation docs to include metrics

aabmass:
Tracking in https://github.com/open-telemetry/opentelemetry-python/issues/2732

This updates the manual instrumentation documentation for python with metrics instrumentation.

### DIFF
--- a/content/en/docs/instrumentation/python/manual.md
+++ b/content/en/docs/instrumentation/python/manual.md
@@ -341,3 +341,16 @@ meter.create_observable_gauge(
     description="The active config version for each configuration",
 )
 ```
+
+## Additional References
+
+- Trace
+    - [Trace Concepts](/docs/concepts/signals/traces/)
+    - [Trace Specification](/docs/reference/specification/overview/#tracing-signal)
+    - [Python Trace API Documentation](https://opentelemetry-python.readthedocs.io/en/latest/api/trace.html)
+    - [Python Trace SDK Documentation](https://opentelemetry-python.readthedocs.io/en/latest/sdk/trace.html)
+- Metrics
+    - [Metrics Concepts](/docs/concepts/signals/metrics/)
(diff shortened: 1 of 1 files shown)


content/en/docs/instrumentation/python/manual.md
```diff
@@ -246,3 +271,73 @@ set_global_textmap(B3Format())
 ```
 
 Note that environment variables will override what's configured in code.
+
+## Metrics
```
cartermp, line 275: Maybe it's worth linking to the language-agnostic document here as well? https://opentelemetry.io/docs/concepts/signals/metrics/

There's basically nothing there right now but that should change in the coming months.

cartermp (approved):
Seems good to me! I think eventually we'll want more subtopics for metrics so that every subconcept is covered and has a copy-pasteable code sample. But this is probably good enough for now.

srikanthccv (approved):

content/en/docs/instrumentation/python/manual.md
```diff
@@ -246,3 +271,86 @@ set_global_textmap(B3Format())
...
+    # count the work being doing
+    work_counter.add(1, {"work.type": work_item.work_type})
+    print("doing some work...")
+```
+
+### Creating and using asynchronous instruments
+
+[Asynchronous instruments](/docs/reference/specification/metrics/api/#synchronous-and-asynchronous-instruments)
+give the user a way to register callback functions, which are invoked on demand to generate
+take measurements. This is useful to periodically measure a value that cannot be instrumented
```
aabmass, line 309:
```suggestion
give the user a way to register callback functions, which are invoked on demand to make
measurements. This is useful to periodically measure a value that cannot be instrumented
```

austinlparker (approved):

Issue #1436, opened 2022-06-11
Which JS SDK should be used in manual instrumentation docs?


cartermp:
The manual instrumentation docs today use `sdk-trace-base`: https://opentelemetry.io/docs/instrumentation/js/instrumentation/

This is helpful insofar as the code here is useful both in Node and in the browser.

However, `sdk-trace-base` doesn't propagate context, and so creating nested spans is kind of awful: https://opentelemetry.io/docs/instrumentation/js/instrumentation/#create-nested-spans

Contrast this with when using the Node SDK (code is from a basic express app):

```js
function doot() {
    tracer.startActiveSpan('doot', span => {
        console.log('doot');
        span.end();
    });
}

app.get('/', (_req, _res) => {
    tracer.startActiveSpan('toot', span => {
        _res.send("toot");
        doot();
        span.end();
    })
});
```

No needing to set the current span in context, get the new context, and use that awful overload to create a nested span. Just creating a span like any other language automatically gives you that parent-child relationship. Nice!

But not everyone reading this document is going to be a Node developer, so the code to initialize tracing, such that you can just use the OTel APIs happily, is not applicable for everyone.

**My proposal** would be to create this content twice on this page: https://opentelemetry.io/docs/instrumentation/js/instrumentation/#initializing-a-tracer

One for node, and one for browser JS. This could be as tabbed content, although we may want to use tabs later on to show JS vs. TS code. I'm fine with whatever folks think works best. But mainly, I'd like to not document the _hard_ way to create nested spans.

@open-telemetry/javascript-approvers what do you think?

austinlparker:
Let's split them out. I think it'd make sense to actually break node and browser into separate files entirely?

cartermp:
So I think that the majority of content can stay the same for both SDKs, including nested spans. Both propagate context (at least within the same process!) and so creating nested spans is easy peasy.

Maybe this can be the layout?

### Initialize tracing
words
#### Node.js
words and code sample
#### Browser
words and code sample
### Create spans
...
etc.

And then maybe a separate doc or section about when you're using neither, which requires you to manually put spans into context if you want a parent-child relationship.

dharesign:
I was reading the docs, which state:
> Don't bother using `startActiveSpan` because it won't do this for you.
and looking into why ended up here as the issue which prompted this addition to the docs.

Why doesn't `startActiveSpan` work?  It seems to call the same methods as the example in the docs calls manually.

[...]

PR #1445, opened 2022-06-14
Reorg manual tracing JS docs based on SDK

cartermp:
This is built on https://github.com/open-telemetry/opentelemetry.io/pull/1435 out of convenience, but I'm unopinionated on which order to pull in things.

Fixes https://github.com/open-telemetry/opentelemetry.io/issues/1436

I had to kind of reorg the manual instrumentation docs a bit for JS here. I also decided to pull in some text from other docs to clarify some sections, and linkify some concepts as well now that we have language-agnostic concepts.

Happy to break stuff out if that helps with review.

Since y'all aren't code owners yet, tagging @open-telemetry/javascript-approvers to take a look. I have no doubt that there may be a few technical inaccuracies here that I'd like to catch before merging 🙂 

---

Preview: https://deploy-preview-1445--opentelemetry.netlify.app/docs/instrumentation/js/instrumentation/

### DIFF
--- a/content/en/docs/instrumentation/js/instrumentation.md
+++ b/content/en/docs/instrumentation/js/instrumentation.md
@@ -50,7 +50,9 @@ Next, ensure that `tracing.js` is required in your node invocation. For example:
 node --require './tracing.js' <other-app-file.js>
 ```
 
-**It is required to `--require` your tracing initialization code to trace.**
+Although it's not technically required to `--require` the tracing initialization
+for only manual instrumentation, it is necessary when you use automatic
+instrumentation. So in practice, it's still likely to be necessary.
 
 ### Browser
 
@@ -59,8 +61,7 @@ First, ensure you've got the right packages:
 ```shell
 npm install \
   @opentelemetry/api \
-  @opentelemetry/sdk-trace-web \
-  @opentelemetry/context-zone
+  @opentelemetry/sdk-trace-web
 ```
 
 Create a `tracing.js` file that initialized the Web SDK, creates a
@@ -69,16 +70,10 @@ Create a `tracing.js` file that initialized the Web SDK, creates a
 ```javascript
 import opentelemetry from "@opentelemetry/api";
 import { WebTracerProvider } from '@opentelemetry/sdk-trace-web';
-import { ZoneContextManager } from '@opentelemetry/context-zone';
 
 const provider = new WebTracerProvider();
 
-provider.register({
-  // Changing default contextManager to use ZoneContextManager - supports asynchronous operations - optional
-  contextManager: new ZoneContextManager(),
-});
-
-// Optionally register some instrumentations
+// Optionally register automatic instrumentation libraries
 registerInstrumentations({
   instrumentations: [],
 });
--- a/content/en/docs/instrumentation/js/instrumentation.md
+++ b/content/en/docs/instrumentation/js/instrumentation.md
@@ -13,6 +13,10 @@ you'll need to have an initialized
 [`TracerProvider`](/docs/concepts/signals/traces/#tracer-provider) that will let
 you create a [`Tracer`](/docs/concepts/signals/traces/#tracer).
 
+If a `TracerProvider` is not created, either explicitly or via SDK APIs, then
+the OpenTelemetry APIs for tracing will use a no-op implementation and fail to
+generate data.
+
 ### Node.js
 
 To initialize tracing with the Node.js SDK, first ensure you have the SDK
@@ -24,15 +28,18 @@ npm install \
   @opentelemetry/sdk-node
 ```
 
-It's recommended to have a separate `tracing.js` file that has all the SDK
-initialization code in it:
+Next, create a separate `tracing.js` file that has all the SDK initialization
+code in it:
 
 ```js
 import opentelemetry from "@opentelemetry/api";
 import nodeOTel from "@opentelemetry/sdk-node";
+import { ConsoleSpanExporter } from '@opentelemetry/sdk-trace-base';
+
+const exporter = new ConsoleSpanExporter();
 
 const sdk = new nodeOTel.NodeSDK({
-  traceExporter: new opentelemetry.tracing.ConsoleSpanExporter(),
+  traceExporter: exporter,
 });
 
 sdk.start()
@@ -43,6 +50,9 @@ export const tracer = opentelemetry.trace.getTracer(
   'name-of-tracer-goes-here'
 );
 ```
+
+Under the covers, this creates a `TracerProvider` that uses the
+`BatchSpanProcessor` to process data in batches asynchronously.
 
 Next, ensure that `tracing.js` is required in your node invocation. For example:
 
@@ -70,8 +80,12 @@ Create a `tracing.js` file that initialized the Web SDK, creates a
 ```javascript
 import opentelemetry from "@opentelemetry/api";
 import { WebTracerProvider } from '@opentelemetry/sdk-trace-web';
+import { ConsoleSpanExporter, BatchSpanProcessor } from '@opentelemetry/sdk-trace-base';
 
 const provider = new WebTracerProvider();
+const exporter = new ConsoleSpanExporter();
+const processor = new BatchSpanProcessor(exporter);
+provider.addSpanProcessor(processor);
 
 // Optionally register automatic instrumentation libraries
 registerInstrumentations({
@@ -85,6 +99,14 @@ export const tracer = opentelemetry.trace.getTracer(
 
 You'll need to bundle this file with your web application to be able to use
 tracing throughought the rest of your web application.
+
+### Picking the right span processor
+
+By default, the Node SDK uses the `BatchSpanProcessor`, and this span processor is also chosen in the Web SDK example. The `BatchSpanProcessor` processes spans in batches before they are exported. This is usually the right processor to use for an application.
+
+In contrast, the `SimpleSpanProcessor` processes spans as they are created. This means that if you create 5 spans, each will be processed an exported before the next span is created in code. This can be helpful in scenarios where you do not want to risk losing a batch, or if you're experimenting with OpenTelemetry in development. However, it also comes with potentially signficant overhead, especially if spans are being exported over a network - each time a call to create a span is made, it would be processed and sent over a network before your app's execution could continue.
+
+In most cases, stick with `BatchSpanProcessor` over `SimpleSpanProcessor`.
 
 ## Create spans
 
@@ -246,15 +268,13 @@ const doWork = () => {
 
 A [Span Event](/docs/concepts/signals/traces/#span-events) is a human-readable
 message on an [`Span`](/docs/concepts/signals/traces/#spans-in-opentelemetry)
-that represents "something happening" during its lifetime. You can think of it
-like a primitive log.
+that represents a discrete event with no duration that can be tracked by a
+single time stamp. You can think of it like a primitive log.
 
 ```js
 span.addEvent('Doing something');
 
 const result = doWork();
-
-span.addEvent('Did something');
 ```
 
 You can also create Span Events with additiona
@@ -388,7 +408,7 @@ const mainWork = () => {
 }
 
 function doWork(parent, i) {
-  // To set a child span, we need to mark the parent span as the active span
+  // To create a child span, we need to mark the current (parent) span as the active span
   // in the context, then use the resulting context to create a child span.
   const ctx = opentelemetry.trace.setSpan(
     opentelemetry.context.active(),
--- a/content/en/docs/instrumentation/js/instrumentation.md
+++ b/content/en/docs/instrumentation/js/instrumentation.md
@@ -24,7 +24,10 @@ package and OpenTelemetry API installed:
 ```shell
 npm install \
   @opentelemetry/api \
-  @opentelemetry/sdk-trace-node
+  @opentelemetry/resources \
+  @opentelemetry/semantic-conventions \
+  @opentelemetry/sdk-trace-node \
+  @opentelemetry/instrumentation
 ```
 
 Next, create a separate `tracing.js` file that has all the SDK initialization
@@ -32,11 +35,23 @@ code in it:
 
 ```js
 const opentelemetry = require("@opentelemetry/api");
+const { Resource } = require("@opentelemetry/resources");
+const { SemanticResourceAttributes } = require("@opentelemetry/semantic-conventions");
 const { NodeTracerProvider } = require("@opentelemetry/sdk-trace-node");
 const { registerInstrumentations } = require("@opentelemetry/instrumentation");
 const { ConsoleSpanExporter, BatchSpanProcessor } = require("@opentelemetry/sdk-trace-base");
 
-const provider = new NodeTracerProvider();
+const resource =
+  Resource.default().merge(
+    new Resource({
+      [SemanticResourceAttributes.SERVICE_NAME]: "service-name-here",
+      [SemanticResourceAttributes.SERVICE_VERSION]: "0.1.0",
+    })
+  )
+
+const provider = new NodeTracerProvider({
+    resource: resource,
+});
 const exporter = new ConsoleSpanExporter();
 const processor = new BatchSpanProcessor(exporter);
 provider.addSpanProcessor(processor);
@@ -47,13 +62,10 @@ registerInstrumentations({
 });
 
 provider.register();
-
-module.exports.tracer = opentelemetry.trace.getTracer(
-  'name-of-tracer-goes-here'
-);
 ```
 
-Next, ensure that `tracing.js` is required in your node invocation. For example:
+Next, ensure that `tracing.js` is required in your node invocation. This is also
+required if you're registering automatic instrumentation libraries. For example:
 
 ```
 node --require './tracing.js' <app-file.js>
@@ -66,6 +78,8 @@ First, ensure you've got the right packages:
 ```shell
 npm install \
   @opentelemetry/api \
+  @opentelemetry/resources \
+  @opentelemetry/semantic-conventions \
   @opentelemetry/sdk-trace-web \
   @opentelemetry/instrumentation
 ```
@@ -75,11 +89,23 @@ Create a `tracing.js` file that initialized the Web SDK, creates a
 
 ```javascript
 const opentelemetry = require("@opentelemetry/api");
+const { Resource } = require("@opentelemetry/resources");
+const { SemanticResourceAttributes } = require("@opentelemetry/semantic-conventions");
 const { WebTracerProvider } = require("@opentelemetry/sdk-trace-web");
 const { registerInstrumentations } = require("@opentelemetry/instrumentation");
 const { ConsoleSpanExporter, BatchSpanProcessor } = require("@opentelemetry/sdk-trace-base");
 
-const provider = new WebTracerProvider();
+const resource =
+  Resource.default().merge(
+    new Resource({
+      [SemanticResourceAttributes.SERVICE_NAME]: "service-name-here",
+      [SemanticResourceAttributes.SERVICE_VERSION]: "0.1.0",
+    })
+  )
+
+const provider = new WebTracerProvider({
+    resource: resource,
+});
 const exporter = new ConsoleSpanExporter();
 const processor = new BatchSpanProcessor(exporter);
 provider.addSpanProcessor(processor);
@@ -90,10 +116,6 @@ registerInstrumentations({
 });
 
 provider.register();
-
-module.exports.tracer = opentelemetry.trace.getTracer(
-  'name-of-tracer-goes-here'
-);
 ```
 
 You'll need to bundle this file with your web application to be able to use
@@ -116,6 +138,28 @@ create a span is made, it would be processed and sent over a network before your
 app's execution could continue.
 
 In most cases, stick with `BatchSpanProcessor` over `SimpleSpanProcessor`.
+
+## Acquiring a tracer
+
+Anywhere in your application where you write manual tracing code should call
+`getTracer` to acquire a tracer. For example:
+
+```js
+const opentelemetry = require("@opentelemetry/api");
+//...
+
+
+const tracer = opentelemetry.trace.getTracer(
+  'my-service-tracer'
+);
+
+// You can now use a 'tracer' to do tracing!
+```
+
+It's generally recommended to call `getTracer` in your app when you need it
+rather than exporting the `tracer` instance to the rest of your app. This
+helps avoid trickier application load issues when other required dependencies are
+involved.
 
 ## Create spans
 
@@ -371,7 +415,7 @@ SDK. The biggest difference, aside from initialization code, is that you'll have
 to manually set spans as active in the current context to be able to create
 nested spans.
 
-### Initilizing tracing
+### Initilizing tracing with `sdk-trace-base`
 
 Initializing tracing is similar to how you'd do it with Node.js or the Web SDK.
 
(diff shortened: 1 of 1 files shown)


content/en/docs/instrumentation/js/instrumentation.md
```diff
@@ -3,135 +3,185 @@ title: Instrumentation
...
+Next, ensure that `tracing.js` is required in your node invocation. For example:
+
+```
+node --require './tracing.js' <other-app-file.js>
+```
 
-As you have learned in the previous [Getting Started](../getting-started) guide you
-need a TracerProvider and an Exporter. Install the dependencies and add them to
-the head of your application code to get started:
+**It is required to `--require` your tracing initialization code to trace.**
```
vmarchaud, line 53: I think i would reword this to take into account two things:
- This only concern CJS module (even though we don't support right now ESM but it will come)
- This is only required if you want auto-instrumentations because they monkey-patch 3rd party modules

content/en/docs/instrumentation/js/instrumentation.md
```diff
@@ -3,135 +3,185 @@ title: Instrumentation
...
+import { ZoneContextManager } from '@opentelemetry/context-zone';
 
-const provider = new BasicTracerProvider();
+const provider = new WebTracerProvider();
 
-// Configure span processor to send spans to the exporter
-provider.addSpanProcessor(new SimpleSpanProcessor(new ConsoleSpanExporter()));
-provider.register();
+provider.register({
+  // Changing default contextManager to use ZoneContextManager - supports asynchronous operations - optional
```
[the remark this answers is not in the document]
vmarchaud: Yes, users will get the most value by using the ZoneContextManager however i think its important to explain that it won't support apps that compiles to >= ES6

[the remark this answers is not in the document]
cartermp: Gotcha. I'm just going to remove this part for now. Maybe it can be handled in a more in-depth way later.

content/en/docs/instrumentation/js/instrumentation.md
```diff
@@ -3,135 +3,185 @@ title: Instrumentation
...
-## Initializing a Tracer
+It's recommended to have a separate `tracing.js` file that has all the SDK
+initialization code in it:
+
+```js
+import opentelemetry from "@opentelemetry/api";
+import nodeOTel from "@opentelemetry/sdk-node";
+
+const sdk = new nodeOTel.NodeSDK({
+  traceExporter: new opentelemetry.tracing.ConsoleSpanExporter(),
```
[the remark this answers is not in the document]
cartermp: For performance-sensitive scenarios, I would assume it's better to use sdk-trace-base?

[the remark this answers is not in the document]
vmarchaud: Indeed but i think most users still want to use the node-sdk because of the auto instrumentation, i was just saying that we might want to document that it has a performance impact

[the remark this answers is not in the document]
cartermp: Are there figures somewhere around what the rough overhead might look like? I'm hesitant to say there's a performance impact without an idea of what kind of impact.

dyladan:
I added comments on #1435 but reviewing two documents which both touch a lot of things is difficult to know which PR the comment "belongs" on. Some of those comments may be addressed here?

cartermp:
Thanks @dyladan - and sorry for the confusion. I'll close #1435 and bring your feedback over to address

content/en/docs/instrumentation/js/instrumentation.md
```diff
@@ -3,135 +3,180 @@ title: Instrumentation
...
-  SimpleSpanProcessor,
-} = require('@opentelemetry/sdk-trace-base');
-const opentelemetry = require('@opentelemetry/api');
+import opentelemetry from "@opentelemetry/api";
+import { WebTracerProvider } from '@opentelemetry/sdk-trace-web';
 
-const provider = new BasicTracerProvider();
+const provider = new WebTracerProvider();
 
-// Configure span processor to send spans to the exporter
```
cartermp, line 49: (needs to be brought back in likely)

> Batch span processor makes sense in some cases (typical server app) but not in others (serverless). It might make sense to add a comment about this because a serverless container may be killed before a batch is exported.

content/en/docs/instrumentation/js/instrumentation.md
```diff
@@ -3,135 +3,180 @@ title: Instrumentation
...
 export const tracer = opentelemetry.trace.getTracer(
-  'example-basic-tracer-node'
+  'name-of-tracer-goes-here'
 );
 ```
 
-This registers a tracer provider with the OpenTelemetry API as the global tracer provider, and exports a tracer instance that you can use to create spans.
-
-If you do not register a global tracer provider, any instrumentation calls will be a no-op, so this is important to do!
+You'll need to bundle this file with your web application to be able to use
```
cartermp, line 86: > This registers a global `TracerProvider` with the OpenTelemetry API, which may be used to create spans.
If no global `TracerProvider` is registered, instrumentations which use the API will use a no-op tracer and no spans will be created or exported.

I'll work out a way to incorporate this.

content/en/docs/instrumentation/js/instrumentation.md
```diff
@@ -292,3 +334,83 @@ try {
...
+  for (let i = 0; i < 3; i += 1) {
+    doWork(parentSpan, i);
+  }
+
+  // Be sure to end the parent span!
+  parentSpan.end();
+}
+
+function doWork(parent, i) {
+  // To set a child span, we need to mark the parent span as the active span
```
cartermp, line 391: From @dyladan:

> Not obvious what "set a child span" means for those not already very familiar with tracing.

content/en/docs/instrumentation/js/instrumentation.md
```diff
@@ -188,27 +227,27 @@ const { SemanticAttributes } = require('@opentelemetry/semantic-conventions');
...
+
+  // Make sure to end the span if you're done with it!
   span.end();
 }
 ```
 
 ## Span events
 
-An event is a human-readable message attached to a span that represents "something happening" during its lifetime. You can think of it like a primitive log.
+A [Span Event](/docs/concepts/signals/traces/#span-events) is a human-readable
```
cartermp, line 247: > IMO "something happening" is too general. I prefer to think of events as an activity that is tracked as a single point in time rather than a start time and and end time. For example, a page load may take time and be represented by a span, but the moment a page becomes interactive is a singular point in time and may be represented as a span event.

content/en/docs/instrumentation/js/instrumentation.md
```diff
@@ -252,37 +295,36 @@ function someFunction(spanToLinkFrom) {
...
+    });
   }
-
-  span.end();
 }
+
+span.end();
 ```
 
+By default, the status for all spans is `Unset` rather than `Ok`. It is
```
cartermp, line 319: > Might be a good idea to explain why this is not necessary?

content/en/docs/instrumentation/js/instrumentation.md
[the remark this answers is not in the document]
Flarna: The overhead depend on the application. If it is using promises/await a lot and has mostly no CPU intensive tasks overhead is higher. Therefore it's impossible to provide any numbers.

> For performance-sensitive scenarios, I would assume it's better to use sdk-trace-base?

Not that instrumentations depend on a working context manager as they get current active span from context. Therefore if one decides to improve performance by using sdk-trace-base they have to be aware that most instrumentations wont work or the user has to take care to implement a context manager.

content/en/docs/instrumentation/js/instrumentation.md
cartermp: I will reword this a bit and then open a separate PR that gives a motivating wordy example in the span event section in language-agnostic conceptual docs.

content/en/docs/instrumentation/js/instrumentation.md
```diff
@@ -252,37 +315,36 @@ function someFunction(spanToLinkFrom) {
...
+typically the job of another component in your telemetry pipeline to interpret
+the `Unset` status of a span, so it's best not to override this unless you're
```
legendecas, line 341: This statement sounds to me like suggesting users to not set the status code OK. Am I understanding this correctly?

legendecas (commented):
Nice work! Just a few comments.

content/en/docs/instrumentation/js/instrumentation.md
```diff
@@ -3,135 +3,202 @@ title: Instrumentation
...
 
-In the following this guide will use the following sample app:
+To start [tracing](/docs/concepts/signals/traces/#tracing-in-opentelemetry),
+you'll need to have an initialized
+[`TracerProvider`](/docs/concepts/signals/traces/#tracer-provider) that will let
+you create a [`Tracer`](/docs/concepts/signals/traces/#tracer).
 
-```javascript
-'use strict';
+If a `TracerProvider` is not created, either explicitly or via SDK APIs, then
```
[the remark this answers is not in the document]
dyladan: I agree

content/en/docs/instrumentation/js/instrumentation.md
```diff
@@ -3,135 +3,202 @@ title: Instrumentation
...
-  // simulate some random work.
-  for (let i = 0; i <= Math.floor(Math.random() * 40000000); i += 1) {}
-}
+To initialize tracing with the Node.js SDK, first ensure you have the SDK
+package and OpenTelemetry API installed:
+
+```shell
+npm install \
+  @opentelemetry/api \
+  @opentelemetry/sdk-node
```
dyladan, line 28: sdk-node is an experimental package. I wouldn't recommend it to be the official website documentation without a disclaimer about that. I would also not remove the documentation which uses the stable packages

content/en/docs/instrumentation/js/instrumentation.md
```diff
@@ -3,135 +3,202 @@ title: Instrumentation
...
+Next, create a separate `tracing.js` file that has all the SDK initialization
+code in it:
 
-As you have learned in the previous [Getting Started](../getting-started) guide you
-need a TracerProvider and an Exporter. Install the dependencies and add them to
-the head of your application code to get started:
+```js
+import opentelemetry from "@opentelemetry/api";
+import nodeOTel from "@opentelemetry/sdk-node";
+import { ConsoleSpanExporter } from '@opentelemetry/sdk-trace-base';
```
dyladan, line 37: importing from a package you did not ask the user to install

content/en/docs/instrumentation/js/instrumentation.md
```diff
@@ -3,135 +3,202 @@ title: Instrumentation
...
-```shell
-npm install @opentelemetry/api
-npm install @opentelemetry/sdk-trace-base
+const exporter = new ConsoleSpanExporter();
+
+const sdk = new nodeOTel.NodeSDK({
+  traceExporter: exporter,
+});
+
+sdk.start()
```
dyladan, line 45: important to know that as of right now the `sdk.start()` method is async and it is possible that if a span is created in the next milliseconds it may not be initialized early enough to catch it.

dyladan (commented):
Thanks for consolidating to a single PR

content/en/docs/instrumentation/js/instrumentation.md
cartermp: A quirk to this is that the getting-started docs use this SDK as of this PR: https://github.com/open-telemetry/opentelemetry.io/pull/697

Should that get reverted as well?

content/en/docs/instrumentation/js/instrumentation.md
cartermp: It's pulled in transitively - should that be called out explicitly? I'd have the same question for `@opentelemetry/sdk-trace-node` should that be used here instead.

content/en/docs/instrumentation/js/instrumentation.md
cartermp: Yep. It's at least my understanding that unless you're an instrumentation library author, you shouldn't set the status to `Ok`, but leave it as `Unset` for a processor or backend to interpret.

content/en/docs/instrumentation/js/instrumentation.md
dyladan: It is the easiest way to get started so I'm not sure. It isn't likely to cause problems, but it is very likely to have changes which break its interface which users may not appreciate if it is presented as the preferred solution. It might be better to have it in a different documentation section explicitly marked as experimental.

cartermp: I'll stick with `sdk-trace-node` for now and leave a note about how you can optionally use the `sdk-node` experimental package. When that package is no longer experimental we can update docs here.

cartermp:
Okay, should be good for (hopefully final?) review. I moved the tracing init stuff to CJS modules. It seems to be the least headache to use right now in the sample apps I'm using as I document stuff. Once ESM is properly supported in the SDK we'll probably want to figure out a way to show both options and instructions for running.

content/en/docs/instrumentation/js/instrumentation.md
```diff
@@ -3,135 +3,211 @@ title: Instrumentation
...
+});
+
+provider.register();
+
+module.exports.tracer = opentelemetry.trace.getTracer(
+  'name-of-tracer-goes-here'
+);
+```
+
+Next, ensure that `tracing.js` is required in your node invocation. For example:
```
austinlparker, line 56: Would it be worth a callout here to note that `tracing` should be imported before any libraries targeted by auto-instrumentation libraries?

content/en/docs/instrumentation/js/instrumentation.md
```diff
@@ -188,37 +258,36 @@ const { SemanticAttributes } = require('@opentelemetry/semantic-conventions');
...
 ```js
 span.addEvent('Doing something');
 
 const result = doWork();
-
-span.addEvent('Did something');
 ```
 
-You can also add an object with more data to go along with the message:
+You can also create Span Events with additiona
```
austinlparker, line 289:
```suggestion
You can also create Span Events with additional
```

austinlparker (approved):
lgtm outside of some nits

content/en/docs/instrumentation/js/instrumentation.md
gareththackeray: Could you elaborate on why that is?  Wouldn't the application dev know whether the span completed or not?

content/en/docs/instrumentation/js/instrumentation.md
```diff
@@ -3,135 +3,211 @@ title: Instrumentation
...
+To initialize tracing with the Node.js SDK, first ensure you have the SDK
+package and OpenTelemetry API installed:
+
+```shell
+npm install \
+  @opentelemetry/api \
+  @opentelemetry/sdk-trace-node
+```
+
+Next, create a separate `tracing.js` file that has all the SDK initialization
```
svrnm, line 30: I am just wondering why the code in `tracing.js` is separated out from the application itself, what's the goal here? Make it easier to swap out the SDK with a vendor specific one? 

If so, the sample app still needs to load the API to work without the `-r tracing.js` I assume? 

This will also be affected from the issue https://github.com/open-telemetry/opentelemetry-js/issues/2912. 

The sample app is also not lined out in the in the document anymore, which makes it hard for the enduser to follow those steps by simply doing a "copy&paste".

@dyladan you know this probably best, any thoughts?

content/en/docs/instrumentation/js/instrumentation.md
```diff
@@ -3,135 +3,211 @@ title: Instrumentation
...
-  console.log('work...');
-  // simulate some random work.
-  for (let i = 0; i <= Math.floor(Math.random() * 40000000); i += 1) {}
-}
+```
+node --require './tracing.js' <app-file.js>
 ```
 
-## Initializing a Tracer
+### Browser
```
svrnm, line 62: This looks similar to https://opentelemetry.io/docs/instrumentation/js/getting-started/browser/, is the goal to replace that one and have browser/node.js in a single file?

svrnm (changes requested):
Added a few comments, my main concerns are the following:

* by removing the example application it's much harder to follow the flow
* see my comment on `tracing.js`, it's not clear why the code is separated out like for auto instrumentation (where a "don't touch my code" experience is nice). If the goal is to separate SDK & API this should be called out to not confuse the user.

content/en/docs/instrumentation/js/instrumentation.md
```diff
@@ -3,135 +3,211 @@ title: Instrumentation
...
+Next, create a separate `tracing.js` file that has all the SDK initialization
+code in it:
+
+```js
+const opentelemetry = require("@opentelemetry/api");
+const { NodeTracerProvider } = require("@opentelemetry/sdk-trace-node");
+const { registerInstrumentations } = require("@opentelemetry/instrumentation");
+const { ConsoleSpanExporter, BatchSpanProcessor } = require("@opentelemetry/sdk-trace-base");
+
+const provider = new NodeTracerProvider();
```
dyladan, line 39: I would include a resource that has at least a service name since that is a required attribute

content/en/docs/instrumentation/js/instrumentation.md
dyladan: Yes I think it is worth it. That is why we use the --register/-r flag

content/en/docs/instrumentation/js/instrumentation.md
```diff
@@ -3,135 +3,211 @@ title: Instrumentation
...
+provider.addSpanProcessor(processor);
+
+// Optionally register automatic instrumentation libraries
+registerInstrumentations({
+  instrumentations: [],
+});
+
+provider.register();
+
+module.exports.tracer = opentelemetry.trace.getTracer(
```
dyladan, line 51: Instead of exporting the tracer I think it would be better to encourage the user to use api.trace.getTracer. If it is exported from here then the main app.js will need to require this file and we may run into odd application load ordering problems if for example the user requires other dependencies first

content/en/docs/instrumentation/js/instrumentation.md
cartermp: I've seen this as a pretty common thing I've seen, to isolate all initialization stuff (and instrumentation registration) into a separate file and then just use the API in the other files in an app where manual instrumentation is done. I'm fine with changing this; whatever comports most with how you'd initialize opentelemetry in a real-world app.

As for a sample, I could also add it back in, but unlike a getting-started flow this doc tends to be more of a "I know I want to do a thing, let me quickly find that subtopic and let me copy-paste some code into my own app and then tweak it so it compiles/runs as-needed" sort of thing. I don't think having a single sample that you run through from top to bottom is as important here.

content/en/docs/instrumentation/js/instrumentation.md
```diff
@@ -292,3 +363,83 @@ try {
...
+// This is what we'll access in all instrumentation code
+export const tracer = opentelemetry.trace.getTracer(
+  'example-basic-tracer-node'
+);
+```
+
+Like the other examples in this document, this exports a tracer you can use
+throughout the app.
+
+## Creating nested spans
```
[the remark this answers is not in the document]
cartermp: It is not, I'll clarify in the title. Creating nested spans is harder in the non-Node/non-browser SDK case, because you need to manually set the current span in context before creating a nested span when using only `sdk-trace-base`. And we should definitely not make that the default recommendation, since it's the non-majority case and more code.

content/en/docs/instrumentation/js/instrumentation.md
cartermp: The goal is to show both of the major ways to initialize tracing. Since each uses a different SDK, I think it's better to have both sections here.

content/en/docs/instrumentation/js/instrumentation.md
Flarna: One very common pitfall regarding instrumentations is that it's needed to first create/register the OTel instrumentation and then require the corresponding module.

By using a dedicated file which is preloaded (or required at the beginning of of the main script) this works more or less automatic.
If you put the OTel init code into your main script one ends up with a Otel section (require + code) and then again 
some require lines and code. Most people don't like require lines after code sections so they tend to resort it - and break their OTel setup.

content/en/docs/instrumentation/js/instrumentation.md
svrnm: OK. Should we then get rid of the browser getting started?

content/en/docs/instrumentation/js/instrumentation.md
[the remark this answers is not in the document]
svrnm: Ah, makes sense, then the issue is the heading level, it probably should be `###`

content/en/docs/instrumentation/js/instrumentation.md
cartermp: I don't think so, no. The goal here is not to be DRY. I think it's fine to have some duplicated code in different contexts.

chalin:
@svrnm - this good to go?

svrnm:
> @svrnm - this good to go?

There are still 2 things open for me:

- minor: `Creating nested spans with sdk-trace-base` needs to be a level 3 heading
- major: I still think that having a sample app people can use to try things out would be [valuable](https://github.com/open-telemetry/opentelemetry.io/pull/1445#discussion_r906001719), I am not going to die on that hill if I am outvoted, but I'd like to have that clarified.

content/en/docs/instrumentation/js/instrumentation.md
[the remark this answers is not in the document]
cartermp: Thanks, missed this!

dyladan:
I appreciate the value of a sample app, but I think this PR is already pretty large and is already such an improvement over the existing document that it should be accepted and a sample can be taken as a follow up task.

chalin:
> I appreciate the value of a sample app, but I think this PR is already pretty large and is already such an improvement over the existing document that it should be accepted and a sample can be taken as a follow up task.

+1

@svrnm - your first point has been addressed, and I agree iwth @dyladan that further changes should be addresses separately. Might you be willing to approve this PR as is?

content/en/docs/instrumentation/js/instrumentation.md
vmarchaud: > Maybe it's worthwhile do have a section explaining exactly that, i.e. using exactly your words why you might like to fork out otel setup into it's own file and then either require it via -r tracing.js or a `require("tracing.js") statement?

I've made a similar comment on the other PR but i understand that we want to avoid to overwhelm readers with too much "gotcha" to know. However it might be worth to add a Q&A section later on to document these

content/en/docs/instrumentation/js/instrumentation.md
```diff
@@ -3,135 +3,255 @@ title: Instrumentation
...
+
+const provider = new NodeTracerProvider({
+    resource: resource,
+});
+const exporter = new ConsoleSpanExporter();
+const processor = new BatchSpanProcessor(exporter);
+provider.addSpanProcessor(processor);
+
+// Optionally register automatic instrumentation libraries
+registerInstrumentations({
```
vmarchaud, line 60: i would move the `registerInstrumentations` before the initialization of the exporter (L55) since 3rd party exporter might require dependency before the instrumentation can patch it (specially http)

vmarchaud (commented):
overall lgtm, nit pick that would better if people modify the init script later on though

svrnm (approved):

cartermp:
I'd also like to think about what a good sample looks like such that each subsection could be meaningfully applied in a way that doesn't feel arbitrary/strange. It might also be good to settle if we want to do something like that for all the manual instrumentation pages. Currently, none of them work off of a common sample/scenario.

svrnm:
I closed my outstanding conversations as well, so we can move that "sample app" discussion to an issue for now.

chalin (approved):
Thanks @cartermp & all for your efforts!

Issue #1194, opened 2022-03-07
Create and use a config/section parameter for release versions for each language


chalin:
For example, #1193 manually updates the OTel-Java release version. The version number for each language should be provided as a (site-config or section-front-matter) parameter.

chalin:
We'd also want to get the collector version, I suppose.

chalin:
We might even want to tie it into the git-submodule version that is fetched, though that could be handled via a separate PR

PR #1468, opened 2022-06-21
Bump collector version to 0.53.0

svrnm:
Update collector getting started with most recent version.

### DIFF
--- a/content/en/docs/collector/getting-started.md
+++ b/content/en/docs/collector/getting-started.md
@@ -22,12 +22,12 @@ $ git clone git@github.com:open-telemetry/opentelemetry-collector-contrib.git; \
 
 ## Docker
 
-Pull a docker image and run the collector in a container. Replace `0.51.0`
+Pull a docker image and run the collector in a container. Replace `0.53.0`
 with the version of the Collector you wish to run.
 
 ```console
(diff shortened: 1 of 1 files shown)


cartermp (approved):
I wonder if there's a way in hugo to define the version up at the top of the page and then splice it into each section in this page? Not a massive deal but I wonder if it's easy to set up. May not with the code blocks.

svrnm:
> I wonder if there's a way in hugo to define the version up at the top of the page and then splice it into each section in this page? Not a massive deal but I wonder if it's easy to set up. May not with the code blocks.

that's actually a great idea, let me check, I think hugo has something build in here (I saw some `{{% param title %}}` at other pages)

svrnm:
Introduced a variable at the top of the page and use it through out for the version, seems to work well. Can we have it like that, thoughts? @cartermp @chalin @austinlparker ?

cartermp:
Nice, I like it a lot.

chalin:
> Introduced a variable at the top of the page and use it through out for the version, seems to work well. Can we have it like that, thoughts? @cartermp @chalin @austinlparker ?

Yep! Great step towards #1194.

chalin (approved):
🎉

PR #1469, opened 2022-06-21
Use 'latest' url to download java agent.

svrnm:
Minor fix for [Java - Getting Started](https://opentelemetry.io/docs/instrumentation/java/getting-started/), this makes use of [Linking to releases](https://docs.github.com/en/repositories/releasing-projects-on-github/linking-to-releases) to always have the latest version of [opentelemetry-javaagent.jar](https://github.com/open-telemetry/opentelemetry-java-instrumentation/releases/latest/download/opentelemetry-javaagent.jar) downloaded

### DIFF
--- a/content/en/docs/instrumentation/java/getting-started.md
+++ b/content/en/docs/instrumentation/java/getting-started.md
@@ -105,6 +105,6 @@ For more:
 [Get the example code.]: https://grpc.io/docs/languages/java/quickstart/#get-the-example-code
 [Java Quick start example]: https://grpc.io/docs/languages/java/quickstart/
 [manual instrumentation]: ../manual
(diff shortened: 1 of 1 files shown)


mateuszrzeszutek (approved):

cartermp (approved):

trask (approved):
👍

laurit (approved):

chalin (commented):
Since the page contains occurrences of the version ID, we might want to consider another approach: use a (Hugo) variable to hold the "current" Java OTel version. This way, the page text and the downloaded version would stay in sync.

If y'all are satisfied with what is being proposed by this PR, that works for me too

chalin:
This is the issue that's related to what I mentioned above:

- #1194

chalin (approved):
Given all the approvals, I'm going to merge this. I've added a note to #1194 in case we'd like to revert this change.

chalin:
Thanks all!

PR #1288, opened 2022-04-08
Adding a howto for building a trace receiver

rquedas:
Creating a PR for the trace receiver tutorial for further review as suggested by @jpkrohling and @austinlparker

PR #1471, opened 2022-06-22
Fixes and normalizes links in Building a trace receiver

chalin:
- Followup to #1288

Preview: https://deploy-preview-1464--opentelemetry.netlify.app/docs/collector/trace-receiver/

/cc @rquedas


cartermp (approved):

chalin:
@rquedas - the main change here is to change external links into the spec, to local links. You had two kinds of external spec links: some pointed to v1.11.0 and other to main. I'm assuming that both of those kinds can be changed to local links into the spec. I'll merge now since I want to make some followup changes. Your comments are always welcome.

chalin:
Thanks for the review @cartermp!

PR #1482, opened 2022-06-27
Markdown fixes for java docs, remove note on agent config

svrnm:
Some fixes for the java docs from markdownlint 

---

Preview, e.g.:

- https://deploy-preview-1482--opentelemetry.netlify.app/docs/instrumentation/java/automatic/agent-config/

PR #1489, opened 2022-06-28
Java agent config page: drop hard-coded TOC and opening sentence

chalin:
Followup to:

- #1482

The "Please report" is being dropped because, IMHO, it should have been dropped at the same time as the Note (#1482). The "Contents" isn't needed since Hugo generates a TOC automatically.


trask (approved):

cartermp (approved):

svrnm (approved):

mateuszrzeszutek (approved):

chalin:
Thanks all for the approvals.

PR #1496, opened 2022-06-29
Vendor page, improve layout on mobile

### LINKED ISSUE #1196: Vendor page: improve vendor list presentation on mobile
Vendor-page specific items of #1172:

- [ ] The table is too wide on mobile
- [ ] Links in the table are raw URLs. Consider using an icon.

Later we might want to consider generating the vendor list table (or list of cards) using a shortcode from a data file.

chalin:
- Closes #1196

This isn't a perfect solution, but it is IMHO a step in the right direction -- enough so to warrant closing #1196. At some point we might prefer to have the table be generated from data, but we can face that later (if at all).

Preview: https://deploy-preview-1491--opentelemetry.netlify.app/vendors/

---

### Screenshots

**Before** (note that the table doesn't fit _and_ the raw links are line-wrapped in some cases):

> image

**After** (the table still doesn't fit on narrow mobile devices, but it does scroll horizontally, and is much narrower than before)

> image

/cc @austinlparker @cartemp @svrnm


svrnm (approved):
looks good to me!

PR #1500, opened 2022-06-30
Remove outdated statement on metrics status

arminru:
The metrics API and SDK are GA in Java and .NET, and in Python there is an RC out already.


pichlermarc (approved):

sharrmander (approved):
true info!

PR #1505, opened 2022-06-30
Update Python exporter docs with console, OTLP, and Prometheus metric exporters

aabmass:
Part of https://github.com/open-telemetry/opentelemetry-python/issues/2732

I tested out the example code to make sure everything is working as well. I also changed around the OTLP sections to present gRPC as the first option to push folks that way and because there isn't an OTLP/HTTP metric exporter yet.

### DIFF
--- a/content/en/docs/instrumentation/python/_index.md
+++ b/content/en/docs/instrumentation/python/_index.md
@@ -55,7 +55,7 @@ $ pip install opentelemetry-instrumentation-{instrumentation}
 ```
 
 These are for exporter and instrumentation packages respectively. The Jaeger,
-Zipkin, OTLP and OpenCensus Exporters can be found in the
+Zipkin, OTLP, OpenCensus, and Prometheus Exporters can be found in the
 [exporter](https://github.com/open-telemetry/opentelemetry-python/blob/main/exporter/)
 directory of the repository. Instrumentations and additional exporters can be
 found in the [Contrib repo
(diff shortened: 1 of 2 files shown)


cartermp:
Trying to dig it up, but we've had some issues specifically with users doing python+grpc export dur to something finnicky with grpc support in python (I believe it's related to swallowing some errors and thus making things undiagnosable). Not necessarily blocking, but this could mean that for traces we do see more python devs with confusing experiences when something goes wrong; whereas http tends not to be as complex

aabmass:
@cartermp lmk if you find anything. I remember something about this from Slack, but I think it was in regard to Honeycomb's OTLP endpoint. There was also https://github.com/open-telemetry/opentelemetry-python/issues/2594 which is now resolved.

Usability concerns aside, I think gRPC is the recommended way regardless? If we can fix bugs on the Python side to make it more usable, that would be preferable

aabmass:
I'll be OOO for the next few weeks. Feel free to push any suggested changes to my branch and get this merged 🙂

cartermp (approved):
I slept on it for a bit and I think I'm fine with grpc by default here. I couldn't dig up the issue on our end that I've seen, but if I can find it then I'll create an issue and we can discuss. Docs changes look great!

srikanthccv (approved):

chalin (approved):
@cartermp gave a 👍 in the comments.

PR #1510, opened 2022-07-05
(js) Update Collector HTTP Port to 4318

pkanal:
Update the default collector HTTP port to `4318` instead of `55681` since the newest versions of the exporter default to `4318`.


chalin (approved):
Thanks all!

PR #1512, opened 2022-07-05
Improve collector getting started guide with tabs.

svrnm:
A few improvements for the collector Getting Started guide:

* Moved the links to reference links (is that how they are called, I don't know)
* Added some details for docker, docker compose & Mac (I don't have windows, so if anyone can provide the details there, this would be appreciated :-) )
* Added tabs to most of the code blocks so we can provide the same for amd64, arm64 & i386

I was thinking about doing even more tabs with otel/otel-contrib, but that would lead to an explosion, so I'd like to keep this as something worth discussing.


cartermp (approved):
Nice, I love it

PR #1514, opened 2022-07-05
Add images to config for 'Twitter Cards'

svrnm:
- Contributes to #1519

---

Just stumbled upon this, when putting a link to opentelemetry.io outside the main page I don't get an image for my "twitter card":

Screen Shot 2022-07-05 at 18 58 08

It looks like the required tags (especially meta og:image) are only added if there is a featured-background.jpg in the same path or an images attribute for that page. I set a default attribute in the config.yml -- I am still trying to validate if this not breaking anything else but it looks good so far. 

Note: We could use the images parameter for individual pages to show a different image!

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -11,6 +11,7 @@ languages:
     title: OpenTelemetry
     description: The OpenTelemetry Project Site
     languageName: English
+    images: ["/featured-background.jpg"]
     weight: 1
 
 imaging:
--- a/config.yaml
+++ b/config.yaml
@@ -11,6 +11,7 @@ languages:
     title: OpenTelemetry
     description: The OpenTelemetry Project Site
     languageName: English
+    # Social media image (e.g., Open Graph). Update the version tag when the image changes.
     images: ["/featured-background.jpg?v1"]
     weight: 1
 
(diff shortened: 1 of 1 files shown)


svrnm:
See the difference here using the preview at ps://deploy-preview-1514--opentelemetry.netlify.app/docs/collector/custom-collector/:

Screen Shot 2022-07-05 at 19 06 33

cartermp:
I like it! I don't know enough about this stuff to really approve it, but I can give approval for the _output_ 🙂

chalin (changes requested):
Based on my experience with other projects, we need to take particular care in how we set this up. I'll mark this as "changes needed" for now / draft-status until I have time to revisit.

chalin (approved):
No that the image URL has a version tag, I'm ok with this change.

chalin:
Thanks again for noticing the issue and submitting this PR @svrnm.

I've done what I consider the minimal required update to the PR so that we can use it as a base solution moving forward -- that is, I added an image "version tag".

The image that you selected is the same as the one used for the homepage (which makes sense). We might want to revisit the image choice later, but this PR improves on our situation, so I'll merge now.

chalin:
Just to followup on this PR: the strategy I was aiming for in preserving social-media images "forever" is currently difficult to implement -- so the tweak I did to this PR wasn't useful 🤷‍♂️ ... and that means more work is needed 😁. For details, see https://github.com/open-telemetry/opentelemetry.io/issues/1519#issuecomment-1204361628.

PR #1518, opened 2022-07-05
Add Google Analytics

### LINKED ISSUE #1517: Google analytics: switch from Netlify injection to Hugo config
GA is currently enabled via Netlify's snippet injection: https://app.netlify.com/sites/opentelemetry/settings/deploys#snippet-injection.

The **main problem** with this approach is that it injects GA into all pages for all servers, including site previews -- which isn't what we want.

Consider switching over to the Docsy/Hugo support for GA.

- [x] Removed Netlify's GA-injected snippet
- [x] Add the corresponding script to generated site files - #1518

chalin:
- Closes #1517 by adding partials containing the Google-recommended snippet (obtained from the GA app)
- Just before this is merged, I'll remove the corresponding **Netlify snippet injection**
- Adds our own partial to handle GA (UA and GA4). Once I can confirm that it is all working, I'll consider looking into using [Hugo's support for GA].
- Note that the layout includes the GA4 measurement ID but doesn't use it yet.

**Note**: We won't be able to see the change implemented in this PR until its deployed to the production server.

[Hugo's support for GA]: https://gohugo.io/templates/internal/#google-analytics


chalin:
Out-of-band, @austinlparker wrote:

> if the template analytics are better than the injected stuff then we should swap over

So I'm going to merge now so that I can test it during the day today. @austinlparker - your feedback is still welcome.

PR #1531, opened 2022-07-07
Update php docs with requirements and otlp export.

svrnm:
When I tried out the PHP Getting Started I got some errors because of the dependencies on `php-http/async-client-implementation` and `psr/http-factory-implementation`, I added those requirements, restructured the doc and also added a section on OTLP.

Preview: https://deploy-preview-1531--opentelemetry.netlify.app/docs/instrumentation/php/getting-started/

### DIFF
--- a/content/en/docs/instrumentation/php/getting-started.md
+++ b/content/en/docs/instrumentation/php/getting-started.md
@@ -152,7 +152,7 @@ trace:
 }
 ```
 
-# Export to collector
+## Export to collector
 
 The next step is to modify the code to send spans to the collector via OTLP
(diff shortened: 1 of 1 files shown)


svrnm:
> Excellent improvement, I like it a lot. 

😊 

> Any desire to show the sample output from the collector as well, and/or how to set it up?

Not with this PR. I think we need that "set up your sample controller" snippet which then can be re-used across languages.

cartermp (approved):
Looks good enough to merge in to me!

chalin (approved):
Thanks all!

Issue #1400, opened 2022-05-27
Node.js metrics API update needed


bruno-bert:
This line of code does not work:
requestCount.bind
it raises an error "bind is not a function"

cartermp:
The issue here is that the code sample is using an older metrics API that has since changed in the lead up to the release candidate. We should update that section (and likely others) soon.

PR #1534, opened 2022-07-08
JS Docs: Temporarily get rid of metrics content in node getting started

cartermp:
Preview: https://deploy-preview-1534--opentelemetry.netlify.app/docs/instrumentation/js/getting-started/nodejs/

This temporarily resolves https://github.com/open-telemetry/opentelemetry.io/issues/1400 (and I believe another issue where someone pointed this out).

The metrics API is still experimental and appears to be undergoing quite a bit of change. Given that there's still churn, I think we should pull the existing (now older, incompatible) docs until metrics stabilize. I moved what was written to a `.old` file so we can re-use most of it and then update code samples to use up to date APIs.

I also added a small "next steps" blurb since that's a common thing we do in other docs.


legendecas (approved):

svrnm (approved):
lgtm

PR #1434, opened 2022-06-09
Update python Getting Started docs to include metrics

aabmass:
Tracking in https://github.com/open-telemetry/opentelemetry-python/issues/2732

This updates the Getting Started documentation for python with metrics instrumentation and the output of commands with metrics added.

PR #1545, opened 2022-07-12
Update getting-started.md

beekeh:
opentelemetry-instrument --metrics_exporter none --traces_exporter console flask run to avoid error Requested component 'otlp_proto_grpc' not found in entry points for 'opentelemetry metrics exporter'

### DIFF
--- a/content/en/docs/instrumentation/python/getting-started.md
+++ b/content/en/docs/instrumentation/python/getting-started.md
@@ -74,7 +74,7 @@ You can now run your instrumented app with `opentelemetry-instrument` and have
 it print to the console for now:
 
 ```shell
-opentelemetry-instrument --traces_exporter console flask run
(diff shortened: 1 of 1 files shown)


cartermp:
Thanks @beekeh -- this should actually also be resolved with https://github.com/open-telemetry/opentelemetry.io/pull/1434 -- although I think it's a little suspect that if you're _not_ using metrics, you need to still specify something about the metrics exporter when using the agent. I'll file an upstream issue.

@chalin do you want to take this in now or await #1434?

chalin:
> do you want to take this in now or await #1434?

I'd wait on #1434.

cartermp:
I think we can actually close this now that #1434 is merged. It's not the exact same thing, but it addresses the issue.

PR #1559, opened 2022-07-24
Add explanation of the trace sample

melbgirl:
To assist OTel (and/or observability) newcomers to better understand.

### DIFF
--- a/content/en/docs/concepts/signals/traces.md
+++ b/content/en/docs/concepts/signals/traces.md
@@ -1,7 +1,8 @@
 ---
-title: "Traces"
+title: Traces
 description: >-
-  Traces give us the big picture of what happens when a request is made by user or an application.
+  Traces give us the big picture of what happens when a request is made by user
+  or an application.
 weight: 1
 ---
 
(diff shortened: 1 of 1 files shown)


melbgirl:
Perhaps it would make sense to describe spans before traces; eg after the sample trace output, explain what spans are, and then go into what makes up a trace in OTel. The context of what is a span is useful for understanding the trace components.

cartermp:
@melbgirl I think this would be better above the example, yep. Also, would you mind formatting the text to be the same width as the rest? I personally use [Rewrap](https://marketplace.visualstudio.com/items?itemName=stkb.rewrap) in vscode to automate that.

melbgirl:
Done. I think it would also be helpful to add some pseudocode showing how to implement a span link - the textual description is great, but I found it hard to then apply it.

austinlparker:
@cartermp is this g2g?

cartermp:
Yarp!

cartermp (approved):

chalin (approved):
(FYI, I ran the markdown formatter.)

Thanks all!

PR #1563, opened 2022-07-26
Fix JaegerExporter Import

viktorb1:
Missing import despite running `pip install opentelemetry-exporter-jaeger`? `pip install opentelemetry-exportert-jaeger` installs `opentelemetry-exporter-jaeger-thrift-1.11.1` by default?

### DIFF
--- a/content/en/docs/instrumentation/python/exporters.md
+++ b/content/en/docs/instrumentation/python/exporters.md
@@ -160,7 +160,7 @@ Then you can configure the exporter when initializing tracing:
 
 ```python
 from opentelemetry import trace
-from opentelemetry.exporter.jaeger import JaegerExporter
+from opentelemetry.exporter.jaeger.thrift import JaegerExporter
 from opentelemetry.sdk.resources import SERVICE_NAME, Resource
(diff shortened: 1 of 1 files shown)


cartermp (approved):
Thanks, yes, this is correct!

cartermp:
@open-telemetry/python-approvers this is the package used in the sample for the API docs in the package that is installed. However, which _should_ be documented as the default, and/or should both be shown?

* `opentelemetry.exporter.jaeger.thrift`
* `opentelemetry.exporter.jaeger.proto`

Or since Jaeger now supports OTLP directly, should we expand the OTLP exporter section and leave something here for "if you're using an older version of jaeger", or some other caveat?

srikanthccv:
I think both should present since they exist for different purposes.

srikanthccv (approved):

cartermp:
Thanks @srikanthccv - I'll queue up a PR that shows both options after this PR gets merged.

chalin (commented):
Thanks all!

chalin:
> Thanks @srikanthccv - I'll queue up a PR that shows both options after this PR gets merged.

Ok, will merge then.

PR #1564, opened 2022-07-26
Fix parameter warning for set_status

yenchenLiu:


cartermp (approved):
Thanks, changes look good (ran locally in a sample app)

srikanthccv (commented):
This is not really needed going forward and I think you are seeing this warning because the change exists in the RC release but your installed package is stable and <= 1.11.

chalin:
> This is not really needed going forward and I think ...

@srikanthccv - so, we should just close this PR (since the problem is transient)?

srikanthccv:
I don't mind merging this because it is still valid.

srikanthccv (approved):

chalin (approved):
Thanks all!

Issue #934, opened 2021-11-17
Creating Spans for New Processes elixir example is wrong


korsmakolnikov:
**What needs to be changed?**
The elixir example in `instrumentation.md` at `Creating Spans for New Processes` doesn't work. I try to propose the correct example in this pr https://github.com/open-telemetry/opentelemetry.io/pull/933.

cartermp:
#1571 should have this resolved, so closing!

PR #1571, opened 2022-07-29
Update instrumentation.md

rahuldimri:
End span outside task process doesn't work #934

### DIFF
--- a/content/en/docs/instrumentation/erlang/instrumentation.md
+++ b/content/en/docs/instrumentation/erlang/instrumentation.md
@@ -156,15 +156,15 @@ span_ctx = OpenTelemetry.Tracer.start_span(<<"child">>)
 ctx = OpenTelemetry.Ctx.get_current()
 
 task = Task.async(fn ->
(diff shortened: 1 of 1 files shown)


cartermp (approved):
Looks good docs-wise, will tag erlang/elixir approvers

cartermp:
@open-telemetry/erlang-approvers could you all take a look? Also, we should probably add y'all as approvers to this subsection of the docs

ferd (approved):
Yep, that sounds like a correct edit. There's a bit of a semantic difference at the core of the Erlang and Elixir snippets: the task in Elixir is synchronous (due to the `Task.await/1` call) and the one in Erlang is asynchronous, so the old and new Elixir ordering of operations are equivalent.

That being said, things look more uniform this way and both example also look more similar, so this is a useful fix.

chalin (approved):
Thanks all!

chalin:
@cartermp et all: does this close #934? If so, can you close it? Thanks!

tsloughter:
Ah yup, thanks! :+1:

PR #1583, opened 2022-08-01
Clarify exporting via OTLP

### LINKED ISSUE #1578: Python getting started mentions modifying code, but it stays as is
**What needs to be changed?**

https://opentelemetry.io/docs/instrumentation/python/getting-started/#modify-the-code-to-export-spans-and-metrics-via-otlp says "[...]  replace the console exporter with an OTLP exporter:" but the code example below it is the same as above. Either some change is missing or the block should be eliminated.

https://opentelemetry.io/docs/instrumentation/python/getting-started/#modify-the-code-to-export-spans-and-metrics-via-otlp

Pyhton -> Getting started -> Modify code to export spans and metrics via OTLP

pilhuhn:
Remove the part that says one has to modify the code and then shows the exact same code as above. 
This is not needed and confuses the reader.

Fixes #1578

### DIFF
--- a/content/en/docs/instrumentation/python/getting-started.md
+++ b/content/en/docs/instrumentation/python/getting-started.md
@@ -462,8 +462,8 @@ To do this, install the OTLP exporter package:
 pip install opentelemetry-exporter-otlp
 ```
 
-As exporters are logically separated from the trace acquiring, we do not 
-need to make any code changes.  
+The `opentelemetry-instrument` agent will detect the package you just installed
+and default to OTLP export when it's run next.
 
 ### Run the application
 
--- a/content/en/docs/instrumentation/python/getting-started.md
+++ b/content/en/docs/instrumentation/python/getting-started.md
@@ -449,11 +449,11 @@ docker run -p 4317:4317 \
     --config=/etc/otel-collector-config.yaml
 ```
 
-You will now have an collector instance running locally, listening on port 4318.
+You will now have an collector instance running locally, listening on port 4317.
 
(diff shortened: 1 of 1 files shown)


srikanthccv (approved):

content/en/docs/instrumentation/python/getting-started.md
```diff
@@ -462,40 +462,8 @@ To do this, install the OTLP exporter package:
...
-        res = randint(1, 6)
-        rollspan.set_attribute("roll.value", res)
-        roll_counter.add(1, {"roll.value": res})
-        return res
-```
-
-By default, it will send telemetry to `localhost:4317`, which is what the collector
-is listening on.
+As exporters are logically separated from the trace acquiring, we do not 
+need to make any code changes.
```
cartermp, line 466:
```suggestion
The `opentelemetry-instrument` agent will detect the package you just installed
and default to OTLP export when it's run next.
```
Suggested rewording to this - the statement as-written wouldn't be true if we weren't using the agent, and that may confuse some people if they think it's a statement that would broadly apply.

pilhuhn: I am all for rewording, but as all this is on context of the agent, they would need to make more changes / be surprised anyway.
My text holds (somewhat) true anyway, that the code does not need to be changed when switching the exporter from console to OTLP. Of course as we don't manually instrument, we also don't need to configure the exporter, which would be required for manual mode.

cartermp (approved):

chalin (approved):
Thanks all!

PR #1592, opened 2022-08-03
Call out that spans can be nested in the conceptual docs

cartermp:
It's kinda Spans 101 level stuff, but it's not explicitly called out and I figured that felt missing.

### DIFF
--- a/content/en/docs/concepts/signals/traces.md
+++ b/content/en/docs/concepts/signals/traces.md
@@ -230,10 +230,9 @@ Sample Span:
 }
 ```
 
-As is implied by the present of a parent span ID, spans can be nested as
-child spans that represent sub-operations. This allows them to accurately
-model work done in an application.
-
+Spans can be nested, as is implied by the presence of a parent span ID:
+child spans represent sub-operations. This allows spans to more accurately
+capture the work done in an application.
 ### Span Context
 
 Span Context is an immutable object on every span that contains the following:
--- a/content/en/docs/concepts/signals/traces.md
+++ b/content/en/docs/concepts/signals/traces.md
@@ -187,6 +187,7 @@ work or operation. Spans are the building blocks of Traces. In OpenTelemetry,
 they include the following information:
 
 - Name
+- Parent span ID
 - Start and End Timestamps
 - [Span Context](#span-context)
 - [Attributes](#attributes)
@@ -228,6 +229,10 @@ Sample Span:
   }
 }
 ```
+
+As is implied by the present of a parent span ID, spans can be nested as
+child spans that represent sub-operations. This allows them to accurately
+model work done in an application.
 
 ### Span Context
 


content/en/docs/concepts/signals/traces.md
```diff
@@ -187,6 +187,7 @@ work or operation. Spans are the building blocks of Traces. In OpenTelemetry,
 they include the following information:
 
 - Name
+- Parent span ID
```
chalin, line 190: Do you want to indicate that this is optional (or can be null or empty)?

content/en/docs/concepts/signals/traces.md
```diff
@@ -229,6 +230,10 @@ Sample Span:
 }
 ```
 
+As is implied by the present of a parent span ID, spans can be nested as
+child spans that represent sub-operations. This allows them to accurately
+model work done in an application.
+
```
chalin, line 236:
```suggestion
Spans can be nested, as is implied by the presence of a parent span ID:
child spans represent sub-operations. This allows spans to more accurately
capture the work done in an application.
```

chalin: Hmm, I'm not quite happy with the last sentence (both the original and the revised one).
Could we just drop it? If not, I'll try to think of something better -- something about capturing operation details or about fine-grainedness?

chalin (commented):
See inline comments, thanks.

content/en/docs/concepts/signals/traces.md
cartermp: I'd prefer not to drop it, since it's a crucial description of why they can be nested. Maybe it needs to be expanded into something more descriptive. Since programs are made up of operations, and some of which are nested (A calls B), we need a way to represent that work in diagnostics. The ability for spans to be nested is what accomplishes that.

content/en/docs/concepts/signals/traces.md
cartermp, line 190:
```suggestion
- Parent span ID (can be null or empty, indicating a root span)
```

content/en/docs/concepts/signals/traces.md
chalin: Ok, then let's keep the last sentence. Are you ok with the edits @cartermp?

content/en/docs/concepts/signals/traces.md
chalin: How about

- Parent span ID (for non-root spans)

content/en/docs/concepts/signals/traces.md
cartermp: Yep! For some reaosn github isn't letting me accept the suggestion, though, even after doing what it says 🙃 

image

content/en/docs/concepts/signals/traces.md
cartermp: So even with root spans there is an ID. It's just empty.

chalin: Ok, how about:

```suggestion
- Parent span ID (empty for root spans)
```

content/en/docs/concepts/signals/traces.md
chalin: Maybe you're logged in using a different ID?

cartermp:
Closing in favor of patch from my fork

PR #1593, opened 2022-08-03
Added missing expressions and file execution path

Yukang-LIAN:


content/en/docs/collector/trace-receiver.md
```diff
@@ -66,6 +66,7 @@ with the following command:
 docker run -d --name jaeger \
   -p 16686:16686 \
   -p 14268:14268 \
+  -p 14250:14250 \
```
mx-psi, line 69: Why is exposing this port necessary?

Yukang-LIAN: https://github.com/open-telemetry/opentelemetry.io/blob/f9c3c0beb66fbd71c6323d55b88d5394227bc569/content/en/docs/collector/trace-receiver.md?plain=1#L83-L107

Because this port `14250` is needed by `jaeger` at line 97 of this config file.

mx-psi (approved):

cartermp (approved):

content/en/docs/collector/custom-collector.md
```diff
@@ -133,7 +133,7 @@ All you need now is to let the `ocb` do it's job, so go to your terminal and
 type the following command:
 
 ```cmd
-ocb --config builder-config.yaml
+./ocb --config builder-config.yaml
```
chalin, line 136: Is this path prefix necessary?

chalin (commented):
See inline question. Thanks.

content/en/docs/collector/custom-collector.md
Yukang-LIAN: `ocb` is an executable file. Without this prefix, an error will occur, so I think it's necessary.

chalin: Adding a path prefix like `./` would only be necessary if you are trying to force the use of a local executable `./obc` vs. any `obc` command that might be picked up from your PATH.

I haven't stepped through the instructions in this page, but IMHO, they probably work without the `./` path prefix.

Thoughts @cartermp et all?

Yukang-LIAN: Mr. Chalin, thanks for your patient reply! I'm just a rookie, so it's up to you to decide whether it's necessary or not. At the same time, I am very happy to learn something from your reply.😄

cartermp: If it's not added to the PATH by default, then we should either document it this way or state that you should add it to the PATH. Compare this with other docs with other executables (e.g., package managers) where the executable is automatically ont he path.

chalin: I just went back and re-read from the start of the page and saw that the command is locally installed (via step 1). Appending `./` to the command path is the simplest solution. So this is good to go!

chalin (approved):
Thanks all!

PR #1611, opened 2022-08-11
Add mypy doc

cartermp:
This came up before, but there was some recent conversation here that made me feel like we could do a quick doc explaining how to configure a mypy project: https://github.com/open-telemetry/opentelemetry-python/issues/2591#issuecomment-1206440086

There's mention of potentially needing this in a project configuration file:

```
[mypy-opentelemetry.sdk.*]
implicit_reexport = True
```

I do not know if this is also necessary, but it should be easy to add if it is.

### DIFF
--- a/content/en/docs/instrumentation/python/mypy.md
+++ b/content/en/docs/instrumentation/python/mypy.md
@@ -3,7 +3,8 @@ title: Using mypy
 ---
 
 If you're using [mypy](http://mypy-lang.org/), you'll need to turn on [namespace
-packages](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-namespace-packages), otherwise `mypy` won't be able to run correctly.
+packages](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-namespace-packages),
+otherwise `mypy` won't be able to run correctly.
 
 To turn on namespace packages, do one of the following:
 
@@ -18,4 +19,15 @@ Or, use a command-line switch:
 
 ```shell
 mypy --namespace-packages
+```
+
+## Using the `strict` option
+
+If you're using the `strict` option with mypy, and you're using the
+OpenTelemetry Python SDK (instead of just the API), you'll need to also set the
+following in your project configuration file:
+
+```toml
+[mypy-opentelemetry.sdk.*]
+implicit_reexport = True
 ```
--- a/content/en/docs/instrumentation/python/mypy.md
+++ b/content/en/docs/instrumentation/python/mypy.md
@@ -1,0 +1,23 @@
+---
+title: Using mypy
+weight: 5
+---
+
+If you're using [mypy](http://mypy-lang.org/), you'll need to turn on [namespace
+packages](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-namespace-packages).
+
+In your project configuration file, add the following:
+
+```toml
+[tool.mypy]
+namespace_packages = true
+```
+
+Or if you'd prefer a command-line switch, use `--namespace-packages`:
+
+```console
+mypy --namespace-packages
+```
+
+If you don't turn on namespace packages, then mypy won't be able to correctly
+run.
--- a/content/en/docs/instrumentation/python/mypy.md
+++ b/content/en/docs/instrumentation/python/mypy.md
@@ -1,6 +1,5 @@
 ---
 title: Using mypy
-weight: 5
 ---
 
 If you're using [mypy](http://mypy-lang.org/), you'll need to turn on [namespace
@@ -8,7 +7,7 @@ packages](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy
 
 To turn on namespace packages, do one of the following:
 
-- Add the following to your project configuration file:
+Add the following to your project configuration file:
 
 ```toml
 [tool.mypy]
@@ -17,9 +16,6 @@ namespace_packages = true
 
 Or, use a command-line switch:
 
-```console
+```shell
 mypy --namespace-packages
 ```
-
-If you don't turn on namespace packages, then mypy won't be able to correctly
-run.
--- a/content/en/docs/instrumentation/python/mypy.md
+++ b/content/en/docs/instrumentation/python/mypy.md
@@ -4,16 +4,18 @@ weight: 5
 ---
 
 If you're using [mypy](http://mypy-lang.org/), you'll need to turn on [namespace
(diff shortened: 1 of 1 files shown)


content/en/docs/instrumentation/python/mypy.md
```diff
@@ -0,0 +1,23 @@
...
+```
+
+Or if you'd prefer a command-line switch, use `--namespace-packages`:
+
+```console
+mypy --namespace-packages
+```
+
+If you don't turn on namespace packages, then mypy won't be able to correctly
+run.
```
chalin, line 23: Given the changes suggested above you can drop this last sentence.

content/en/docs/instrumentation/python/mypy.md
```diff
@@ -0,0 +1,23 @@
...
+In your project configuration file, add the following:
+
+```toml
+[tool.mypy]
+namespace_packages = true
+```
+
+Or if you'd prefer a command-line switch, use `--namespace-packages`:
+
+```console
```
chalin, line 18: Since you're not adding a leading prompt, you can make the block language `shell` (or add `$` as a prompt).

content/en/docs/instrumentation/python/mypy.md
```diff
@@ -0,0 +1,23 @@
+---
+title: Using mypy
+weight: 5
```
chalin, line 3: Should this "note" page be at the end of the Python-section page list?

(If you drop the weight it might migrate to the end of the list.)

content/en/docs/instrumentation/python/mypy.md
```diff
@@ -0,0 +1,23 @@
...
+---
+
+If you're using [mypy](http://mypy-lang.org/), you'll need to turn on [namespace
+packages](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-namespace-packages).
```
chalin, line 7:
```suggestion
If you're using [mypy](http://mypy-lang.org/), you'll need to turn on [namespace
packages](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-namespace-packages), otherwise `mypy` won't be able to run correctly.
```

content/en/docs/instrumentation/python/mypy.md
```diff
@@ -0,0 +1,23 @@
...
+
+In your project configuration file, add the following:
```
chalin, line 9:
```suggestion
To turn on namespace packages, do one of the following:

- Add the following to your project configuration file:
```

If you agree with this change you'll need to indent the toml code block below.

content/en/docs/instrumentation/python/mypy.md
```diff
@@ -0,0 +1,23 @@
...
+
+```toml
+[tool.mypy]
+namespace_packages = true
+```
+
+Or if you'd prefer a command-line switch, use `--namespace-packages`:
```
chalin, line 16:
```suggestion
Or, use a command-line switch:
```

If you agree with this change you'll need to indent the code block below.

chalin (commented):
See inline copyedit comments. Thanks!

chalin:
Broader comment: As it stands the added page is _very_ specific. Might it make sense to add the content to another page (such as the Python landing page), or to generalize the page title? Just a thought, I'm ok if this lands with the current page addition if others are ok with that too.

cartermp:
I'd like to get the @open-telemetry/python-approvers take on this being a specific page. One way to look at mypy is that it's analogous to TypeScript, which is to say it's a dialect of the same core/family of language. And as it turns out, there are some unique things to keep in mind when using it. We have the same thing to consider for the JS docs long-term as well.

sk-:
> There's mention of potentially needing this in a project configuration file:
> 
> ```
> [mypy-opentelemetry.sdk.*]
> implicit_reexport = True
> ```
> 
> I do not know if this is also necessary, but it should be easy to add if it is.

This is absolutely necessary if you are using the `strict` option, because it enables the flag `--no-implicit-reexport`.  

This seems to be required only for the `sdk` because it programmatically constructs the `__all__` variable, as opposed to the `api` where the `__all__` variable is explicitly defined.

Without this option one gets errors like the following:

```
app/otel.py:7: error: Module "opentelemetry.sdk.metrics" does not explicitly export attribute "MeterProvider"; implicit reexport disabled  [attr-defined]
app/otel.py:8: error: Module "opentelemetry.sdk.metrics.export" does not explicitly export attribute "PeriodicExportingMetricReader"; implicit reexport disabled  [attr-defined]
app/otel.py:13: error: Module "opentelemetry.sdk.metrics.view" does not explicitly export attribute "ExplicitBucketHistogramAggregation"; implicit reexport disabled  [attr-defined]
```

content/en/docs/instrumentation/python/mypy.md
```diff
@@ -0,0 +1,25 @@
...
+title: Using mypy
+weight: 5
+---
+
+If you're using [mypy](http://mypy-lang.org/), you'll need to turn on [namespace
+packages](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-namespace-packages), otherwise `mypy` won't be able to run correctly.
+
+To turn on namespace packages, do one of the following:
+
+- Add the following to your project configuration file:
```
svrnm, line 11: Remove the bullet point, or add one below (Or ...)

svrnm (changes requested):

cartermp:
@chalin and @svrnm your feedback should be addressed.

@open-telemetry/python-approvers any feedback on the doc?

chalin (approved):

svrnm (approved):
lgtm

chalin:
@open-telemetry/python-approvers - any feedback on this?

ocelotl:
> @open-telemetry/python-approvers - any feedback on this?

Looks good to me initially. Just a few things:

Is the underlying intention of this documentation to explain how to use `mypy` in a "standalone" way? What I mean is that we use `mypy` as part of our regular testing process, we do this by runnning `tox -e mypy`. I can imagine someone wants to run `mypy` by themselves on OTel Python by I wonder what is the context behind doing so.

If the intent of running `mypy` is to make sure everything is ok in order to submit a PR to collaborate with the project (I imagine that is not the case), then this information should be [here](https://github.com/open-telemetry/opentelemetry-python/blob/main/CONTRIBUTING.md).

Regarding [this](https://github.com/open-telemetry/opentelemetry.io/pull/1611#issuecomment-1212693395) comment, we actually should update the API to also dynamically generate the contents of `__all__` to be consistent with the SDK.

ocelotl (approved):

chalin (approved):
Thanks all!

Issue #1612, opened 2022-08-11
Accessibility add link name


Arhell:
**What happened?**
![Screenshot_2](https://user-images.githubusercontent.com/26163841/184125626-d808dc3d-84c4-4b61-9018-bc5375477821.png)

**What did you expect would happen?**
no errors in accessibility app

/assign

PR #1614, opened 2022-08-11
Remove elipsis from home page for accessibility

cartermp:
fixes https://github.com/open-telemetry/opentelemetry.io/issues/1612

image


svrnm:
an alternative is using an aria-label, not great but might help. not sure if or how well this works with hugo ...

chalin:
> an alternative is using an aria-label

Done, see https://github.com/open-telemetry/opentelemetry.io/pull/1617/files.

PR #1617, opened 2022-08-11
Address accessibility issue of link to instrumentation

chalin:
- I've tested this with a screen reader under Safari and it works as intended, when focus hits the button with the elipsis label, the reader reads "All languages"
- Closes #1612
- If @cartermp agrees, this also closes #1614

Preview: https://deploy-preview-1617--opentelemetry.netlify.app

PR #1619, opened 2022-08-11
Accessibility fix link

Arhell:


cartermp:
@chalin any comments here? I don't see anything wrong with the change, but the link still works today without these changes.

Arhell:
This is small fix after update https://github.com/open-telemetry/opentelemetry.io/pull/1617
need to add this attribute and there will be no errors in the application

chalin (approved):
validator.w3.org seems happy with this change :)
Thanks!

PR #1623, opened 2022-08-16
Homepage: accessibility link attribute update

Arhell:
This attribute fixes the problem but is not appropriate for this block.
changed to a more suitable item
all check

https://github.com/open-telemetry/opentelemetry.io/pull/1619
https://github.com/open-telemetry/opentelemetry.io/pull/1617


chalin (approved):
I haven't tested with a screen reader, but this seems good to me.

PR #1625, opened 2022-08-17
Remove prometheus exporter for JavaScript

svrnm:
Removing the prometheus metric exporter from javascript docs until we can provide an updated & correct version

See https://github.com/open-telemetry/opentelemetry.io/pull/1621#discussion_r947656433 for details, cc: @pichlermarc , @cartermp


pichlermarc (approved):
thanks for removing this :slightly_smiling_face:

chalin (approved):
Thanks all!

Issue #1026, opened 2022-01-06
Gitpod support


cartermp:
Gitpod is a great way to make contributing easy for repositories. It's similar to GitHub Codespaces, but more mature, free for OSS projects, and priced better when you want the more premium features. I've been using it for most OSS projects I help maintain and it's been a joy - no `node_modules` black hole on any of my machines, fresh environments each time I do something, and no complex environments to maintain and twiddle with on my end.

Adding support to this repo would just be a `.gitpod.yml` file with an init command to build and run the docs when a workspace is opened: https://www.gitpod.io/docs/config-gitpod-file

Would it be acceptable to support gitpod for this repo?

chalin:
There's already a link to open this repo in GitPod, see 
https://github.com/open-telemetry/opentelemetry.io/blob/b7a2d4cef70e10c26499232b2f77d95d707252c0/CONTRIBUTING.md?plain=1#L5:

> Follow these steps for local or [cloud-IDE][] (via [Gitpod.io][]) development:
> ...
> [cloud-IDE]: https://gitpod.io/#https://github.com/open-telemetry/opentelemetry.io

I've chosen to not add a custom `.gitpod.yml` file because the default one that is generated works fine.

chalin:
Otherwise I agree, I've had good experiences so far using GitPod.
Can we consider this issue resolved, or is there anything else that you'd want addressed?

cartermp:
Yep, resolved!

chalin:
Reopening: to request that we either add a custom `.gitpod.yml` file or add that file to `.gitignore`

cartermp:
I'd personally be in favor of adding it just so that we can specialize what the init experience is (for example, spawn a second shell for git operations but keep the first process for running `npm run serve` so you can see your changes live as you edit things).

But this isn't a particularly strong opinion.

musingvirtual:
As someone who uses GitPod because I don't do dev work in my current role so I don't have a highly customized local environment, strongly seconding this request that we consider git ignore. 

I have to delete these files literally every single time I make even the most minor change which may be 3-5 times during the process of trying to land a single pull request for site copy etc.  If I do not do it someone will request that I do.

If we can't do git ignore, another possibility that would make it simpler for GitPod users would be to ask reviewers not to request that GitPod committers delete this file midway through, and have a final checkpoint once all other reviews have been passed to remove these files.

chalin:
[...]

PR #1629, opened 2022-08-17
Gitpod minimal config

chalin:
- Addresses #1026, maybe even closes it if @cartermp thinks that it's good enough.

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,0 +1,4 @@
+# Gitpod config. For details, see https://www.gitpod.io/docs/config-gitpod-file.
+
+tasks:
+  - init: npm install && npm run check-links


.gitpod.yml
```diff
@@ -0,0 +1,4 @@
+# Gitpod config. For details, see https://www.gitpod.io/docs/config-gitpod-file.
+
+tasks:
+  - init: npm install && npm run check-links
```
cartermp, line 4: It might be good to also use `npm run serve` so that folks can have a dev version of the site built and running in their environment. But I'm also fine if that's not included.

cartermp (approved):

.gitpod.yml
chalin: My preference would be to first commit this minimal version and then we can experiment and grow it over time. Thanks for the review and approval. Will merge this now as is.

PR #1631, opened 2022-08-18
Update blog archetype

svrnm:
Based on the previous few blog posts I suggest that we update the archetype to

* contain an example for the linked author name
* ask people to use prettier.


cartermp (approved):

chalin (approved):
All: I made some edits (for example to fix the author name syntax error). PTAL and let me know if you're ok with the changes. Otherwise, it LGTM!

chalin:
@svrnm - could you test this updated archetype to ensure that it doesn't generate errors out-of-the-box?

svrnm:
> @svrnm - could you test this updated archetype to ensure that it doesn't generate errors out-of-the-box?

tried it out, looked good to me, no errors.

chalin (approved):
Thanks!

PR #1652, opened 2022-08-23
Update new-end-user-resources.md to mention the OTel in practice talks

musingvirtual:
Adding a quick section about the talks to this post so that everything is a complete circle and we can easily link here to represent the WG's full public activities. The section is copied directly from the other blog post. I'm experimenting with doing this directly in github rather than gitpod so let's see how it goes!

### DIFF
--- a/content/en/blog/2022/new-end-user-resources.md
+++ b/content/en/blog/2022/new-end-user-resources.md
@@ -17,10 +17,10 @@ solutions to achieving ubiquitous observability. Share successes and failures,
 discover best practices, and meet others who are also on a journey to implement
 observability powered by OpenTelemetry.
 
-### OpenTelemetry In Practice Talk Series
+### OpenTelemetry In Practice
 
-Welcome to the **OpenTelemetry in Practice** series! This is a new experiment by
-some OpenTelemetry contributors in the End User Working Group.
+**OpenTelemetry in Practice** is an experimental **series** of talks initiated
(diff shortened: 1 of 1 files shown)


cartermp (approved):

content/en/blog/2022/new-end-user-resources.md
```diff
@@ -17,6 +17,30 @@ solutions to achieving ubiquitous observability. Share successes and failures,
 discover best practices, and meet others who are also on a journey to implement
 observability powered by OpenTelemetry.
 
+### OpenTelemetry In Practice Talk Series
+
+Welcome to the **OpenTelemetry in Practice** series! This is a new experiment by
+some OpenTelemetry contributors in the End User Working Group.
```
chalin, line 23:
```suggestion
### OpenTelemetry In Practice

**OpenTelemetry in Practice** is an experimental **series** of talks initiated
by some members of the End User Working Group.
```

chalin (approved):
See inline suggestion. Otherwise LGTM.

We should decide what to do (if anything) with posts that get edited after they are initially posted. But we have have that conversation outside of the scope of this PR. /cc @svrnm

chalin:
> We should decide what to do (if anything) with posts that get edited after they are initially posted. But we have have that conversation outside of the scope of this PR. /cc @svrnm

Related: https://github.com/google/docsy/issues/1196

Issue #1636, opened 2022-08-19
Address GDPR compliance requirements


chalin:

chalin:
@svrnm @cartermp - how about something like this as a starting point?

> <img width="615" alt="image" src="https://user-images.githubusercontent.com/4140793/186043539-5e22e9d9-0dbd-4b5b-a4e0-53440a925a0a.png">

I'm linking to the [Linux Foundation Cookie policy], which covers the websites of all its projects.

/cc @caniszczyk 

[Linux Foundation Cookie policy]: https://www.linuxfoundation.org/cookies/

chalin:
I'll reply to https://github.com/open-telemetry/opentelemetry.io/pull/1653#issuecomment-1223879691 here.

If I understand correctly, you (@svrnm) would be ok with the following:

- Assume opt-in, like [cncf.io] does
- Support do-not-track for GA and OTel
- Have the dialog be non-modal and a bit more discrete, like [cncf.io]

Is that right?

[cncf.io]: https://cncf.io

svrnm:
> * Assume opt-in, like [cncf.io](https://cncf.io) does

cncf.io neither provides an Opt-In, nor an Opt-Out, it implicitly sets the cookies and only provides an "Accept" button to acknowledge the cookies & tracking. Per my understanding, the following is required:

* Do not set (non-functional) cookies or enable any tracking without the consent of the user (which is equivalent to the user clicking on decline)
* There needs to be an "accept" and a "decline" button
* If and only if the user clicks on accept the tracking & setting the non-functional cookies is allowed
* If the user clicks "decline", it's okay to set a cookie that stores that information, since this can be declared as a functional cookie.

I have no legal expertise on that, so please do not assume that this is correct.

> * Support do-not-track for GA and OTel

As "nice to have", yes.

> * Have the dialog be non-modal and a bit more discrete, like [cncf.io](https://cncf.io)

As "nice to have, yes

chalin:
FYI, @caniszczyk confirmed on 08/24 that we have a thumbs-up to go with the cncf.io approach for now. So I'll codify that as a first step and we can take it from there.

cartermp:
Hmmmm. What if we just didn't use google analytics? I don't personally feel I'll ever need it for the site (users by geo-region doesn't change things, user journeys are nice in theory but I never look at 'em, etc.) And we already capture anonymized requests by http route which is a more accurate count (even if it's a blunt instrument that can be spoofed).

[...]

PR #1653, opened 2022-08-23
Adds cookie- / privacy-policy consent support

chalin:
- Contributes to #1636
- Features:
  - Displays a modal dialog that is centered in the page (works well on desktop and mobile).
  - The only way to dismiss the dialog is by clicking the "accept" button (the green check mark). So, the site cannot be used without a consent.
  - Tab-focus goes to the accept button.
- A first attempt at addressing GDPR requirements. If there is general agreement on moving forward with this I'll make some followup changes such as (most items, other than the first, are mainly a note-to-self):
  - **Only enabling this for production builds. WDYT?**
  - Moving the currently hard-coded styling into our SCSS files, and 
  - Switch to using FA 6 icon names.
  - I might also propose this as an upstream change to Docsy.

Preview:

- Any page from the preview server, such as:
  - Homepage: https://deploy-preview-1653--opentelemetry.netlify.app
  - Docs page, e.g.: https://deploy-preview-1653--opentelemetry.netlify.app/docs/instrumentation/go/
  - Blog page, e.g.: https://deploy-preview-1653--opentelemetry.netlify.app/blog/2022/demo-announcement/
- You can reset the cookie through you browser settings, but also through the JS console using the following commands:
  ```
  Cookies.get('linux-foundation-org-cookie-consent')
  Cookies.set('linux-foundation-org-cookie-consent', '')
  ```
  Or from, say, Chrome DevTools >> Application >> Storage >> Cookies.

/cc @caniszczyk @austinlparker @cartermp

### DIFF
--- a/assets/js/privacy-notice.js
+++ b/assets/js/privacy-notice.js
@@ -1,0 +1,12 @@
+$(function () {
+  const key = 'linux-foundation-org-cookie-consent';
+  const consentValue = 'true'
+
+  if (Cookies.get(key) === consentValue) { return; }
+
+  $('#privacyNotice')
+    .on('hide.bs.modal', function () {
+      Cookies.set(key, consentValue, { sameSite: 'strict', expires: 90 });
+    })
+    .modal('show');
+});
--- a/layouts/partials/hooks/body-end.html
+++ b/layouts/partials/hooks/body-end.html
@@ -1,1 +1,6 @@
 {{ partial "script.html" (dict "src" "js/tracing.js") -}}
+
+<script src="https://cdn.jsdelivr.net/npm/js-cookie@3.0.1/dist/js.cookie.min.js" integrity="sha256-0H3Nuz3aug3afVbUlsu12Puxva3CP4EhJtPExqs54Vg=" crossorigin="anonymous"></script>
+{{ partial "script.html" (dict "src" "js/privacy-notice.js") -}}
+
+{{ partialCached "privacy-notice.html" . -}}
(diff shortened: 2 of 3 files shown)


svrnm:
Thanks for creating this @chalin! I see some discussion here and at #1636, should we have it here or there?

A few thoughts:

(*question*) the only cookie I see is the google analytics cookie. Besides google analytics we are using otel itself for tracking. As far as I see it from other websites there needs(?) to be an option to not allow those cookies (and probably also to disable GA & otel tracking). Actually if I remember correctly there's a requirement to have "Opt-In" being the default.

(*nice-to-have*) Related: I saw some websites recently to start respecting the "Do-Not-Track" header being set, so if we provide a way to disable cookies/tracking, this would be a feature lots of people appreciate.

(*nice-to-have*) I like the way how the cncf.io solves it having it in the bottom right corner. Is it possible to have it that way? It's getting less into the way of the users.

Screen Shot 2022-08-23 at 12 19 35

chalin:
Sure, let's move this broad "how to" discussion to the main issue. See my response here: https://github.com/open-telemetry/opentelemetry.io/issues/1636#issuecomment-1224013483.

cartermp:
My main issue with this implementation is:

> So, the site cannot be used without a consent.

I'd rather we "lose" views in GA than _actually lose_ people because they couldn't decline.

chalin:
As mentioned in the SIG meeting, this is still on my radar, but it seems like we might need to have a broader discussion first?

svrnm:
> As mentioned in the SIG meeting, this is still on my radar, but it seems like we might need to have a broader discussion first?

Yes: 

I agree with @cartermp (_I'd rather we "lose" views in GA than actually lose people because they couldn't decline._), but right now we have nothing at all, and since CNCF/LF "force" us to use google analytics, I would say the solution with only the accept button is better than nothing.

PR #1655, opened 2022-08-23
[bikeshedding] adjusted status page layout

cartermp:
We chatted about this a little bit in the comms SIG, where it was pondered if it was better to have language SDK stuff first. So I did that! And did some color and words changing stuff too. It's a bikeshed that I'm painting.

Preview: https://deploy-preview-1655--opentelemetry.netlify.app/status/

### DIFF
--- a/content/en/status.md
+++ b/content/en/status.md
@@ -9,18 +9,13 @@ description: Maturity-level of the main OpenTelemetry components
 
 ## {{% param title %}}
 
-The development status, or maturity level, of the [main OpenTelemetry
-components][main-comp] are as follows:
+OpenTelemetry is made up of several components, each having different meanings
+for "status". For example, the status of a signal in the specification may not
+be the same as its status in a particular language SDK.
 
+The development status, or maturity level, of the [per-language
+SDKs](/docs/instrumentation/) is as follows:
 
-<div class="l-get-started-buttons">
-
-- [Specification status](/docs/reference/specification/status/)
-- [Collector status](/docs/collector/#status-and-releases)
-
-</div>
-
-The development status, or maturity level, of the [per-language SDKs](/docs/instrumentation/) is as follows:
 <div class="l-primary-buttons mt-5">
 
 - [C++](/docs/instrumentation/cpp/#status-and-releases) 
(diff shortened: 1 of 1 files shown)


content/en/status.md
```diff
@@ -9,19 +9,14 @@ description: Maturity-level of the main OpenTelemetry components
...
+be the same as its status in a particular language SDK.
 
-<div class="l-get-started-buttons">
-
-- [Specification status](/docs/reference/specification/status/)
-- [Collector status](/docs/collector/#status-and-releases)
-
-</div>
+The development status, or maturity level, of the [per-language
+SDKs](/docs/instrumentation/) is as follows:
```
chalin, line 17:
```suggestion
For the development status, or maturity level, of a [language
SDK](/docs/instrumentation/), see the status section of that language:
```

content/en/status.md
```diff
@@ -37,6 +32,15 @@ The development status, or maturity level, of the [per-language SDKs](/docs/inst
 
 </div>
 
+The development status, or maturity level, of the [language-agnostic
+components][main-comp] are as follows:
```
chalin, line 36:
```suggestion
The development status, or maturity level, of the [collector](/docs/collector/) and [specification](/docs/reference/specification/) are as follows:
```

content/en/status.md
```diff
@@ -9,19 +9,14 @@ description: Maturity-level of the main OpenTelemetry components
 
 ## {{% param title %}}
 
-The development status, or maturity level, of the [main OpenTelemetry
-components][main-comp] are as follows:
-
+OpenTelemetry is made up of several components, each having different meanings
+for "status". For example, the status of a signal in the specification may not
+be the same as its status in a particular language SDK.
```
chalin, line 14: Oh! I don't really like the phrase _each having different meanings for "status"._

Hmm, after think about this for a bit, ... how about:

```suggestion
OpenTelemetry is made up of [several components][main-comp], some
language-specific and others language-agnostic. When looking for a status, make
sure to look for the status from the right component page. For example, the
status of a signal in the specification may not be the same as the signal status
in a particular language SDK.
```

chalin (commented):
Much more than bike shedding here IMHO :)

See inline comments. Thanks.

content/en/status.md
```diff
@@ -37,6 +34,14 @@ The development status, or maturity level, of the [per-language SDKs](/docs/inst
 
 </div>
 
+The development status, or maturity level, of the [collector](/docs/collector/) and [specification](/docs/reference/specification/) are as follows:
```
svrnm, line 37: Can we replace this as well with a similar wording as above:

> For the development status, or maturity level, of the [collector](/docs/collector/) and [specification](/dosc/reference/specification/), see the status section pages:

chalin: Hmm, that wording doesn't quite work (the collector has a status _section_, but the spec doesn't -- the entire page is dedicated to its status). How about:

> For the development status, or maturity level, of the [collector](/docs/collector/) and [specification](/docs/reference/specification/), see the following:

cartermp: @chalin I went with your suggestion.

chalin (approved):
🚀

svrnm:
lgtm :)

PR #1659, opened 2022-08-24
Put "Go Web App" blog post before "nginx"

svrnm:
Looking at blog.opentelemetry.io right now the nginx blog post is ahead of the go blog post, although the latter is the most recent one. Moving the date by a day I think we should put it ahead. cc @chalin

### DIFF
--- a/content/en/blog/2022/go-web-app-instrumentation/index.md
+++ b/content/en/blog/2022/go-web-app-instrumentation/index.md
@@ -1,7 +1,7 @@
 ---
 title: Go Web-app Instrumentation
 linkTitle: Go App Instrumentation
-date: 2022-08-22
+date: 2022-08-23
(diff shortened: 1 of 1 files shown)


cartermp (approved):

svrnm:
Actually, the order is right in the left side menu but not in the "Posts in 2022".

chalin:
Do you want this merged now and/or for me to investigate the differences in order first?

chalin:
FYI, I created the following issue to track the problem:

- #1660

chalin (approved):
Let's just merge this and I'll track down the problem separately now that I have a clear repro.

svrnm:
I am ok with merging this. To me this looks like a corner case, how often will we have blog post published at the same day?

PR #1674, opened 2022-08-28
`exporter` ENDPOINT default schema

Petrie:
In [specification/protocol/exporter](https://opentelemetry.io/docs/reference/specification/protocol/exporter/#configuration-options). 
> SDKs SHOULD default endpoint variables to use http scheme unless they have good reasons to choose https scheme for the default (e.g., for backward compatibility reasons in a stable SDK release).

So I think, the example of `ENDPOINT` should use `http`. If not, newcomer maybe confuse with should I use `https` by default.


svrnm:
@cartermp: if I remember correctly you created those docs, can you check, thanks.

cartermp (approved):

PR #1685, opened 2022-08-30
Iconography.zip should contain icons only, not diagrams

### LINKED ISSUE #664: Iconography.zip should contain icons only, not diagrams
It currently contains two diagrams. /cc @jkowall

svrnm:
Fix #664


jkowall:
I am not sure why we need the zip really :)

svrnm:
We currently need it because there is a link to it on the page footer. We can remove that or have it don’t differently but until then we need it, I raised a follow up ticket here:

https://github.com/open-telemetry/opentelemetry.io/issues/1693

PR #1690, opened 2022-08-31
New icons added and removed zip

### LINKED ISSUE #392: Additional icon requests
Loving the icons! I was wondering if a few more could be created:

- gRPC (to go with HTTP)
- service and extension (to go with receivers/processors/exporters). These may be challenging...
- Fluent Bit (to go with Jaeger and Prometheus)

jkowall:
Added new icons per request and cleaned up the old zipfile which is not useful. 

Closes #392 

Signed-off-by: Jonah Kowall <jkowall@gmail.com>


svrnm:
> Added new icons per request

Thanks!

> cleaned up the old zipfile which is not useful.

It is linked on the otel page footer, that's why we have it. I am not really sure if this is needed or if we just point to the folder but should be addressed in another issue/PR.

svrnm (approved):

svrnm:
let me merge and then reupload the zip, in #1685 I have an updated version without the diagrams prepared.

PR #1704, opened 2022-09-02
Changes for jaeger otlp support in .NET

### LINKED ISSUE #1567: Jaeger now supports OTLP directly - how should we document it for the SDKs?
SDKs have all built Jaeger exporters because historically, Jaeger has not supported OTLP. But now they're not technically necessary: https://medium.com/jaegertracing/introducing-native-support-for-opentelemetry-in-jaeger-eb661be8183c

> With this new capability, it is no longer necessary to use the Jaeger exporters with the OpenTelemetry SDKs, or to run the OpenTelemetry Collector in front of the Jaeger backend. Using the OTLP exporter, the SDKs can be configured to send the data directly to the Jaeger backend. The OTLP receiver accepts data via gRPC and HTTP endpoints (gRPC mode had an issue that was patched in 1.35.1).

Given this, I think it's best to migrate our jaeger exporter documentation sections. I'd actually rather just get rid of the current pieces of docs that tell you how to install and use a jaeger exporter, and instead just have each subsection tell you how to install and run jaeger, then use the otlp exporter to talk to jaeger. An alternative is to document both, but I'd also rather not show both ways unless we clearly note that the non-otlp way of doing things is for older jaeger versions.

Gaurang-Patel:
This PR resolves #1567 for .NET

Preview:
.NET : https://deploy-preview-1704--opentelemetry.netlify.app/docs/instrumentation/net/exporters/#otlp-endpoint


content/en/docs/instrumentation/net/exporters.md
```diff
@@ -105,66 +104,28 @@ AppContext.SetSwitch("System.Net.Http.SocketsHttpHandler.Http2UnencryptedSupport
...
-## Jaeger
+### Jaeger
 
 If you are using [Jaeger](https://www.jaegertracing.io/) to visualize trace
-data, you'll need to set it up first. This is how to run it in a docker
-container:
+data, This is how to run it in a docker container:
 
 ```shell
 $ docker run -d --name jaeger \
```
Kielek, line 113:
```suggestion
docker run -d --name jaeger \
```

content/en/docs/instrumentation/python/exporters.md
```diff
@@ -136,71 +135,28 @@ OTLPSpanExporter(endpoint="<traces-endpoint>/v1/traces")
...
+### Jaeger
 
 If you are using [Jaeger](https://www.jaegertracing.io/) to visualize trace
-data, you'll need to set it up first. This is how to run it in a docker
-container:
+data, This is how to run it in a docker container:
 
-```console
+```shell
 $ docker run -d --name jaeger \
```
Kielek, line 144:
```suggestion
docker run -d --name jaeger \
```

Kielek (approved):

content/en/docs/instrumentation/net/exporters.md
```diff
@@ -105,66 +104,28 @@ AppContext.SetSwitch("System.Net.Http.SocketsHttpHandler.Http2UnencryptedSupport
 
 If you are using .NET 5 or higher, the previous code sample is not required.
 
-## Jaeger
+### Jaeger
 
 If you are using [Jaeger](https://www.jaegertracing.io/) to visualize trace
-data, you'll need to set it up first. This is how to run it in a docker
-container:
+data, This is how to run it in a docker container:
```
nrcventura, line 110: Just a capitalization recommendation.
```suggestion
data, this is how to run it in a docker container:
```

nrcventura (approved):

cartermp (approved):

cartermp:
Will leave this open for a bit for @open-telemetry/python-approvers to take a look, but otherwise will merge sometime next week if there's no objections (as this is as-per the docs plan)

content/en/docs/instrumentation/python/exporters.md
```diff
@@ -136,71 +135,28 @@ OTLPSpanExporter(endpoint="<traces-endpoint>/v1/traces")
...
-from opentelemetry.sdk.trace.export import BatchSpanProcessor
-
-resource = Resource(attributes={
-    SERVICE_NAME: "your-service-name"
-})
-
-jaeger_exporter = JaegerExporter(
-    agent_host_name="localhost",
-    agent_port=6831,
-)
```
srikanthccv, line 188: Why are we removing this altogether? While the newer version of jaeger support OTLP directly, this is still important for users who would want to use other protocol such as thrift/UDP or thrift/HTTP. I would keep this and add a new note that jaeger now supports OTLP directly and one can just use OTLP exporter to send data directly.

srikanthccv (changes requested):

content/en/docs/instrumentation/net/exporters.md
```diff
@@ -105,66 +104,28 @@ AppContext.SetSwitch("System.Net.Http.SocketsHttpHandler.Http2UnencryptedSupport
 
 If you are using .NET 5 or higher, the previous code sample is not required.
 
-## Jaeger
+### Jaeger
 
 If you are using [Jaeger](https://www.jaegertracing.io/) to visualize trace
```
svrnm, line 109: `If you are using [Jaeger](https://www.jaegertracing.io/) to visualize trace data, this is how to run it in a docker container:`

This sentence is confusing (you use jaeger but we ask you to install jaeger?)

I suggest something like:

`To try out the OTLP exporter, you can run  [Jaeger](https://www.jaegertracing.io/) as an OTLP endpoint and for trace visualization in a docker container`

content/en/docs/instrumentation/python/exporters.md
```diff
@@ -136,71 +135,28 @@ OTLPSpanExporter(endpoint="<traces-endpoint>/v1/traces")
 
 There is not currently an OTLP/HTTP metric exporter.
 
-## Jaeger
+### Jaeger
 
 If you are using [Jaeger](https://www.jaegertracing.io/) to visualize trace
```
svrnm, line 140: Copy comment from above:

`If you are using [Jaeger](https://www.jaegertracing.io/) to visualize trace data, this is how to run it in a docker container:`

This sentence is confusing (you use jaeger but we ask you to install jaeger?)

I suggest something like:

`To try out the OTLP exporter, you can run  [Jaeger](https://www.jaegertracing.io/) as an OTLP endpoint and for trace visualization in a docker container`

content/en/docs/instrumentation/python/exporters.md
cartermp: @srikanthccv we felt that just moving to OTLP for documenting was adequate as per here: https://github.com/open-telemetry/opentelemetry.io/issues/1567 -- I don't personally believe that the intersection of people who need to know how to send data to Jaeger _and_ want to explicitly use the thrift protocol is high enough to justify adding two ways to do things to docs. I don't have any data to back that up, just that in general non-defaults like this drop off like a cliff in in terms of usage.

srikanthccv: Many probably prefer using OTLP directly with Jaeger. I am not against documenting that. My point is why do we need to remove the existing documentation which shows how to send using thrift to jaeger to convey that. This PR is getting rid of jaeger exporter section totally and adding a sub section under OTLP. If we accept this change where do we send the people who come looking for thrift documentation (for whatever reason they might have)? I view this page as a source of guide for end user rather than recommendation page backed by data.

content/en/docs/instrumentation/net/exporters.md
```diff
@@ -41,11 +41,10 @@ using var tracerProvider = Sdk.CreateTracerProviderBuilder()
     .Build();
 ```
 
-## OTLP endpoint or Collector
+## OTLP endpoint
 
-To send data to an OTLP endpoint or the [OpenTelemetry
-Collector](/docs/collector/getting-started/), you'll want to configure an OTLP
-exporter that sends to your endpoint.
+To send trace data to a OTLP endpoint (like the [collector](/docs/collector) or
```
alanwest, line 46: nit:
```suggestion
To send trace data to an OTLP endpoint (like the [collector](/docs/collector) or
```

content/en/docs/instrumentation/python/exporters.md
```diff
@@ -58,11 +58,10 @@ provider = MeterProvider(resource=resource, metric_readers=[reader])
 metrics.set_meter_provider(provider)
 ```
 
-## OTLP endpoint or Collector
+## OTLP endpoint
 
-To send data to an OTLP endpoint or the [OpenTelemetry
-Collector](/docs/collector/getting-started/), you'll want to configure an OTLP
-exporter that sends to your endpoint.
+To send trace data to a OTLP endpoint (like the [collector](/docs/collector) or
```
alanwest, line 63:
```suggestion
To send trace data to an OTLP endpoint (like the [collector](/docs/collector) or
```

alanwest (approved):
Just a couple nits

content/en/docs/instrumentation/python/exporters.md
cartermp: So I'm of two minds for this:

* I agree in principle that having reference documentation for edge cases like this matters
* I can't think of a good reason why someone, outside of using an older version of jaeger, would need or want to use thrift rather than OTLP, and I'd much rather encourage them to switch to OTLP. But I am not used to operating environments that use thrift so I don't know how realistic that stance is

FWIW there are docs for the exporter itself here: https://opentelemetry-python.readthedocs.io/en/latest/exporter/jaeger/jaeger.html

So I'm fine with leaving the Python docs as-is, or at least linking to this, for a time.

@svrnm what do you think?

srikanthccv: > I can't think of a good reason why someone, outside of using an older version of jaeger, would need or want to use thrift rather than OTLP, and I'd much rather encourage them to switch to OTLP. But I am not used to operating environments that use thrift so I don't know how realistic that stance is

In my experience, It's not the version but rather the deployment pattern. In cases where jaeger runs as an agent thrift/UDP is preferred and where it's not possible to run the agent alongside then other ways are preferred. The thrift exporter on PyPI gets significant enough downloads each month https://pypistats.org/packages/opentelemetry-exporter-jaeger-thrift. I want to reiterate my concern. I am fine with actively encouraging people to move to direct OTLP but that shouldn't mean we remove the documentation for thrift/UDP or thrift/HTTP exporters.

>FWIW there are docs for the exporter itself here: https://opentelemetry-python.readthedocs.io/en/latest/exporter/jaeger/jaeger.html

We have been in the process of getting rid of them and asking users to consider opentelemetry.io docs as one true source and only use the python readthedocs.io for the component API documentation.

cartermp: @Gaurang-Patel can you revert the changes to Python for now? I think we can figure out a good way forward, but let's unblock .NET for now, which is approved.

Gaurang-Patel: @cartermp I'll do it in a day or two.

cartermp (approved):
Thank you @Gaurang-Patel

PR #1732, opened 2022-09-09
Unify page head for language instrumentations.

svrnm:
Related: #1656 & other discussions around reusable status.

I did some reading on hugo and [Data Templates](https://gohugo.io/templates/data-templates/), this seems to be "the thing" we should be using for the status. Eventually this should allow us to have per-language as well as a consolidated view. 

I could not apply it to all pages yet because of differences in style and the `content_modules`

@cartermp, @chalin: what do you think?


tsloughter (approved):
Giving approval since the change to the Erlang page should be fine. But I don't see where the template for the head is. Does it already exist or did you miss adding it?

svrnm:
> Giving approval since the change to the Erlang page should be fine. But I don't see where the template for the head is. Does it already exist or did you miss adding it?

Thanks😅, I added them.

content/en/docs/instrumentation/cpp/_index.md
```diff
@@ -8,22 +8,7 @@ description: >
   A language-specific implementation of OpenTelemetry in C++.
 ---
 
-This is the OpenTelemetry for C++ documentation. OpenTelemetry is an
```
cartermp, line 11: Is there a way to do a docs comment or something that won't render? Since language SIG owners are already used to going into these pages to update status, we should leave a comment here asking them to update it in the yaml file.

svrnm: Good point, I added a small comment to each page, we probably can/should do the same throughout the docs

cartermp (approved):
I'll merge in a few days if there's no objections.

Issue #1737, opened 2022-09-11
content goes outside the container


Arhell:
**What happened?**
https://opentelemetry.io/registry/
![Screenshot_2](https://user-images.githubusercontent.com/26163841/189518840-86d8e00b-bec4-446c-bbfc-0cc99718020b.png)

/assign

cartermp:
@Arhell if you're up for a contribution I'm happy to review!

PR #1738, opened 2022-09-12
fix content goes outside the container

Arhell:
resolve https://github.com/open-telemetry/opentelemetry.io/issues/1737
after fix
Screenshot_1
when there is enough space it becomes as it was in a line
Screenshot_2


cartermp (approved):
Excellent, thank you @Arhell. This does look better on a smaller mobile device, and indeed will become a single line on a larger one (e.g., my iphone).

PR #1742, opened 2022-09-14
Reformat erlang/elixir testing docs

cartermp:
Just formats the text to line wrap like the rest of the docs.


cartermp:
Haha, oh dear, the extension I use to reformat doesn't recognize the hugo blocks. Will close and do this differently.

PR #1743, opened 2022-09-14
fix typo in CODEOWNERS.

svrnm:


svrnm:
oh... that group is named `mainteiners` ... then, never mind ...

PR #1744, opened 2022-09-14
Add instructions for web tracer & exporting

svrnm:
When people try to setup opentelemetry with a browser application they commonly run into the following issues:

* They try to use GRPC
* They don't have a CSP set up for their collector
* They don't have CORS headers send by their collector
* They are not aware that they need to expose their collector to the public

While trying to [bring back the otel browser extension](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1152) I ran into those issues myself, so I documented what I have done and with that I tried to address them here with a few words as well. 

Not perfect yet, but hopefully a starting point.

### DIFF
--- a/content/en/docs/instrumentation/js/exporters.md
+++ b/content/en/docs/instrumentation/js/exporters.md
@@ -3,8 +3,8 @@ title: "Exporters"
 weight: 5
 ---
 
-In order to visualize and analyze your traces, you will need to
-export them to a backend such as [Jaeger](https://www.jaegertracing.io/) or
+In order to visualize and analyze your traces, you will need to export them to a
+backend such as [Jaeger](https://www.jaegertracing.io/) or
 [Zipkin](https://zipkin.io/). OpenTelemetry JS provides exporters for some
 common open source backends.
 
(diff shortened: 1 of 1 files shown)


content/en/docs/instrumentation/js/exporters.md
```diff
@@ -67,6 +67,109 @@ $ docker run -d --name jaeger \
...
+
+Below you will find instructions to use the right exporter, to configure your
+CSPs and CORS headers and what precautions you have to take when exposing
+your collector.
+
+#### Use OTLP exporter with HTTP
+
+[OpenTelemetry Collector Exporter with grpc][] does only work with Node.JS,
+therefor you are limited to use the
+[OpenTelemetry Collector Exporter with http][]. This is not an issue, but you
```
cartermp, line 89: I believe this defaults to http/json and not http/proto. Is that worth mentioning?

content/en/docs/instrumentation/js/exporters.md
```diff
@@ -67,6 +67,109 @@ $ docker run -d --name jaeger \
   jaegertracing/all-in-one:latest
 ```
 
+### Usage with the WebTracer
+
+When you use the OTLP exporter in a browser-based application, you need to note
+that:
+
+1. grpc-based exporting is not supported
```
cartermp, line 75: I believe http/proto is also not supported (today) either, so that means if you wish to send to some other OTLP endpoint (let's assume it doesn't require any authentication so api keys or whatever aren't a concern) - then you'll need to ensure that backend also supports http/json

content/en/docs/instrumentation/js/exporters.md
svrnm: let me double check on that and update accordingly.

content/en/docs/instrumentation/js/exporters.md
svrnm: You are right, there's a third package for protobuf. Let me fix that.

pichlermarc (approved):
Looks good! :slightly_smiling_face:

cartermp (approved):

PR #1745, opened 2022-09-14
docs: make events an array in example span

ttd2089:
Based on the example trace above it looks like the events property of a span is supposed to be an array an objects rather than an object.


cartermp (approved):
Thanks! You are correct.

PR #1747, opened 2022-09-14
Replace ruby with rust in repo structure example

svrnm:
Let's be honest, the `content/en/docs/concepts/instrumenting.md` needs a major rework, but before that I fixed that minor issue: ruby has a contrib repo now, so the statement is not true. I replaced it with rust for the time being.


cartermp (approved):
Heh, this is almost like developer documentation rather than end-user documentation. Woof. Deffo needs a rewrite. Anyways, good update!

PR #1762, opened 2022-09-19
Initialize npm for the node.js example

svrnm:
This PR adds an `npm init -f` to the node.js example, while not strictly needed, I ran into the bug that npm installed packages into a parent folder which had a node_modules folder already, which is confusing. This makes sure that everything is contained in the same folder. It will also help in another addition I plan to do (resource detection)


pichlermarc (approved):

cartermp (approved):

psk001:
Does any such example exist for ES6 ?

PR #1786, opened 2022-09-26
Add Aria by VMware to vendor list

oppegard:


cartermp (commented):
I'm prepared to approve this, but in the link you give there's no mention of "Aria" anywhere. Is that intentional?

oppegard:
Thanks for taking a look @cartermp. Our product within VMware is undergoing a [rebranding](https://blogs.vmware.com/management/2022/08/introducing-vmware-aria.html), but it's going to take some time before our various websites are updated. We'd like our listing on the OTel vendors page to reflect what the name will be in a few months from now.

Ideally, our entry would temporarily say "Aria by VMware (Wavefront)". We would come back later to remove the Wavefront part. Is that acceptable?

cartermp:
That sounds good @oppegard - with that change, I'll gladly approve and merge 🙂

chalin (approved):
LGTM, and thanks for adding cSpell entries!

PR #1787, opened 2022-09-26
Clarify Wavefront in vendor list

oppegard:
#1786 was merged before I could make an additional change, so that "Aria by VMware" is instead "Aria by VMware (Wavefront)". We intend this to be a stop-gap until a rebranding at VMware is finished.


oppegard:
@cartermp and @chalin: I made a small tweak to #1786, as discussed in that PR.

cartermp (approved):
Thanks!

PR #1792, opened 2022-09-30
Add registry entry for the official .NET OTLP exporter

joaopgrassi:
I noticed we didn't have a registry entry for the OTLP exporter in .NET. 

Also: I _think_ the `isThirdParty` field had a misleading comment so I modified that as well. I hope it's ok to bundle this together with this change.

### DIFF
--- a/content/en/registry/exporter-dotnet-otlp.md
+++ b/content/en/registry/exporter-dotnet-otlp.md
@@ -1,0 +1,16 @@
+---
+title: OTLP Exporter
+registryType: exporter
+isThirdParty: false
+language: dotnet
+tags:
+  - .NET
+  - C#
+  - dotnet
+  - exporter
+repo: https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/src/OpenTelemetry.Exporter.OpenTelemetryProtocol
+license: Apache 2.0
+description: This library allows to export data to the OpenTelemetry Collector using the OpenTelemetry Protocol.
+authors: OpenTelemetry Authors
+otVersion: latest
+---
--- a/templates/registry-entry.md
+++ b/templates/registry-entry.md
@@ -1,7 +1,7 @@
 ---
 title: My OpenTelemetry Integration // the name of your project
 registryType: <exporter/core/instrumentation> // the type of integration; is this an exporter, plugin, API package, or something else?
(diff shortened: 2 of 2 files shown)


content/en/registry/exporter-dotnet-otlp.md
```diff
@@ -0,0 +1,16 @@
...
+isThirdParty: false
+language: dotnet
+tags:
+  - .NET
+  - C#
+  - dotnet
+  - exporter
+repo: https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/src/OpenTelemetry.Exporter.OpenTelemetryProtocol
+license: Apache 2.0
+description: This library allows to export data to the OpenTelemetry Collector using the OpenTelemetry Protocol.
```
joaopgrassi, line 13: I noticed this text is used in all other "OTLP exporter entries". But it's not completely correct, as OTLP exporters also allow sending data to other places outside the collector (e.g., directly to an o11y back-end).

Maybe something like this:
> This library allows to export telemetry data in the OpenTelemetry Protocol (OTLP) format.

I can modify all the other exporters to have the same text, once we agree on the description.

svrnm: agreed, we had similar statements across the docs and changed that there as well.

> This library allows to export telemetry data in the OpenTelemetry Protocol (OTLP) format.

I am good with that, but maybe we can also add:

> This library allows to export telemetry data in the OpenTelemetry Protocol (OTLP) format to the OpenTelemetry Collector and OTLP-compliant backends

svrnm:
> Also: I _think_ the `isThirdParty` field had a misleading comment so I modified that as well. I hope it's ok to bundle this together with this change.

Yeah, that makes sense. I am ok with bundling it here.

content/en/registry/exporter-dotnet-otlp.md
joaopgrassi: Nice! We could swap `backends` to `receivers`, both are fine with me. I opened an issue for it, I guess we could move on with this PR as is, then I can fix all the others in one go. 

https://github.com/open-telemetry/opentelemetry.io/issues/1793

cartermp (approved):
Thank you! And yes, please feel free to modify the other exporter descriptions after this is merged in.

Issue #1773, opened 2022-09-21
[Meta] Add pages about resources & resource detection to concept & languages


svrnm:
Add a new page to “concepts“ that talks about resources & detection in a general way 

If available&applicable, add a page about resources & resource detection for each language, similar to [this one](https://opentelemetry.io/docs/instrumentation/js/resources/):

* What resources are detected automatically?
* Adding resources via OTEL_RESOURCE_ATTRIBUTES
* Adding resources in code
* Adding a (non default) resource detector (e.g. container)

**TODO**:
- [ ] Concept 
- [ ] C++
- [x] .NET
- [x] Erlang/Elixir
- [x] Go
- [x] Java
- [x] [JavaScript](https://opentelemetry.io/docs/instrumentation/js/resources/)
- [x] PHP
- [ ] Python
- [ ] Ruby
- [ ] Rust
- [ ] Swift

Note, I added all languages to get started, if this is not applicable for some of them, we can remove them.

svrnm:
Added also that we need a page that talks about the concept of resources and how they are detected

svrnm:
> Wonder if we could pin this one and simlar, @svrnm

I list them here: https://github.com/open-telemetry/opentelemetry.io/issues/3686, but we can also pin some of them directly

PR #1795, opened 2022-10-02
Add resources doc for .NET

cartermp:
Contributes to https://github.com/open-telemetry/opentelemetry.io/issues/1773

Similar to the JS one, but with some notable changes:

* No process detection section (can add later if there's a good way to do it)
* Code sample is different, as there is no semantic attributes package for .NET yet
* Elided adding docker resources, but could add later if there's a package for that

### DIFF
--- a/content/en/docs/instrumentation/net/resources.md
+++ b/content/en/docs/instrumentation/net/resources.md
@@ -128,8 +128,8 @@ Resource associated with Activity:
     os.version: 21.6.0
 ```
 
-**Note**: If you set your resource attributes via environment variable and code,
-any values set via the environment variable take precedence.
+**Note**: If you set resource attributes with both environment variables and
+code, the values in code take precedence.
 
 ## Next steps
 
--- a/content/en/docs/instrumentation/net/resources.md
+++ b/content/en/docs/instrumentation/net/resources.md
@@ -23,7 +23,7 @@ Follow the instructions in the [Getting Started][], so that you have a running
 
 You can use the `OTEL_RESOURCE_ATTRIBUTES` environment variable to inject
 resources into your application. The .NET SDK will automatically detect these
-resrouces.
+resources.
 
 The following example adds [Service][], [Host][] and [OS][] resource attributes
 via environment variables, running unix programs like `uname` to generate the
--- a/content/en/docs/instrumentation/net/resources.md
+++ b/content/en/docs/instrumentation/net/resources.md
@@ -14,9 +14,6 @@ investigate interesting behavior. For example, if your trace or metrics data
 indicate latency in your system, you can narrow it down to a specific container,
 pod, or kubernetes deployment.
 
-Below you will find some introductions on how to setup resource detection with
-the Node.JS SDK.
-
 ## Setup
 
 Follow the instructions in the [Getting Started][], so that you have a running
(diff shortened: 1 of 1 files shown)


pellared (approved):
Just left a comment regarding more options on how to set `service.name` as it is the only required resource attribute.

content/en/docs/instrumentation/net/resources.md
```diff
@@ -0,0 +1,148 @@
...
+
+using var tracerProvider = Sdk.CreateTracerProviderBuilder()
+    .AddSource(sourceName)
+    .SetResourceBuilder(resourceBuilder)
+    .AddConsoleExporter()
+    .Build();
+
+var MyActivitySource = new ActivitySource(sourceName);
+
+using var activity = MyActivitySource.StartActivity("SayHello");
```
Kielek, line 93: I think that we should promote API where the Kind is specified. It is based on a bug we have had in AutoInstrumentation for .NET.

```suggestion
using var activity = MyActivitySource.StartActivity("SayHello", ActivityKind.Internal);
```

svrnm:
> Elided adding docker resources, but could add later if there's a package for that

@swetharavichandrancisco created an extension for that: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main/src/OpenTelemetry.Extensions.Docker -- does it have all we need?

content/en/docs/instrumentation/net/resources.md
cartermp: Hmmm. Is this not the default though? So long as it's possible to specify a different kind of `ActivityKind` via .NET APIs then I'm not sure I'd agree with having this overload (specifying something that's a default) for every call to `StartActivity` in the docs.

content/en/docs/instrumentation/net/resources.md
```diff
@@ -0,0 +1,148 @@
...
+## Setup
+
+Follow the instructions in the [Getting Started][], so that you have a running
+.NET app exporting data to the console.
+
+## Adding resources with environment variables
+
+You can use the `OTEL_RESOURCE_ATTRIBUTES` environment variable to inject
+resources into your application. The .NET SDK will automatically detect these
+resrouces.
```
cijothomas, line 26:
```suggestion
resources.
```

content/en/docs/instrumentation/net/resources.md
cijothomas: Internal is the default, so no need of explcitily specifying it.

content/en/docs/instrumentation/net/resources.md
```diff
@@ -0,0 +1,148 @@
...
+    service.version: 1.0
+    service.instance.id: 28976A1C-BF02-43CA-BAE0-6E0564431462
+    host.name: pcarter
+    host.type: arm64
+    os.name: Darwin
+    os.version: 21.6.0
+```
+
+**Note**: If you set your resource attributes via environment variable and code,
+any values set via the environment variable take precedence.
```
cijothomas, line 132: i do not think this is true.

cartermp: I see, yes, it is actually the inverse. Fixing!

content/en/docs/instrumentation/net/resources.md
```diff
@@ -0,0 +1,148 @@
...
+pod, or kubernetes deployment.
+
+## Setup
+
+Follow the instructions in the [Getting Started][], so that you have a running
+.NET app exporting data to the console.
+
+## Adding resources with environment variables
+
+You can use the `OTEL_RESOURCE_ATTRIBUTES` environment variable to inject
```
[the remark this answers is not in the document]
cartermp: I filed https://github.com/open-telemetry/opentelemetry.io/issues/1830 for this -- for now I'll keep this since it's also in the JS doc, and we can address these uniformly.

cartermp:
@pellared @cijothomas PTAL! Should be all addressed.

svrnm:
@cartermp, thoughts on adding a section for docker container id with https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main/src/OpenTelemetry.Extensions.Docker ? We can do this as a follow up.

svrnm (approved):

pellared (approved):
👍

content/en/docs/instrumentation/net/resources.md
```diff
@@ -0,0 +1,148 @@
...
+using System.Collections.Generic;
+
+using OpenTelemetry;
+using OpenTelemetry.Trace;
+using OpenTelemetry.Resources;
+
+var resourceBuilder =
+    ResourceBuilder
+        .CreateDefault()
+        .AddAttributes(new Dictionary<string, object>
```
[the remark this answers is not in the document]
svrnm: then let's have a followup issue!

Issue #1809, opened 2022-10-05
Host community pages via repo submodule


svrnm:
via https://github.com/open-telemetry/community/pull/1056:

> Discussed in the GC call. GC recommendation is to use the same publishing mechanism for Community repo that we already use for including Specification into the website pages.

> My personal suggestion: instead of using a submodule, use a web hook and pull the dependency repos at build time. This will eliminate the need for manual submodule hash updates and will keep the docs up-to-date with the Specs/Community automatically (one possible downside is if commits in those repos end up breaking the build of the website, e.g. by using invalid link).

chalin:
We're using submodules already, I'd rather continue using submodules (easier to reproduce builds IMHO).

chalin:
@svrnm - are you ok with a change in title for this issue, something like: **Host community pages via repo submodule**?

svrnm:
> @svrnm - are you ok with a change in title for this issue, something like: **Host community pages via repo submodule**?

Yes, let's change it to that.

chalin:
Closing, since the community repo is now a submodule. For the mission page work, see:

- #1289

PR #1731, opened 2022-09-09
First draft for a persona based documentation.

svrnm:
Preview: https://deploy-preview-1731--opentelemetry.netlify.app/

Based on the discussion we had yesterday, here's a draft for a persona based documentation. 

**The purpose of this draft is to kick off a discussion and is incomplete!**

Some notes: follow the flow from the main page to the "Getting Started" page, there's a list of links per persona, there the flow stops right now and/or requires some content that does not yet exist. 

A few months back we had a related discussion & document: https://docs.google.com/document/d/1Bs9nOyopvjwGqJWAH1VZwMfEmUSGwyjOo2T8EBEnxxM/edit#heading=h.xtwnq2rnmk86

PR #1825, opened 2022-10-07
Move mission page to community

svrnm:
As discussed yesterday we move the mission, vision, values below /community

#1809 will not change because of that, later we will source that page from the submodule.

### DIFF
--- a/content/en/community/mission.md
+++ b/content/en/community/mission.md
@@ -1,13 +1,7 @@
 ---
-title: Mission
+title: Mission, vision, and values
+aliases: [/mission, /vision, /values]
 ---
-{{% blocks/lead color="primary" %}}
-
-# Mission, vision, and values
-
-{{% /blocks/lead %}}
-
-{{% blocks/section type="section" color="white" %}}
 
 ## Mission &mdash; our overall north star as a community
 
--- a/content/en/community/mission.md
+++ b/content/en/community/mission.md
@@ -1,6 +1,6 @@
 ---
 title: Mission, vision, and values
-aliases: [/mission, /vision, /values]
+aliases: [/mission]
 ---
 
 ## Mission &mdash; our overall north star as a community
--- a/content/en/_index.md
+++ b/content/en/_index.md
@@ -14,7 +14,7 @@ developer_note:
 <div class="l-primary-buttons mt-5">
 
 - [Learn more](/docs/concepts/)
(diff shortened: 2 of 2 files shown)


chalin (changes requested):
I'm having second thoughts about this.

IMHO, the **OTel** mission statement _does_ make sense at the top level, `/mission`.

I think that the TOC/GC _happened_ to place the OTel-project mission statement in the community repo because OTel doesn't have a single organization repo, but it isn't a "community" resource, it is an OTel project-wide statement.

content/en/community/mission.md
```diff
@@ -1,13 +1,7 @@
 ---
-title: Mission
+title: Mission, vision, and values
+aliases: [/mission, /vision, /values]
```
chalin, line 3: If ever there is a decision to move ahead with this page move, I'd vote against introducing two new aliases for pages that did not exist before, namely `/vision`, and `/values`.

svrnm:
> I'm having second thoughts about this.
> 
> IMHO, the **OTel** mission statement _does_ make sense at the top level, `/mission`.
> 
> I think that the TOC/GC _happened_ to place the OTel-project mission statement in the community repo because OTel doesn't have a single organization repo, but it isn't a "community" resource, it is an OTel project-wide statement.

Good point.

There are 2 problems I wanted to fix with that PR:

* Right now the mission page is only linked via the link on the home page, so it's really hard to find
* https://github.com/open-telemetry/opentelemetry.io/pull/1731 is going to remove that link, so there is no link at all to the mission

I am open to find another home for that page.

content/en/community/mission.md
svrnm: fixed

reyang (approved):

melbgirl:
I do think the mission/vision page doesn't add much value as top-level item. The most important things when coming to opentelemetry.io are to see a) what the project is about and b) how to learn more. Mission/vision is a tertiary interest.
That being said, we still need a way to get to the mission/vision, and a home for it. IMO it naturally lives in the Community section of the site, and should be linked from the "About OTel" page.

svrnm:
I think that the mission, vision & values are very important and should have prominent (places), but the main links on the home page are geared towards new people, so yes it doesn't have to be that prominent on that page.

cartermp (approved):

chalin (approved):
I'm willing to go with the majority! ✨

PR #1838, opened 2022-10-10
Add workflow to auto-update collector version

svrnm:
Goal of this workflow is to regularly check for collector version updates and then provide a PR to upgrade them.

Most of the implementation is taken from @jpkrohling's work referenced here: https://github.com/open-telemetry/opentelemetry-collector-releases/issues/196


scripts/update-collector-version.sh
```diff
@@ -0,0 +1,64 @@
...
+
+# here we don't like having different versions, so we skip creating the PR if contrib & core are not equal
+if [ $latest_core_version != $latest_contrib_version ]; then
+    echo "The contrib and core versions aren't matching. This might be OK, but perhaps there's a release in process?"
+    exit -1
+fi
+
+branch="auto-update/collector_version_${latest_core_version}"
+
+# git checkout -b "${branch}" main
```
jpkrohling, line 46: You might indeed need to use a new branch out of main, otherwise you'll get failures during the push.

.github/workflows/auto-update-collector-version
```diff
@@ -0,0 +1,32 @@
+name: Auto-update Collector Version
```
chalin, line 1: I wouldn't call this auto-update, since the workflow doesn't, for example, both check and then automatically commit the new version to `main`.

This script both checks the version and submits a PR if necessary. That's a bit of a mouthful, so maybe just name this "Check collector version".

```suggestion
name: Check collector version
```

If you agree, then the file name should be updated too.

.github/workflows/auto-update-collector-version
```diff
@@ -0,0 +1,32 @@
...
+      - name: Auto-update
+        run: ./scripts/update-collector-version.sh
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GIT_AUTHOR_NAME: Auto-update CI workflow
+          GIT_COMMITTER_NAME: Auto-update CI workflow
+
+          # Set svrnm's email, otherwise CLA checks will fail
+          GIT_AUTHOR_EMAIL: neumanns@cisco.com
+          GIT_COMMITTER_EMAIL: neumanns@cisco.com
```
chalin, line 32: We should get an official OTel bot account for this. WDYT @austinlparker @cartermp?

chalin (commented):
Cool. See inline comments, thanks.

.github/workflows/auto-update-collector-version
svrnm: agreed

.github/workflows/auto-update-collector-version
svrnm: Here I just copied from @jpkrohling's work, it looks like that there's some issues with CLA, not sure if we can work around that easily?

scripts/update-collector-version.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
svrnm: Agreed, I will rename this after the other conversations are resolved, so they don't get outdated

.github/workflows/auto-update-collector-version
trask: see discussion in https://github.com/open-telemetry/community/issues/809 and https://github.com/open-telemetry/community/issues/954

TLDR, there have been no objections to using the @opentelemetrybot account which has signed the CLA. It's on me to officially add this to the community repository assets page, I'll send a PR to the community repo later today.

```
git config user.name opentelemetrybot
git config user.email 107717825+opentelemetrybot@users.noreply.github.com
```

trask: sent community PR to (hopefully) make this approach official: https://github.com/open-telemetry/community/pull/1285

svrnm:
I updated the email addresses to use the bot (is this the right approach, or is there more I need to do @trask ?)

I am wondering if I should split the PR into the script and then adding the workflow in a second step?

.github/workflows/auto-update-collector-version
```diff
@@ -0,0 +1,32 @@
...
+      - name: Auto-update
+        run: ./scripts/update-collector-version.sh
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GIT_AUTHOR_NAME: Auto-update CI workflow
+          GIT_COMMITTER_NAME: Auto-update CI workflow
+
+          # Set svrnm's email, otherwise CLA checks will fail
+          GIT_AUTHOR_EMAIL: 107717825+opentelemetrybot@users.noreply.github.com
+          GIT_COMMITTER_EMAIL: 107717825+opentelemetrybot@users.noreply.github.com
```
trask, line 32:
```suggestion
          GIT_AUTHOR_NAME: opentelemetrybot
          GIT_COMMITTER_NAME: opentelemetrybot

          GIT_AUTHOR_EMAIL: 107717825+opentelemetrybot@users.noreply.github.com
          GIT_COMMITTER_EMAIL: 107717825+opentelemetrybot@users.noreply.github.com
```

trask (commented):
hey @svrnm, I sent some suggestions as a PR to your branch: https://github.com/svrnm/opentelemetry.io/pull/1

one other thought that I didn't include since it made the diff harder to follow, is to move the script to the `.workflow/scripts` folder, this is where I've been sticking scripts that are only used for supporting workflows

scripts/update-collector-version.sh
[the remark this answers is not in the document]
svrnm: will rename now.

.github/workflows/auto-update-collector-version
svrnm: fixed via PR

.github/workflows/auto-update-collector-version.yml
```diff
@@ -0,0 +1,30 @@
...
+          go-version: 1.18
+
+      - name: Auto-update
+        run: ./.github/workflows/scripts/update-collector-version.sh
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GIT_AUTHOR_NAME: opentelemetrybot
+          GIT_COMMITTER_NAME: opentelemetrybot
+
+          # Set svrnm's email, otherwise CLA checks will fail
```
trask, line 28:
```suggestion
          # Set opentelemetrybot's email, otherwise CLA checks will fail
```

trask (approved):
👍

cartermp (approved):
I think this all looks kosher, @svrnm and @chalin any additional thoughts?

cartermp:
Will merge - we can make more changes if we need to later 🙂

PR #1839, opened 2022-10-10
Addition SDK env configs 

### LINKED ISSUE #1783: Missing other exporters configuration in the SDK configuration page
I was looking at the SDK configuration documentation here https://opentelemetry.io/docs/concepts/sdk-configuration/ and noticed that other than general SDK related configs, it also lists configuration specific for the OTLP exporter.
I was wondering if it would need to list the main two other exporters as well like Jaeger and Zipkin because their configuration seems to be listed here https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md but not in the documentation itself.
So someone who is using Jaeger and looking for configuration can't find content on the doc but has to go to the SDK env vars in the specification repo.

siddhant-khisty:
Addition the environment variable configuration to the SDK configuration page.

I included the [SDK environment reference](https://opentelemetry.io/docs/reference/specification/sdk-environment-variables/) to the [SDK configuration page](https://opentelemetry.io/docs/concepts/sdk-configuration). Currently, I have only created the structure of the file. I will add more details to it about the jaeger and Zipkin exporters if this format is alright

Fixes #1783


content/en/docs/concepts/sdk-configuration/sdk-environment-variables.md
```diff
@@ -0,0 +1,142 @@
...
+ Maximum time (in milliseconds) the Jaeger exporter will wait for each batch export 
+ example: 10000
+- OTEL_EXPORTER_JAEGER_USER    
+ Username to be used for HTTP basic authentication 
+- OTEL_EXPORTER_JAEGER_PASSWORD | Password to be used for HTTP basic authentication 
+
+
+Environment variables specific for the `udp/thrift.compact` transport protocol:
+
+-  OTEL_EXPORTER_JAEGER_AGENT_HOST
```
melbgirl, line 100: Variables are expressed as
## 'OTEL_EXPORTER_JAEGER_AGENT_HOST'
not
- 'OTEL_EXPORTER_JAEGER_AGENT_HOST'

content/en/docs/concepts/sdk-configuration/sdk-environment-variables.md
```diff
@@ -0,0 +1,142 @@
+# OpenTelemetry Environment Variable Specification
+
+**Status**: [Mixed](document-status.md)
+
+The goal of this specification is to unify the environment variable names between different OpenTelemetry SDK implementations. SDKs MAY choose to allow configuration via the environment variables in this specification, but are not required to. If they do, they SHOULD use the names listed in this document.
+
+## Parsing empty value
+
+**Status**: [Stable](document-status.md)
```
melbgirl, line 9: We tend not to add the Status within the docs here, I believe in order to avoid having to update multiple sources. 
I'm curious to see what others think about this? Could this be beneficial to readers to avoid them having to switch over to the SDK language to view the environment variable status?

content/en/docs/concepts/sdk-configuration/sdk-environment-variables.md
```diff
@@ -0,0 +1,142 @@
...
+  example: 10000
+
+Additionally, the following environment variables are reserved for future
+usage in Zipkin Exporter configuration:
+
+- `OTEL_EXPORTER_ZIPKIN_PROTOCOL`
+
+This will be used to specify whether or not the exporter uses v1 or v2, json,
+thrift or protobuf.  As of 1.0 of the specification, there
+_is no specified default, or configuration via environment variables_.
```
melbgirl, line 141: is
_no default specified, or configuration via environment variables_.

content/en/docs/concepts/sdk-configuration/sdk-environment-variables.md
```diff
@@ -0,0 +1,142 @@
...
+
+The goal of this specification is to unify the environment variable names between different OpenTelemetry SDK implementations. SDKs MAY choose to allow configuration via the environment variables in this specification, but are not required to. If they do, they SHOULD use the names listed in this document.
+
+## Parsing empty value
+
+**Status**: [Stable](document-status.md)
+
+The SDK MUST interpret an empty value of an environment variable the same way as when the variable is unset.
+
+## Special configuration types
```
melbgirl, line 13: Should this say "Special data types"?

content/en/docs/concepts/sdk-configuration/sdk-environment-variables.md
```diff
@@ -0,0 +1,142 @@
+# OpenTelemetry Environment Variable Specification
+
+**Status**: [Mixed](document-status.md)
+
+The goal of this specification is to unify the environment variable names between different OpenTelemetry SDK implementations. SDKs MAY choose to allow configuration via the environment variables in this specification, but are not required to. If they do, they SHOULD use the names listed in this document.
```
melbgirl, line 5: Unless there's no impact to using different variable names, it seems that this would be a 'must' vs a 'should':
"If they do, they MUST use the variable names listed in this document" ?

content/en/docs/concepts/sdk-configuration/sdk-environment-variables.md
```diff
@@ -0,0 +1,142 @@
...
+## Special configuration types
+
+**Status**: [Stable](document-status.md)
+
+### Boolean value
+
+Any value that represents a Boolean MUST be set to true only by the case-insensitive string `"true"`, meaning `"True"` or `"TRUE"` are also accepted, as true.
+An SDK MUST NOT extend this definition and define additional values that are interpreted as true.
+Any value not explicitly defined here as a true value, including unset and empty values, MUST be interpreted as false.
+If any value other than a true value, case-insensitive string `"false"`, empty, or unset is used, a warning SHOULD be logged to inform users about the fallback to false being applied.
```
melbgirl, line 22: Same comment as above, wondering why use 'should' here vs 'must'. If the content here is trying to inform best practices, it's clearer to outline them as 'must' (especially so in the case of error messages)

content/en/docs/concepts/sdk-configuration/sdk-environment-variables.md
```diff
@@ -0,0 +1,142 @@
...
+
+The SDK MUST interpret an empty value of an environment variable the same way as when the variable is unset.
+
+## Special configuration types
+
+**Status**: [Stable](document-status.md)
+
+### Boolean value
+
+Any value that represents a Boolean MUST be set to true only by the case-insensitive string `"true"`, meaning `"True"` or `"TRUE"` are also accepted, as true.
```
melbgirl, line 19: Please follow the structure in the other SDK configuration files, ie

Accepted values for Boolean are:
- empty
- unset
- `"true"`, `"True"` or `"TRUE"`: a case insensitive string. **Note** an SDK MUST NOT extend this definition and define additional values interpreted as true.
- `"false"`, `"False"` or `"FALSE"`, empty, unset: any value not expressed as `true` MUST be interpreted as false.

If a value other than the above is used, a warning MUST be logged informing users about the fallback applied.

content/en/docs/concepts/sdk-configuration/sdk-environment-variables.md
```diff
@@ -0,0 +1,142 @@
...
+
+Any value that represents a duration, for example a timeout, MUST be an integer representing a number of
+milliseconds. The value is non-negative - if a negative value is provided, the SDK MUST generate a warning,
+gracefully ignore the setting and use the default value if it is defined.
+
+For example, the value `12000` indicates 12000 milliseconds, i.e., 12 seconds.
+
+
+
+## Jaeger Exporter
```
melbgirl, line 47: ## Jaeger Exporter Environment Variables

melbgirl (changes requested):

content/en/docs/concepts/sdk-configuration/sdk-environment-variables.md
siddhant-khisty: I believe 'should' refers to using these variable names as best practices, but could be done with others as well?

content/en/docs/concepts/sdk-configuration/sdk-environment-variables.md
siddhant-khisty: I'll change all instances for 'should' to 'must' as we are trying to enforce the best practices for naming convention

content/en/docs/concepts/sdk-configuration/sdk-environment-variables.md
siddhant-khisty: Is there a styling guide that I could refer to?

content/en/docs/concepts/sdk-configuration/sdk-environment-variables.md
```diff
@@ -0,0 +1,164 @@
+# OpenTelemetry Environment Variable Specification
+
+**Status**: [Mixed](document-status.md)
+
+The goal of this specification is to unify the environment variable names between different OpenTelemetry SDK implementations. SDKs MAY choose to allow configuration via the environment variables in this specification, but are not required to. If they do, they MUST use the names listed in this document.
```
cartermp, line 5: This is more-or-less just copying text from the specification. I don't think we should do that. The audience for the specification is not end users, it is implementers of opentelemetry SDKs. For end users we need to phrase things differently, or in some cases, simply elide some text because it's just not relevant for anyone other than implementers of opentelemetry SDKs.

cartermp (changes requested):
Thanks for starting this, but it will need quite a bit of work. Much of the text here is just lifted from the spec, and that's not appropriate for end users of opentelemetry. It needs significant adjustments to the text so that it's more like the existing conceptual docs we have for SDK configuration and not just lifted from the specification.

content/en/docs/concepts/sdk-configuration/sdk-environment-variables.md
siddhant-khisty: I agree with that point. I plan to go through each of the environmental variables and add specific and relevant information to them.

melbgirl: Thanks @siddhant-khisty (and @cartermp)!

cartermp:
Closing due to inactivity. If you'd like to get these changes in, please feel free to ping here or resubmit and we'll take another look.

Issue #1820, opened 2022-10-06
Switch to using Docy's tabbed panes


chalin:
Docy has a flexible implementation of [tabbed panes](https://www.docsy.dev/docs/adding-content/shortcodes/#tabbed-panes) that, for example supports both code and text in a tab. I suggest that we drop our own custom implementation of tabs and upgrade to Docsy's.

WDYT @svrnm @cartermp @austinlparker? I'd like your thumbs up before I start on this cleanup.

We only have about 4 files so the upgrade wouldn't be too painful, but more files are being updated (e.g., #1800) to use tabs, so if we're going to do this I think it would be better sooner than latter :).

Tasks:

- [x] #2527
- [x] #2528
- [x] #2529

Related:

- #2525

cartermp:
Yes, please! This would allow for use of rewrap on the relevant files if code moves to code blocks.

svrnm:
Yes, let's do that!

chalin:
Great! Ok, I'll start that as soon as I can (and I'll add a comment to #1772) -- I'm waiting on some Docsy updates before proceeding.

chalin:
Waiting on:

- https://github.com/google/docsy/issues/1271

PR #1841, opened 2022-10-10
Upgrade Docsy to 7489c35

chalin:
- Upgrade Docsy to 7489c35, which is effectively a 0.5.1-dev release.
- This is in support of #1820


chalin:
Further upgrades are needed for Docsy, but for #1820, this is good enough. Merging.

PR #1844, opened 2022-10-11
ruby registry: fix links to instrumentations moved to -contrib repo

tsloughter:


cartermp (approved):
Oh boy, indeed it looks like these moved. Thanks!

PR #1849, opened 2022-10-11
Auto link in stability table

svrnm:
Small improvement for the tables with the stability matrix: if the value is "Stable" or "Experimental" it will automatically link the definition in the spec like some of the pages are doing it already.

### DIFF
--- a/layouts/shortcodes/lang_instrumentation_index_head.md
+++ b/layouts/shortcodes/lang_instrumentation_index_head.md
@@ -1,7 +1,18 @@
 {{ $data := index $.Site.Data.instrumentation.languages (.Get 0) }}
 {{ $name := $data.name }}
 {{ $relUrl := printf "https://github.com/open-telemetry/opentelemetry-%s/releases" (.Get 0) -}}
-
+{{ $tracesStatus := $data.status.traces | humanize }}
+{{ $metricsStatus := $data.status.metrics | humanize }}
+{{ $logsStatus := $data.status.logs | humanize }}
+{{ if in "Stable Experimental" $tracesStatus }}
+    {{ $tracesStatus = printf "[%s](/docs/reference/specification/versioning-and-stability/#%s)" $tracesStatus $data.status.traces }}
+{{ end }}
+{{ if in "Stable Experimental" $metricsStatus }}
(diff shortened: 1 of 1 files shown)


layouts/shortcodes/lang_instrumentation_index_head.md
```diff
@@ -1,7 +1,18 @@
...
+{{ $tracesStatus := $data.status.traces | humanize }}
+{{ $metricsStatus := $data.status.metrics | humanize }}
+{{ $logsStatus := $data.status.logs | humanize }}
+{{ if in "Stable Experimental" $tracesStatus }}
+    {{ $tracesStatus = printf "[%s](/docs/reference/specification/versioning-and-stability/#%s)" $tracesStatus $data.status.traces }}
+{{ end }}
+{{ if in "Stable Experimental" $metricsStatus }}
+    {{ $metricsStatus = printf "[%s](/docs/reference/specification/versioning-and-stability/#%s)" $metricsStatus $data.status.metrics }}
+{{ end }}
+{{ if in "Stable Experimental" $logsStatus }}
```
cartermp, line 13: Should "Development" also be added?

melbgirl:
What will the change look like @svrnm? TIA

svrnm:
> Details

Sure, compare the preview with the current docs on some of the languages where that table is used:

https://deploy-preview-1849--opentelemetry.netlify.app/docs/instrumentation/cpp/
vs
https://opentelemetry.io/docs/instrumentation/cpp/

layouts/shortcodes/lang_instrumentation_index_head.md
svrnm: To what URL should it point then?

https://opentelemetry.io/docs/reference/specification/versioning-and-stability/

knows Stable, Experimental, Deprecated, Removed

cartermp: Not sure, but this is the status for JS + logs, for example.

svrnm: yeah, there are some terms used which are not officially supported, that's why I check for Stable & Experimental and skip the rest.

cartermp (approved):

PR #1851, opened 2022-10-11
Update specification to 1.14.0, and fix broken links

### LINKED ISSUE #1846: All schema files are returning 404
**What happened?**
The schema file URLs that used to work are now returning 404. E.g. https://opentelemetry.io/schemas/1.7.1 or https://opentelemetry.io/schemas/1.14.0

**What did you expect would happen?**
It should be a 200 and return the content of relevant file from https://github.com/open-telemetry/opentelemetry-specification/tree/main/schemas

**What is the name + path of the page where you encountered the issue?**
`https://opentelemetry.io/schemas/*`

**What browser, OS, and platform were you using when you encountered the problem?**
Chrome Version 105.0.5195.125 (Official Build) (x86_64), Mac OS 12.5

chalin:
Updates the spec to v1.14.0 from v1.12.0 (compare spec changes via https://github.com/open-telemetry/opentelemetry-specification/compare/v1.12.0...v1.14.0):

- Closes #1846
- Fixes broken links into the updated spec pages.

Preview:

- https://deploy-preview-1851--opentelemetry.netlify.app/docs/reference/specification
- New schema files:
  -  https://deploy-preview-1851--opentelemetry.netlify.app/schemas/1.13.0
  -  https://deploy-preview-1851--opentelemetry.netlify.app/schemas/1.14.0
- Redirect test:
  - https://deploy-preview-1851--opentelemetry.netlify.app/docs/reference/specification/metrics/datamodel/#temporality

/cc @tigrannajaryan 

---

```console
$ (cd content-modules/opentelemetry-specification && git branch -v | grep \*)
* (HEAD detached at v1.14.0) 8c73c16 Release 1.14.0 (#2851)
```

### DIFF
--- a/content/en/blog/2022/exponential-histograms/index.md
+++ b/content/en/blog/2022/exponential-histograms/index.md
@@ -290,13 +290,13 @@ _A version of this article was [originally posted][] on the New Relic blog._
 [aggregation]: /docs/reference/specification/metrics/sdk/#aggregation
 [type of instrument]: /docs/reference/specification/metrics/api/#histogram
 [histogram-aggregation]:
-  /docs/reference/specification/metrics/sdk/#histogram-aggregation-common-behavior
+  /docs/reference/specification/metrics/sdk/#histogram-aggregations
 [explicit bucket histogram]:
   /docs/reference/specification/metrics/sdk/#explicit-bucket-histogram-aggregation
 [explicit bucket histogram proto]:
   https://github.com/open-telemetry/opentelemetry-proto/blob/724e427879e3d2bae2edc0218fff06e37b9eb46e/opentelemetry/proto/metrics/v1/metrics.proto#L382
 [exponential bucket histogram]:
-  /docs/reference/specification/metrics/sdk/#exponential-histogram-aggregation
+  /docs/reference/specification/metrics/sdk/#exponential-bucket-histogram-aggregation
 [exponential bucket histogram proto]:
   https://github.com/open-telemetry/opentelemetry-proto/blob/724e427879e3d2bae2edc0218fff06e37b9eb46e/opentelemetry/proto/metrics/v1/metrics.proto#L463
 [example code]:
--- a/content/en/docs/instrumentation/python/exporters.md
+++ b/content/en/docs/instrumentation/python/exporters.md
@@ -37,7 +37,7 @@ trace.set_tracer_provider(provider)
 
 Use a [`PeriodicExportingMetricReader`][pemr] to periodically print metrics to the console.
 `PeriodicExportingMetricReader` can be configured to export at a different interval, change the
-[temporality](/docs/reference/specification/metrics/datamodel/#temporality) for each instrument
+[temporality](/docs/reference/specification/metrics/data-model/#temporality) for each instrument
 kind, or change the default aggregation for each instrument kind.
 
 ```python
--- a/layouts/index.redirects
+++ b/layouts/index.redirects
@@ -23,3 +23,6 @@
 {{ $og_image_current := `/img/social/logo-wordmark-001.png` -}}
 
 /featured-background.jpg  {{ $og_image_current }} {{/* homepage og:image used prior to 2022/08 */}}
+
+{{/* Temporary until alias is added to spec - https://github.com/open-telemetry/opentelemetry.io/issues/1850 */ -}}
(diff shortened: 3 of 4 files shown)


chalin:
I'm going to merge now, especially given that I have followup changes to make anyways. If there are any issues with the PR, I can address them at the same time.

tigrannajaryan:
I don't see any approvals. Are you able to merge without approvals? Nothing wrong with this PR, but I want to understand how it was possible to merge.

chalin:
> Are you able to merge without approvals? 

Yes, at my discretion, I can merge without approvals -- in this (website) repo only. 

> Nothing wrong with this PR, 

Good.

> but I want to understand how it was possible to merge.

Ok.

tigrannajaryan:
> Yes, at my discretion, I can merge without approvals -- in this (website) repo only.

I don't think we should be doing this. Our [policies say](https://github.com/open-telemetry/community/blob/main/docs/how-to-configure-new-repository.md#policies) that even Admins should not merge without approvals. Unless we obtained an exception from GC/TC we need to stop the practice.

I also see the entire doc-maintainers team has Admin rights on the repo. This is also against the [current policies](https://github.com/open-telemetry/community/blob/main/docs/how-to-configure-new-repository.md#permissions).

@open-telemetry/technical-committee please chime in. Unless we believe we need an exception for this repo/this team I suggest to apply our regular policies.

austinlparker:
I actually have an open request for parma-admin (https://github.com/open-telemetry/community/issues/1205) that I need to follow up on... generally, though, the website has different needs than one of the code repositories, I'd say? There's a lot more instances of non-functional PRs (like this one - we just need to update the spec on the website to reflect reality) that don't need to wait for multiple approvers. In the past, when we _didn't_ have elevated permissions, these sort of updates would languish because, while OpenTelemetry may be _part_ of the maintainers job, it is not their _only_ job.

In addition, this repo has more regular need for changes to things like branch protection rules or checks as it's more likely to need to modify or update its workflows. Unlike the Java SIG, for example, it could be the case that we may wish to integrate or remove checks somewhat frequently based on changes to the underlying functional website.

cartermp:
I would say that this warrants a post-hoc exception to the rule given the severity of the issue that it resolved + lack of significant potential downstream impacts (it's easy enough to just roll a website back some commits)

reyang:
Sharing something that OpenTelemetry .NET has https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/CONTRIBUTING.md#how-to-get-prs-merged in case it helps:

"Urgent fix can take exception as long as it has been actively communicated."

Aneurysm9:
> "Urgent fix can take exception as long as it has been actively communicated."

Go has similar language, but it only applies to merging with a single approval or before 24h have elapsed from the time the PR was created.  We cannot merge a PR without at least one approval from someone other than the author.  Allowing bypass of the branch protection settings that prevent that seems like a bad idea.

austinlparker:
I mean, this specific situation isn't even "merging something that's unapproved", this change was specifically made to fix a p0 (reported here: https://github.com/open-telemetry/opentelemetry.io/issues/1846) and the only update was to pull in things that were already updated and validated. Would you have preferred that @chalin go poke approvers/maintainers on Slack until one of them rubber-stamped it? If the website is critical infrastructure, then I think allowing for a general exception here is fine when there's a critical issue.

Respectfully, I'll go back to my earlier point - the website isn't the same as other artifacts of the project, and it doesn't make sense to apply the same rationale that you'd apply to other SIG repos. If the concern is that said exception isn't in writing somewhere, then that's solvable, but I'm really just trying to understand the question behind the question at this point. @chalin isn't a volunteer, he's contracted by the CNCF and we're very fortunate that there _is_ somebody who has website maintenance as a for-real paid duty and can get things done when other maintainers aren't available.

melbgirl:
Perhaps there should be discussion and a decision about how we meet these needs. There are a number of cases I can think of where the website being out of date - or wrong - will impact our end users (and potential users) negatively. The risk of merging a website change is also lower impact than it is to the other repos. 
I agree with Philip and Austin’s points above, particularly in the case of resolving a P0.
I could also see that we may want to expand and document  the requirements and responsibilities for those in the Comms SIG inline with their ability to make approval-less commits.

tigrannajaryan:
I made a proposal to clarify the general repo policy here https://github.com/open-telemetry/community/pull/1259
Let's discuss there.

PR #1856, opened 2022-10-12
Change misleading statements in blog post, fixes #1790

### LINKED ISSUE #1790: Misleading statements in the blog post
Hey folks. This is regarding #1721.

Nice blog post. While this does not change the gist of the post, I found your explanation of the actual bug rather confusing.

First, the comparison of `tracestate` as multiple headers vs. `traceparent` is misleading. `tracestate` is a comma-separated list of entries, and since that's how RFC7230 defines header concatenation, it works as you described. But `traceparent` is not a list, it's a single unbreakable string, so the reasoning for header concatenation does not apply. Also your reasoning about Java vs. other SDKs seems backwards - one would argue it was Java SDK that was non-compliant with Trace-Context spec because it was supposed to fail to parse clearly malformed `traceparent`, not just take the first value.

Second, the screenshot shows that your `traceparent` has a list of a whole number of entries, each representing completely different traces. Where were they all coming from? Obviously not from the caller, the caller would only send one trace-id (unless the caller instrumentation was broken). I am guessing the issue may have been some static storage in nginx module where it was accumulating and adding all previously seen traceparent entries to each outbound request, instead of the single entry coming in the inbound request, but the discussion of the blog talks about different things.

Since this is a blog post on the official OTEL website, could you please correct these mis-statements?

cc @svrnm @sanketmehta28 @kpratyus

svrnm:
prettier introduced some more changes, but the relevant part is in line 210 onwards.

@yurishkuro does this sound good to you? TY

### DIFF
--- a/content/en/blog/2022/debug-otel-with-otel/index.md
+++ b/content/en/blog/2022/debug-otel-with-otel/index.md
@@ -217,7 +217,7 @@ Java application. And, without reading into the source code of the OTel Java
 SDK, it looks like that Java is flexible in taking a `traceparent` with multiple
 values, even though such format is invalid per the W3C Trace Context specification. 
 So propagation from nginx to the Java service worked, while in contrast,
-python (and other languages) do not provide that flexibility and propagation
+Python (and other languages) do not provide that flexibility and propagation
 from nginx to the downstream service silenty fails.
 
 Note, that we are not suggesting that the other languages should have the same
--- a/content/en/blog/2022/debug-otel-with-otel/index.md
+++ b/content/en/blog/2022/debug-otel-with-otel/index.md
@@ -215,7 +215,8 @@ headers again and again, and since having multi value headers is covered by
 We tested the capability to correlate from nginx to a downstream service with a
 Java application. And, without reading into the source code of the OTel Java
 SDK, it looks like that Java is flexible in taking a `traceparent` with multiple
-values. So propagation from nginx to the Java service worked, while in contract,
+values, even though such format is invalid per the W3C Trace Context specification. 
+So propagation from nginx to the Java service worked, while in contrast,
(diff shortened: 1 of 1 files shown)


cartermp (approved):
I like the changes. @yurishkuro?

content/en/blog/2022/debug-otel-with-otel/index.md
```diff
@@ -208,15 +208,20 @@ backend:
...
+headers again and again, and since having multi value headers is covered by
+[RFC7230][], this didn't lead to an issue immediately.
 
-Doing some research on that topic, we found out, that this is coved by
-[RFC7230][] and is [allowed and defined][] for `tracestate`, but the [definition
-for `traceparent`][] does not provide details on that corner-case.
+We tested the capability to correlate from nginx to a downstream service with a
+Java application. And, without reading into the source code of the OTel Java
+SDK, it looks like that Java is flexible in taking a `traceparent` with multiple
+values. So propagation from nginx to the Java service worked, while in contract,
```
yurishkuro, line 218:
```suggestion
values, even though such format is invalid per the W3C Trace Context specification. 
So propagation from nginx to the Java service worked, while in contrast,
```

content/en/blog/2022/debug-otel-with-otel/index.md
```diff
@@ -208,15 +208,20 @@ backend:
...
+python (and other languages) do not provide that flexibility and propagation
```
yurishkuro, line 219:
```suggestion
Python (and other languages) do not provide that flexibility and propagation
```

svrnm:
thanks @yurishkuro , I committed your suggestions.

yurishkuro (approved):
thanks!

PR #1861, opened 2022-10-13
Use status table shortcode for python index

svrnm:
Introducing the new short code for python as well, I moved down the links to the API&SDK, but later we should have it like java with a link in the menu

### DIFF
--- a/content/en/docs/instrumentation/python/_index.md
+++ b/content/en/docs/instrumentation/python/_index.md
@@ -8,36 +8,18 @@ description: >-
 aliases: [/python, /python/metrics, /python/tracing]
 weight: 22
 ---
+<!--
+You can see & update the `lang_instrumentation_index_head` shortcode in
+/layouts/shortcodes/lang_instrumentation_index_head.md
 
-This is the OpenTelemetry for Python documentation. OpenTelemetry is an
-observability framework -- an API, SDK, and tools that are designed to aid in
-the generation and collection of application telemetry data such as metrics,
-logs, and traces. This documentation is designed to help you understand how to
-get started using OpenTelemetry for Python.
+The data (name, status) is located at
+/data/instrumentation.yaml
+-->
+{{% lang_instrumentation_index_head "python" %}}
 
 ## Version support
 
 OpenTelemetry-Python supports Python 3.6 and higher.
-
-## Status and Releases
-
-The current status of the major functional components for OpenTelemetry Python
-is as follows:
-
-- **Traces** ([API][api/t], [SDK][sdk/t]): [Stable][]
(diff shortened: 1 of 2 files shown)


content/en/docs/instrumentation/python/_index.md
```diff
@@ -8,37 +8,19 @@ description: >-
...
+/layouts/shortcodes/lang_instrumentation_index_head.md
 
-This is the OpenTelemetry for Python documentation. OpenTelemetry is an
-observability framework -- an API, SDK, and tools that are designed to aid in
-the generation and collection of application telemetry data such as metrics,
-logs, and traces. This documentation is designed to help you understand how to
-get started using OpenTelemetry for Python.
+The data (name, status) is located at
+/data/instrumentation.yaml
+-->
```
chalin, line 17: I'd vote to drop the comment about the shortcode.

content/en/docs/instrumentation/python/_index.md
```diff
@@ -87,14 +69,22 @@ $ pip install -e ./opentelemetry-sdk
 
 ## Learn more
 
-- [API reference][]
+- [API & SDK reference][]
```
chalin, line 72: These are all what we call "API reference" docs:

```suggestion
- [API reference][]
```

chalin: If you agree to this change, you'll need to edit the markdown link definition below as well.

chalin (approved):
See inline comments for suggested changes. Otherwise LGTM.

content/en/docs/instrumentation/python/_index.md
svrnm: Someone asked for having it, so that people who are new to our docs can find it easily, I have no strong preference for keeping or removing it. @cartermp , thoughts?

chalin: To share a bit more about this ... We use shortcodes to avoid duplication and to _encapsulate_ certain information. So, for example, this shortcode will help save in the overhead of updating the opening paragraph text, which was previously duplicated (and still partially is) across multiple files.

But what if you change the location / name / format of the data or datafile, you'll be left with the chore of updating such a comment in every file where the shortcode is used.

It is easy enough, IMHO, to do an GH search for the shortcode name, or, look at the blame/history of a file to see the full context of a change that introduced.

Would you add a comment just before a function call to say which data file it uses inside? Probably not. The same logic applies here IMHO.

svrnm: fair point.

svrnm: removed.

cartermp (approved):

Issue #1073, opened 2022-01-22
.NET Docs: document how to start independent spans


cartermp:
Sourced from this discussion: https://cloud-native.slack.com/archives/C01N3BC2W7Q/p1642882241025300

When creating a new activity or span, you need to explicitly ~not~ associate it with a parent, such as calling this API: https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.activitysource.createactivity?view=net-6.0#system-diagnostics-activitysource-createactivity(system-string-system-diagnostics-activitykind-system-string-system-collections-generic-ienumerable((system-collections-generic-keyvaluepair((system-string-system-object))))-system-collections-generic-ienumerable((system-diagnostics-activitylink))-system-diagnostics-activityidformat)

```csharp
using var act = activitySource.StartActivity("asdf", ActivityKind.Internal, parentId: someId)
```

We should probably document the best way to do this and give it an appropriate scenario description.

NinoFloris:
Relevant https://github.com/dotnet/runtime/issues/65528

PR #1867, opened 2022-10-14
.NET manual: independent and new root docs

cartermp:
This fixes https://github.com/open-telemetry/opentelemetry.io/issues/1073 and brings the manual instrumentation bit up to par with JS.

### DIFF
--- a/content/en/docs/instrumentation/net/manual.md
+++ b/content/en/docs/instrumentation/net/manual.md
@@ -376,7 +376,7 @@ catch (Exception ex)
 
 After you've setup manual instrumentation, you may want to use [instrumentation
 libraries](/docs/instrumentation/net/libraries). Instrumentation libraries will
-instrument relevant libraries you're using and generate data for things like
+instrument relevant libraries you're using and generate spans (activities )for things like
 inbound and outbound HTTP requests and more.
 
 You'll also want to configure an appropriate exporter to [export your telemetry
--- a/content/en/docs/instrumentation/net/manual.md
+++ b/content/en/docs/instrumentation/net/manual.md
@@ -214,7 +214,7 @@ In the preceding example, `childOperation` is ended because the scope of the
 ## Creating independent Activities
 
 The previous examples showed how to create Activities that follow a nested
-heirarchy. In some cases, you'll want to create independent Activities that are
+hierarchy. In some cases, you'll want to create independent Activities that are
 siblings of the same root rather than being nested.
 
 ```csharp
@@ -290,7 +290,7 @@ activity?.SetTag("operation.other-stuff", new int[] { 1, 2, 3 });
 
 An [event](/docs/concepts/signals/traces/#span-events) is a human-readable
 message on an `Activity` that represents "something happening" during its
-lifetime. You can think of it like a primitive log.
+lifetime.
 
 ```csharp
 using var myActivity = MyActivitySource.StartActivity("SayHello");
@@ -375,8 +375,8 @@ catch (Exception ex)
 ## Next steps
 
 After you've setup manual instrumentation, you may want to use [instrumentation
-libraries](/docs/instrumentation/net/libraries). Instrumentation libraries will
-instrument relevant libraries you're using and generate spans (activities )for things like
+libraries](/docs/instrumentation/net/libraries). As the name suggests, they will
+instrument relevant libraries you're using and generate spans (activities) for things like
 inbound and outbound HTTP requests and more.
 
 You'll also want to configure an appropriate exporter to [export your telemetry
--- a/content/en/docs/instrumentation/net/manual.md
+++ b/content/en/docs/instrumentation/net/manual.md
@@ -10,15 +10,15 @@ application.
 ## A note on terminology
 
 .NET is different from other languages/runtimes that support OpenTelemetry.
-[Tracing](/docs/concepts/signals/traces/#tracing-in-opentelemetry) is implemented
-by the
+[Tracing](/docs/concepts/signals/traces/#tracing-in-opentelemetry) is
+implemented by the
 [System.Diagnostics](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics)
 API, repurposing existing constructs like `ActivitySource` and `Activity` to be
 OpenTelemetry-compliant under the covers.
 
 However, there are parts of the OpenTelemetry API and terminology that .NET
-developers must still know to be able to instrument their applications, which are
-covered here as well as the `System.Diagnostics` API.
+developers must still know to be able to instrument their applications, which
+are covered here as well as the `System.Diagnostics` API.
 
 If you prefer to use OpenTelemetry APIs instead of `System.Diagnostics` APIs,
 you can refer to the [OpenTelemetry API Shim docs for tracing]({{< relref "shim"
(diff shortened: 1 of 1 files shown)


content/en/docs/instrumentation/net/manual.md
```diff
@@ -210,6 +211,53 @@ In the preceding example, `childOperation` is ended because the scope of the
...
+
+```csharp
+public static void DoWork()
+{
+    using var currentActivity = MyActivitySource.StartActivity("MyActivity");
+
+    using var sibling1 = MyActivitySource.StartActivity(
+        "Sibling1",
+        ActivityKind.Internal, 
+        parentId: currentActivity?.Id);
```
cartermp, line 228: Should this be using parentId or parentContext? Both are functionally equivalent in my sample app.

content/en/docs/instrumentation/net/manual.md
```diff
@@ -10,15 +10,15 @@ application.
 ## A note on terminology
 
 .NET is different from other languages/runtimes that support OpenTelemetry.
-[Tracing](/docs/concepts/signals/traces/#tracing-in-opentelemetry) is implemented
-by the
+[Tracing](/docs/concepts/signals/traces/#tracing-in-opentelemetry) is
```
cijothomas, line 13:
```suggestion
[Tracing API](/docs/concepts/signals/traces/#tracing-in-opentelemetry) is
```

content/en/docs/instrumentation/net/manual.md
```diff
@@ -210,6 +211,53 @@ In the preceding example, `childOperation` is ended because the scope of the
 `using` block is explicitly defined, rather than scoped to `DoWork` itself like
 `parentOperation`.
 
+## Creating independent Activities
+
+The previous examples showed how to create Activities that follow a nested
+heirarchy. In some cases, you'll want to create inactive Activities that are
```
cijothomas, line 217: inactive?

content/en/docs/instrumentation/net/manual.md
cijothomas: please using context.

content/en/docs/instrumentation/net/manual.md
cartermp: Thanks, fixed

content/en/docs/instrumentation/net/manual.md
cartermp: Done. One thing that sorta sucks is because the system.diagnostics api returns nullable values, this requires handling null. Currently I'm doing it inline rather than check up top or use the unsafe `!` operator.

content/en/docs/instrumentation/net/manual.md
```diff
@@ -210,6 +211,53 @@ In the preceding example, `childOperation` is ended because the scope of the
...
+
+```csharp
+public static void DoWork()
+{
+    using var currentActivity = MyActivitySource.StartActivity("MyActivity");
+
+    using var sibling1 = MyActivitySource.StartActivity(
+        "Sibling1",
+        ActivityKind.Internal, 
+        parentContext: currentActivity is null ? default : currentActivity.Context);
```
cartermp, line 228: This could arguably also be done as a check before creating the other activities.

....although this is really just akwardness with the system.diagnostics API.

content/en/docs/instrumentation/net/manual.md
```diff
@@ -210,6 +211,53 @@ In the preceding example, `childOperation` is ended because the scope of the
 `using` block is explicitly defined, rather than scoped to `DoWork` itself like
 `parentOperation`.
 
+## Creating independent Activities
+
+The previous examples showed how to create Activities that follow a nested
+heirarchy. In some cases, you'll want to create independent Activities that are
```
svrnm, line 217:
```suggestion
hierarchy. In some cases, you'll want to create independent Activities that are
```

content/en/docs/instrumentation/net/manual.md
```diff
@@ -239,9 +288,9 @@ activity?.SetTag("operation.other-stuff", new int[] { 1, 2, 3 });
 
 ## Adding events
 
-An [event](/docs/concepts/signals/traces/#span-events) is a human-readable message
-on an `Activity` that represents "something happening" during its lifetime. You
-can think of it like a primitive log.
+An [event](/docs/concepts/signals/traces/#span-events) is a human-readable
+message on an `Activity` that represents "something happening" during its
+lifetime. You can think of it like a primitive log.
```
svrnm, line 293:
```suggestion
lifetime.
```

content/en/docs/instrumentation/net/manual.md
```diff
@@ -324,14 +374,13 @@ catch (Exception ex)
 
 ## Next steps
 
-After you've setup manual instrumentation, you may want to use
-[instrumentation libraries](/docs/instrumentation/net/libraries). Instrumentation
-libraries will instrument relevant libraries you're using and
-generate data for things like inbound and outbound HTTP requests and more.
+After you've setup manual instrumentation, you may want to use [instrumentation
+libraries](/docs/instrumentation/net/libraries). Instrumentation libraries will
+instrument relevant libraries you're using and generate spans (activities )for things like
```
svrnm, line 379:
```suggestion
instrument relevant libraries you're using and generate spans (activities) for things like
```

content/en/docs/instrumentation/net/manual.md
```diff
@@ -324,14 +374,13 @@ catch (Exception ex)
 
 ## Next steps
 
-After you've setup manual instrumentation, you may want to use
-[instrumentation libraries](/docs/instrumentation/net/libraries). Instrumentation
-libraries will instrument relevant libraries you're using and
-generate data for things like inbound and outbound HTTP requests and more.
+After you've setup manual instrumentation, you may want to use [instrumentation
+libraries](/docs/instrumentation/net/libraries). Instrumentation libraries will
```
svrnm, line 378:
```suggestion
libraries](/docs/instrumentation/net/libraries). They will
```
or
```suggestion
libraries](/docs/instrumentation/net/libraries). As the name suggests, they will
```

content/en/docs/instrumentation/net/manual.md
reyang: Why would one want to do this? It seems to be complicated and error prone.

I think the common scenario should be:

```csharp
using (var parent = DemoSource.StartActivity("Parent"))
{
    using (var child1 = DemoSource.StartActivity("Child1"))
    {
    }
    using (var child2 = DemoSource.StartActivity("Child2"))
    {
    }
    using (var child3 = DemoSource.StartActivity("Child3"))
    {
    }
}
```

reyang (changes requested):
The code example seems to be very complicated. I feel that we should encourage best practices in the doc which helps to create a pit of success, please find my suggestion here https://github.com/open-telemetry/opentelemetry.io/pull/1867/files#r997371774

content/en/docs/instrumentation/net/manual.md
cartermp: Huh, not sure why I didn't think to do it that way. Yes, that's a better code sample. The one downside is it means you can't use `using var` without the bigger API call, but I think that's going to be very unlikely for independent/sibling spans that track independent work in the same scope.

cartermp: I actually used `using var` for the parent in the updated sample, let me know what you think. Happy to also have that one nest as well; I just generally prefer to use more modern C# features in code samples if they fit the scenario.

cartermp:
@reyang Sorry for not pinging, thought I had. Can you take another look to see if the changes are suitable?

cartermp:
@reyang any chance to take another look? Looking to get through our open PRs this week.

chalin (approved):
Minor comments inline. Thx

reyang (approved):
LGTM.

PR #1868, opened 2022-10-16
Add independent and root span section to .NET shim

cartermp:
This is #1867 but also applied to the tracing shim doc.

### DIFF
--- a/content/en/docs/instrumentation/net/shim.md
+++ b/content/en/docs/instrumentation/net/shim.md
@@ -65,8 +65,9 @@ endpoint, you'll have to use a different exporter.
 
 ### ASP.NET Core
 
-To start [tracing](/docs/concepts/signals/traces/#tracing-in-opentelemetry) in an
-ASP.NET Core-based app, use the OpenTelemetry extensions for ASP.NET Core setup.
+To start [tracing](/docs/concepts/signals/traces/#tracing-in-opentelemetry) in
+an ASP.NET Core-based app, use the OpenTelemetry extensions for ASP.NET Core
+setup.
 
 First, ensure that you have the right packages:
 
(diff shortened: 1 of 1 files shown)


svrnm (approved):
lgtm

svrnm:
@open-telemetry/dotnet-approvers please take a look at this.

content/en/docs/instrumentation/net/shim.md
```diff
@@ -242,6 +244,36 @@ In the preceding example, `childSpan` is ended because the scope of the `using`
...
+nested heirarchy. In some cases, you'll want to create independent Spans that
+are siblings of the same root rather than being nested.
+
+```csharp
+public static void DoWork(Tracer tracer)
+{
+    using var sibling1 = tracer.StartSpan("sibling1");
+    using var sibling2 = tracer.StartSpan("sibling2");
+
+    // Both 'sibling1' and 'sibling2' share the same parent
```
cijothomas, line 260: how ?

reyang: I think sibling spans should never overlap.

This is the pattern I would recommend:

```
using (var parent = tracer.StartSpan("parent")
{
    using (var child1 = tracer.StartSpan("child1")
    {
    }

    using (var child2 = tracer.StartSpan("child2")
    {
    }
}
```

reyang (changes requested):
Please find my suggestion https://github.com/open-telemetry/opentelemetry.io/pull/1868/files#r1008249446

content/en/docs/instrumentation/net/shim.md
cartermp: > how ?

In this case I suppose this isn't true because as-written there's no explicit parent, and so both would be root spans. But in general they will share whatever the parent is.

> I think sibling spans should never overlap.

Fair enough, good feedback, will add!

cartermp:
@reyang Adjusted!

reyang (approved):
LGTM.

cijothomas (approved):

PR #1869, opened 2022-10-17
Add metrics to cpp docs

svrnm:
Shamelessly copied from https://github.com/open-telemetry/opentelemetry-cpp/tree/main/examples/metrics_simple

We want to have more docs on Metrics for Kubecon, I think that example already has everything we need for a great start, so I took the chance to have one for C++, hope that's ok @open-telemetry/cpp-approvers 

Preview: https://deploy-preview-1869--opentelemetry.netlify.app/docs/instrumentation/cpp/manual/#metrics

### DIFF
--- a/content/en/docs/instrumentation/cpp/manual.md
+++ b/content/en/docs/instrumentation/cpp/manual.md
@@ -7,7 +7,9 @@ weight: 3
 Manual instrumentation is the process of adding observability code to your
 application.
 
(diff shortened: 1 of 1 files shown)


cartermp (approved):
(in the interest of getting something here, rather than something perfect, I whole-heartedly approve)

svrnm:
TY

PR #1876, opened 2022-10-18
Fix dates for KubeCon NA

svrnm:
cc @alolita @djaglowski 

I visited all the links and reviewed the dates, but if some one could do another round of checks that would be appreciated


djaglowski (approved):

cartermp (approved):

alolita (commented):
Many thanks @svrnm for fixing the event dates. Appreciate it!

Issue #1895, opened 2022-10-24
Broken Registry Links


svrnm:
Some links are broken from within the registry, 301/302 can be replaced with the new target, 404s require either a new link or the entry should be removed.

- [ ] collector-exporter-aws-prometheus-remote-write.md: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/awsprometheusremotewriteexporter, error 404
- [ ] collector-exporter-elastic.md: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/elasticexporter, error 404
- [ ] collector-exporter-jaeger.md: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/jaegerexporter, error 301
- [ ] exporter-go-google-cloud-monitoring.md: https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/tree/master/exporter/metric, error 302
- [ ] exporter-go-google-cloud-trace.md: https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/tree/master/exporter/trace, error 302
- [ ] exporter-java-google-cloud.md: https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/tree/master/exporters/trace, error 302
- [ ] exporter-js-google-cloud-monitoring.md: https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/tree/master/packages/opentelemetry-cloud-monitoring-exporter, error 302
- [ ] exporter-js-google-cloud-trace.md: https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/tree/master/packages/opentelemetry-cloud-trace-exporter, error 302
- [ ] exporter-python-azure: https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/monitor/azure-monitor-opentelemetry-exporter, error 301
- [ ] exporter-python-azure-monitor.md: https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/monitor/azure-monitor-opentelemetry-exporter, error 301
- [ ] exporter-python-datadog.md: https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/exporter/opentelemetry-exporter-datadog, error 404
- [ ] instrumentation-java-akka-context-propagation.md: https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/instrumentation/akka/akka-actor-2.5, error 404
[...]

PR #1896, opened 2022-10-24
Add a script to scan registry repo urls.

svrnm:
Nothing to fancy, just a small script that outputs a list of URLs in the registry that are not working which can be copied into a github issue (https://github.com/open-telemetry/opentelemetry.io/issues/1895)

### DIFF
--- a/scripts/check-registry-urls
+++ b/scripts/check-registry-urls
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+REPO_DIR=$(dirname $0)/..
+
+REGISTRY_FILES=$REPO_DIR/content/en/registry/*
+
+for FILE in ${REGISTRY_FILES} ; do
+    REPO_URL=$(grep "repo: " ${FILE})
+    REPO_URL=${REPO_URL#repo: }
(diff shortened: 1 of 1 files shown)


cartermp:
Should we fix/comment out these in one sweep and turn on external link checks for CI?

svrnm:
> Should we fix/comment out these in one sweep and turn on external link checks for CI?

I would have preferred to fix them where possible, but yes, we could just throw them out (at least those with a 40x error)

cartermp:
Yeah. I'm okay with pulling this in as a link checker utility, but I'd like to use the blunt instrument of a CI link checker at some point soon, provided we fix/remove bad URLs.

svrnm:
> Yeah. I'm okay with pulling this in as a link checker utility, but I'd like to use the blunt instrument of a CI link checker at some point soon, provided we fix/remove bad URLs.

The idea behind this was exactly that, have something intermediate for now like this script we can run manually once in a while and later something more automatic that does the job for us. I thought I just put the script into the repo so anyone can use it besides me.

cartermp (approved):

PR #1903, opened 2022-10-24
Update banner to point at demo GA release

cartermp:
GC election is all done, and the GA release of the demo is far more interesting!


svrnm (approved):
yeah

PR #1913, opened 2022-10-25
Batch add registry entries for collector exporters

svrnm:
And the third one, adding all the missing exporters.

Preview: https://deploy-preview-1913--opentelemetry.netlify.app/registry/?language=collector&component=exporter

(Still has some duplicates, need to fix that)


svrnm:
fixed the duplicates, again I moved existing files to the common naming pattern, so the scanner will not fail

cartermp (approved):

PR #1929, opened 2022-10-27
Improve linkTitles for .NET

chalin:
The original intent of this PR was to propose improved linkTitles (that show up in the sidenav) for two pages. Here are the linkTitle changes

- _.NET Framework instrumentation configuration_
  - Old linkTitle: .NET Framework
  - New linkTitle: Framework config
  - Rational: we're already under the .NET section so the prefix ".NET" isn't that helpful in a short linkTitle. Added config to the linkTitle (and renamed the page to match), since that's what the page seems to be about
- _OpenTelemetry Tracing Shim_
  - Old linkTitle: OpenTelemetry API Shim
  - New linkTitle: Tracing Shim
  - Rational: the entire site is about OTel so the prefix "OpenTelemetry" isn't that helpful. "Tracing Shim" seems like the most appropriate short title.

Also did other minor front-matter cleanup and fixed capitalization.

### DIFF
--- a/content/en/docs/instrumentation/net/framework-config.md
+++ b/content/en/docs/instrumentation/net/framework-config.md
@@ -1,6 +1,7 @@
 ---
-title: ".NET Framework instrumentation configuration"
-linkTitle: ".NET Framework"
+title: .NET Framework instrumentation configuration
+linkTitle: Framework config
+aliases: [/docs/instrumentation/net/netframework]
 weight: 5
 ---
 
@@ -50,7 +51,7 @@ public class WebApiApplication : HttpApplication
     {
         _tracerProvider = Sdk.CreateTracerProviderBuilder()
             .AddAspNetInstrumentation()
-            
+
             // Other configuration, like adding an exporter and setting resources
             .AddConsoleExporter()
             .AddSource("my-service-name")
--- a/content/en/docs/instrumentation/net/shim.md
+++ b/content/en/docs/instrumentation/net/shim.md
@@ -1,6 +1,6 @@
 ---
 title: OpenTelemetry Tracing Shim
-linkTitle: OpenTelemetry API Shim
+linkTitle: Tracing Shim
 weight: 5
 ---
 
@@ -167,7 +167,7 @@ If you're not using ASP.NET Core or would rather not inject an instance of a
 var tracer = tracerProvider.GetTracer(serviceName);
 
 // Assign it somewhere globally
- 
+
 //...
 ```
 
--- a/content/en/docs/instrumentation/net/exporters.md
+++ b/content/en/docs/instrumentation/net/exporters.md
@@ -1,5 +1,5 @@
 ---
-title: "Exporters"
+title: Exporters
 weight: 4
 ---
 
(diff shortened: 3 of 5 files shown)


content/en/docs/instrumentation/net/framework-config.md
```diff
@@ -1,6 +1,7 @@
 ---
-title: ".NET Framework instrumentation configuration"
-linkTitle: ".NET Framework"
+title: .NET Framework instrumentation configuration
+linkTitle: Framework config
+aliases: [/docs/instrumentation/net/netframework]
```
chalin, line 4: Note that this file has been renamed. This alias is to the old page's path.

content/en/docs/instrumentation/net/framework-config.md
```diff
@@ -1,6 +1,7 @@
 ---
-title: ".NET Framework instrumentation configuration"
-linkTitle: ".NET Framework"
+title: .NET Framework instrumentation configuration
+linkTitle: Framework config
```
Kielek, line 3: I would say, that it is rare to use  just Framework even if the context is known. It is usually used as a full proper name _.NET Framework_.

content/en/docs/instrumentation/net/shim.md
```diff
@@ -1,6 +1,6 @@
 ---
 title: OpenTelemetry Tracing Shim
-linkTitle: OpenTelemetry API Shim
+linkTitle: Tracing Shim
```
pellared, line 3: What is the plan regarding docs for metrics? Do you want to create a separate page like `Metrics Shim`?

cartermp: Is there an otel api -compatible shim like there is for tracing? If so then perhaps, although it might be more like an "opentelemetry API shim" doc instead

content/en/docs/instrumentation/net/framework-config.md
cartermp: Yes, this should say `.NET Framework`. It refers to a different runtime/toolset (.NET is a family of runtimes, tools, compilers, etc.)

content/en/docs/instrumentation/net/shim.md
pellared: I thought there, but it seems I was mistaken 😬 
Reference: https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/src/OpenTelemetry.Api/Metrics
I am resolving this comment.

content/en/docs/instrumentation/net/framework-config.md
chalin: Got it, thanks!

So should I just revert all changes to this file @cartermp?
Or maybe just close the PR?

cartermp: I would say revert the changes to this file and it should be good. If it's easier to just do another PR then that's good too 🙂

chalin: Ok, so I've reverted the title and linkTitle changes.

As for the website URL, would it make sense to use the following:

- `.../docs/instrumentation/net/framework`

Or do you want to keep the **_stuttered_** `net` that we had before, namely:

- <code>.../docs/instrumentation/<b>net/net</b>framework</code>

cartermp: I'd prefer the latter. Even though it seems kinda weird, `netframework` is unambiguous about what it supports, whereas `framework` can be ambiguous.

chalin:
Changes to `content/en/docs/instrumentation/net/netframework.md` have been reverted. PTAL @cartermp et al.

chalin:
@pellared - is this good to go now?

Kielek (approved):

PR #1932, opened 2022-10-27
Remove "automatic" in JS Instrumentation page

svrnm:
"Automatic Instrumentation Libraries" to "Instrumentation Libraries"


chalin (approved):
LGTM, but if the code excerpts come from files that are in some other repo, it might be a good idea to change the original sources so that when the code excerpts are updated here, we don't lose these edits. WDYT @svrnm?

cartermp (approved):

svrnm:
> LGTM, but if the code excerpts come from files that are in some other repo, it might be a good idea to change the original sources so that when the code excerpts are updated here, we don't lose these edits. WDYT @svrnm?

yes, let me check if there's some place where this is coming from.

Issue #1808, opened 2022-10-05
Consider how best to surface API references for each language


cartermp:
Today, if you want to get to an API reference for a language's SDK, we have a link near the bottom of each language's index page on the website if there's something to link to. Example: https://opentelemetry.io/docs/instrumentation/js/#further-reading

While this technically fulfills the need to link to this kind of material, it's not terribly discoverable despite being fairly important for anyone doing manual instrumentation.

Instead, we should figure out a way, for each language, to make the API reference more prominent so that end-users can more easily find them.

It is not currently a goal to host or design a given language's API reference. If a SIG would prefer us to design and host their API docs we can consider what that may look like, but given that most languages have ways to generate reference docs themselves, and end-users often expect reference material to look a certain way (e.g., Java developers and javadoc), this is out of scope for now.

---

**Edit** (@chalin and @cartermp)

- [ ] Collector?
- [x] C++ - #1928
- [x] Erlang/Elixir: https://github.com/open-telemetry/opentelemetry.io/pull/1943
- [x] Go
  - [x] https://github.com/open-telemetry/opentelemetry-go/pull/3280
  - [x] Merge in updated Go content-module - #1882
- [x] Java API: - #1859
- [x] JS: https://github.com/open-telemetry/opentelemetry.io/pull/1936
- [x] Python: 
  - [x] #1934
  - [x] #1946
- [x] Ruby: https://github.com/open-telemetry/opentelemetry-ruby/pull/1394
- [x] .NET: https://github.com/open-telemetry/opentelemetry.io/pull/1947
- [x] Link-checker ignore-path regex for placeholder pages across language SIGs: #1950

chalin:
This is how I set things up on grpc.io:

- Each language menu has an (external) **API** reference link (Dart language shown in the screenshot).
- There is also a link to the **API** reference in the language's landing page -- in particular because the side-nav doesn't appear by default on mobile.

> <img width="800" alt="gRPC.io Dart API" src="https://user-images.githubusercontent.com/4140793/194089720-d9b19bd0-e136-435a-a3c9-b659b1e6f170.png">

Would that work for y'all?

cartermp:
Yes, that would be great I think! Will let @svrnm weigh in, though.

chalin:
Can you provide links to the API references for the various languages here? Or should they be relatively easy to find?

svrnm:
Let's go with the GRPC way of doing things, I like that. Here are some languages:

[...]

PR #1936, opened 2022-10-27
Add JS API ref to sidebar

cartermp:
Contributes to https://github.com/open-telemetry/opentelemetry.io/issues/1808

### DIFF
--- a/content/en/docs/instrumentation/js/api.md
+++ b/content/en/docs/instrumentation/js/api.md
@@ -1,0 +1,8 @@
+---
+title: API reference
+linkTitle: API
+redirect: https://open-telemetry.github.io/opentelemetry-js/
+manualLinkTarget: _blank
+_build: { render: link }
+weight: 50
+---
--- a/content/en/docs/instrumentation/js/api/_index.md
+++ b/content/en/docs/instrumentation/js/api/_index.md
@@ -1,7 +1,0 @@
----
-title: API
-description: OpenTelemetry JavaScript API Documentation
-weight: 1
----
-
-Here you can find in-depth explanation of OpenTelemetry JavaScript APIs and concepts.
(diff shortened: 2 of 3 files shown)


svrnm:
I think this should point to https://open-telemetry.github.io/opentelemetry-js-api/, because https://open-telemetry.github.io/opentelemetry-js is for the SDK

svrnm:
> What we need is a single API-reference page that would allow users to visit either of the following links:
> 
> * https://open-telemetry.github.io/opentelemetry-js
> * https://open-telemetry.github.io/opentelemetry-js-api
> 
> That _could_ be setup over https://open-telemetry.github.io, but it might be easier (and more under our control), to keep the **API** page (which you removed in this PR), and from that page offer these two API-reference links. WDYT?

not ideal since it's different from the other languages then, but yes, let's do it that way.

chalin:
> not ideal since it's different from the other languages

Agreed.

Then maybe we can try it the way @cartermp has it here ... because https://open-telemetry.github.io/opentelemetry-js/ does have a link to the "API" API-reference in the page header. Maybe that's enough for now? We give it a try and collect user feedback. WDYT?

cartermp:
I'm down like a clown! And also happy to see if there's a way to improve the rendering of the API reference(s) in the page we link do over in the JS repo.

chalin:
> I'm down like a clown! And also happy to see if there's a way to improve the rendering of the API reference(s) in the page we link do over in the JS repo.

Coolio. If you end up opening an issue over the JS repo for that, please link it here. Thanks!

chalin (approved):
✨

cartermp:
Filed here, we'll see what can be done https://github.com/open-telemetry/opentelemetry-js/issues/3362

svrnm (approved):
👍

Issue #1315, opened 2022-04-25
Document Node.js context propagation


paulius-valiunas:
**What needs to be changed?**
Node.js (or just Javascript) documentation should have at least a short introduction to Context Propagation and how it works in the JS API. Practical usage examples would be nice too.

**What is the name + path of the page that needs changed?**
`instrumentation/js/getting-started/nodejs/`

**Additional context**
Examples:
https://opentelemetry.io/docs/instrumentation/java/manual/#context-propagation
https://opentelemetry.io/docs/instrumentation/ruby/context-propagation/
https://opentelemetry.io/docs/instrumentation/go/manual/#propagators-and-context

cartermp:
What sort of content would you expect to be here, showing how to use B3 propagation instead of W3C?

I mainly ask because in Go, context propagation isn't on by default and so it needs to be explicitly called out.

paulius-valiunas:
> What sort of content would you expect to be here, showing how to use B3 propagation instead of W3C?
> 
> I mainly ask because in Go, context propagation isn't on by default and so it needs to be explicitly called out.

well, neither W3C nor B3 propagation is even mentioned there, so documenting either of them would be nice. I've been told the Express.js auto-instrumentation includes some form of context propagation, but even that wasn't mentioned in the Node.js documentation. I'd say anything that would put someone who doesn't use auto-instrumentation (or if auto-instrumentation is not available for the packages they're using) on the right track would be helpful.

cartermp:
In this case of JS (and most languages), context propagation is one by default ([source](https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-sdk-trace-base/src/BasicTracerProvider.ts#L55-L62)) and it's set to W3C because that's the default format chosen for OpenTelemetry. You may be able to find a way to trace without context propagation, but I think you'd have to really go out of your way to do that.

Would it be helpful to just explicitly call this out?

paulius-valiunas:
wait, so what does this default do, how do I use it? I mean, someone has to take the trace id from HTTP headers or wherever else it's being sent through and feed it to a propagator.

cartermp:
When you instrument either via exclusively automatic instrumentation, or you manually use one of `BasicTracerProvider`/`NodeTracerProvider`/`WebTracerProvider` then context propagation will be on by default (Node and Web both subclass `BasicTracerProvider`). These are the ways we document how to instrument, and what is used in all examples.

[...]

PR #1951, opened 2022-10-29
Add JS context propagation doc

cartermp:
This is a cut at solving https://github.com/open-telemetry/opentelemetry.io/issues/1315. I imagine this could be improved greatly, but it's better than nothing, and that's our bar right now. Heh.

### DIFF
--- a/content/en/docs/instrumentation/js/propagation.md
+++ b/content/en/docs/instrumentation/js/propagation.md
@@ -26,8 +26,8 @@ context.** Although it is possible to propagate context manually, if your system
 uses libraries to communicate between services, use a matching instrumentation
 library to propagate context.
 
-Refer to [Libraries](libraries.md) to learn more about instrumentation libraries
-and how to use them.
+Refer to [Libraries](/docs/instrumentation/js/libraries) to learn more about
+instrumentation libraries and how to use them.
 
 ## Manual context propagation
 
@@ -37,7 +37,7 @@ you're using to have services communicate with one another. Or you many have
 requirements that instrumentation libraries cannot fulfill, even if they exist.
 
 When you must propagate context manually, you can use the [context
-api](context.md).
+api](/docs/instrumentation/js/context).
 
 The following generic example demonstrates how you can propagate trace context
 manually.
@@ -92,5 +92,5 @@ trace.setSpan(activeContext, span)
 From there, when you have a deserialized active context, you can create spans
 that will be a part of the same trace from the other service.
 
-You can also use the [Context](context.md) API to modify or set the deserialized
-context in other ways.
+You can also use the [Context](/docs/instrumentation/js/context) API to modify
+or set the deserialized context in other ways.
--- a/content/en/docs/instrumentation/js/propagation.md
+++ b/content/en/docs/instrumentation/js/propagation.md
@@ -1,0 +1,96 @@
+---
+title: Propagation
+description: Context propagation for the JS SDK
+aliases: [/docs/instrumentation/js/api/propagation]
+weight: 8
+---
+Propagation is the mechanism that moves data between services and processes.
+Although not limited to tracing, it is what allows traces to build causal
+information about a system across services that are arbitrarily distributed
+across process and network boundaries.
+
+## Context propagation with libraries
+
+For the vast majority of use cases, context propagation is done with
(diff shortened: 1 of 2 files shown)


content/en/docs/instrumentation/js/propagation.md
```diff
@@ -0,0 +1,96 @@
...
+// You can then treat this context as the active context for your
+// traces.
+let activeContext = propagation.extract(context.active(), input)
+
+
+let tracer = trace.getTracer('app-name')
+
+let span = tracer.startSpan(spanName, {
+  attributes: {}
+}, activeContext)
```
cartermp, line 86: A possible alternative here could be to use `context.setGlobalContextManager` like in the JS tests? https://github.com/open-telemetry/opentelemetry-js/blob/main/api/test/api/api.test.ts#L55-L64

Would be great to get some @open-telemetry/javascript-approvers thoughts on this.

cartermp:
@open-telemetry/javascript-approvers would love to get a review here.

svrnm:
from a quick read it lgtm, just one additional thought: I am always a big fan of working examples. Now, the problem here is, that it is extremely dangerous to give an example with a real protocol like HTTP, MQ, etc, since people **will** ignore the first part you wrote and copy that code for custom propagation. My idea is that we do an example with the most stupid protocol you can have: writing to a file. So there's app 1 writing to a file, and there's app2 reading that file and we add the propagation headers, etc. into that file. wdyt?

svrnm (approved):

content/en/docs/instrumentation/js/propagation.md
```diff
@@ -0,0 +1,96 @@
...
+const output = {}
+
+// Serialize the traceparent and tracestate from context into
+// an output object.
+//
+// This example uses the active trace context, but you can
+// use whatever context is appropriate to your scenario.
+propagation.inject(context.active(), output)
+
+const { traceparent, tracestate } = output
```
legendecas, line 59: There are no guarantees that the globally registered propagation would inject these fields. We should treat the output as opaque data transportation and send it as is.

content/en/docs/instrumentation/js/propagation.md
legendecas: I would find the current example more self-contained. Context APIs can be documented in their dedicated section.

content/en/docs/instrumentation/js/propagation.md
legendecas: This is a question about which propagator you are using. IIUC, these fields are defined in W3C trace context spec. That is to say, when you are using the Jeager propagator as the globally registered one, you won't get these fields.

cartermp: Aha, gotcha, makes sense. Will think about how to document this.

cartermp:
Okay, I decided to scope this to just W3C and updated the name as such for now. I think in the future we could update this in one of two ways:

* Name the section back to "Context Propagation" and use tabbed content for different formats
* Add a new section for B3 and others

However, since the majority of users use W3C context now (at least as far as I could measure from my employer's telemetry) I am comfortable with scoping it to just that for now and adding stuff afterwards.

cartermp:
@legendecas could you give it another look-over to see if this meets the bar?

content/en/docs/instrumentation/js/propagation.md
```diff
@@ -0,0 +1,96 @@
...
+
+**It is highly recommend that you use instrumentation libraries to progagate
+context.** Although it is possible to propagate context manually, if your system
+uses libraries to communicate between services, use a matching instrumentation
+library to propagate context.
+
+Refer to [Libraries](/docs/instrumentation/js/libraries) to learn more about
+instrumentation libraries and how to use them.
+
+## Manual W3C Trace Context Propagation
```
legendecas, line 32: It might be worth mentioning that official support for the W3C propagator is available (the `W3CTraceContextPropagator` from the package `@opentelemetry/core`) and this section demonstrates manual W3C propagation.

legendecas (approved):
LGTM. Thank you for working on this!

cartermp:
Filed this to track more work: https://github.com/open-telemetry/opentelemetry.io/issues/2054

Issue #1942, opened 2022-10-27
Migrate Ruby docs to this repo


chalin:
Contributes to #730. This is the otel.io portion of the work to be done for:

- https://github.com/open-telemetry/opentelemetry-ruby/issues/1395

Tasks:

- [x] Update otel-ruby submodule on website
- [x] Copy `website-docs` content from submodule into `content/en/docs/instrumentation/ruby` (requires replacing the symlink with a real folder)
- [x] Drop the otel-ruby submodule
- [x] Ensure that building the site generates the same site files
- [x] Adjust `github_repo`-related front-matter so that links into GH work once again
- [ ] [CODEOWNERS](https://github.com/open-telemetry/opentelemetry.io/blob/main/.github/CODEOWNERS): grant @open-telemetry/ruby-approvers and @open-telemetry/ruby-contrib-approvers access to the `ruby` docs folder - https://github.com/open-telemetry/opentelemetry.io/pull/1966

PR #1966, opened 2022-11-03
Add ruby docs approvers to relevant section in CODEOWNERS

cartermp:
Resolves https://github.com/open-telemetry/opentelemetry.io/issues/1942


chalin (approved):
Works for me.

(I'm unsure if we want to insist that all language SIGs follow the same or a similar pattern of permitted code owners?)

cartermp:
Heh, not sure either. I am fine with things how they are right now, though.

PR #1971, opened 2022-11-04
Make span kind description softer and closer the the spec

lmolkova:
Spec language is soft in regards to clients/server span relationships:

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#spankind

Server span does not need to have a parent, it could also be that multiple server instrumentations that are not aware of each other create server->server spans. While not ideal, this is a reality.

Client spans can be nested under other client spans.  E.g.  DB ORM creates a client span and the underlying DB driver creates a client span. They are unaware of each other and have no knowledge if they are both enabled in the same application.

### DIFF
--- a/content/en/docs/concepts/signals/traces.md
+++ b/content/en/docs/concepts/signals/traces.md
@@ -326,7 +326,7 @@ back-end that processes spans can now assign a final status.
 
 ### Span Kind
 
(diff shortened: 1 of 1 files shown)


cijothomas (approved):

cartermp (approved):
Thanks!

PR #1974, opened 2022-11-07
Add Serverless section with AWS Lambda docs 

samimusallam:
Added a new Serverless section in `content/docs/instrumentation/js/` with AWS Lambda documentation.

One thing to note is that I used Serverless Framework for the deployment section since it makes things a lot easier, and supports other cloud vendors (can be used for Azure docs, etc.). Let me know if it makes sense to keep this in the docs.

### DIFF
--- a/content/en/docs/instrumentation/js/serverless.md
+++ b/content/en/docs/instrumentation/js/serverless.md
@@ -11,7 +11,7 @@ The following show how to use OpenTelemetry to instrument AWS Lambda functions t
 
 ### Dependencies
 
-Create an empty package.json:
+First, create an empty package.json:
 
 ```sh
 npm init -y
--- a/content/en/docs/instrumentation/js/serverless.md
+++ b/content/en/docs/instrumentation/js/serverless.md
@@ -86,8 +86,8 @@ More details can be found in the instrumentation
 
 ### AWS Lambda function handler
 
-Now that you have a Lambda wrapper, you can create a simple handler that will
-serve as a Lambda function. Save the following code as `handler.js`.
+Now that you have a Lambda wrapper, create a simple handler that servers
+as a Lambda function. Save the following code as `handler.js`.
 
 ```javascript
 /* handler.js */
--- a/content/en/docs/instrumentation/js/serverless.md
+++ b/content/en/docs/instrumentation/js/serverless.md
@@ -106,7 +106,7 @@ module.exports.hello = async (event) => {
 };
 ```
 
-## Deployment
+### Deployment
 
 There are multiple ways of deploying, but we will be using [Serverless Framework](https://github.com/serverless/serverless)
 for ease of use. Create a file called `serverless.yml`:
--- a/content/en/docs/instrumentation/js/serverless.md
+++ b/content/en/docs/instrumentation/js/serverless.md
@@ -96,14 +96,32 @@ serve as a Lambda function. Save the following code as `handler.js`.
 
 const https = require('https');
 
-module.exports.hello = async (event) => {
-  https.get('https://opentelemetry.io/', res => {
-    return {
-      statusCode: res.statusCode,
-    };
-  }).on('error', err => {
-    console.log('Error: ', err.message);
+function getRequest() {
+  const url = 'https://opentelemetry.io/';
+
+  return new Promise((resolve, reject) => {
+    const req = https.get(url, res => {
+      resolve(res.statusCode);
+    });
+
+    req.on('error', err => {
+      reject(new Error(err));
+    });
   });
+}
+
+exports.handler = async (event) => {
+  try {
+    const result = await getRequest();
+    return {
+      statusCode: result,
+    };
+  } catch (error) {
+    return {
+      statusCode: 400,
+      body: error.message,
+    };
+  }
 };
 ```
 
--- a/content/en/docs/instrumentation/js/serverless.md
+++ b/content/en/docs/instrumentation/js/serverless.md
@@ -94,15 +94,16 @@ serve as a Lambda function. Save the following code as `handler.js`.
 
 "use strict";
 
-const http = require("http");
+const https = require('https');
 
 module.exports.hello = async (event) => {
-  http.get("https://opentelemetry.io/");
-
-  return {
-    statusCode: 200,
-    body: "Success!",
-  };
+  https.get('https://opentelemetry.io/', res => {
+    return {
+      statusCode: res.statusCode,
+    };
+  }).on('error', err => {
+    console.log('Error: ', err.message);
+  });
 };
 ```
 
--- a/content/en/docs/instrumentation/js/serverless.md
+++ b/content/en/docs/instrumentation/js/serverless.md
@@ -17,7 +17,7 @@ Create an empty package.json:
 npm init -y
 ```
 
-Install dependencies used by the example.
+Then install the required dependencies:
 
 ```sh
 npm install \
--- a/content/en/docs/instrumentation/js/serverless.md
+++ b/content/en/docs/instrumentation/js/serverless.md
@@ -128,7 +128,9 @@ exports.handler = async (event) => {
 ### Deployment
 
 There are multiple ways of deploying, but we will be using [Serverless Framework](https://github.com/serverless/serverless)
-for ease of use. Create a file called `serverless.yml`:
+for ease of use. More details can be found in the [Setting Up Serverless Framework guide](https://www.serverless.com/framework/docs/getting-started).
+
+Create a file called `serverless.yml`:
 
 ```yaml
 service: lambda-otel-native
--- a/content/en/docs/instrumentation/js/serverless.md
+++ b/content/en/docs/instrumentation/js/serverless.md
@@ -1,13 +1,15 @@
 ---
-title: AWS Lambda
-aliases: [/docs/instrumentation/js/api/tracing]
-weight: 4
+title: Serverless
+weight: 8
 ---
 
-This guide will show you how to get started with tracing AWS Lambda Nods.js
-functions using OpenTelemetry instrumentation libraries.
+This guide will show you how to get started with tracing serverless functions using OpenTelemetry instrumentation libraries.
 
-### Getting started
+## AWS Lambda
+
+Following the next steps will allow your AWS Lambda functions to use OpenTelemetry to send traces to the configured backend.
+
+### Dependencies
 
 Create an empty package.json:
 
@@ -71,8 +73,8 @@ registerInstrumentations({
 });
 ```
 
-Replace `<backend_url>` with the URL of your favorite backend to send all
-tracing data to it.
+Replace `<backend_url>` with the URL of your favorite backend to export all
+traces to it. If you don't have one setup already, you can check out [Jaeger](https://www.jaegertracing.io/) or [Zipkin](https://zipkin.io/).
 
 Note that `disableAwsContextPropagation` is set to true. The reason for this is
 that the Lambda instrumentation tries to use the X-Ray context headers by
--- a/content/en/docs/instrumentation/js/serverless.md
+++ b/content/en/docs/instrumentation/js/serverless.md
@@ -127,8 +127,13 @@ exports.handler = async (event) => {
 
 ### Deployment
 
-There are multiple ways of deploying, but we will be using [Serverless Framework](https://github.com/serverless/serverless)
-for ease of use. More details can be found in the [Setting Up Serverless Framework guide](https://www.serverless.com/framework/docs/getting-started).
+There are multiple ways of deploying your Lambda function:
+* [AWS Console](https://aws.amazon.com/console/)
+* [AWS CLI](https://aws.amazon.com/cli/)
+* [Serverless Framework](https://github.com/serverless/serverless)
+* [Terraform](https://github.com/hashicorp/terraform)
+
+Here we will be using Serverless Framework, more details can be found in the [Setting Up Serverless Framework guide](https://www.serverless.com/framework/docs/getting-started).
 
 Create a file called `serverless.yml`:
 
--- a/content/en/docs/instrumentation/js/serverless.md
+++ b/content/en/docs/instrumentation/js/serverless.md
@@ -7,7 +7,7 @@ This guide shows how to get started with tracing serverless functions using Open
 
 ## AWS Lambda
 
-Following the next steps will allow your AWS Lambda functions to use OpenTelemetry to send traces to the configured backend.
+The following show how to use OpenTelemetry to instrument AWS Lambda functions that send traces to a configured backend.
 
 ### Dependencies
 
--- a/content/en/docs/instrumentation/js/serverless.md
+++ b/content/en/docs/instrumentation/js/serverless.md
@@ -78,7 +78,7 @@ traces to it. If you don't have one setup already, you can check out [Jaeger](ht
 
 Note that `disableAwsContextPropagation` is set to true. The reason for this is
 that the Lambda instrumentation tries to use the X-Ray context headers by
-default, this results in a non-sampled context, which creates a
+default, unless active tracing is enabled for this function, this results in a non-sampled context, which creates a
 `NonRecordingSpan`.
 
 More details can be found in the instrumentation
--- a/content/en/docs/instrumentation/js/serverless.md
+++ b/content/en/docs/instrumentation/js/serverless.md
@@ -3,7 +3,7 @@ title: Serverless
 weight: 8
 ---
 
-This guide will show you how to get started with tracing serverless functions using OpenTelemetry instrumentation libraries.
+This guide shows how to get started with tracing serverless functions using OpenTelemetry instrumentation libraries.
 
 ## AWS Lambda
 
--- a/content/en/docs/instrumentation/js/serverless.md
+++ b/content/en/docs/instrumentation/js/serverless.md
@@ -7,7 +7,9 @@ This guide shows how to get started with tracing serverless functions using Open
 
 ## AWS Lambda
 
-The following show how to use OpenTelemetry to instrument AWS Lambda functions that send traces to a configured backend.
+The following show how to use Lambda wrappers with OpenTelemetry to instrument AWS Lambda functions and send traces to a configured backend.
+
+Check out [OpenTelemetry Lambda Layers](https://github.com/open-telemetry/opentelemetry-lambda), if you are interested in a plug and play user experience.
 
 ### Dependencies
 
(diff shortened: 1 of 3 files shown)


svrnm:
Preview: https://deploy-preview-1974--opentelemetry.netlify.app/docs/instrumentation/js/serverless/

svrnm (changes requested):

svrnm:
@open-telemetry/docs-approvers , in accordance with our plans to have more docs and not blog posts, I asked @samimusallam to take his blog post on lambda and turn it into a doc.

content/en/docs/instrumentation/js/serverless/_index.md
```diff
@@ -0,0 +1,5 @@
+---
+title: Serverless
+aliases: [/docs/js/serverless]
```
chalin, line 3: This alias should be dropped. The other similar aliases that are present in older pages are there for historical reasons. We don't create new aliases to pages to the non-existent section `/docs/js`.

content/en/docs/instrumentation/js/serverless/aws-lambda.md
```diff
@@ -0,0 +1,146 @@
...
+
+```javascript
+/* handler.js */
+
+"use strict";
+
+const http = require("http");
+
+module.exports.hello = async (event) => {
+  http.get("http://aws.amazon.com");
```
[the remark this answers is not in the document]
samimusallam: Will use opentelemetry.io here

content/en/docs/instrumentation/js/serverless/aws-lambda.md
```diff
@@ -0,0 +1,146 @@
...
+  return {
+    statusCode: 200,
+    body: "Success!",
+  };
+};
+```
+
+## Deployment
+
+There are multiple ways of deploying, but we will be using Serverless Framework
```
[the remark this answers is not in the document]
samimusallam: done :)

content/en/docs/instrumentation/js/serverless/_index.md
samimusallam: Thanks for letting me know! dropped.

content/en/docs/instrumentation/js/serverless.md
```diff
@@ -0,0 +1,148 @@
...
+Install dependencies used by the example.
+
+```sh
+npm install \
+  @opentelemetry/api \
+  @opentelemetry/auto-instrumentations-node \
+  @opentelemetry/exporter-trace-otlp-http \
+  @opentelemetry/instrumentation \
+  @opentelemetry/sdk-trace-base \
+  @opentelemetry/sdk-trace-node
```
svrnm, line 29: I think you don't need all of them because they have some interdependencies, can you verify what the smallest set of packages is, that you really need?

content/en/docs/instrumentation/js/serverless.md
```diff
@@ -0,0 +1,148 @@
...
+
+const http = require("http");
+
+module.exports.hello = async (event) => {
+  http.get("https://opentelemetry.io/");
+
+  return {
+    statusCode: 200,
+    body: "Success!",
+  };
```
svrnm, line 105: this looks surprising, shouldn't there be some dependency between the http.get and the return of that function? like return the statusCode of the http.get in your return object?

svrnm (changes requested):
two more changes

content/en/docs/instrumentation/js/serverless.md
samimusallam: Just verified that this is the smallest number of packages that I need

content/en/docs/instrumentation/js/serverless.md
```diff
@@ -0,0 +1,149 @@
...
+
+module.exports.hello = async (event) => {
+  https.get('https://opentelemetry.io/', res => {
+    return {
+      statusCode: res.statusCode,
+    };
+  }).on('error', err => {
+    console.log('Error: ', err.message);
+  });
+};
```
svrnm, line 107:
```suggestion
module.exports.hello = async (event) => {
  return https.get('https://opentelemetry.io/', res => {
    return {
      statusCode: res.statusCode,
    };
  }).on('error', err => {
    console.log('Error: ', err.message);
  });
};
```
Not a lambda expert, but I assume you need to return something from the handler?

content/en/docs/instrumentation/js/serverless.md
```diff
@@ -0,0 +1,167 @@
...
+  environment:
+    NODE_OPTIONS: --require lambda-wrapper
+functions:
+  lambda-otel-test:
+    handler: handler.hello
+```
+
+For OpenTelemetry to work properly, `lambda-wrapper.js` must be included before
+any other file. That's why we have added the environment variable
+`NODE_OPTIONS:--require lambda-wrapper` which preloads the wrapper at startup.
```
chalin, line 149:
```suggestion
any other file: the `NODE_OPTIONS` setting ensures this.
```

chalin (approved):
See inline comment.

content/en/docs/instrumentation/js/serverless.md
samimusallam: Looks a lot better, thanks! :)

svrnm (approved):

content/en/docs/instrumentation/js/serverless.md
```diff
@@ -0,0 +1,166 @@
...
+api.diag.setLogger(new api.DiagConsoleLogger(), api.DiagLogLevel.ALL);
+
+const provider = new NodeTracerProvider();
+const collectorOptions = {
+  url: "<backend_url>",
+};
+
+const spanProcessor = new BatchSpanProcessor(
+  new OTLPTraceExporter(collectorOptions)
+);
```
cartermp, line 60: Hmmm. I am not sure about this. My understanding is that batches can get lost on AWS Lambda. Don't know about other serverless environments. But I'm not an expert here either.

svrnm: Probably there needs to be a [force flush](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#forceflush) somewhere?

cartermp: I think that can also hurt performance in general. I'd be okay with getting this out there as a first cut but we can't have things documented that drop spans or significantly screw up performance.

Maybe @Aneurysm9 has some suggestions on what to document here?

samimusallam: The BatchSpanProcessor does flush on [shutdown](https://github.com/open-telemetry/opentelemetry-js/blob/27c4f762948c9b8d693d5ce1c2df5aaa5b0e92ee/packages/opentelemetry-sdk-trace-base/src/export/BatchSpanProcessorBase.ts#L99), so it should send all traces before the lambda shutdown.
I'm not sure how big of a performance difference it is compared to a SimpleSpanProcessor for example, but the BatchSpanProcessor is also used in the [OpenTelemetry AWS Lambda Layer](https://github.com/open-telemetry/opentelemetry-lambda/blob/3971af6d6f47e9dc138f9d45448880e3ba8ae37f/nodejs/packages/layer/src/wrapper.ts#L101)

Aneurysm9: It is generally recommended to use the `BatchSpanProcessor` everywhere but in testing where blocking export on span end can be important and tolerated.  With Lambda functions it is generally advised to call `ForceFlush` on the `(Trace|Meter)Provider` at the end of your function invocation to ensure that any telemetry is emitted before the runtime can be frozen.

There are [known issues](https://github.com/open-telemetry/opentelemetry-lambda/issues/224) with telemetry getting dropped by the Lambda layers, but they relate to extension lifecycle management issues in the collector extension layer and should not be affected by choice of span processor in the function space.

content/en/docs/instrumentation/js/serverless.md
```diff
@@ -0,0 +1,166 @@
...
+});
+```
+
+Replace `<backend_url>` with the URL of your favorite backend to export all
+traces to it. If you don't have one setup already, you can check out [Jaeger](https://www.jaegertracing.io/) or [Zipkin](https://zipkin.io/).
+
+Note that `disableAwsContextPropagation` is set to true. The reason for this is
+that the Lambda instrumentation tries to use the X-Ray context headers by
+default, this results in a non-sampled context, which creates a
+`NonRecordingSpan`.
```
Aneurysm9, line 82: I believe this is only true if active tracing is disabled on your function.  When it is enabled then a sampled trace context will be available.

content/en/docs/instrumentation/js/serverless.md
```diff
@@ -0,0 +1,166 @@
...
+  @opentelemetry/exporter-trace-otlp-http \
+  @opentelemetry/instrumentation \
+  @opentelemetry/sdk-trace-base \
+  @opentelemetry/sdk-trace-node
+```
+
+### AWS Lambda wrapper code
+
+This file will contain all the OpenTelemetry logic, which will enable tracing.
+Please save the following code as `lambda-wrapper.js`.
```
Aneurysm9, line 35: This involves reproducing a substantial portion of the OTel Lambda layers.  Should the documentation provide guidance on how to use those layers instead of directing the user to create their own wrapper?

content/en/docs/instrumentation/js/serverless.md
cartermp: Makes sense! Yeah, in this case ensuring there's a call at the end of the function invocation to forceFlush should then suffice.

content/en/docs/instrumentation/js/serverless.md
```diff
@@ -0,0 +1,166 @@
+---
+title: Serverless
+weight: 8
+---
+
+This guide will show you how to get started with tracing serverless functions using OpenTelemetry instrumentation libraries.
+
+## AWS Lambda
+
+Following the next steps will allow your AWS Lambda functions to use OpenTelemetry to send traces to the configured backend.
```
cartermp, line 10:
```suggestion
The following show how to use OpenTelemetry to instrument AWS Lambda functions that send traces to a configured backend.
```

content/en/docs/instrumentation/js/serverless.md
```diff
@@ -0,0 +1,166 @@
...
+
+### Dependencies
+
+Create an empty package.json:
```
cartermp, line 14:
```suggestion
First, creat an empty package.json:
```

content/en/docs/instrumentation/js/serverless.md
```diff
@@ -0,0 +1,166 @@
...
+
+```sh
+npm init -y
+```
+
+Install dependencies used by the example.
```
cartermp, line 20:
```suggestion
Then install the required dependencies:
```

content/en/docs/instrumentation/js/serverless.md
```diff
@@ -0,0 +1,166 @@
...
+default, this results in a non-sampled context, which creates a
+`NonRecordingSpan`.
+
+More details can be found in the instrumentation
+[documentation](https://www.npmjs.com/package/@opentelemetry/instrumentation-aws-lambda).
+
+### AWS Lambda function handler
+
+Now that you have a Lambda wrapper, you can create a simple handler that will
+serve as a Lambda function. Save the following code as `handler.js`.
```
cartermp, line 90:
```suggestion
Now that you have a Lambda wrapper, create a simple handler that servers
as a Lambda function. Save the following code as `handler.js`.
```

content/en/docs/instrumentation/js/serverless.md
```diff
@@ -0,0 +1,166 @@
...
+  @opentelemetry/auto-instrumentations-node \
+  @opentelemetry/exporter-trace-otlp-http \
+  @opentelemetry/instrumentation \
+  @opentelemetry/sdk-trace-base \
+  @opentelemetry/sdk-trace-node
+```
+
+### AWS Lambda wrapper code
+
+This file will contain all the OpenTelemetry logic, which will enable tracing.
```
cartermp, line 34:
```suggestion
This file contains all the OpenTelemetry logic, which enables tracing.
```

content/en/docs/instrumentation/js/serverless.md
cartermp: Hmmm. Would there be any benefit to _not_ configuring and using the layer? I would say yes, we should show that, but I would want a good accounting of the pros and cons.

content/en/docs/instrumentation/js/serverless.md
```diff
@@ -0,0 +1,166 @@
...
+serve as a Lambda function. Save the following code as `handler.js`.
+
+```javascript
+/* handler.js */
+
+"use strict";
+
+const https = require('https');
+
+function getRequest() {
```
cartermp, line 99: Is `async` and `await` possible to use in lambda? Would prefer to show that instead of promises if possible

content/en/docs/instrumentation/js/serverless.md
```diff
@@ -0,0 +1,166 @@
...
+functions:
+  lambda-otel-test:
+    handler: handler.hello
+```
+
+For OpenTelemetry to work properly, `lambda-wrapper.js` must be included before
+any other file: the `NODE_OPTIONS` setting ensures this.
+
+Note if you are not using Serverless Framework to deploy your Lambda function,
+you must manually add this environment variable using the AWS Console UI.
```
cartermp, line 151: This seems to me like we need an additional section for how to properly package and deploy the app for Lambda when not using serverles.

cartermp (changes requested):
Thanks! Left a review pass and some changes that need to be made. Overall I like this content a lot.

content/en/docs/instrumentation/js/serverless.md
samimusallam: I tried making the documentation as simple as possible to get the user started with tracing their Lambdas, letting them know the hows and the whys.
Using the layer is also possible, but it has to be compiled and deployed by the user.

content/en/docs/instrumentation/js/serverless.md
```diff
@@ -0,0 +1,166 @@
...
+      statusCode: 400,
+      body: error.message,
+    };
+  }
+};
+```
+
+### Deployment
+
+There are multiple ways of deploying, but we will be using [Serverless Framework](https://github.com/serverless/serverless)
```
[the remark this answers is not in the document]
samimusallam: I'll link the Serverless Framework's Getting Started guide.
Other than using AWS Console or AWS CLI, I'm not familiar with other deployments methods

content/en/docs/instrumentation/js/serverless.md
samimusallam: I went with this as a reference since I want to wait for the response and return it
https://docs.aws.amazon.com/lambda/latest/dg/nodejs-handler.html#nodejs-handler-async

content/en/docs/instrumentation/js/serverless.md
samimusallam: True, since by default active tracing is not enabled, I've set this option to true

content/en/docs/instrumentation/js/serverless.md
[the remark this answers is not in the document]
svrnm: As long as we clarify that this is _one_ of multiple ways of doing it, I have no issue with being prescriptive in a first version of the doc. Later we still can provide options. @samimusallam besides having the links to the serverless framework maybe call out those other options without documenting them, e.g.

```suggestion
There are multiple ways of deploying your lambda function:

  * AWS Console
  * AWS CLI
  * [Serverless Framework](https://github.com/serverless/serverless)
  * ...
  
Here we will be using Serverless Framework
```

content/en/docs/instrumentation/js/serverless.md
svrnm: Via the [AWS Distro docs](https://aws-otel.github.io/docs/getting-started/lambda/lambda-js) the description _"Users can enable and disable OpenTelemetry for their Lambda function without changing code."_ sounds like we have a "manual vs automatic instrumentation" situation here? 

Agreed with @cartermp we need to have a good understanding of the Pros/Cons of using the one vs the other, that makes it easier to decide if we should have both or only one of them here in the docs. 

If the answer is "both" for the sake of closing this PR we could start with an extra section that calls out the existence of the layer, would that work?

content/en/docs/instrumentation/js/serverless.md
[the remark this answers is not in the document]
samimusallam: I've added a list of other options with their respective links

cartermp (approved):
Looks good to me as a first version of the doc.

content/en/docs/instrumentation/js/serverless.md
Aneurysm9:
```suggestion
Note that `disableAwsContextPropagation` is set to true. The reason for this is
that the Lambda instrumentation tries to use the X-Ray context headers by
default, this results in a non-sampled context unless active tracing is enabled for this function, which creates a
`NonRecordingSpan`.
```

That should be made explicit in the text.  Readers may be confused and think it is not at all possible to get a recording span or to use the AWS context propagation carrier.

content/en/docs/instrumentation/js/serverless.md
Aneurysm9: I think this documentation should at the least call out the existence of the layers and note that it duplicates a subset of their functionality.  It's good to have the DIY option spelled out, but I expect that many users are not going to want to maintain this themselves if an alternative is available.

svrnm: > I think this documentation should at the least call out the existence of the layers and note that it duplicates a subset of their functionality. It's good to have the DIY option spelled out, but I expect that many users are not going to want to maintain this themselves if an alternative is available.

Makes sense. @samimusallam can you put a pointer to the layers for now. In a follow up we can maybe have both in the docs.

content/en/docs/instrumentation/js/serverless.md
samimusallam: You're right, it should be noted. I pushed an update!

content/en/docs/instrumentation/js/serverless.md
samimusallam: I've added a reference to it at the beginning of the docs.

cartermp:
@Aneurysm9 any final thoughts on the latest changes?

Aneurysm9 (approved):
LGTM!

cartermp:
Thanks you @samimusallam!

PR #1976, opened 2022-11-08
Added Crowdstrike

melbgirl:
Falcon Logscale has added native OTLP support

### DIFF
--- a/content/en/vendors.md
+++ b/content/en/vendors.md
@@ -24,6 +24,7 @@ OpenTelemetry in their commercial products.
 | Aspecto                    | Yes               | Yes         | [aspecto.io](https://www.aspecto.io)                                                                                                                  |
 | AWS                        | Yes               | No          | [aws-otel.github.io](https://aws-otel.github.io)                                                                                                      |
 | Azure                      | Yes               | No          | [docs.microsoft.com/...](https://docs.microsoft.com/azure/azure-monitor/app/opentelemetry-overview)                                                   |
+| Crowdstrike                | No                | Yes         | [docs.crowdstrike.com/...](https://library.humio.com/)
 | Datadog                    | Yes               | Yes         | [docs.datadoghq.com/...](https://docs.datadoghq.com/tracing/setup_overview/open_standards)                                                            |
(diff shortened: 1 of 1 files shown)


content/en/vendors.md
```diff
@@ -24,6 +24,7 @@ OpenTelemetry in their commercial products.
 | Aspecto                    | Yes               | Yes         | [aspecto.io](https://www.aspecto.io)                                                                                                                  |
 | AWS                        | Yes               | No          | [aws-otel.github.io](https://aws-otel.github.io)                                                                                                      |
 | Azure                      | Yes               | No          | [docs.microsoft.com/...](https://docs.microsoft.com/azure/azure-monitor/app/opentelemetry-overview)                                                   |
+| Crowdstrike                | No                | Yes         | [docs.crowdstrike.com/...](https://library.humio.com/)
```
chalin, line 27: - `docs.crowdstrike.com` isn't a valid domain name AFAICT, so we shouldn't be using that as the link text
- Also, usually we want to link to an OTel specific landing page, maybe something like: https://library.humio.com/humio-server/log-shippers-other-log-shippers.html#log-shippers-other-log-shippers-otel? WDYT

```suggestion
| Crowdstrike                | No                | Yes         | [library.humio.com/...](https://library.humio.com/humio-server/log-shippers-other-log-shippers.html#log-shippers-other-log-shippers-otel)
```

svrnm: @melbgirl how long will it take until this rebrand is resolved? Maybe you can hold back a few more days? The "short link" and the link below should match, so otherwise just have Crowdstrike left and [library.humio.com] for the time being and update later on.

TBH, I really don't like the way the vendor page looks like today, maybe we can do something better? Maybe remove the "Learn More" link and directly put the link over the vendor name?

chalin: Thanks for the context @melbgirl.
I agree with @svrnm that maybe it's worth waiting a bit.

I agree that the vendor page design could be improved. I don't think that we should drop the Learn-more column and just add the link over the vendor name. IMHO, that would be misleading. I'd vote to keep the "Learn more" or equivalent.

If the display were cards, then we could link the text "Learn more" to the docs.

svrnm:
@melbgirl any updates?

svrnm:
@melbgirl any updates?

cartermp:
Closing due to inactivity, but definitely ping here or resubmit if you'd like to get it updated in accordance with the recommended changes. More than happy to take another look.

PR #1986, opened 2022-11-10
updated swift instrumentation status

bryce-b:
updated swift sdk status

### DIFF
--- a/data/instrumentation.yaml
+++ b/data/instrumentation.yaml
@@ -63,5 +63,5 @@ languages:
     name: Swift
     status:
       traces: stable
-      metrics: in development
+      metrics: experimental
       logs: in development
--- a/data/instrumentation.yaml
+++ b/data/instrumentation.yaml
@@ -63,5 +63,5 @@ languages:
     name: Swift
     status:
       traces: stable
-      metrics: outdated
+      metrics: in development
       logs: in development
--- a/data/instrumentation.yaml
+++ b/data/instrumentation.yaml
@@ -62,6 +62,6 @@ languages:
   swift:
     name: Swift
     status:
-      traces: beta
-      metrics: alpha
-      logs: not yet implemented
+      traces: stable
+      metrics: outdated
+      logs: in development


data/instrumentation.yaml
```diff
@@ -62,6 +62,6 @@ languages:
   swift:
     name: Swift
     status:
-      traces: beta
-      metrics: alpha
-      logs: not yet implemented
+      traces: stable
+      metrics: outdated
```
cartermp, line 66: hmmm, how should I interpret this as an end-user? Is it something I can count on getting up to date soon?

svrnm: for .NET we have a „mixed *“ and then a description what that means, we could do the same here, put a * besides the outdated and a comment on the index page what this means. Not perfect tbh but it’s temporary…

Otherwise you could keep it as „in development“. This was the situation for JS for some time, they also had done some major reworks.

cartermp: Yeah, I would say that if there's nothing special to say about it here, then "in development" would be my preference. @bryce-b any thoughts?

bryce-b: Sure

cartermp (approved):

data/instrumentation.yaml
alolita: "outdated" is confusing. Please use the project's stability guarantee terminology - see https://opentelemetry.io/docs/reference/specification/versioning-and-stability/

svrnm: > "outdated" is confusing. Please use the project's stability guarantee terminology - see https://opentelemetry.io/docs/reference/specification/versioning-and-stability/

So instead of "in development" to "experimental"? We have used "in development" before as well: for me it has the nuance of "we are working on that, but there's not a lot you can _experiment_ with yet", so it's somewhere between "not available" and "experimental"

svrnm: then let's go for experimental

svrnm (approved):

PR #2007, opened 2022-11-15
Collector Version Checker Workflow fixes

### LINKED ISSUE #2006: GH actions / cron jobs should not be run for forks
Cron jobs should run only if executed from the https://github.com/open-telemetry org using a conditional something like the follow: `if: ${{ github.repository_owner == 'open-telemetry' }}`. This avoids issues with GH actions failing for various reasons when run from forks. WDYT @svrnm et al.?

svrnm:
* Run collector checker only daily,
* Run only on fixed repos (fixes #2006)
* check for existing PRs

cc: @chalin

### DIFF
--- a/.github/workflows/auto-update-collector-version.yml
+++ b/.github/workflows/auto-update-collector-version.yml
@@ -3,10 +3,12 @@ name: Check collector version
 on:
   workflow_dispatch:
   schedule:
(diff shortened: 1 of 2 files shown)


chalin (approved):
Nice!

PR #2010, opened 2022-11-15
Add textlint tooling

svrnm:
After playing around with vale.sh for some time, I figured out that textlint might be the better tool for us, since it's config neatly fits into the package.json, so we don't need to add a huge list of additional files. Also there's some capability for "auto-fixing" which is nice if used in the visual code UI.

I still want to build this incrementally, so instead of pulling in a HUGE set of rules I wrote some terminology settings from scratch, mainly following the [CNCF style guide](https://github.com/cncf/foundation/blob/main/style-guide.md), e.g. "open-source" vs "open source" or "K8s" vs "K8S" and the suggested check how CNCF projects prefer their naming (gRPC vs grpc vs GRPC).

There's two issues right now with textlint:
* ~~A bug that it does not catch wrong use of terms at the end of the line (will raise an issue for that with the project)~~ now it works 🤷 
* It works well with markdown but has issues again with hugo 

Additionally I added a cspell.json, so instead of having each and every word ignored in the file we can maintain a global set of words.


cspell.json
```diff
@@ -0,0 +1,15 @@
...
+        "OTLP",
+        "OpenTelemetry",
+        "OTel",
+        "daemonset",
+        "zipkin",
+        "protobuf",
+        "parentbased",
+        "tabpane"
+    ]
+}
```
chalin, line 15: We already have a cSpell file under https://github.com/open-telemetry/opentelemetry.io/blob/main/.vscode/cspell.json. I'd suggest making additions there.

chalin (commented):
See inline comment about cSpell file.

cspell.json
svrnm: missed that! Will fix it.

svrnm:
I removed the changes for the cpp docs

.vscode/cspell.json
```diff
@@ -24,8 +24,11 @@
     "otelcol",
     "otlp",
     "OTLP",
+    "protobuf",
+    "parentbased",
```
chalin, line 28: The string "parentbased" is not a proper spelling, but I suspect that you've added it to the list because it appears as an identifier. Is that correct?

If so, I suggest adding this word locally to the page(s) that contain the identifier. (Search for cSpell and you'll see how I've done this for other pages.

chalin (commented):
See inline comment.

.vscode/cspell.json
svrnm: yes, I thought I can add it because we hav relref and tabpane there as well, although "parentbased" and "parent based" both might appear, so let me remove it.

chalin (approved):

chalin:
Yeah, I was on the fence for tabpane, but it feels ok for it to stay in the global file.

PR #1642, opened 2022-08-22
Update docs to 1.17.0, point to new annotation coordinates

jack-berg:

PR #2016, opened 2022-11-16
Update annotation enablement property

trask:


cartermp (approved):
Change looks good to me -- is this just the case that the changed setting is an older value that's no longer applicable?

trask:
ya, this changed when we moved the annotations package, we missed this particular update in doc PR #1642

PR #2023, opened 2022-11-16
Update first paragraph of Instrumenting page

svrnm:
I fixed the first paragraph of the page on Instrumenting, since the details on repositories are not really relevant to the end user and also some of the information is misleading and again the term "Instrumentation Libraries" is used ambiguous.


cartermp (approved):
I wish I could hit a "super approve" button because that previous paragraph reads to me like "opentelemetry may or may not work, haha, have fun!"

PR #2028, opened 2022-11-18
Add link for Mastodon verification

austinlparker:
This PR adds an empty link to the body in order to verify the OpenTelemetry Mastodon account.


chalin:
Once the verification is done, will this be removed? If not, I'd suggest that we put it somewhere other that that base layout file.

Issue #2029, opened 2022-11-19
Manual Instrumentation in Java is missing dependencies and imports and this makes it a time waster


odupuy:
On https://opentelemetry.io/docs/instrumentation/java/manual The "set up" code below (first and last line) requires a lot of top level dependencies.
```
Resource resource = Resource.getDefault()
...
  .buildAndRegisterGlobal();
```
While you can find them it is a lengthy and very painful process.
Please provide the corresponding maven "top level" dependencies below (verified with `mvn dependency:analyze`). Use a collapsible section if you want to keep the things short and sweet if this was the goal. Note that `opentelemetry-semconv` does not exist (yet) as `1.20.1` but only as `1.20.1-alpha`
```
</project>
    <dependencyManagement>
        <dependencies>
            <dependency>
                <groupId>io.opentelemetry</groupId>
                <artifactId>opentelemetry-bom</artifactId>
                <version>1.20.1</version>
                <type>pom</type>
                <scope>import</scope>
            </dependency>
        </dependencies>
    </dependencyManagement>

    <dependencies>
        <dependency>
            <groupId>io.opentelemetry</groupId>
            <artifactId>opentelemetry-api</artifactId>
        </dependency>
        <dependency>
            <groupId>io.opentelemetry</groupId>
            <artifactId>opentelemetry-sdk</artifactId>
        </dependency>
        <dependency>
            <groupId>io.opentelemetry</groupId>
            <artifactId>opentelemetry-sdk-trace</artifactId>
        </dependency>
        <dependency>
            <groupId>io.opentelemetry</groupId>
            <artifactId>opentelemetry-sdk-metrics</artifactId>
        </dependency>
        <dependency>
            <groupId>io.opentelemetry</groupId>
            <artifactId>opentelemetry-sdk-common</artifactId>
        </dependency>
        <dependency>
            <groupId>io.opentelemetry</groupId>
            <artifactId>opentelemetry-exporter-otlp</artifactId>
        </dependency>
        <dependency>
            <groupId>io.opentelemetry</groupId>
            <artifactId>opentelemetry-context</artifactId>
        </dependency>
        <dependency>
            <groupId>io.opentelemetry</groupId>
            <artifactId>opentelemetry-semconv</artifactId>
            <version>1.20.1-alpha</version>
        </dependency>
    </dependencies>
</project>
```
And add then as well the imports for the setup code (collapsible if needed)
```
import io.opentelemetry.api.OpenTelemetry;
[...]

PR #2030, opened 2022-11-19
Manual setup example in Java is incomplete

odupuy:
See https://github.com/open-telemetry/opentelemetry.io/issues/2029 Manual Instrumentation example in Java is missing dependencies and imports and this makes it a time waster. Both are added as a collapsible section.

### DIFF
--- a/content/en/docs/instrumentation/java/manual.md
+++ b/content/en/docs/instrumentation/java/manual.md
@@ -7,7 +7,6 @@ aliases:
   - /docs/instrumentation/java/manual_instrumentation
 weight: 5
 javaVersion: 1.20.1
-javaVersionSemConv: 1.20.1-alpha
 ---
 
 **Libraries** that want to export telemetry data using OpenTelemetry MUST only
@@ -73,7 +72,7 @@ the most important piece of telemetry source-identifying info.
 ```
 
 ### Gradle
-```
+```kotlin
 dependencies {
     implementation 'io.opentelemetry:opentelemetry-api:{{% param javaVersion %}}'
     implementation 'io.opentelemetry:opentelemetry-sdk:{{% param javaVersion %}}'
--- a/content/en/docs/instrumentation/java/manual.md
+++ b/content/en/docs/instrumentation/java/manual.md
@@ -35,6 +35,87 @@ the most important piece of telemetry source-identifying info.
 
 For example:
 
+<details>
+  <summary>Maven top level dependencies. Click me.</summary>
+  Note that opentelemetry-semconv does not exist (yet) as 1.20.1 but only as 1.20.1-alpha.
+  
+  ```
+  
+  <project>
+      <dependencyManagement>
+          <dependencies>
+              <dependency>
+                  <groupId>io.opentelemetry</groupId>
+                  <artifactId>opentelemetry-bom</artifactId>
+                  <version>1.20.1</version>
+                  <type>pom</type>
+                  <scope>import</scope>
+              </dependency>
+          </dependencies>
+      </dependencyManagement>
+      
+      <dependencies>
+          <dependency>
+              <groupId>io.opentelemetry</groupId>
+              <artifactId>opentelemetry-api</artifactId>
+          </dependency>
+          <dependency>
+              <groupId>io.opentelemetry</groupId>
+              <artifactId>opentelemetry-sdk</artifactId>
+          </dependency>
+          <dependency>
+              <groupId>io.opentelemetry</groupId>
+              <artifactId>opentelemetry-sdk-trace</artifactId>
+          </dependency>
+          <dependency>
+              <groupId>io.opentelemetry</groupId>
+              <artifactId>opentelemetry-sdk-metrics</artifactId>
+          </dependency>
+          <dependency>
+              <groupId>io.opentelemetry</groupId>
+              <artifactId>opentelemetry-sdk-common</artifactId>
+          </dependency>
+          <dependency>
+              <groupId>io.opentelemetry</groupId>
+              <artifactId>opentelemetry-exporter-otlp</artifactId>
+          </dependency>
+          <dependency>
+              <groupId>io.opentelemetry</groupId>
+              <artifactId>opentelemetry-context</artifactId>
+          </dependency>
+          <dependency>
+              <groupId>io.opentelemetry</groupId>
+              <artifactId>opentelemetry-semconv</artifactId>
+              <version>1.20.1-alpha</version>
+          </dependency>
+      </dependencies>
+  </project>
+  
+  ```
+</details>
+
+<details>
+  <summary>Java imports. Click me.</summary>
+  
+  ```java
+  
+  import io.opentelemetry.api.OpenTelemetry;
+  import io.opentelemetry.api.common.Attributes;
+  import io.opentelemetry.api.trace.propagation.W3CTraceContextPropagator;
+  import io.opentelemetry.context.propagation.ContextPropagators;
+  import io.opentelemetry.exporter.otlp.metrics.OtlpGrpcMetricExporter;
+  import io.opentelemetry.exporter.otlp.trace.OtlpGrpcSpanExporter;
+  import io.opentelemetry.sdk.OpenTelemetrySdk;
+  import io.opentelemetry.sdk.metrics.SdkMeterProvider;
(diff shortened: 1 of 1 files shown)


odupuy:
I spent a good 30 minutes on imports and dependencies, then 30 minutes on the fork and finding the right syntax for collapsible.
I hope that it will work for you and will save time to others.

content/en/docs/instrumentation/java/manual.md
```diff
@@ -35,6 +35,87 @@ the most important piece of telemetry source-identifying info.
...
+              <artifactId>opentelemetry-sdk-common</artifactId>
+          </dependency>
+          <dependency>
+              <groupId>io.opentelemetry</groupId>
+              <artifactId>opentelemetry-exporter-otlp</artifactId>
+          </dependency>
+          <dependency>
+              <groupId>io.opentelemetry</groupId>
+              <artifactId>opentelemetry-context</artifactId>
+          </dependency>
```
mateuszrzeszutek, line 85: `opentelemetry-api` includes `opentelemetry-context`, there should be no need to add that dependency manually. 
```suggestion
```

content/en/docs/instrumentation/java/manual.md
```diff
@@ -35,6 +35,87 @@ the most important piece of telemetry source-identifying info.
...
+              <artifactId>opentelemetry-sdk-trace</artifactId>
+          </dependency>
+          <dependency>
+              <groupId>io.opentelemetry</groupId>
+              <artifactId>opentelemetry-sdk-metrics</artifactId>
+          </dependency>
+          <dependency>
+              <groupId>io.opentelemetry</groupId>
+              <artifactId>opentelemetry-sdk-common</artifactId>
+          </dependency>
```
mateuszrzeszutek, line 77: These three are already imported by `opentelemetry-sdk`
```suggestion
```

content/en/docs/instrumentation/java/manual.md
```diff
@@ -35,6 +35,87 @@ the most important piece of telemetry source-identifying info.
 
 For example:
 
+<details>
+  <summary>Maven top level dependencies. Click me.</summary>
```
mateuszrzeszutek, line 39: We always post both maven and gradle examples, e.g. see https://opentelemetry.io/docs/instrumentation/java/#maven

Can you add a gradle sample too, and make it look similar to the other pages? While I think it makes perfect sense to keep imports hidden in a `<details>` tag, we should be consistent with the build system samples.

content/en/docs/instrumentation/java/manual.md
```diff
@@ -35,6 +35,87 @@ the most important piece of telemetry source-identifying info.
...
+  
+  ```
+  
+  <project>
+      <dependencyManagement>
+          <dependencies>
+              <dependency>
+                  <groupId>io.opentelemetry</groupId>
+                  <artifactId>opentelemetry-bom</artifactId>
+                  <version>1.20.1</version>
```
mateuszrzeszutek, line 50: On other pages, we use variables to store the Java SDK version: https://github.com/open-telemetry/opentelemetry.io/blob/main/content/en/docs/instrumentation/java/_index.md?plain=1#L10
There is an automatic process that updates these variables after each Java SDK/agent release.

content/en/docs/instrumentation/java/manual.md
svrnm: @mateuszrzeszutek I am wondering if that section you linked (https://opentelemetry.io/docs/instrumentation/java/#maven) should be moved to the /manual page? It is not needed for automatic instrumentation right? So I find it kind of disconnected, I was not even aware that it's there until recently...

mateuszrzeszutek: Hmm, yeah, it probably should -- although it might be needed in automatic instrumentation if you want to supplement it with your own API usage.
Maybe we should have a small, separate section about using the API together with the agent? E.g. that you need to use `GlobalOpenTelemetry` and so on. 
Definitely out of scope for this PR, though.

svrnm: yes, out of scope, but worth to look into later.

content/en/docs/instrumentation/java/manual.md
odupuy: Regarding the dependencies listed here.

I refer you to https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html
and this comment "Maven also provides [dependency:analyze](https://maven.apache.org/plugins/maven-dependency-plugin/analyze-mojo.html) plugin goal for analyzing the dependencies: it helps making this best practice more achievable."

This is exactly what I did. Your maven file should include all the dependencies which materialize as an "import" in the Java code.
I explicitly ran `mvn dependency:analyze` to see what was missing or extra. If you have never used it, please give it a try.
Then the list present is complete, nothing missing, nothing extra.

content/en/docs/instrumentation/java/manual.md
odupuy: same reply on top level dependencies to declare

content/en/docs/instrumentation/java/manual.md
cartermp: I think the scenario described in that article is exceptionally rare for opentelemetry and unlikely to ever pass. Not only are breaking changes like that simply not an acceptable outcome for the project, but context in particular is such a foundational building block for the API that removing it would be akin to removing the notion of arithmetic in a programming language.

trask: hey @odupuy! I think in these docs it's beneficial to keep things simple and not overwhelm new users with the modular breakdown of the API.

Experienced maven users can always run `dependency:analyze` and add the transitive dependencies if they are familiar with / follow this practice.

odupuy (commented):
It should be now done according to all your comments.

Note that you still have to deal with 2 versions, `javaVersion/1.20.1` for most of the libraries, and `javaVersioSemConv/1.20.1-alpha` for semconv.
The first libraries do not exist as 1.20.1-alpha while semconv does not exist as 1.20.1.

Nothing that I can fix ;-)

cartermp (approved):
I like the updates/changes. @open-telemetry/javascript-maintainers?

legendecas:
@cartermp I assume you were mentioning @open-telemetry/java-maintainers.

cartermp:
Agh, sorry, yea, errant auto-complete!

content/en/docs/instrumentation/java/manual.md
mateuszrzeszutek: Yep, the intention here is to have a minimal maven/gradle example that users can copy-paste to get started with OpenTelemetry. `dependency:analyze` and pinning the transitive dep versions are exercises left to the reader.

content/en/docs/instrumentation/java/manual.md
```diff
@@ -35,6 +37,86 @@ the most important piece of telemetry source-identifying info.
...
+              <artifactId>opentelemetry-sdk</artifactId>
+          </dependency>
+          <dependency>
+              <groupId>io.opentelemetry</groupId>
+              <artifactId>opentelemetry-exporter-otlp</artifactId>
+          </dependency>
+          <dependency>
+              <groupId>io.opentelemetry</groupId>
+              <artifactId>opentelemetry-semconv</artifactId>
+              <version>{{% param javaVersionSemConv %}}</version>
```
mateuszrzeszutek, line 74: WDYT about using the `javaVersion` param for now?
```suggestion
                  <version>{{% param javaVersion %}}-alpha</version>
```
It'll continue to be alpha for a long while anyway.

content/en/docs/instrumentation/java/manual.md
mateuszrzeszutek: @odupuy can you make the maven & gradle examples look similar to other pages (e.g. [this one](https://opentelemetry.io/docs/instrumentation/java/#maven))?

odupuy (commented):
all your comments applied

svrnm:
@open-telemetry/java-maintainers can you please review/approve?

odupuy:
Hi @open-telemetry/java-maintainers 
Anything blocking this to be merged?
Thanks

trask (approved):
thanks @odupuy!

PR #2033, opened 2022-11-21
fix: typescript example application

blumamir:
fixing 2 minor issues I spotted in typescript [Example Application](https://opentelemetry.io/docs/instrumentation/js/getting-started/nodejs/#example-application):

1. `npm install` lists the `express` dependency twice, which is redundant. Removed one of these.
2. when calling `app.listen()` on the express `app`, the arrow function that prints a message to log should actually be a second parameter to the `listen` function. The parentheses are currently closed too early (after the `PORT` parameter), causing no message to be output to the console as intended.
3. `parseInt` function on the `PORT` is invoked in different patterns in js and ts examples (when extracting env variable in ts, and when invoking `app.listen` in js). I aligned them so the call is done in the same place for both examples

### DIFF
--- a/content/en/docs/instrumentation/js/getting-started/nodejs.md
+++ b/content/en/docs/instrumentation/js/getting-started/nodejs.md
@@ -29,7 +29,7 @@ Install dependencies used by the example.
 {{< tabpane lang=shell >}}
 
 {{< tab TypeScript >}}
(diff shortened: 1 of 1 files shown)


legendecas (approved):

svrnm (approved):
lgtm

PR #2034, opened 2022-11-21
Update custom-collector.md

andrewsjg:
The example config fails (at least on ARM Macs) unless the collector version is at least v0.64.0


cartermp (approved):
Thanks!

PR #2038, opened 2022-11-21
Add brief note about using --prerelease flag

reese-lee:
This PR adds a brief note about the use of the `--prerelease` flag when adding OTel instrumentation packages via the CLI. For additional context, see this [Slack thread](https://cloud-native.slack.com/archives/C02UN96HZH6/p1669067199829659).

PR #2039, opened 2022-11-21
Add note about using the --prerelease flag in Manual section

reese-lee:
Opened a separate PR to add a note about using the `--prerelease` flag here in case it's considered redundant due to https://github.com/open-telemetry/opentelemetry.io/pull/2038/files, which adds the same note to the [Getting Started page](https://opentelemetry.io/docs/instrumentation/net/getting-started/#aspnet-core). Alternatively, could link back to the Getting Started section if verbosity/repetition is a concern.

### DIFF
--- a/content/en/docs/instrumentation/net/manual.md
+++ b/content/en/docs/instrumentation/net/manual.md
@@ -83,7 +83,11 @@ dotnet add package OpenTelemetry.Extensions.Hosting --prerelease
 dotnet add package OpenTelemetry.Exporter.Console --prerelease
 ```
 
-And then configure it in your ASP.NET Core startup routine where you have access
+Note that the `--prerelease` flag is required for all instrumentation packages 
+as they are all are pre-release, and using the flag is the only way to get the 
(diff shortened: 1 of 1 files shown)


content/en/docs/instrumentation/net/manual.md
```diff
@@ -83,7 +83,11 @@ dotnet add package OpenTelemetry.Extensions.Hosting --prerelease
 dotnet add package OpenTelemetry.Exporter.Console --prerelease
 ```
 
-And then configure it in your ASP.NET Core startup routine where you have access
+Note that the `--prerelease` flag is required for all instrumentation packages 
+as they are all are pre-release, and using the flag is the only way to get the 
+latest version via the CLI.
```
cartermp, line 88:
```suggestion
Note that the `--prerelease` flag is required for all instrumentation packages 
because they are all are pre-release.
```

cartermp (approved):
Similar comment as in your other PR - LGTM though!

PR #2043, opened 2022-11-22
fix instrumentation example json output

rahulsalgare:
1. fields like name, http.target, http.route had wrong value '/roll' which should be '/rolldice' 
2. updated resource object structure


cartermp (approved):
Thanks!

PR #2055, opened 2022-11-28
Fix bug in check-collector-version.sh

svrnm:
🙄


cartermp (approved):
Nothing like string interpolation syntax being different from language to language.

Issue #1807, opened 2022-10-05
Should we have a page for "Other Languages"?


svrnm:
There are a few "inofficial" implementations of OpenTelemetry out there, I saw some for perl, haskell, deno, etc.

I am wondering if we should have a page http://localhost:1313/docs/instrumentation/other/ that calls out that opentelemetry is not limited to the languages we have, that the community is open to add further languages and that there are "inofficial" implementations (and we link them, or we add them to the registry and point to that?)

cartermp:
I think so, yes. This is anecdotal, but the company I work for actually has more customers using Haskell + OTel than Rust + OTel even though Haskell is not an officially-supported language!

chalin:
We can always start by listing **all** of the unofficial implementations on the [Instrumentation](https://opentelemetry.io/docs/instrumentation/) page and then promote them progressively to their own language-landing page. 

We might need some (at least minimal) guidelines for deciding which languages get promoted.
The GC might need to be involved? WDYT @austinlparker?

svrnm:
Related: https://github.com/open-telemetry/opentelemetry.io/pull/1922

svrnm:
Also related (Lua): https://github.com/open-telemetry/community/issues/1276

svrnm:
Also related, dart: https://github.com/open-telemetry/community/issues/552#issuecomment-1311135028

svrnm:
perl https://github.com/open-telemetry/community/issues/828

trask:
Julia: https://github.com/open-telemetry/community/issues/898

chalin:
Can this issue be closed @svrnm, now that the "Other languages" page has been published?

svrnm:
yes, we can close this, ty

PR #2062, opened 2022-11-30
Add page for other languages

svrnm:
Addresses #1807

Preview: 
* https://deploy-preview-2062--opentelemetry.netlify.app/docs/instrumentation/other/
* Registry examples: [Perl](https://deploy-preview-2062--opentelemetry.netlify.app/registry/?s=perl&component=&language=), [Julia](https://deploy-preview-2062--opentelemetry.netlify.app/registry/?s=julia&component=&language=)

copying your comments from there @chalin:

> For unofficial implementations that are sufficiently well developed, we could grant them their own page.

This sounds like a potential step 2, but we need to see if those sufficiently well developed projects move to become an "official" implementation eventually.

> For all others, I'd mention them on the [Instrumentation](https://opentelemetry.io/docs/instrumentation/) page.

Putting it there prominently enough without taking space for other important concerns was not possible, so I created a standalone page

> We can always start by listing all of the unofficial implementations on the [Instrumentation](https://opentelemetry.io/docs/instrumentation/) page and then promote them progressively to their own language-landing page.

I thought about having them in the registry first. This is already a place where we list inofficial things.

> We might need some (at least minimal) guidelines for deciding which languages get promoted.
The GC might need to be involved? WDYT @austinlparker?

I solved that by pointing out that there are already 3 issues for languages that want to have a SIG and saying that if you have interest, please raise another one.

### DIFF
--- a/content/en/docs/instrumentation/other/_index.md
+++ b/content/en/docs/instrumentation/other/_index.md
@@ -10,7 +10,7 @@ Implementing the OpenTelemetry specification is not limited to the languages you
 will find in our documentation. OpenTelemetry is designed in a way that it is
 possible to implement it in every language you like.
 
-For some languages inofficial implementations exist, you can find them in the
+For some languages unofficial implementations exist, you can find them in the
 [registry](/registry). If you know about an implementation not listed there,
 please [add it to the registry][].
 
--- a/content/en/docs/instrumentation/other/_index.md
+++ b/content/en/docs/instrumentation/other/_index.md
@@ -1,0 +1,29 @@
+---
+linkTitle: Other
+title: Other languages
+weight: 99
+description: >-
+  Language-specific implementation of OpenTelemetry for other languages.
+---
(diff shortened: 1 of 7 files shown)


content/en/docs/instrumentation/other/_index.md
```diff
@@ -0,0 +1,29 @@
...
+weight: 99
+description: >-
+  Language-specific implementation of OpenTelemetry for other languages.
+---
+
+Implementing the OpenTelemetry specification is not limited to the languages you
+will find in our documentation. OpenTelemetry is designed in a way that it is
+possible to implement it in every language you like.
+
+For some languages inofficial implementations exist, you can find them in the
```
tedsuo, line 13:
```suggestion
For some languages unofficial implementations exist, you can find them in the
```

cartermp (approved):
Will approve as-is!

PR #2075, opened 2022-12-03
Update nodejs.md

richseviora:
There isn't a troubleshooting step and if you're coming in here having followed instructions from elsewhere, it is really easy to miss that the example shows you explicitly how to setup the otel logger.


cartermp (approved):
Thanks!

svrnm (approved):
lgtm. 

We had that troubleshooting blog post for Node.JS a long while back, is it worthwhile to link to that for more info or should we copy some more from there into the docs (in a follow up)

pichlermarc (approved):

dyladan (approved):

Issue #2104, opened 2022-12-15
Remove or relocate Mastodon verification link


chalin:
_Originally posted by @chalin in https://github.com/open-telemetry/opentelemetry.io/issues/2028#issuecomment-1320211835_

> Once the verification is done, will this be removed? If not, I'd suggest that we put it somewhere other that that base layout file.

chalin:
If it needs to be kept, does it need to appear in every site file or only the homepage? /cc @austinlparker

austinlparker:
I don't know if it re-checks periodically, but it only needs to be on the homepage.

chalin:
Thanks for confirming @austinlparker. Do you know if it needs to be visible or only present in the homepage HTML?

austinlparker:
It needs to be present... I didn't think it was visible right now? It's just an empty link.

chalin:
Right, I just scratched out my question when I noticed it was empty 🙃. Thx!

svrnm:
- See #2115

chalin:
For more context see, https://github.com/open-telemetry/opentelemetry.io/pull/2115#issuecomment-1357717511.

PR #2115, opened 2022-12-19
Add Mastodon link to footer

svrnm:
Related: #2104

the best place for having the link to the mastodon account would be the footer where we have all those other links.

Only challenge: I can not set the `rel="me"` easily, this is something that probably needs to be fixed in docs, @chalin?

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -81,6 +81,10 @@ params:
         url: https://fosstodon.org/@opentelemetry
         icon: fab fa-mastodon
         desc: Follow us on Mastodon to get the latest news!
+      - name: Twitter
+        url: https://twitter.com/opentelemetry
+        icon: fab fa-twitter
+        desc: Follow us on Twitter to get the latest news!
       - name: Stack Overflow
         url: https://stackoverflow.com/questions/tagged/open-telemetry
         icon: fab fa-stack-overflow
--- a/config.yaml
+++ b/config.yaml
@@ -77,10 +77,10 @@ params:
         url: https://github.com/open-telemetry/community#mailing-lists
         icon: fa fa-envelope
         desc: List of mailing lists that the project uses.
-      - name: Twitter
-        url: https://twitter.com/opentelemetry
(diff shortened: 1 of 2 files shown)


chalin:
> the best place for having the link to the mastodon account would be the footer where we have all those other links.

Agreed.

> Only challenge: I can not set the `rel="me"` easily, this is something that probably needs to be fixed in docs, @chalin?

Yes, I can work on that.

config.yaml
```diff
@@ -77,10 +77,10 @@ params:
...
         icon: fa fa-envelope
         desc: List of mailing lists that the project uses.
-      - name: Twitter
-        url: https://twitter.com/opentelemetry
-        icon: fab fa-twitter
-        desc: Follow us on Twitter to get the latest news!
+      - name: Mastodon
+        url: https://fosstodon.org/@opentelemetry
+        icon: fab fa-mastodon
+        desc: Follow us on Mastodon to get the latest news!
```
chalin, line 83: I'd suggest leaving in Twitter, at least initially. You might want to put Mastodon first in the list.

chalin:
I think that there's value in adding the Mastodon link to the footer -- without removing Twitter, and without removing the "verification" link from it's current place. If you'd like @svrnm, I'll deal with #2104 (probably in two steps: first relocating it, then making the enhancement to the Docsy template). WDYT?

config.yaml
svrnm: done :)

svrnm:
> I think that there's value in adding the Mastodon link to the footer -- without removing Twitter, and without removing the "verification" link from it's current place. If you'd like @svrnm, I'll deal with #2104 (probably in two steps: first relocating it, then making the enhancement to the Docsy template). WDYT?

Let's do it that way. Having it as part of the docsy template will then also help other projects to have their mastodon handles verified.

chalin (approved):
✨

cartermp (approved):

PR #2116, opened 2022-12-19
Add cli options & go scan to registry scanner

svrnm:
Some improvements for the registry scanner.


cartermp (approved):
Dig it! Also curious why Go has to be so different

PR #2122, opened 2022-12-19
CONTRIBUTING: document how to upgrade node using nvm

chalin:
When using nvm, whether your are installing the active LTS version of Node afresh, or upgrading, the instructions are the same. Also, we can simplify the command (drop the LTS argument), if you run the install command from within the repo directory. This PR updates the contrib instructions accordingly.

Preview: https://github.com/chalin/opentelemetry.io/blob/chalin-im-contrib-node-upgrade-2022-12-19/CONTRIBUTING.md

/cc @cartermp @svrnm

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -29,12 +29,15 @@ files.
 
 ### Local setup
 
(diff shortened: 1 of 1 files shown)


cartermp (approved):
I dig it.

Issue #2065, opened 2022-12-01
Traces


jlderidder:
It is not clear in this documentation how to discriminate between "trace context", "span context" and "baggage."  Could you please provide explicit differentiation - or else call out which terms are synonymous, and what is the preferred term?

cartermp:
Hello @jlderidder, which documentation are you referring to? We do describe each as independent in these pages:

https://opentelemetry.io/docs/concepts/signals/traces/#trace-context
https://opentelemetry.io/docs/concepts/signals/traces/#span-context
https://opentelemetry.io/docs/concepts/signals/baggage/

The phrase "context" may be overloaded a bit here and I don't think we've cross-linked everyhwere we'd like, though.

Stepping back a bit - is there a particular scenario you're after, and is it confusing which _thing_ is best-suited towards your scenario?

jlderidder:
Thank you for reaching out.
We're trying to clarify our terminology around telemetry at my company, and
I am unable to tell the difference between these.
It appears that they are all information about the event in a span, that is
then shared to provide context.

"Trace Context is metadata about trace spans that provides correlation
between spans across service and process boundaries."
"Baggage refers to contextual information that's passed between spans.
" (appears
to be the same as above)
And Span Context just tells me what's in there, not what it is, though it's
clearly contextual information.

Do you have any definitions that clearly delineate the difference between
these?  Or are the terms simply interchangeable?

Thanks!
--jody

cartermp:
Thanks for the context! I'll see if I can get some docs updates queued up to clarify. For now:

* Baggage a bag of key:value pairs, but its job is to actually propagate arbitrary metadata across services/process/etc. And so if you capture data in ServiceA and want to also have it be present in ServiceB (for whatever reason) it's a way to that.
* Span Context is an object that sits on each span that stores the trace ID, ID of the span it's for, and some other minor information. This is the data that is used to tell which spans are a part of the same trace, and it's propagated across service/process boundaries. The data in here is how tracing "works", basically, and it's not something you use yourself to store data into.

"Trace Context" is referring to the Context object, which is itself also an object that contains some key:value pairs, including data about what the "current/active span" is. While you can technically also use it to inject other data and propagate it, it's not _really_ designed for that and instead you should use Baggage. There are several propagation formats that OpenTelemetry supprts; the defualt (W3C), B3 (from zipkin), and any other arbitrary format that you have a propagator for. In general it's useful to just think of this as "how stuff works behind the scenes".

[...]

PR #2125, opened 2022-12-21
Change/clean up uses of 'Context' in conceptual stuff

cartermp:
This fixes https://github.com/open-telemetry/opentelemetry.io/issues/2065 (or at least tries to)

I removed a section that was sorta useless and I think ultimately confusing because it's describing abstract concepts with slightly different words than concrete ones. Might need some more tweaks. I also clarified some parts of Baggage and tried to get rid of unrelated stuff and the use of the phrase "trace context".

### DIFF
--- a/content/en/docs/concepts/signals/baggage.md
+++ b/content/en/docs/concepts/signals/baggage.md
@@ -5,40 +5,38 @@ description: >-
 weight: 4
 ---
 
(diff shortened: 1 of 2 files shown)


svrnm (approved):
Lgtm, good stuff 👍

PR #2131, opened 2022-12-23
[Java] Add docs to enable only specific instrumentation

tylerbenson:
Also:
- Added docs explaining the mapping of system properties to environment variables. 
- Moved the section for "Enable manual instrumentation only" to be grouped better.
- Added `JAVA_TOOL_OPTIONS` setup instructions.

### DIFF
--- a/content/en/docs/instrumentation/java/automatic/agent-config.md
+++ b/content/en/docs/instrumentation/java/automatic/agent-config.md
@@ -194,8 +194,8 @@ This may be desireable to reduce startup overhead or to have more control of whi
 - Disable all instrumentation in the agent using `-Dotel.instrumentation.common.default-enabled=false`
 (or using the equivalent environment variable `OTEL_INSTRUMENTATION_COMMON_DEFAULT_ENABLED=false`).
 - Enable each desired instrumentation individually using `-Dotel.instrumentation.[name].enabled=true`
-(or using the equivalent environment variable `OTEL_INSTRUMENTATION_[NAME]_ENABLED`) where `name`
-(`NAME`) is the corresponding instrumentation `name` below.
+(or using the equivalent environment variable `OTEL_INSTRUMENTATION_[NAME]_ENABLED`) where `[name]`
+(`[NAME]`) is the corresponding instrumentation `name` below.
 
 ### Enable manual instrumentation only
 
--- a/content/en/docs/instrumentation/java/automatic/agent-config.md
+++ b/content/en/docs/instrumentation/java/automatic/agent-config.md
@@ -197,6 +197,10 @@ This may be desireable to reduce startup overhead or to have more control of whi
 (or using the equivalent environment variable `OTEL_INSTRUMENTATION_[NAME]_ENABLED`) where `[name]`
 (`[NAME]`) is the corresponding instrumentation `name` below.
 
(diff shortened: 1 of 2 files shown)


content/en/docs/instrumentation/java/automatic/agent-config.md
```diff
@@ -173,6 +186,23 @@ span link connecting it to the producer trace.
...
+(or using the equivalent environment variable `OTEL_INSTRUMENTATION_COMMON_DEFAULT_ENABLED=false`).
+- Enable each desired instrumentation individually using `-Dotel.instrumentation.[name].enabled=true`
+(or using the equivalent environment variable `OTEL_INSTRUMENTATION_[NAME]_ENABLED`) where `name`
+(`NAME`) is the corresponding instrumentation `name` below.
+
+### Enable manual instrumentation only
+
+You can suppress all auto instrumentations but have support for manual
+instrumentation with `@WithSpan` and normal API interactions by using
+`-Dotel.instrumentation.common.default-enabled=false -Dotel.instrumentation.opentelemetry-api.enabled=true -Dotel.instrumentation.opentelemetry-instrumentation-annotations.enabled=true`
```
cartermp, line 204: Interesting, so the other two are also considered off if you disable `default-enabled`?

tylerbenson: There are a small subset of integrations that don't follow the "default enabled" flag by manually returning a boolean.  ([For example](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/internal/internal-url-class-loader/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/internal/urlclassloader/UrlClassLoaderInstrumentationModule.java#L23-L26).) 

Otherwise, everything else follows the default config value.

laurit: See https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/7201 we'll probably change it so that these wouldn't be affected by `default-enabled`

content/en/docs/instrumentation/java/automatic/agent-config.md
```diff
@@ -173,6 +186,23 @@ span link connecting it to the producer trace.
...
 equivalent environment variable `OTEL_JAVAAGENT_ENABLED=false`).
 
+### Enable only specific instrumentation
+
+You can disable all default auto instrumentation and selectively re-enable individual instrumentation.
+This may be desireable to reduce startup overhead or to have more control of which instrumentation is applied.
+
+- Disable all instrumentation in the agent using `-Dotel.instrumentation.common.default-enabled=false`
+(or using the equivalent environment variable `OTEL_INSTRUMENTATION_COMMON_DEFAULT_ENABLED=false`).
+- Enable each desired instrumentation individually using `-Dotel.instrumentation.[name].enabled=true`
```
laurit, line 196: Would it make sense to add a warning here that sometimes instrumentations depend on other instrumentations. It is up to the people who use this flag to discover such dependencies and make sure that the instrumentation works as expected.

tylerbenson: Added.

cartermp (approved):
Thanks!

mateuszrzeszutek (approved):

PR #2134, opened 2022-12-28
Update OTel in Practice Meetup group link

codyborders:
The previous link sends users to the DE version of Meetup.


cartermp (approved):
Great catch, looks like we had a localization thingie happen and that wasn't removed in the URL.

PR #2136, opened 2022-12-29
[collector] Use ${env:ENV} style syntax for env variable expansion

mx-psi:
Uses new `${env:ENV}` style syntax for environment variable expansion. This is the recommended way as of recent Collector versions.

Relates to open-telemetry/opentelemetry-collector-contrib/issues/17299

### DIFF
--- a/content/en/docs/collector/configuration.md
+++ b/content/en/docs/collector/configuration.md
@@ -467,8 +467,8 @@ service:
 
 ### Configuration Environment Variables
 
-The use and expansion of environment variables is supported in the Collector
-configuration using PowerShell-style syntax. For example:
+The use and expansion of environment variables is supported in the Collector configuration. 
+For example to use the values stored on the `DB_KEY` and `OPERATION` environment variables you can write the following:
 
 ```yaml
 processors:
--- a/content/en/docs/collector/configuration.md
+++ b/content/en/docs/collector/configuration.md
@@ -468,14 +468,14 @@ service:
 ### Configuration Environment Variables
 
 The use and expansion of environment variables is supported in the Collector
-configuration using shell-style syntax. For example:
+configuration using PowerShell-style syntax. For example:
 
 ```yaml
 processors:
   attributes/example:
     actions:
-      - key: "${DB_KEY}"
-        action: "$OPERATION"
+      - key: "${env:DB_KEY}"
+        action: "${env:OPERATION}"
 ```
 
 Use `$$` to indicate a literal `$`. For example, representing


cartermp (approved):
Thanks!

Do you think we should document this more thoroughly, like showing another example with a key different from `env`? I don't have a real sense for this but I wonder if some folks will get confused by the syntax since they may not be familiar with pwsh.

svrnm:
> Thanks!
> 
> Do you think we should document this more thoroughly, like showing another example with a key different from `env`? I don't have a real sense for this but I wonder if some folks will get confused by the syntax since they may not be familiar with pwsh.

As someone who is not familiar with pwsh I am confused, so yes, let's have another example or link to some external docs that describe that syntax

mx-psi:
I think it makes sense to add other config providers examples (probably after rewording this section to be a bit more general), but would like to hear what @bogdandrutu thinks first

Aneurysm9:
Is this actually PowerShell syntax, or simply something that shares some cosmetic similarities in the same way that the previous examples shared some cosmetic similarities with Bourne Shell expansion syntax but were not?  If it is not actually PowerShell syntax (including defaults, substitutions, slicing, etc), then we should not mention PowerShell.

cartermp:
Hmm, I think I would be fine with saying something like "similar to PowerShell", but in general the structure of the syntax should get documented with at least one other example than `env:SOMETHING`.

mx-psi:
> Is this actually PowerShell syntax, or simply something that shares some cosmetic similarities in the same way that the previous examples shared some cosmetic similarities with Bourne Shell expansion syntax but were not?

Definitely the latter (at least at this point); I assumed if the previous wording was okay then this should also be (since we didn't support defaults, substitution or slicing with the previous format either), but probably the amount of people familiar with PowerShell syntax is much smaller and this is more confusing than helpful.

Does something like *"The use and expansion of environment variables is supported in the Collector configuration. For example to use the values stored on the `DB_KEY` and `OPERATION` environment variables you can write the following:"* sound better to you?

> Hmm, I think I would be fine with saying something like "similar to PowerShell", but in general the structure of the syntax should get documented with at least one other example than env:SOMETHING.

I think that needs a bigger refactor on the docs, I would rather get the syntax updated first on this example and then open a second PR to reword this in terms of providers. I am happy to work on that second PR myself.

Aneurysm9:
> Does something like _"The use and expansion of environment variables is supported in the Collector configuration. For example to use the values stored on the `DB_KEY` and `OPERATION` environment variables you can write the following:"_ sound better to you?

Yes, I could get behind that.  I just don't want to inadvertently give the impression that the collector has capabilities it doesn't.

mx-psi:
Reworded as mentioned on https://github.com/open-telemetry/opentelemetry.io/pull/2136#issuecomment-1367809167

cartermp:
I think this is good enough for now (it could continue to improve but we can handle that later)

cartermp:
Thanks!

PR #2140, opened 2023-01-04
Replace swap scraper with the paging scrape

matej-g:
After playing around with the example configuration, the collector was complaining that the swap scraper for `hostmetrics` receiver does not exist, it might have been the initial naming for the paging scraper. Fixing this in the configuration doc.


cartermp (approved):
Thanks!

PR #2153, opened 2023-01-10
Add semantic conventions to the concepts TOC in docs

cartermp:
Adds it so you can navigate to this very important concept easily.

It should have more commentary and all that, but for now let's surface it more.

### DIFF
--- a/content/en/docs/concepts/semantic-conventions.md
+++ b/content/en/docs/concepts/semantic-conventions.md
@@ -1,0 +1,14 @@
+---
+title: Semantic Conventions
+description: >-
+  Common names for different kinds of operations and data
+weight: 30
+---
+
+OpenTelemetry defines Semantic Conventions (sometimes called Semantic Attributes) that specify common names for different kinds of operations and data. The benefit to using Semantic Conventions is in following a common naming scheme that can be standardized across a codebase, libraries, and platforms.
+
+Currently, Semantic Conventions are split by signal:
+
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/semantic-conventions.md
```diff
@@ -0,0 +1,14 @@
+---
+title: Semantic Conventions
+description: >-
+  Common names for different kinds of operations and data
+weight: 30
+---
+
+OpenTelemetry defines Semantic Conventions (sometimes called Semantic Attributes) that specify common names for different kinds of operations and data. The benefit to using Semantic Conventions is in following a common naming scheme that can be standardized across a codebase, libraries, and platforms.
+
+Currently, Semantic Conventions are split by signal:
```
svrnm, line 10: ppl might get confused that there is a "Resource Semantic Convention" then, maybe

```suggestion
Currently, Semantic Conventions are available for traces, metrics and resources:
```

svrnm (changes requested):

chalin (commented):
Wouldn't it make the most sense to have such a page under `/docs/reference/specification`?

cartermp:
I don't think so, no. Eventually this page should have commentary and link out to the reference as needed. I don't want to start adding docs-like things that summarize and rephrase concepts to the spec. Additionally, I think we lose discovery by making it part of the spec. In general I've found a deep reluctance to dive into the spec when talking to end users.

svrnm (approved):

PR #2166, opened 2023-01-16
Fix some style issues in otlp-exporter-configuration.md

windsonsea:
- Fix typo
- Add some spaces
- For `Example`: if only an example is listed, putting it into a single line might be more natural


svrnm (approved):
lgtm

Issue #2158, opened 2023-01-12
Consolidate 3 demo pages — there should be only one


chalin:
The fact that we have 3 demo pages with the same opening text is a bad smell, not great for SEO, potentially confusing for readers.

For context:

- The page duplication grew out of https://github.com/open-telemetry/opentelemetry.io/issues/2137
- Also see https://github.com/open-telemetry/opentelemetry.io/pull/2139/files#r1064388801

IMHO, it would be preferable to:

- Have a **single demo** page
- If we want the demo page more prominent, we can do so by adding menu entries that link to the canonical page.

Thoughts @svrnm @cartermp?

---

- [x] Eliminate page duplication - #2171
- [x] Consider dropping the /docs demo link - #2174
- [x] Consider adding a link to the demo from the homepage - #2174
- [x] Consider dropping the getting-started demo link in favor of an in-page CTA button - #2257
- ~Update title and/or description to state that the demo is a "reference implementation"~

chalin:
The repo demo title says that it's a "community" app, so the canonical link should be under community. I suggest dropping the top-level entry, and consider adding a link to the home page instead. My 2 cents.

cartermp:
For now I think the community page could be considered canonical, but it's not just a community app - it's a canonical reference implementation of opentelemetry. I think that elevates its importance tremendously, especially for an architect or someone in an IC leadership position at an organization who's tasked with understanding how OTel can fit into their systems more broadly. I would want to at least keep a menu entry under Getting Started for that purpose alone. Top-level is probably also fine I suppose, is there additional concern with having multiple menu entries?

chalin:
> For now I think the community page could be considered canonical

Ok

> it's not just a community app - it's a canonical reference implementation of opentelemetry

- [ ] Ok, that's important to know, and should be added to the page description and/or title.

> I think that elevates its importance tremendously

Agreed.

> I would want to at least keep a menu entry under Getting Started for that purpose alone

Ok.

> Top-level is probably also fine I suppose,

I'd avoid adding it to the Docs top level. Instead, I'd propose adding a link to the demo to either:

- The homepage
- The top nav

I'd vote for the former. Thoughts?

> is there additional concern with having multiple menu entries?

No.

chalin:
FYI, I'm going to address the changes to this issue incrementally. First, I'll eliminate the page duplication, then I can address adding mention of the demo being a "reference implementation", etc.

chalin:
Closed by #2257

PR #2174, opened 2023-01-17
Add demo link to homepage and drop docs placeholder

chalin:
- Contributes to #2158
- Proposes to drop the demo sidebar placeholder under `/docs` in favor of a homepage link to the demo. Part of the rational here is to try to keep the docs top-level as short and coherent as possible. This is just a proposal, feedback is welcome.

### Screenshot

Screen Shot 2023-01-17 at 08 54 05

/cc @cartermp


cartermp (approved):
I love this.

svrnm:
let's do that:-)

PR #2175, opened 2023-01-18
fix(swift): add missing comma in Setup code example

blumamir:
I followed the swift instructions from the OpenTelemetry website and there is a comma missing in the Setup example:

image

From [the swift docs](https://docs.swift.org/swift-book/LanguageGuide/Functions.html):
> You call the greet(person:alreadyGreeted:) function by passing it both a String argument value labeled person and a Bool argument value labeled alreadyGreeted in parentheses, separated by commas

According to the swift website, when calling a function, there should be a comma


nachoBonafonte (approved):

svrnm (approved):

svrnm:
Thanks!

PR #2179, opened 2023-01-18
Introduce prettier workflow

svrnm:
Based on our slack discussion the goal of this is to introduce a workflow that runs prettier on each PR/push. I added an extensive prettierignore file, that we can gradually reduce by files that have been linted.


.github/prettierignore
```diff
@@ -0,0 +1,29 @@
...
+# iconography
+layouts
+# netlify.toml
+package.json 
+# resources
+scripts
+static
+templates
+themes
+tmp
```
TylerHelmuth, line 29:
```suggestion
tmp

```

TylerHelmuth (commented):
It would be nice if we can create a package.json script and/or CONTRIBUTING.md entry on how to run this command locally.  Best case scenario is users are able to run the exact command this workflow is running before submitting the PR.

.github/workflows/prettier.yml
```diff
@@ -0,0 +1,19 @@
...
+    name: Prettier
+    runs-on: ubuntu-latest
+    container:
+      image: node:18
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: "18"
+      - run: npx prettier --ignore-path=./.github/prettierignore -w .
```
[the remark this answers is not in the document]
svrnm: I am figured that out right now .. have to replace the `-w` with a `-c` and bring that output over to the PR view.

chalin (commented):
Thanks for putting this together.

I have a question though: as I had mentioned during one of our Comms meetings, Prettier doesn't always play nice with .md files containing Hugo syntax -- like tabbed panes. Has this been addressed? (You, @svrnm, might have said that that isn't a problem anymore, but I'm not sure.)

.github/prettierignore
```diff
@@ -0,0 +1,29 @@
+ISSUE_TEMPLATE
+PULL_REQUEST_TEMPLATE
+workflows
+dependabot.yml
+.gitpod.yml
+# .netlify
```
chalin, line 6: Shouldn't `. netlify` also be ignored?

chalin:
I agree with @TylerHelmuth that it should be easy for contributors to run Prettier on single files as well as over the entire repo (as is being done in the GH action).

.github/prettierignore
```diff
@@ -0,0 +1,29 @@
+ISSUE_TEMPLATE
```
chalin, line 1: In support of making it easier for contributors to run Prettier themselves, shouldn't this file be at the top-level of the repo, named `.prettierignore`? (Cf. https://prettier.io/docs/en/ignore.html)

.github/prettierignore
```diff
@@ -0,0 +1,29 @@
...
+PULL_REQUEST_TEMPLATE
+workflows
+dependabot.yml
+.gitpod.yml
+# .netlify
+.htmltest.yml
+.vscode
+# CONTRIBUTING.md
+# LICENSE
+#LICENSE-CODE
```
chalin, line 11: I think that we should be ignoring the LICENSE files.

.github/prettierignore
chalin: (I understand that this file's content is preliminary and that the main purpose of this PR is to setup the workflow. So feel free to ignore this and the next comment about tweaking the lines in this file. Though I still think that it should be moved to the top-level and renamed :))

svrnm:
I was so sure that I commented on this issue, anways, I just tried the following with another PR:

https://github.com/open-telemetry/opentelemetry.io/pull/2191/commits/b964a43550c4c296655ea238a4b73962491c75ed

Using that would allow us to force the use of prettier at some point, wdyt?

chalin:
> Using that would allow us to force the use of prettier at some point, wdyt?

Yes, `prettier-ignore-*` feature is just what we need! Note that sometimes there is a way to format markdown pages containing shortcodes so that Prettier plays well, but not always. This ignore feature will allow us to cover the cases where it doesn't play well.

> I didn't create a .prettierignore file, because I wanted to have this seperated from people manually running npx prettier -w on their file.

Eventually we should only ignore files that we really don't want to have checked/reformatted. In which case, having the ignore file at the top level shouldn't prevent users from running it on valid markdown files, etc. My 2 cents.

This is a great start! I'd vote to keep it simple, but get something committed. We can then iteratively improve the checking.

chalin:
For an alternative base config and setup, see the following PR that won't require a separate script:

- #2286

PR #2183, opened 2023-01-19
Add connect-rpc to registry

joshcarp:
Adds https://github.com/bufbuild/connect-opentelemetry-go to registry. 

This instrumentation implements both [grpc trace/metrics](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/rpc.md#grpc) standard plus the [connect rpc](https://github.com/open-telemetry/opentelemetry-specification/pull/3116) conventions which is why it has both `grpc` and `connect` tags


cartermp (approved):
Very nice, thank you!

PR #2186, opened 2023-01-21
Tweak long lines and add some periods

windsonsea:
- Tweak long lines
- Add a period at end if it is missing
- Add blank line before or after a heading


cartermp (approved):
Thanks!

PR #2194, opened 2023-01-24
Update logs.md

atai:
Duplication of a sentence.

### DIFF
--- a/content/en/docs/concepts/signals/logs.md
+++ b/content/en/docs/concepts/signals/logs.md
@@ -1,12 +1,11 @@
 ---
 title: "Logs"
 description: >-
-  A log is a timestamped text record, either structured (recommended) or unstructured, with metadata. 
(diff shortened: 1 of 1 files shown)


chalin (changes requested):
@svrnm @cartermp et al.: we should discuss this change. Every signal page has a similar setup where the page `description` pretty much matches a phrase/sentence in the opening paragraph.

@atai - I agree that such an obvious "duplication" feels odd.

IMHO, we should:

1. Leave the opening paragraphs as they are
2. Rewrite page descriptions.
  A **page description** should describe the page content, not provide a definition of a signal kind.

cartermp:
Yes, agreed with 1 and 2.

chalin:
@atai - thanks again for bringing the problem to our attention! I've created an issue to track this:

- #2203

Closing this PR. @atai, feel free to comment on the issue if you'd like to work on it, otherwise someone else will pick it up. Either way is fine.

Issue #2199, opened 2023-01-25
Accessibility issues with Mastodon rel=me link


Arhell:
**What happened?**
fosstodon.org link need some accessibilit link

/assign

chalin:
Btw, it occurred to me yesterday that even empty or invisible elements can need accessibility attributes. Thanks!

PR #2206, opened 2023-01-26
Address accessibility issues of Mastodon rel=me link

Arhell:
resolve https://github.com/open-telemetry/opentelemetry.io/issues/2199

### DIFF
--- a/layouts/_default/baseof.html
+++ b/layouts/_default/baseof.html
@@ -6,7 +6,7 @@
     {{ partial "head.html" . }}
   </head>
   <body class="td-{{ .Kind }}{{ with .Page.Params.body_class }} {{ . }}{{ end }}">
-    <a rel="me" href="https://fosstodon.org/@opentelemetry"></a>
+    <a rel="me" href="https://fosstodon.org/@opentelemetry" aria-hidden="true" tabindex="-1"></a>
     <header>
       {{ partial "navbar.html" . }}
     </header>


chalin (changes requested):
As I mentioned in the issue, this link isn't visible and so this addition is unnecessary.

layouts/_default/baseof.html
```diff
@@ -6,7 +6,7 @@
     {{ partial "head.html" . }}
   </head>
   <body class="td-{{ .Kind }}{{ with .Page.Params.body_class }} {{ . }}{{ end }}">
-    <a rel="me" href="https://fosstodon.org/@opentelemetry"></a>
+    <a rel="me" href="https://fosstodon.org/@opentelemetry" aria-label="fosstodon.org/@opentelemetry"></a>
```
chalin, line 9: If we must add an ARIA attribut, then this element should be declared as [hidden](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-hidden):

```suggestion
    <a rel="me" href="https://fosstodon.org/@opentelemetry" aria-hidden="true"></a>
```

Arhell: Done thanks

chalin (approved):
Ah yes, we do need to deal with tabbing too. Thanks!

Now that I've had a bit of time to look into this more, I see that further changes are needed. I'll deal with them while working on #2104.

PR #2207, opened 2023-01-26
Add Last9 to vendors list

prathamesh-sonpatki:


cartermp (commented):
Hmmm, I'm not seeing a link to a distro of either an SDK or the collector in the docs for Last9. Can you show where that would be?

prathamesh-sonpatki:
Thanks @cartermp Please check https://docs.last9.io/docs/integration-opentelemetry if that satisfies the requirement.

cartermp:
That wouldn't cover requirements, no, since it's not a vendor distribution of the collector being used and the endpoint does not appear to support OTLP natively.

But it's an interesting point, because the Collector is designed to allow prometheus export, which the endpoint does appear to support. I'm not sure if that warrants inclusion. My gut feel is no, but curious what @chalin and @svrnm think.

chalin:
My initial feeling was "no" as well, but we may need to discuss and clarify admission criteria.

cartermp:
Yeah, I don't think we have any, heh

svrnm:
> That wouldn't cover requirements, no, since it's not a vendor distribution of the collector being used and the endpoint does not appear to support OTLP natively.

TBH I am not sure if this applies to all the vendors we have in that list. We really need to fix that vendors page (and the integrators, adopters, etc.). I will try to find some time to come up with a suggestion

cartermp:
I will say no to this for now, as it's not using OTLP.

PR #2213, opened 2023-01-27
Add BullMQ instrumentation to the registry

jenniferplusplus:
A few months ago I wrote an auto-instrumentation for the BullMQ javascript package. It's been in regular use by several projects since then, and I'd like to add it to the registry to hopefully make it easier for others to find.


cartermp (approved):
Thanks!

PR #2243, opened 2023-02-01
Python remove Mypy work around from the docs

jenshnielsen:
With the merge of https://github.com/open-telemetry/opentelemetry-python/pull/3143 this work around is no longer required and can be removed from the docs.

This should probably not be merged until that pr is in a release but posting here for cross reference


cartermp:
Thanks!

@open-telemetry/python-maintainers, what's a good way to keep track of a release, assuming that the upstream work gets approved and merged?

srikanthccv:
It usually happens during the second week of every month. It is due the next week. I see the PR already has one approval so we may get this in for the next release.

jenshnielsen:
The relevant change landed in [v1.16.0/0.37b](https://github.com/open-telemetry/opentelemetry-python/releases/tag/v1.16.0) so I think this should be safe to land

cartermp (approved):
Thanks!

PR #2245, opened 2023-02-01
Clarify instrumentation definition

MattDodsonEnglish:
The original version implied that the code itself emits metrics, which I don't think is precisely true. Isn't it the system that emits?

Great docs, by the way!

### DIFF
--- a/content/en/docs/concepts/instrumenting.md
+++ b/content/en/docs/concepts/instrumenting.md
@@ -8,7 +8,6 @@ weight: 40
 
 In order to make a system observable, it must be **instrumented**: That is, code from the
 system's components must emit
-system must emit
 [traces](/docs/concepts/observability-primer/#distributed-traces),
 [metrics](/docs/concepts/observability-primer/#reliability--metrics), and
 [logs](/docs/concepts/observability-primer/#logs).
--- a/content/en/docs/concepts/instrumenting.md
+++ b/content/en/docs/concepts/instrumenting.md
@@ -6,11 +6,12 @@ description: >-
 weight: 40
 ---
 
-In order to make a system observable, it must be **instrumented**: That is, the
+In order to make a system observable, it must be **instrumented**: That is, code from the
+system's components must emit
 system must emit
 [traces](/docs/concepts/observability-primer/#distributed-traces),
 [metrics](/docs/concepts/observability-primer/#reliability--metrics), and
-[logs](/docs/concepts/observability-primer/#logs) when code runs on it.
+[logs](/docs/concepts/observability-primer/#logs).
 
 Without being required to modify the source code you can collect telemetry from
 an application using [Automatic Instrumentation][]. If you previously used an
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/instrumenting.md
```diff
@@ -7,10 +7,10 @@ weight: 40
 ---
 
 In order to make a system observable, it must be **instrumented**: That is, the
```
chalin, line 9: Thanks for the proposal. I'd prefer something like:

```suggestion
In order to make a system observable, it must be **instrumented**: That is, code from the
system's components must emit
```
Thoughts @cartermp et al.

If you agree, then drop the other changes.

svrnm:
@MattDodsonEnglish I am not sure if I understand your concern: Instrumentation is added into the code (either manually or by means to inject it) so that the application (code) is able to emit telemetry. What are you referring to when you say "system"?

MattDodsonEnglish:
>  What are you referring to when you say "system"?

Would it be more precise to say "The observed system" or "the instrumented application"?

> the application (code) is able to emit telemetry.

This is my essential confusion: I don't think that application is equivalent to code, since an application also has hardware that runs the code. Code itself can't emit anything; it's just lines of text.

For a little more context, I'm about to review a docs PR for another OSS tool, k6 (developed by Grafana). These docs use the word "instrument" often, and I see "Instrument" and "instrumentation" all over text about observability and observability tools. But, it was hard to find definitions. Someone pointed me here, and I really like the simplicity and economy of it. However it still gives me a slight pause to read "code emits". I suppose it's not so easy to have a clear, concise definition. Another attempt:

> When the application runs, the components with instrumented code emit metrics...

If this all seems too pedantic, I won't feel bad if the issue gets closed :-)

svrnm:
I see what you mean. I think it makes sense, @chalin, @cartermp, @austinlparker: wdyt?

svrnm (approved):

chalin:
I think that there is room for improvement, though I'd be in favor of the suggested change I proposed earlier: which refers to "code from the system's components". Any comment on that?

svrnm:
@MattDodsonEnglish please apply @chalin's suggestion and then I think we should be good to go

content/en/docs/concepts/instrumenting.md
```diff
@@ -7,10 +7,10 @@ weight: 40
...
-code must emit
+system must emit
 [traces](/docs/concepts/observability-primer/#distributed-traces),
 [metrics](/docs/concepts/observability-primer/#reliability--metrics), and
-[logs](/docs/concepts/observability-primer/#logs).
+[logs](/docs/concepts/observability-primer/#logs) when code runs on it.
```
MattDodsonEnglish, line 13:
```suggestion
[logs](/docs/concepts/observability-primer/#logs).
```

content/en/docs/concepts/instrumenting.md
```diff
@@ -6,8 +6,9 @@ description: >-
 weight: 40
 ---
 
-In order to make a system observable, it must be **instrumented**: That is, the
-code must emit
+In order to make a system observable, it must be **instrumented**: That is, code from the
+system's components must emit
+system must emit
```
MattDodsonEnglish, line 11:
```suggestion

```

cartermp (approved):
Thanks!

PR #2279, opened 2023-02-06
check-collector-version.sh: fix for Linux sed

chalin:
Followup to:

- #2266

Which broke the script under Linux, despite my attempt at addressing the `sed` differences between Linux and macOS. The result was action failures:
  - https://github.com/open-telemetry/opentelemetry.io/actions/runs/4101553729/jobs/7073465659

Take 2. Tested script under macOS. Tested command change under Linux. 🤞 😄


svrnm (approved):

chalin:
@svrnm - thanks for double checking that the [script works now](https://github.com/open-telemetry/opentelemetry.io/actions/runs/4107613557/jobs/7087314733).

PR #2284, opened 2023-02-07
Add C++ packages to registry

svrnm:
Since I created one of them, I thought it would be great to have these different options to install OpenTelemetry C++ via package managers listed in our registry.

Source: https://github.com/open-telemetry/opentelemetry-cpp/blob/main/INSTALL.md#using-package-managers

### DIFF
--- a/data/registry/tools-cpp-alpine-apk.yml
+++ b/data/registry/tools-cpp-alpine-apk.yml
@@ -9,6 +9,6 @@ tags:
   - package
 repo: https://pkgs.alpinelinux.org/packages?name=opentelemetry-cpp-* 
 license: Apache 2.0
-description: If you are using alpine linux you can install opentelemetry-cpp with these packages.
+description: Alpine Linux packages in support of opentelemetry-cpp.
 authors: Severin Neumann
 otVersion: unknown
--- a/data/registry/tools-cpp-alpine-apk.yml
+++ b/data/registry/tools-cpp-alpine-apk.yml
@@ -1,5 +1,5 @@
 title: Alpine Linux Packages for OpenTelemetry C++
-registryType: core
+registryType: core exporter
 isThirdParty: true
 language: cpp
 tags:
--- a/data/registry/tools-cpp-alpine-apk.yml
+++ b/data/registry/tools-cpp-alpine-apk.yml
@@ -1,4 +1,4 @@
-title: Alpine Linux Package for OpenTelemetry C++
+title: Alpine Linux Packages for OpenTelemetry C++
 registryType: core
 isThirdParty: true
 language: cpp
--- a/data/registry/tools-cpp-alpine-apk.yml
+++ b/data/registry/tools-cpp-alpine-apk.yml
@@ -7,7 +7,7 @@ tags:
   - alpine linux
   - apk
   - package
-repo: https://pkgs.alpinelinux.org/packages?name=opentelemetry-cpp-* 
+repo: https://pkgs.alpinelinux.org/packages?name=opentelemetry-cpp-*
 license: Apache 2.0
 description: Alpine Linux packages in support of opentelemetry-cpp.
 authors: Severin Neumann
--- a/data/registry/tools-cpp-alpine-apk.yml
+++ b/data/registry/tools-cpp-alpine-apk.yml
@@ -1,0 +1,14 @@
+title: Alpine Linux Package for OpenTelemetry C++
+registryType: core
+isThirdParty: true
+language: cpp
+tags:
+  - cpp
+  - alpine
+  - apk
+  - package
+repo: https://pkgs.alpinelinux.org/packages?name=opentelemetry-cpp-* 
+license: Apache 2.0
+description: If you are using alpine linux you can install opentelemetry-cpp with these packages.
+authors: Severin Neumann
+otVersion: unknown
--- a/data/registry/tools-cpp-alpine-apk.yml
+++ b/data/registry/tools-cpp-alpine-apk.yml
@@ -4,7 +4,7 @@ isThirdParty: true
 language: cpp
 tags:
   - cpp
-  - alpine
+  - alpine linux
   - apk
   - package
 repo: https://pkgs.alpinelinux.org/packages?name=opentelemetry-cpp-* 
--- a/data/registry/tools-cpp-conan.yml
+++ b/data/registry/tools-cpp-conan.yml
@@ -9,5 +9,5 @@ tags:
 repo: https://conan.io/center/opentelemetry-cpp
 license: MIT
 description: Conan package for `opentelemetry-cpp`.
-authors: Others
+authors:
 otVersion: unknown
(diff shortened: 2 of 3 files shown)


chalin (commented):
Would it make sense to split the Alpine Linux entry into two entires, one for core packages and one for exporters? IMHO that would make things clearer. WDYT?

data/registry/tools-cpp-alpine-apk.yml
```diff
@@ -0,0 +1,14 @@
+title: Alpine Linux Package for OpenTelemetry C++
+registryType: core
+isThirdParty: true
+language: cpp
+tags:
+  - cpp
+  - alpine
```
chalin, line 7: Maybe:

```suggestion
  - alpine linux
```

svrnm:
> Would it make sense to split the Alpine Linux entry into two entires, one for core packages and one for exporters? IMHO that would make things clearer. WDYT?

Interesting suggestion, I have to think about that... it would make it easier to explore those packages, but on the other hand we have "exporters" already as their own category pointing to the language specific repos, so this might get confusing...

cartermp:
I don't feel too strongly either way, but I think I prefer things as-is. Something about having the "hub" of alpine packages in one spot feels nice to link to. I know we're not really doing that for other things, though.

chalin:
That is what confused me at first. If an entry is a hub to a bunch of packages, then description should make that clear -- which I don't find these new entries make clear. My 2 cents. I don't feel strongly about it either.

svrnm:
I didn't want to introduce "package" or something like that as a new category, because what we actually would need is something like an entry for OpenTelemetry C++ and it's components and then multiple links to the package managers, similar could be done with other languages (e.g. instead of linking to github we could link to npmjs for nodejs packages, etc.)

My goal is more to advertise the existence of things via the registry ... some day in the future we can consider making it better&different.

data/registry/tools-cpp-alpine-apk.yml
```diff
@@ -0,0 +1,14 @@
+title: Alpine Linux Package for OpenTelemetry C++
```
chalin, line 1:
```suggestion
title: Alpine Linux Packages for OpenTelemetry C++
```

data/registry/tools-cpp-alpine-apk.yml
```diff
@@ -0,0 +1,14 @@
...
+isThirdParty: true
+language: cpp
+tags:
+  - cpp
+  - alpine linux
+  - apk
+  - package
+repo: https://pkgs.alpinelinux.org/packages?name=opentelemetry-cpp-* 
+license: Apache 2.0
+description: If you are using alpine linux you can install opentelemetry-cpp with these packages.
```
chalin, line 12: IMHO it'll be clearer if we don't ask a question in a description:

```suggestion
description: Alpine Linux packages in support of opentelemetry-cpp.
```

data/registry/tools-cpp-alpine-apk.yml
```diff
@@ -0,0 +1,14 @@
+title: Alpine Linux Package for OpenTelemetry C++
+registryType: core
```
chalin, line 2: I see two options here. Either list all that apply, or we could use a convention like `*` or `mixed`:

```suggestion
registryType: core exporter
```

This doesn't break anything. In a followup PR we can adapt the filter to expect more than one keyword. WDYT?

data/registry/tools-cpp-alpine-apk.yml
```diff
@@ -0,0 +1,14 @@
...
+isThirdParty: true
+language: cpp
+tags:
+  - cpp
+  - alpine linux
+  - apk
+  - package
+repo: https://pkgs.alpinelinux.org/packages?name=opentelemetry-cpp-*
```
chalin, line 10:
```suggestion
repo: https://pkgs.alpinelinux.org/packages?name=opentelemetry-cpp-*
```

data/registry/tools-cpp-conan.yml
```diff
@@ -0,0 +1,13 @@
...
+registryType: core
+isThirdParty: true
+language: cpp
+tags:
+  - cpp
+  - conan
+  - package
+repo: https://conan.io/center/opentelemetry-cpp
+license: MIT
+description: If you are using Conan to manage your dependencies, add opentelemetry-cpp/x.y.z to your conanfile's requires.
```
chalin, line 11:
```suggestion
description: Conan package for `opentelemetry-cpp`.
```

data/registry/tools-cpp-conan.yml
```diff
@@ -0,0 +1,13 @@
...
+authors: Others
```
chalin, line 12: This sets a precedent. I'd rather leave it blank (or put unknown):
```suggestion
authors:
```

data/registry/tools-cpp-vcpkg.yml
```diff
@@ -0,0 +1,13 @@
+title: vcpkg package for OpenTelemetry C++
+registryType: core
+isThirdParty: true
+language: cpp
+tags:
+  - cpp
+  - vcpkg
+  - package
+repo: https://github.com/microsoft/vcpkg/tree/master/ports/opentelemetry-cpp
```
chalin, line 9:
```suggestion
repo: https://github.com/microsoft/vcpkg/tree/master/ports/opentelemetry-cpp
```

data/registry/tools-cpp-vcpkg.yml
```diff
@@ -0,0 +1,13 @@
...
+registryType: core
+isThirdParty: true
+language: cpp
+tags:
+  - cpp
+  - vcpkg
+  - package
+repo: https://github.com/microsoft/vcpkg/tree/master/ports/opentelemetry-cpp 
+license: MIT 
+description: If you are using vcpkg, then you can install opentelemetry-cpp with this package.
```
chalin, line 11:
```suggestion
description: A vcpkg package for opentelemetry-cpp. 
```

data/registry/tools-cpp-vcpkg.yml
```diff
@@ -0,0 +1,13 @@
...
+isThirdParty: true
+language: cpp
+tags:
+  - cpp
+  - vcpkg
+  - package
+repo: https://github.com/microsoft/vcpkg/tree/master/ports/opentelemetry-cpp 
+license: MIT 
+description: If you are using vcpkg, then you can install opentelemetry-cpp with this package. 
+authors: Others
```
chalin, line 12:
```suggestion
authors:
```

chalin (commented):
Ok, thanks for explaining your intent. Makes sense.

In the meantime, I think that we can make things a bit clearer. See inline comments.

data/registry/tools-cpp-alpine-apk.yml
svrnm: makes sense :-)

svrnm:
@chalin thanks for the cleanups :)

chalin (approved):
✨

Issue #1063, opened 2022-01-20
CI: check markdown formatting and common gotchas


chalin:
Now that we starting to have more and more contributors, the project would benefit from standardizing Markdown formatting, and a prettifier and linter could help in this respect.

- I've (@chalin) started experimenting with [Prettier](https://prettier.io), e.g., see:
  - #1290
  - https://github.com/open-telemetry/opentelemetry.io/pulls?q=is%3Apr+prettier+is%3Aclosed
- Note that there's module for formatting Go templates. [Read more](https://discourse.gohugo.io/t/formatter-for-go-templates/38403)

---

- [x] #2286
- [x] #2292
- [x] #2302
- [x] #2309
- [x] #2311
- [x] #2315
- [x] #2316
- [x] #2338
- [x] #2363
- [x] #2368
- [x] #2377
- [x] #2372
- [x] Go docs
  - [x] #2394
  - [x] https://github.com/open-telemetry/opentelemetry-go/pull/3762
  - [x] #2380
- [x] #2389
- [x] #2390
- [x] #2392
- [x] #2393
- [x] #2398
- [x] #2399
- [x] #2408
- [x] #2412
- [x] Community pages: #2414
- [x] Data files and more: #2414

chalin:
@svrnm - I've noticed that lately you've been running a markdown linter/prettifier and submitting PRs. Thanks for those.

It would be good if we could agree on the same tool. I'd like to suggest [Prettier](https://prettier.io). I'm open to discussing this at the next Comms meeting if you have a strong preference for another tool.

You can run Prettier as follows:

```console
$ npx prettier --write path/to/the/markdown/file.md
```

/cc @cartermp @austinlparker

austinlparker:
I'm fine with Prettier

svrnm:
> @svrnm - I've noticed that lately you've been running a markdown linter/prettifier and submitting PRs. Thanks for those.
> 
> It would be good if we could agree on the same tool. I'd like to suggest [Prettier](https://prettier.io). I'm open to discussing this at the next Comms meeting if you have a strong preference for another tool.

I'm fine with prettier, I just had setup markdownlint to check MD files automatically for me when I open them up, but I can work with prettier as well!

svrnm:
@chalin -- if we add a `.prettierrc.json` to the project vscode and other editors should be able to pick it up and use prettier (instead of markdownlint in my case). Sounds like a non-invasive first step, wdyt?

[...]

PR #2286, opened 2023-02-07
Prettier base config and NPM scripts in preparation for CI integration

chalin:
- Preparation for use of Prettier as part of CI
  - Sets up a base config for use of Prettier, that makes it easy for contributors to use
  - Defines NPM scripts for checking all files
- Offers an alternative base to #2179
- Contributes to #1063
- Includes minor Prettier fixes for a few top-level files

---

The `check-all` script will then be useful for a GH action since it fails which a human friendly message when formatting needs to be applied. For example:

```console
$ npm run prettier:check-all ; echo $?        
...
Checking formatting...
[warn] README.md
[warn] Code style issues found in the above file. Forgot to run Prettier?
1
```

To fix the issues add the `-w` flag:

```console
$ npm run prettier:check-all -- -w    
...
Checking formatting...
[warn] README.md
[warn] Code style issues fixed in the above file.
```

Now check again, and all is good:

```console
$ npm run prettier:check-all ; echo $?
...
Checking formatting...
All matched files use Prettier code style!
0
```

To run Prettier on a file that is ignored by default, use either of the following commands:

```console
$ npm run prettier:no-ignore -- -w assets/scss/_tabs.scss
...
assets/scss/_tabs.scss 39ms
$ npx prettier --ignore-path '' -w assets/scss/_tabs.scss
assets/scss/_tabs.scss 37ms
```

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,23 @@
+# Base
+
+/.netlify
+/.vscode
+/content-modules
+/iconography
+/layouts
+package-lock.json
+/public
+/themes
+/tmp
+
+# Temporary while we ramp up use of Prettier
+
+/.github
+/archetypes
+/assets
+/content
+/data
+/resources
+/scripts
+/static
+/templates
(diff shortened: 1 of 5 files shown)


cartermp (approved):
I prefer this (base) config. Approving since it works and I assume it's what you'd prefer as well?

svrnm (approved):

svrnm:
I prefer this one as well :)

Issue #2287, opened 2023-02-07
Wrong timestamp for span event


sungam3r:
> A Span Event can be thought of as a structured log message (or annotation) on a Span, typically used to denote a meaningful, singular point in time **during the Span’s duration**.

But in example from https://opentelemetry.io/docs/concepts/signals/traces/#tracing-in-opentelemetry timestamp of the second event is greater than `end_time` of the span itself.

![изображение](https://user-images.githubusercontent.com/21261007/217281734-6a74ccb4-6a95-4909-8ce1-19358e27c40d.png)

Neilblaze:
@cartermp Can I take this up? I think replacing the positions of those timestamps will fix the issue. Thanks! :)

svrnm:
@Neilblaze, sure, please go ahead and raise a PR

PR #2313, opened 2023-02-09
updated trace timestamps

Neilblaze:
### Target :
In docs under `Sample Trace` in [#tracing-in-opentelemetry](https://opentelemetry.io/docs/concepts/signals/traces/#tracing-in-opentelemetry) timestamp of the *second event* is greater than the *end_time* of the span itself. 

> Hence, it has been updated!

fixes : #2287 
cc : @svrnm @cartermp


cartermp (approved):
Thanks!

PR #2322, opened 2023-02-10
docs: fix grammar in observability primer

Go-pye:
Fix for broken grammar. "it" can also be replaced with "observability"


chalin (approved):
Thanks for submitting a fix. Will merge after you sign the CLA.

PR #2325, opened 2023-02-10
docs: fix grammar in observability primer

Go-pye:
Duplicate of #2322 

I had to close the above PR to fix EasyCLA signing


chalin (approved):
Thanks for resubmitting!

PR #2326, opened 2023-02-10
add telemetryhub to vendors table

quinnmil:
Adds TelemetryHub (https://telemetryhub.com/) to vendors list.


cartermp:
Thanks! I think I'd prefer a link to the docs instead - we're not exactly consistent here but it seems to be the most common pattern.

cartermp:
wtf easycla

cartermp (approved):
Thanks!

PR #2330, opened 2023-02-12
Avoid usages of obsolete APIs

sungam3r:


julianocosta89:
Hello @sungam3r, thanks for this PR.
The code you are suggesting the update is based on the implementation used within the services on the demo.
To update the snippets here we also need to update the demo.
Would you be willing to send the PR updating the demo?
https://github.com/open-telemetry/opentelemetry-demo

sungam3r:
done

cartermp (changes requested):
This is a larger change that requires having updated packages (basically everything as the RC) and some additional changes, like how Resources are specified. If you'd like to continue with this PR then we'll also need to get those updates to the docs too.

sungam3r:
Do you mean changes in https://github.com/open-telemetry/opentelemetry-demo/pull/746 ?

cartermp:
Not quite - you'll want to also update to use `WithResource` like here: https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/examples/AspNetCore/Program.cs#L48-L50

And ensure that the right packages are specified in the prerequisites across the relevant pages.

sungam3r:
https://github.com/open-telemetry/opentelemetry-demo/blob/main/src/cartservice is broken in main, it does not compile.

cartermp:
Not sure I understand? This is the docs repo, not the demo repo.

sungam3r:
From https://github.com/open-telemetry/opentelemetry.io/pull/2330#issuecomment-1426983132 : 
> To update the snippets here we also need to update the demo.

sungam3r:
I mean it is difficult to change the documentation based on an example if the example does not compile due to the lack of the necessary files.

cartermp:
I think there may be some confusion here.

The obsolete APIs you refere to were made obsolete by changes in non-core packages that correspond to the 1.4.x-rc core package.

Just updating code to use the newer, preview APUs is not sufficient for neither demo code nor the docs here. To do that, you'll need to also update the packages used in the demo app and prerequisite sections in all docs. There are also further changes to consider (namely changes in how Resources are declared), so as to be fully up to date with the latest preview releases of both the core package and non-core packages.

Unfortunately it is a bit of a confusing state, but hopefully soon everything will be out of preview, making it easier to document and demonstrate in the demo. Does that help?

sungam3r:
Nope. I have no problems with updating .NET packages. I'm talking about different thing. I have a problem with the demo code for .NET example itself in `main`. It does not compile at all.
изображение

Where are all classes from `Oteldemo` namespace?

cartermp:
My guess it's some grpc prebuild magic going on, but I'd ask over on the demo repo.

sungam3r:
I'll update this PR once https://github.com/open-telemetry/opentelemetry-demo/pull/746 is being merged.

puckpuck:
This is cleaned up on the Otel-demo side. @cijothomas we still need a changelog entry on that side, but this PR should be unblocked now.

sungam3r:
> Not quite - you'll want to also update to use `WithResource` like here: https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/examples/AspNetCore/Program.cs#L48-L50

@cartermp  I don't see `WithResource` in the provided link.

> And ensure that the right packages are specified in the prerequisites across the relevant pages.

What does _right_ mean? Do you want me just to append `--prerelease` to each of `dotnet add package OpenTelemetry.XXX` line everywhere in examples?

cijothomas (approved):

cijothomas:
> And ensure that the right packages are specified in the prerequisites across the relevant pages.
> 
> What does right mean? Do you want me just to append --prerelease to each of dotnet add package OpenTelemetry.XXX line everywhere in examples?

I do not think any change is required . The stable packages are correctly listed without "--prerelease". The instrumentations already have "--prerelease", so no need to change that.
The only exception is to remove the "--prerelease" from OpenTelemetry.Extensions.Hosting (as it'll be having stable release tomorrow), but that should not be blocking this PR.

cijothomas:
> @cartermp I don't see WithResource in the provided link.

Probably meant `ConfigureResource`. Again, I do not think its a blocker. The current code is valid. `ConfigureResource` is an alternative, which we should be using, but dont think it must be part of this PR, which is more about fixing the broken docs.

cijothomas:
@cartermp Could you re-review? It looks good to me, and would work with new version of OTel.Extensions.Hosting (which is going stable tomorrow).

I agree with the other draft PR doing similar, but I'd suggest to merge this in if you see no other blockers, and follow up with more improvements as needed.

Kielek (approved):

julianocosta89 (approved):
LGTM

cartermp (approved):

cartermp:
LGTM for now. We can use #2374 to hammer out the ideal way to structure things.

PR #2254, opened 2023-02-02
Add Honeycomb.io Go and Node Distros to Registry

vreynolds:
Add Honeycomb.io OpenTelemetry Go and Node.js Distros to Registry

PR #2343, opened 2023-02-15
[registry] streamline Honeycomb distro language

vreynolds:
Follow-up from #2254

Streamline title and authors language across Honeycomb distros in the Registry.


chalin (approved):
Great, thanks for the followup PR!

PR #2351, opened 2023-02-16
Auto-update more versions

trask:
Let me know if you'd like to change anything. After you merge this, it should pick up this lagging `javaVersion` that we haven't been updating as a good test that it's working (I've also tested it in my fork).

https://github.com/open-telemetry/opentelemetry.io/blob/0f45ace76c31107d48e767a37447c7ca58097b1e/content/en/docs/instrumentation/java/manual.md?plain=1#L9


chalin (commented):
Oh! I'm so looking forward to reviewing this more closely, but it might not be until later today or tomorrow. If another maintainer approves before then, I'm ok if this lands sooner than later.

I do have some comments about making the script runnable locally, in particular under macOS. I'd made improvements to `.github/workflows/scripts/check-collector-version.sh` so that it could be run locally _and_ in a "dry-run" mode. I'd like to preserve that ability. I don't mind submitting followup changes.

.github/workflows/scripts/auto-update-version.sh
```diff
@@ -0,0 +1,40 @@
...
+  sed -i -e "s/$variable_name: .*/$variable_name: $latest_version/" "$file_name"
+done
+
+if git diff --quiet; then
+  echo "Already at the latest version."
+  exit 0
+fi
+
+message="Update $repo version to $latest_version"
+body="Update $repo version to \`$latest_version\`."
```
chalin, line 23: It would be helpful if the body could contain, at a minimum, a link to the repo's releases; or even better, a list to the specific version. Or both :)

.github/workflows/scripts/auto-update-version.sh
```diff
@@ -0,0 +1,40 @@
...
+file_names=("${@:3}") # remaining args
+
+latest_version=$(gh api -q .tag_name "repos/open-telemetry/$repo/releases/latest" | sed 's/^v//')
+
+echo "Repo: $repo"
+echo "Latest version: $latest_version"
+
+for file_name in "${file_names[@]}"
+do
+  sed -i -e "s/$variable_name: .*/$variable_name: $latest_version/" "$file_name"
```
chalin, line 14: It would be beneficial if this failed in the case that no match is found. It would help detect when a version field is (possibly) inadvertently rename or (re)moved from one of the target files. WDYT @trask?

chalin:
Just sharing a few ideas of the top. Will share more later.

svrnm:
I like that! 

@chalin I am wondering if we should move those versions into a data/ file and then use from there at the different places, wdyt? (Does not have to be part of this PR)

.github/workflows/auto-update-versions.yml
```diff
@@ -0,0 +1,27 @@
...
+
+      - name: Use CLA approved github bot
+        run: |
+          git config user.name opentelemetrybot
+          git config user.email 107717825+opentelemetrybot@users.noreply.github.com
+
+      - name: Auto-update
+        run: |
+          ./.github/workflows/scripts/auto-update-version.sh "opentelemetry-collector-releases" "collectorVersion" "content/en/docs/collector/_index.md"
+          ./.github/workflows/scripts/auto-update-version.sh "opentelemetry-java" "javaVersion" "content/en/docs/instrumentation/java/_index.md" "content/en/docs/instrumentation/java/manual.md"
```
chalin, line 25: Following #2352, we can drop the second file:

```suggestion
          ./.github/workflows/scripts/auto-update-version.sh "opentelemetry-java" "javaVersion" "content/en/docs/instrumentation/java/_index.md"
```

.github/workflows/auto-update-versions.yml
```diff
@@ -0,0 +1,27 @@
...
+        uses: actions/checkout@v3
+
+      - name: Use CLA approved github bot
+        run: |
+          git config user.name opentelemetrybot
+          git config user.email 107717825+opentelemetrybot@users.noreply.github.com
+
+      - name: Auto-update
+        run: |
+          ./.github/workflows/scripts/auto-update-version.sh "opentelemetry-collector-releases" "collectorVersion" "content/en/docs/collector/_index.md"
```
chalin, line 24: Why the leading `./` on these lines?

.github/workflows/scripts/auto-update-version.sh
```diff
@@ -0,0 +1,40 @@
...
+message="Update $repo version to $latest_version"
+body="Update $repo version to \`$latest_version\`."
+
+existing_pr_count=$(gh pr list --state all --search "in:title $message" | wc -l)
+if [ "$existing_pr_count" -gt 0 ]; then
+    echo "PR for this version was already created."
+    exit 0
+fi
+
+branch="opentelemetrybot/auto-update-$repo-$latest_version"
```
chalin, line 31: I'd drop opentelemetrybot, because I want this to be executable locally if/when we might feel the need for an immediate update. If you prefer having a branch of the form `foo/bar` then maybe it can be `auto-update/...`. Otherwise a simple branch seems fine, WDYT?

```suggestion
branch="auto-update-$repo-$latest_version"
```

.github/workflows/scripts/auto-update-version.sh
```diff
@@ -0,0 +1,40 @@
...
+  echo "Already at the latest version."
+  exit 0
+fi
+
+message="Update $repo version to $latest_version"
+body="Update $repo version to \`$latest_version\`."
+
+existing_pr_count=$(gh pr list --state all --search "in:title $message" | wc -l)
+if [ "$existing_pr_count" -gt 0 ]; then
+    echo "PR for this version was already created."
```
chalin, line 27:
```suggestion
    echo "PR(s) for version $latest_version of $repo already exist:"
    gh pr list --state all --search "in:title $message"
    echo "So we won't create another. Exiting."
```

.github/workflows/scripts/auto-update-version.sh
```diff
@@ -0,0 +1,40 @@
...
+    exit 0
+fi
+
+branch="opentelemetrybot/auto-update-$repo-$latest_version"
+
+git checkout -b "$branch"
+git commit -a -m "$message"
+git push --set-upstream origin "$branch"
+
+echo "Creating a pull request on your behalf."
```
chalin, line 37:
```suggestion
echo "Submitting auto-update PR '$message'."
```

.github/workflows/auto-update-versions.yml
chalin: ⚠️ If `auto-update-version.sh` runs twice, won't the second run inherit the branch from the first run?

chalin: If so, the script can save the branch when it starts and change back to that branch when it's done.

.github/workflows/auto-update-collector-version.yml
```diff
@@ -1,27 +0,0 @@
...
-      - name: Auto-update
-        run: ./.github/workflows/scripts/check-collector-version.sh
-        env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          GIT_AUTHOR_NAME: opentelemetrybot
-          GIT_COMMITTER_NAME: opentelemetrybot
-
-          # Set opentelemetrybot's email, otherwise CLA checks will fail
-          GIT_AUTHOR_EMAIL: 107717825+opentelemetrybot@users.noreply.github.com
-          GIT_COMMITTER_EMAIL: 107717825+opentelemetrybot@users.noreply.github.com
```
chalin, line 27: Do any of these env vars need to be set in the new workflow too?

.github/workflows/auto-update-collector-version.yml
```diff
@@ -1,27 +0,0 @@
...
-    name: Check collector version
-    runs-on: ubuntu-20.04
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v3
-
-      - name: Auto-update
-        run: ./.github/workflows/scripts/check-collector-version.sh
-        env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
chalin, line 21: If we're no longer using `GITHUB_TOKEN` (vs. `OPENTELEMETRYBOT_GITHUB_TOKEN ` in the new workflow), then `GITHUB_TOKEN` should be deleted from the settings, right?

chalin (changes requested):
Great improvement. See inline comments and questions, thanks.

.github/workflows/scripts/auto-update-version.sh
trask: > if/when we might feel the need for an immediate update

this workflow supports `workflow_dispatch` (see top), which means you can trigger it manually in GitHub Actions any time you want (this is also a really handy way to test the workflow on your fork, push to `main` on your fork, and then run manually trigger the workflow)

.github/workflows/auto-update-collector-version.yml
trask: `GITHUB_TOKEN` is a built-in env var (not one that you add in your settings)

.github/workflows/auto-update-collector-version.yml
trask: no, these are handled by

```
          git config user.name opentelemetrybot
          git config user.email 107717825+opentelemetrybot@users.noreply.github.com
```

.github/workflows/auto-update-versions.yml
trask: removed

.github/workflows/auto-update-versions.yml
trask: fixed by resetting between each update

.github/workflows/scripts/auto-update-version.sh
chalin: Got it, thanks!

.github/workflows/scripts/auto-update-version.sh
trask: done

.github/workflows/auto-update-versions.yml
```diff
@@ -0,0 +1,30 @@
...
+        run: |
+          git config user.name opentelemetrybot
+          git config user.email 107717825+opentelemetrybot@users.noreply.github.com
+
+      - name: Auto-update
+        run: |
+          .github/workflows/scripts/auto-update-version.sh "opentelemetry-collector-releases" "collectorVersion" "content/en/docs/collector/_index.md"
+          git reset --hard origin/main
+          .github/workflows/scripts/auto-update-version.sh "opentelemetry-java" "javaVersion" "content/en/docs/instrumentation/java/_index.md"
+          git reset --hard origin/main
```
chalin, line 27: We probably don't need this second reset since the workflow is done after that, right?

.github/workflows/scripts/auto-update-version.sh
chalin: Is this a change you'll be adding in this or a followup PR @trask (or maybe have someone else do it)?

.github/workflows/scripts/auto-update-version.sh
chalin: Is this a change you'll be adding in this or a followup PR @trask (or maybe have someone else do it)?

.github/workflows/auto-update-versions.yml
trask: ya, just trying to help the next person who adds another line here 😄

.github/workflows/scripts/auto-update-version.sh
trask: ya, I'll add it in this PR

.github/workflows/auto-update-versions.yml
trask:
```suggestion
```

chalin: It's your call, but it is currently "dead code" so I'd drop it.

.github/workflows/scripts/auto-update-version.sh
chalin: Oh, sorry, I missed that. Thanks!

.github/workflows/scripts/auto-update-version.sh
trask: done and tested 👍

.github/workflows/auto-update-versions.yml
trask: ^ leaving this suggestion in case you want to commit it no problem

trask:
> I do have some comments about making the script runnable locally, in particular under macOS. I'd made improvements to `.github/workflows/scripts/check-collector-version.sh` so that it could be run locally _and_ in a "dry-run" mode. I'd like to preserve that ability. I don't mind submitting followup changes.

ya, sorry about that, I generally just test jobs in my fork, but I totally support adding those features back

.github/workflows/scripts/auto-update-version.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash -e
+
+repo=$1
+variable_name=$2
+file_names=("${@:3}") # remaining args
```
trask, line 5: this could be simplified if there is only a single file per repo version

.github/workflows/auto-update-collector-version.yml
```diff
@@ -1,27 +0,0 @@
...
-
-on:
-  workflow_dispatch:
-  schedule:
-    - cron: '0 1 * * *'
-
-jobs:
-  build:
-    # Make sure that the workflow does not run on forks (add yourself as exception if you work on the workflow)
-    if: github.repository_owner == 'open-telemetry' || github.repository_owner == 'svrnm'
```
chalin, line 11: I think that we should have a guard like this in the new workflow. At least the `if: github.repository_owner == 'open-telemetry'` part.

chalin (approved):
Left an inline comment about adding a guard. Otherwise, this LGTM. Thanks!

.github/workflows/auto-update-collector-version.yml
trask: this kind of guard used to be helpful, but github disables jobs on (new) forks now, so I haven't been adding them anymore (and you have to remove the guard if you do want to test it on your fork). or is your concern different?

cartermp (approved):
Approving on account of the question about the guard being addressed

.github/workflows/auto-update-collector-version.yml
chalin: Ok, that's good enough for me. Thanks!

PR #2372, opened 2023-02-20
Enable Prettier check over instrumentation/js

chalin:
- Contributes to #1063
- Enables format checking over `instrumentation/js` section
- Replaces `relref` link(s) by standard markdown link(s)
- Copyedits (minor)

⚠️ **Note**: code excerpts in code blocks also get reformatted per Prettier rules.

Preview: https://deploy-preview-2372--opentelemetry.netlify.app/docs/instrumentation/js/

PR #2290, opened 2023-02-07
CI: enable Prettier checking

chalin:
- Enables Prettier checking over all non-ignored resources to CI by creating NPM scripts that contributors can run too
- Closes #2179 by superseding it
- Closes #1063, though we'll have to progressively relax the `.prettierignore` rules

✅ **Note that this PR was seeded with a formatting issue** in the `README.md` file. The build did fail (for details, see the comment below). ~If the build fails, then that will confirm that the changes in this PR are working. I'll then remove the `README.md` file change.~

PR #2377, opened 2023-02-21
Prettier: switch to single-quotes by default

chalin:
My original inclination (supported by @svrnm and @cartermp) was to opt for single quotes (https://github.com/open-telemetry/opentelemetry.io/pull/2290#discussion_r1098882987), but I reverted after seeing that YAML code blocks were being changed. Well, in retrospect (and given the situation we're facing in https://github.com/open-telemetry/opentelemetry.io/pull/2372#pullrequestreview-1306451171), I'd rather opt for single quotes for all (including JS code blocks, given that it's part of the OTel-JS standard).

WDYT @svrnm @cartermp?

This PR:

- Contributes to #1063
- Enables Prettier's single-quotes option
- Drops quotes from YAML code blocks in those situations where quotes are unnecessary

Two situations where I wasn't sure if the YAML key value type was important, so I didn't drop the quotes (cc @TylerHelmuth):

- https://github.com/open-telemetry/opentelemetry.io/commit/208befc5131ab827bd1c0da41dcca5e063c9bba4#r101481607
- https://github.com/open-telemetry/opentelemetry.io/commit/01c2c7fb0ef7089bf3ea511c7eb0c8bd52d9333b#r101482142

**Edit**: I added comments to this PR matching the two listed above. Sorry for the duplication!

Preview: https://deploy-preview-2377--opentelemetry.netlify.app

### DIFF
--- a/content/en/docs/collector/scaling.md
+++ b/content/en/docs/collector/scaling.md
@@ -171,7 +171,7 @@ kind: Pod
 metadata:
   name: my-microservice
   annotations:
-    sidecar.opentelemetry.io/inject: "true"
+    sidecar.opentelemetry.io/inject: 'true'
 spec:
   containers:
     - name: my-microservice
--- a/content/en/docs/k8s-operator/automatic.md
+++ b/content/en/docs/k8s-operator/automatic.md
@@ -164,7 +164,7 @@ spec:
     - baggage
   sampler:
     type: parentbased_traceidratio
-    argument: "1"
+    argument: '1'
   java:
     env:
       - name: OTEL_INSTRUMENTATION_KAFKA_ENABLED
@@ -234,7 +234,7 @@ spec:
     - baggage
   sampler:
     type: parentbased_traceidratio
-    argument: "1"
+    argument: '1'
   python:
     env:
       - name: OTEL_PYTHON_DISABLED_INSTRUMENTATIONS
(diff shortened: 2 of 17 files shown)


content/en/docs/collector/scaling.md
```diff
@@ -171,7 +171,7 @@ kind: Pod
 metadata:
   name: my-microservice
   annotations:
-    sidecar.opentelemetry.io/inject: "true"
+    sidecar.opentelemetry.io/inject: 'true'
```
chalin, line 174: @TylerHelmuth - can we drop the quotes here?

content/en/docs/k8s-operator/automatic.md
```diff
@@ -164,7 +164,7 @@ spec:
     - baggage
   sampler:
     type: parentbased_traceidratio
-    argument: "1"
+    argument: '1'
```
chalin, line 167: @TylerHelmuth - can we drop the quotes here (and at line 237)?

cartermp (approved):
I don't personally see issues with this as-is, but would like confirmation of the single quote messes with the otel yaml or not

content/en/docs/collector/scaling.md
TylerHelmuth: No, but single quotes work.

content/en/docs/k8s-operator/automatic.md
TylerHelmuth: I'm not sure if they can be dropped, but single quotes will work.

reyang (approved):

Issue #2396, opened 2023-02-24
Fix broken link.


deepakbhtt2023:
There is a broken link in the page - https://github.com/open-telemetry/opentelemetry.io/blob/main/content/en/_index.md

![image](https://user-images.githubusercontent.com/126073934/221141795-4223e7ac-f11c-4907-9172-54a29ca71a6e.png)

chalin:
Hi, which **website** URL are you visiting that leads to a 404? By a website URL, I mean a URL that starts with <https://opentelemetry.io>.

It seems that you are viewing the markdown file directly from the repository, it which case, links aren't expected to be valid when previewed from GitHub.

deepakbhtt2023:
@chalin I was going through the document and I found below link is broken -

![image](https://user-images.githubusercontent.com/126073934/221151696-2edfa733-84b5-4bcc-9157-90cf5390c1a1.png)

chalin:
Thanks for clarifying. What you've shown in the screenshot above is a preview of the markdown file given by GitHub. As I mentioned before, previews aren't expected to be valid. That page is actually hosted on the website here: https://opentelemetry.io/. You should go through the document by using the website, not GH. Happy reading on https://opentelemetry.io!

PR #2397, opened 2023-02-24
Updated opentelemetry-horizontal-color.svg link

deepakbhtt2023:
#2396 - Fixing the broken link on opentelemetry-horizontal-color.svg


chalin:
As you can see from the build log for this PR, the change you are suggesting breaks the build because it introduces an invalid link. For an explanation, see https://github.com/open-telemetry/opentelemetry.io/issues/2396#issuecomment-1443443960.

Issue #2203, opened 2023-01-25
Concepts > Signals: update descriptions


chalin:
For context and justification, see:

- #2194

Tasks:

- [x] #2409
- [ ] Go docs:
  - [x] https://github.com/open-telemetry/opentelemetry-go/pull/3807
  - [ ] #2411

PR #2405, opened 2023-02-27
Removing duplicate lines

ChillOrb:
Replacing redundant line


svrnm (approved):

svrnm (changes requested):
Please run prettier workflow to fix all CI issues.

chalin (changes requested):
Thanks for the PR @ChillOrb, but this is an issue that we've faced before (#2203), and that we need a better solution to. I'll submit a PR soon.

PR #2416, opened 2023-02-28
Set correct stability for Rust

hdost:
* From what I can tell Rust's status was incorrectly marked as stable. 
** Introduced in #1732 as what might be a typo.

Opened ticket on open-telemetry/opentelemetry-rust#977 to verify.


TommyCpp (approved):
Good catch 👍

svrnm (approved):
thanks!

PR #2433, opened 2023-03-01
Fix registry scanner and include two more entries

svrnm:
Fixes the registry scanner to work with the YML files.


chalin (commented):
Haven't tried it, but LGTM :)

Consider updating the PR title to say that it also contains two new entries.

cartermp (approved):

PR #2444, opened 2023-03-03
OTel-in-focus: link core components to website section

chalin:
What you think about **linking core components** to their **website landing pages** -- @austinlparker @svrnm @cartermp? This makes it easier for readers to find out more about the component.

### Screenshot

Here's a screenshot of what it could look like:

> image

### Before

In contrast, this is what is was before this PR:

> image

Note that I moved the PHP news into the Releases and Updates section.

### DIFF
--- a/content/en/blog/2023/otel-in-focus-02.md
+++ b/content/en/blog/2023/otel-in-focus-02.md
@@ -6,6 +6,10 @@ author: >-
   [Austin Parker](https://github.com/austinlparker)
 ---
 
+<style>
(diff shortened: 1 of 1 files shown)


chalin:
Btw, unfortunately, Prettier doesn't handle definitions lists yet, which is why I added the ignore directives. For details, see:

- https://github.com/prettier/prettier/issues/10701

cartermp (approved):
I like the adjustment

svrnm:
lgtm

Issue #2448, opened 2023-03-03
Make build and CI more contributor friendly


chalin:
Tasks:

- [x] #2445
- [x] #2504
- [x] #3747
- [x] Consider GH action on push, or something similar, for the **refcache**; although the following might be better:
  #2388
- [x] Check for use of external links to site-local pages.
  Related: #2497

Related (CI enhanced checks):

- Check spelling via GH action? - #948
- #2387

chalin:
All done, closing.

PR #2449, opened 2023-03-03
PR-actions workflow with support for `/fix:format`

chalin:
- A first attempt at a comment-triggerable bot action to reformat files in a PR.
- Contributes to #2448

### DIFF
--- a/.github/workflows/pr-actions.yml
+++ b/.github/workflows/pr-actions.yml
@@ -1,0 +1,46 @@
+name: PR actions
+
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  fix-format:
+    name: /fix:format
+    runs-on: ubuntu-latest
+
+    if: |
+      github.event.issue.pull_request &&
+      contains(github.event.comment.body, '/fix:format')
+    permissions:
+      contents: write
+
+    env:
+      PR_NUM: ${{ github.event.issue.number }}
+      COMMENT: ${{ github.event.comment.body }}
+
+    steps:
+      - name: Context info
+        run: |
+          echo $PR_NUM
+          echo $COMMENT
+
+      - uses: actions/checkout@v3
+
+      - run: gh pr checkout $PR_NUM
+        env:
+          GH_TOKEN: ${{ github.token }}
+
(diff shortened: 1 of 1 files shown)


chalin:
Dunno if it'll trigger on this PR. Here goes ... 
/fix:format

chalin:
Hey @trask: what's the best way to test & debug new GH actions (that trigger on comments) like the one in this PR?

trask:
push the PR to the `main` branch on your fork, then open a PR on your fork (targeting `main` on your fork as opposed to targeting the otel repo), then comment on that PR

btw, we have similar here: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/.github/workflows/comment-driven-automation.yml

the problem which I don't have solution for yet is that when it pushes the commit back to the PR, it doesn't trigger workflow, and the only way to trigger workflow is to manually close and re-open the PR, e.g. https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/7881#issuecomment-1440887920

chalin:
Oh, very cool that y'all have a script like that already. Thanks for guidance @trask, and the heads up regarding the (non-)triggering issue.

chalin:
@cartermp @svrnm (cc @trask): this is ready for review. It's a first cut, but it's functional. 

I tested it from another repo. As mentioned in [Commits made by this Action do not trigger new Workflow runs](https://github.com/stefanzweifel/git-auto-commit-action#commits-made-by-this-action-do-not-trigger-new-workflow-runs):

> The resulting commit will not trigger another GitHub Actions Workflow run.

But we can always manually re-trigger the "Check format" action. Maybe we'll find a better solution down the line.

cartermp (approved):

cartermp:
Let's go with this for now and we can adjust if needed!

chalin:
@cartermp - FYI, I've tested the workflow and it goes through all the paces but the pushed commit doesn't make it back into the PR 🤷🏼‍♂️. I'll investigate as soon as I can, but it might not be until late next week.

chalin:
@cartermp - hey, it did work! Just took longer than I expected 🤷🏼‍♂️.

Issue #2457, opened 2023-03-07
Sampling doc: describe head sampling


cartermp:
Explain it, the benefits, how to use it, etc. Elaborate on comparing it to tail sampling.

PR #2463, opened 2023-03-08
Add head sampling section + elaborate tail sampling

cartermp:
Fixes https://github.com/open-telemetry/opentelemetry.io/issues/2457

This still leaves our sampling docs in a very, very incomplete state. But in the name of incremental progress, here's something.

Preview: https://deploy-preview-2463--opentelemetry.netlify.app/docs/concepts/sampling/

### DIFF
--- a/content/en/docs/concepts/sampling/index.md
+++ b/content/en/docs/concepts/sampling/index.md
@@ -73,10 +73,10 @@ within them are sampled. For this, you need Tail Sampling.
 
 ## Tail Sampling
 
-Tail Sampling is where the decision to sample a trace happens _after_ all the
-spans in a request have been completed. Tail Sampling gives you the option to
-sample your traces based on specific criteria, which isn’t an option with Head
-Sampling.
+Tail Sampling is where the decision to sample a trace takes place by considering
+all or most of the spans within the trace. Tail Sampling gives you the option to
+sample your traces based on specific criteria derived from different parts of a
+trace, which isn’t an option with Head Sampling.
 
 ![Illustration shows how spans originate from a root span. After the spans are complete, the tail sampling processor makes a sampling decision.](tail_sampling_process.png)
 
@@ -85,7 +85,8 @@ Some examples of how you can use Tail Sampling include:
 - Always sampling traces that contain an error
 - Sampling traces based on overall latency
 - Sampling traces based on the presence or value of specific attributes on one
-  or more spans in a trace; for example, sampling more traces originating from a newly deployed service
+  or more spans in a trace; for example, sampling more traces originating from a
+  newly deployed service
 - Applying different sampling rates to traces based on certain criteria
 
 As you can see, Tail Sampling allows for a much higher degree of sophistication.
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/sampling/index.md
```diff
@@ -30,18 +30,93 @@ want to sample. You might want to customize your sampling strategy to:
...
+
+As you can see, Tail Sampling allows for a much higher degree of sophistication.
+For larger systems that must sample telemetry, it is almost always necessary to
+use Tail Sampling to balance data volume with usefulness of that data.
+
+There are three primary downsides to Tail Sampling today:
+
+- Tail Sampling can be difficult to implement. Depending on the kind of sampling
+  techniques available to you, it is not always a "set and forget" kind of
+  setting. As your systems change, so too will your sampling strategies. For a
```
kentquirk, line 100:
```suggestion
  thing. As your systems change, so too will your sampling strategies. For a
```

content/en/docs/concepts/sampling/index.md
```diff
@@ -30,18 +30,93 @@ want to sample. You might want to customize your sampling strategy to:
...
+  large and sophisticated distributed system, rules that implement sampling
+  strategies can also be large and sophisticated
```
kentquirk, line 102:
```suggestion
  strategies can also be large and sophisticated.
```

chalin (commented):
I'd avoid capitalizing "head sampling" and "tail sampling" (unless the context asks for it like a heading or the start of a sentence).

content/en/docs/concepts/sampling/index.md
```diff
@@ -30,18 +30,93 @@ want to sample. You might want to customize your sampling strategy to:
...
+state that they are "sampling out data" or that data not processed or exported
+is considered "sampled". These are incorrect statements.
+
+## Head Sampling
+
+Head sampling is a sampling technique used to make a sampling decision as early
+as possible. A decision to sample or drop a span or trace is not made by
+inspecting the trace as a whole.
+
+For example, the most common form of head sampling is Consistent Probability
```
svrnm, line 54: can we link to the spec of Consistent Probability Sampling here?

svrnm (approved):
non-blocking comment, great addition:)

content/en/docs/concepts/sampling/index.md
cartermp: Doneski!

cartermp:
Hmmm, after I hit the button I wondered if the head sampling bit should not use the Consistent Probability sampler terminology, since technically SDKs don't implement this yet.

...but, like, the concept is still the same from the perspective of an end-user. Deterministically taking a fraction of traces is what both do, but the CPS propagates metadata that tells you information about the total population that the sample comes from. This makes the CPS superior because it lets a telemetry backend re-weight counts so that any aggregation you derive from tracing data (like total # of requests) is actually correct w.r.t. the total population compared to the sample.

...anyways, I'm really not too worried at the moment, as it's easy enough to change. Frankly this is the direction all the head samplers should be heading anyways.

jmacd (commented):
LGTM

PR #2469, opened 2023-03-08
Order node.js over browser for TOC

cartermp:
[Node is hit a lot more in our docs](https://play.honeycomb.io/opentelemetry-web/datasets/opentelemetry.io/result/ENZskzNonnB?useStackedGraphs). From my own work, I can also say that teams using node are ~4.5x more numerous than teams sending instrumentation from the browser. So let's have the TOC reflect that.

image

### DIFF
--- a/content/en/docs/instrumentation/js/getting-started/browser.md
+++ b/content/en/docs/instrumentation/js/getting-started/browser.md
@@ -1,6 +1,7 @@
 ---
 title: Browser
 aliases: [/docs/js/getting_started/browser]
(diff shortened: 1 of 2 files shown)


chalin (approved):
LGTM, though it might be a good idea to multiply the weights by 10, to give us some room (though this section might not have new pages). As is, is ok too.

cartermp:
Yeah, I doubt there'd be more in this subfolder, but it's easy enough to tweak later.

PR #2497, opened 2023-03-13
CI: warn about external links to local pages rather than abort with an error

chalin:
- The site build fails under Windows, apparently because of the difference in behavior of the [adjust-pages.pl](https://github.com/open-telemetry/opentelemetry.io/blob/main/scripts/content-modules/adjust-pages.pl) script.
- For now, turn those errors into warnings. I'll reintroduce the errors separately, via: #2448
- Contributes to #2448

/cc @svrnm @tsloughter


tsloughter:
It isn't a failure under Windows (I'm on Fedora) but this does work and allows me to run `npm serve` successfully. Thanks!

svrnm (approved):

tsloughter (approved):

PR #2505, opened 2023-03-14
Add ruby head sampling doc

cartermp:
More in service towards https://github.com/open-telemetry/opentelemetry.io/pull/2463

@open-telemetry/ruby-approvers I'm not 100% confident in this being the ideal way to configure a sampler in code, does this look right? Couldn't find any examples online to pull from and it appears you need to new up a tracerprovider yourself to do it.

https://deploy-preview-2505--opentelemetry.netlify.app/docs/instrumentation/ruby/sampling/


content/en/docs/instrumentation/ruby/sampling.md
```diff
@@ -0,0 +1,51 @@
...
+## Default behavior
+
+By default, all spans are sampled, and thus, 100% of traces are sampled. If you
+do not need to manage data volume, don't bother setting a sampler.
+
+## TraceIDRatioBased Sampler
+
+When sampling, the most common head sampler to use is the TraceIdRatioBased
+sampler. It deterministically samples a percentage of traces that you pass in as
+a parameter.
```
chalin, line 19:
```suggestion
The most common head sampler to use is the [TraceIdRatioBased][]
sampler. It deterministically samples a percentage of traces that you pass in as
a parameter.

[TraceIdRatioBased]: https://open-telemetry.github.io/opentelemetry-ruby/opentelemetry-sdk/latest/OpenTelemetry/SDK/Trace/Samplers/TraceIdRatioBased.html
```

content/en/docs/instrumentation/ruby/sampling.md
```diff
@@ -0,0 +1,51 @@
...
+
+### Environment Variables
+
+You can configure the TraceIdRatioBasedSampler with environment variables:
```
chalin, line 23:
```suggestion
You can configure a `TraceIdRatioBased` sampler with environment variables:
```

content/en/docs/instrumentation/ruby/sampling.md
```diff
@@ -0,0 +1,51 @@
...
+```shell
+export OTEL_TRACES_SAMPLER="traceidratio"
+export OTEL_TRACES_SAMPLER_ARG="0.1"
+```
+
+This tells the SDK to sample spans such that only 10% of traces get exported.
+
+### Configuration in Code
+
+You can also configure the TraceIdRatioBased sampler in code. Here's an example:
```
chalin, line 34:
```suggestion
You can also configure a `TraceIdRatioBased` sampler in code. Here's an example:
```

chalin (approved):

content/en/docs/instrumentation/ruby/sampling.md
```diff
@@ -0,0 +1,53 @@
...
+weight: 6
+---
+
+[Sampling](/docs/concepts/sampling/) is a process that restricts the amount of
+traces that are generated by a system. The Ruby SDK offers several
+[head samplers](/docs/concepts/sampling#head-sampling).
+
+## Default behavior
+
+By default, all spans are sampled, and thus, 100% of traces are sampled. If you
```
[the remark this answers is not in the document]
cartermp: That's correct, although it's also the same outcome as always sampling (since shouldSample always returns true for the root if you don't configure anything). I can elaborate in the text if you think that will help, though.

content/en/docs/instrumentation/ruby/sampling.md
```diff
@@ -0,0 +1,54 @@
...
+do not need to manage data volume, don't bother setting a sampler.
+
+## TraceIDRatioBased Sampler
+
+The most common head sampler to use is the [TraceIdRatioBased][] sampler. It
+deterministically samples a percentage of traces that you pass in as a
+parameter.
+
+[TraceIdRatioBased]:
+  https://open-telemetry.github.io/opentelemetry-ruby/opentelemetry-sdk/latest/OpenTelemetry/SDK/Trace/Samplers/TraceIdRatioBased.html
```
cartermp, line 22: @chalin for some reason this url gives a 404 in CI. But the site is it points to is live and works?

chalin: That sometimes happens when there are too many redirects or the redirects introduce too much latency. I'm investigating.

Could we get #2504 reviewed and merged so that the Netlify build can pass even if there are link-check issues (as is the case here)?

chalin: Nope, in this case the Ruby docs are _actually_ returning a 404, then there's a script in the page that dynamically changes the location to point to the current version. It's probably not a good idea to return 404 in that case.

chalin, line 22: Let's use the current version for now. We can adjust later, WDYT?

```suggestion
  https://open-telemetry.github.io/opentelemetry-ruby/opentelemetry-sdk/v1.2.0/OpenTelemetry/SDK/Trace/Samplers/TraceIdRatioBased.html
```

chalin: FYI, I've filed the following issue:

- https://github.com/open-telemetry/opentelemetry-ruby/issues/1435

chalin:
FYI, I've filed the following issue:

- https://github.com/open-telemetry/opentelemetry-ruby/issues/1435

cartermp:
@arielvalentin I elaborated a bit on the default -- mind giving this another look?

content/en/docs/instrumentation/ruby/sampling.md
```diff
@@ -0,0 +1,64 @@
...
+You can also configure a `TraceIdRatioBased` sampler in code. Here's an example:
+
+```ruby
+require 'opentelemetry/sdk'
+require 'opentelemetry/exporter/otlp'
+require 'opentelemetry/instrumentation/all'
+
+# sample 10% of traces
+sampling_percentage = 0.1
+sampler = Samplers.trace_id_ratio_based(sampling_percentage)
```
arielvalentin, line 56:
```suggestion
sampler = OpenTelemetry::SDK::Trace::Samplers.trace_id_ratio_based(sampling_percentage)
```

content/en/docs/instrumentation/ruby/sampling.md
```diff
@@ -0,0 +1,64 @@
...
+
+tracer_provider = OpenTelemetry.tracer_provider(sampler: sampler)
```
arielvalentin, line 58: This example is incorrect. You have to create a new object instance to set on the configurator block below. 

I do not recommend documenting the programmatic configuration for a use case where the users will be setting up default samplers. 

The author would have to initialize the tracer provider themselves and I do not recommend we document it unless folks want to write their own custom samplers: 

https://github.com/open-telemetry/opentelemetry-ruby/blob/main/sdk/lib/opentelemetry/sdk/trace/tracer_provider.rb#L30

Users can do something like this in code to achieve the same results for default samplers

```ruby
ENV['OTEL_TRACES_SAMPLER'] ||= "traceidratio"
ENV['OTEL_TRACES_SAMPLER_ARG'] ||= "0.1"
```

arielvalentin (changes requested):

content/en/docs/instrumentation/ruby/sampling.md
cartermp: Thanks! Yeah, I was worried about this since the Ruby SDK does a great job of letting you easily set things up programmatically, but this one thing all of a sudden sets you down a hard mode path that doesn't feel right. Maybe at some point we can have docs for custom samplers and other stuff, but for now I'm more than happy to nix this whole section.

arielvalentin: 🙇🏼 thank you 🙇🏼

cartermp:
@arielvalentin one last review of the commentary and I think we're good!

cartermp:
@arielvalentin ping! 🙂

arielvalentin (approved):

cartermp:
No worries - we're all super busy 🙂

PR #2510, opened 2023-03-15
add red hat to vendor list

frzifus:


svrnm:
Hey, this looks more like an integration (https://opentelemetry.io/ecosystem/integrations/)? Vendors are right now (mostly commercial) backends.

frzifus:
hm.. get it - would it then be something like this? 
https://catalog.redhat.com/software/operators/detail/5ec54a5c78e79e6a879fa271

But still its a product that a customer deploys on its own. :sweat_smile:

svrnm:
tbh the whole ecosystem pages are still a mess, I really want to find some time to improve them, but so far I was not able to do that.

We had a similar situation with Apache Skywalking recently, where we asked them to add themselves to the integrators not the vendors, since vendors so far are commercial observability backends providers.

@cartermp wdyt?

cartermp:
Yeah, I agree with this. Even though this blurs the lines between ecosystem integration and commercial product, the current vendors page is really mostly focused on backends and mechanisms by which you send data to them (e.g., may not support OTLP but does support an exporter in a distro, or vice/versa).

cartermp:
I would say that this counts as "collector" for an entry in that table.

PR #2514, opened 2023-03-15
SDK.start doesn't seem to return a promise

sadikassistivlabs:
I ran into this earlier while using vanilla JS (it crashed on the `then`)

And just now when using TS it says `start` returns void.


cartermp (approved):
Ahhhhh good catch, this looks like it was recently changed about a month ago. Used to be async/promise-returning, now it's sync (like it probably should have been. Thank you!

sadikassistivlabs:
> @sadikassistivlabs if you can sign the CLA, then we can merge this.

Yup just got through I think

cartermp:
Yep, you got it, thanks again!

PR #2539, opened 2023-03-23
Fixed path to config.yaml

agronholm:
The path is now correct for the opentelemetry-collector-contrib image.


cartermp (approved):
Thanks, yes, this is correct

PR #2553, opened 2023-03-29
Docs: fix span links in python example

mikek:
Actually create two independent spans and just "link" them from the second span. Th original example used nested spans that effectively made them parent/child, which contradicted the description and the point of using a link.


cartermp (approved):
Thanks, yeah, this is a good change. The original example was confusing.

PR #2568, opened 2023-04-02
Auto-update java instrumentation version

trask:
we'll turn off the auto-update that is triggered from the instrumentation repo after this is merged


mateuszrzeszutek (approved):

cartermp (approved):

.github/workflows/auto-update-versions.yml
```diff
@@ -24,6 +24,7 @@ jobs:
           .github/workflows/scripts/auto-update-version.sh "opentelemetry-collector-releases" "collectorVersion" "content/en/docs/collector/_index.md"
           git reset --hard origin/main
           .github/workflows/scripts/auto-update-version.sh "opentelemetry-java" "javaVersion" "content/en/docs/instrumentation/java/_index.md"
+          .github/workflows/scripts/auto-update-version.sh "opentelemetry-java-instrumentation" "javaInstrumentationVersion" "content/en/docs/instrumentation/java/automatic/annotations.md"
```
chalin, line 27: @trask - just double checking: was it intentional to **_not_** do a hard reset between the Java and Java-instrumentation version-update commands? If both are updated, the second PR will have two file updates (the one from the first PR and the java-instrumentation update.

If it wasn't intentional, then here's a fix:

- #2665

trask: oops, no, thx!

PR #2593, opened 2023-04-11
Also add format commands

cartermp:
Inspired by my recent misadventure of running:

```
npm run format
npm run format:write
npm run prettier:write
```

because I don't read instructions, I expect others won't, so here's an attempt to help things

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,8 @@
     "cp:spec": "./scripts/content-modules/cp-pages.sh",
     "diff:check": "git diff --name-only --exit-code || (echo; echo 'WARNING: the files above have not been committed'; echo)",
     "diff:fail": "git diff --name-only --exit-code || (echo; echo 'ERROR: the files above have changed. Locally rerun `npm run test` and commit changes'; echo; git diff | head -100; exit 1)",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -32,6 +32,8 @@
     "cp:spec": "./scripts/content-modules/cp-pages.sh",
     "diff:check": "git diff --name-only --exit-code || (echo; echo 'WARNING: the files above have not been committed'; echo)",
     "diff:fail": "git diff --name-only --exit-code || (echo; echo 'ERROR: the files above have changed. Locally rerun `npm run test` and commit changes'; echo; git diff | head -100; exit 1)",
+    "format": "npm run _check:formatting -- --write",
+    "format:write": "npm run _check:formatting -- --write",
```
chalin, line 36: I'd vote for at most one script alias (I'd even eliminate `prettier:write`, but I can do that in a followup PR since there's some cleanup I'd like to do at the same time):

```suggestion
    "format": "npm run _check:formatting -- --write",
```

chalin (approved):
LGTM with the given suggestions

PR #2597, opened 2023-04-13
Fix collector download URL for macOS

gouthamve:
The existing URL returns a not found. I have verified that the new URL is valid and correct in both bash and ZSH.


jpkrohling (approved):

chalin (approved):
Thanks for the fix!

PR #2599, opened 2023-04-13
Install @opentelemetry/api when instrumenting app

Symbianx:
The auto instrumentation docs are broken because at some point, the `@opentelemetry/api` dependency was changed to a peer dependency in the upstream packages.
This causes packages install to fail.

This PR adds the api package to the install command.


cartermp (approved):

cartermp:
Thanks!

Issue #2604, opened 2023-04-14
The example code for Getting Started in NodeJS has broken code


platoyuval:
Two issues:
1. 
The installation step for dependencies is missing `@opentelemetry/sdk-trace-node`

2. Cannot find name 'opentelemetry'. line 9

```
exporter: new opentelemetry.metrics.ConsoleMetricExporter()
```

This fixed code works:
```
import { NodeSDK } from '@opentelemetry/sdk-node'
import { ConsoleSpanExporter } from '@opentelemetry/sdk-trace-node'
import { getNodeAutoInstrumentations } from '@opentelemetry/auto-instrumentations-node'
import { PeriodicExportingMetricReader, ConsoleMetricExporter } from '@opentelemetry/sdk-metrics'

const sdk = new NodeSDK({
  traceExporter: new ConsoleSpanExporter(),
  metricReader: new PeriodicExportingMetricReader({
    exporter: new ConsoleMetricExporter()
  }),
  instrumentations: [getNodeAutoInstrumentations()]
})

sdk.start()
```

cartermp:
Fixed by https://github.com/open-telemetry/opentelemetry.io/pull/2605

For issue (1), I can't repreduce that. That package is installed by `sdk-node` transitively so it just works.

For issue (2), yeah, definitely noticed that. I saw some other discrepancies and cleaned things up more.

platoyuval:
One thing that might be different is that I used `pnpm`, not `npm`

PR #2605, opened 2023-04-15
Node getting started fixes

cartermp:
fixes https://github.com/open-telemetry/opentelemetry.io/issues/2604

Also cleans up some import/qualifications, and quotes on strings

Validated with a local app I created from scratch.

### DIFF
--- a/content/en/docs/instrumentation/js/getting-started/nodejs.md
+++ b/content/en/docs/instrumentation/js/getting-started/nodejs.md
@@ -48,12 +48,6 @@ npm install express
 
 ### Code
 
-If you are using TypeScript, then run following command:
-
-```shell
-tsc --init
-```
-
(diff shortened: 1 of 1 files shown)


chalin (approved):

platoyuval (approved):
LGTM.

In another PR, maybe worth adding some section/note for developers using `pnpm` (e.g. pnpm exec instead of npx, and maybe an explicit package dependency needed)

svrnm (approved):

svrnm:
> platoyuval

A note would be fine, but overall we need to compromise at some point between giving options and readability. The TypeScript/JavaScript tabs are already making things much more complicated than for other languages. @platoyuval if you provide a PR with a suggestion I am happy to review it.

PR #2606, opened 2023-04-15
Add Teletrace to vendor list

ophiryael:
[Teletrace](https://github.com/teletrace/teletrace) is an open-source tracing platform. In short, the project architecture includes an [OTel distribution](https://github.com/teletrace/teletrace/tree/main/teletrace-otelcol), query backend, and a UI.


cartermp (commented):
I'm not quite sure where to place this one. Is teletrace a commerical vendor, as in, is there a commercial offering? Otherwise I feel like it maybe belongs under Ecosystem.

ophiryael:
Hi @cartermp, thanks for the response.
Teletrace does not have a commercial offering. It is a completely free and open-source project.
When you say that it belongs under Ecosystem, do you mean the "Integrations" -> "Within OpenTelemetry" section? 
I can move it there, but I think that putting it in a list next to tools such as Kafka and Elasticsearch doesn't make much sense, as there is no boundary between tools and more complete systems such as the ones in the vendor's list. Adding a new section/column for open-source projects such as Jaeger and Teletrace under the vendor's page could be a better solution for the future.

Let me know if you want me to add it to the integrations section or elsewhere.

cartermp (approved):
I'm gonna not let perfect be the enemy of good here. We need to clean up this page anyways, and it seems like Teletrace is more like a vendor than an integration, so when we clean stuff up we can divvy out the OSS stuff from the commercial stuff.

PR #2660, opened 2023-05-05
Blog post: histograms vs summaries

dyladan:
Follow-up blog to #2658 

This one is even less otel-centric and is likely to need some copy changes. I set the date to a week in the future to allow some time between posts and some time for copy edits.

---

Preview: https://deploy-preview-2660--opentelemetry.netlify.app/blog/2023/histograms-vs-summaries/

PR #2658, opened 2023-05-05
Blog: why histograms?

dyladan:
Guest post from https://dyladan.me/histograms/2023/05/02/why-histograms/

This is the first in a series of 3 posts. The second explains the differences and tradeoffs between histograms and summaries (https://dyladan.me/histograms/2023/05/03/histograms-vs-summaries/). The third explains exponential histograms in detail (https://dyladan.me/histograms/2023/05/04/exponential-histograms/).

Please feel free to share any editing or copy suggestions.

---

**Preview**: https://deploy-preview-2658--opentelemetry.netlify.app/blog/2023/why-histograms/

PR #2661, opened 2023-05-05
Blog: exponential histograms

dyladan:
Follow up to #2658 and #2660 

Link checker will fail until #2660 is merged.

Set date for the 22nd to allow another week after the second post before publishing.


content/en/blog/2023/exponential-histograms.md
```diff
@@ -0,0 +1,196 @@
...
+title: Exponential Histograms
+date: 2023-05-22
+author: '[Daniel Dyla](https://github.com/dyladan)'
+spelling: cSpell:ignore Dyla quantiles subsetting Ruslan Vovalov Ganesh Vernekar
+canonical_url: https://dyladan.me/histograms/2023/05/04/exponential-histograms/
+---
+
+Previously, in [Why Histograms?][] and [Histograms vs Summaries][], I have gone
+over the basics of histograms and summaries and the tradeoffs, benefits, and
+limitations of each of them. Because they're easy to understand and demonstrate,
```
chalin, line 11:
```suggestion
Previously, in [Why Histograms?][] and [Histograms vs Summaries][], I went
over the basics of histograms and summaries, explaining the tradeoffs, benefits, and
limitations of each. Because they're easy to understand and demonstrate,
```

content/en/blog/2023/exponential-histograms.md
```diff
@@ -0,0 +1,196 @@
...
+those posts focused on so-called explicit bucket histograms. The exponential
+bucket histogram, also referred to as native histogram in Prometheus, is a
+low-cost, efficient alternative to explicit bucket histograms. In this post, I
+will go through what they are, how they work, and the problems they solve that
```
chalin, line 15:
```suggestion
go through what they are, how they work, and the problems they solve that
```

content/en/blog/2023/exponential-histograms.md
```diff
@@ -0,0 +1,196 @@
...
+explicit bucket histograms struggle with.
+
+# Types of histograms
```
chalin, line 18: Remember that headings start at level 2 in our docs and blog pages. Please make the update here and for all other headings in this post.

```suggestion
## Types of histograms
```

content/en/blog/2023/exponential-histograms.md
```diff
@@ -0,0 +1,196 @@
...
+SLOs.
+
+When using linear interpolation, the expected relative error is half the bucket
+width divided by the bucket midpoint. Because the relative error is the same
+across all buckets, we can use the first bucket with the upper bound of the base
+to make the math easy. An example is shown below using a scale of 3.
+
+```
+scale = 3
+# see above for base calculation
```
chalin, line 121:
```suggestion
# For base calculation, see above
```

content/en/blog/2023/exponential-histograms.md
```diff
@@ -0,0 +1,196 @@
...
+_A version of this article was [originally posted][] to the author's blog._
+
+[originally posted]: {{% param canonical_url %}}
+
+[Why Histograms?]: {{% relref "why-histograms" %}}
+
+[Histograms vs Summaries]: {{% relref "histograms-vs-summaries" %}}
+
+[Using OpenTelemetry’s Exponential Histograms in Prometheus]:
+  https://www.youtube.com/watch?v=W2_TpDcess8
```
chalin, line 196:
```suggestion
[Why Histograms?]: ../why-histograms/
[Histograms vs Summaries]: ../histograms-vs-summaries/
[Using OpenTelemetry’s Exponential Histograms in Prometheus]:
  https://www.youtube.com/watch?v=W2_TpDcess8

[originally posted]: {{% param canonical_url %}}
```

chalin (commented):
See inline comments. 

**Reminder**: headings start at level 2 in our docs and blog pages. Please update all headings appropriately in this new post. Thanks!

content/en/blog/2023/exponential-histograms.md
dyladan: When I did this before the formatter got mad. It tries to combine all 3 link definition lines into a single paragraph which breaks the markdown

dyladan: This is what `npm run format` does to the markdown in your suggestion:

```
[Why Histograms?]: {{% relref "why-histograms" %}} [Histograms vs Summaries]:
{{% relref "histograms-vs-summaries" %}} [Using OpenTelemetry’s Exponential
Histograms in Prometheus]: https://www.youtube.com/watch?v=W2_TpDcess8

[originally posted]: {{% param canonical_url %}}
```

dyladan: Found a way to make it nicer using prettier ignore directives

svrnm (approved):

content/en/blog/2023/exponential-histograms.md
```diff
@@ -0,0 +1,196 @@
...
+boundaries using an exponential growth function. This means each consecutive
+bucket is larger than the previous bucket and ensures a constant relative error
+for every bucket.
+
+## Exponential histograms
+
+In OpenTelemetry exponential histograms, buckets are calculated automatically
+from an integer _scale factor_, with larger scale factors offering smaller
+buckets and greater precision. It is important to select a scale factor that is
+appropriate for the scale of values you are collecting in order to minimize
```
oertl, line 35: maybe "distribution of values" instead of "scale of values"?

content/en/blog/2023/exponential-histograms.md
```diff
@@ -0,0 +1,196 @@
...
+single histogram with exactly the precision of the least precise histogram being
+combined. For example, histogram _A_ with scale 3 and histogram _B_ with scale 2
+can be combined into a single histogram _C_ with scale 2 by first summing each
+pair of neighboring buckets in _A_ to form histogram _A'_ with scale 2. Then,
+each bucket in _A'_ is summed with the corresponding bucket of the same index in
+_B_ to make _C_.
+
+## Error rate
+
+The error rate of a histogram is defined as the average relative error when
```
oertl, line 105: The error of a histogram is defined as the maximum relative error when...

content/en/blog/2023/exponential-histograms.md
```diff
@@ -0,0 +1,196 @@
...
+estimating the value of a particular-ranked data point. This is important
+because it is how φ-quantiles are estimated. In a histogram with x data points,
+the nth percentile is the data point at rank `n / 100 * x`. For an example of
```
oertl, line 108: n-th

content/en/blog/2023/exponential-histograms.md
```diff
@@ -0,0 +1,196 @@
...
+this calculation, see my previous post [Histograms vs Summaries][]. When
+estimating ɸ-quantiles it is very important to know the expected relative error
```
oertl, line 110: I think you can drop mentioning the expected relative error here. More important is the maximum relative error, for which the calculation is shown below.

content/en/blog/2023/exponential-histograms.md
```diff
@@ -0,0 +1,196 @@
...
+rates and maximum relative error rates so that you can effectively monitor your
```
oertl, line 111: I would always write  "relative error" instead of "relative error rate"

content/en/blog/2023/exponential-histograms.md
```diff
@@ -0,0 +1,196 @@
...
+SLOs.
+
+When using linear interpolation, the expected relative error is half the bucket
```
oertl, line 114: The relative error is minimized, if the harmonic mean of the bucket boundaries is used for estimation.
For bucket with index `i`, the lower and upper boundaries are given by `l = base^i` and `u = base^(i + 1)`, respectively. The estimated value `x` given by the harmonic mean would be
```
x = 2*l*u / (l + u)
  = (2 * base^i * base^(i + 1)) / (base^i + base^(i + 1))
  = 2*base^(i+1) / (1 + base)
```
The worst case estimation error appears, if the original value was either close to the lower boundary `l` or the upper boundary `u`. We can calculate the relative error for both cases:
```
(x - l) / l = x/l -1 = (2*base) / (1 + base) - 1 = (base - 1) / (base + 1)
(u - x) / u = 1 - x/u = 1 - 2 / (1 + base) = (base - 1) / (base + 1)
```
Since the relative error is the same in both extreme cases, `(base - 1) / (base + 1)` also defines the maximum relative estimation error. For `scale = 3` with corresponding `base  = 1.090508` (see base calculation above) the maximum relative error would be 
```
(1.090508 - 1) / (1.090508 + 1) = 0.04329 = 4.329%
```

content/en/blog/2023/exponential-histograms.md
dyladan: Isn't it possible to have an absolute error of the entire bucket width?

If a bucket goes from 1 to 2, and you have 100 values of 1.0001, using linear interpolation to select the 99th point in the bucket would result in 1.99 when the actual value is 1.0001. All of the calculations used in the OTEP and spec use the bucket midpoint for the error calculation. To me, that indicates it is the average or expected relative error, but not the maximum.

cartermp (approved):
I think this looks good to me, unless there's no objections let's merge this early in the week.

cartermp:
Thanks for everything @dyladan! This will be going live now.

Issue #2387, opened 2023-02-23
Add GH action to pull in latest release of the OTel specification


chalin:
@open-telemetry/specs-approvers - is this something that you'd like?

This is a followup to

- #1963 

which was closed since the collector (and Java) scripts have been updated (via #2351).

The processing of the spec will be different, since the check out of a submodule is involved, but I'm glad to put something together if it's automation that you'd like to see implemented. If there's another workflow that you had in mind, let me know.

- [x] #2666
- [x] #2667
- [x] #2674
  - [x] Add the spec version update to the automation config
  - [x] Also update the spec submodule, checking it out at the latest version
- [x] https://github.com/open-telemetry/opentelemetry-specification/pull/3481

/cc @trask

svrnm:
@open-telemetry/specs-approvers please take a look

tigrannajaryan:
Just to be clear, this is about automating something that is manual currently? It would be definitely nice to do!

cartermp:
Yes, correct, it's still manual today.

chalin:
Closed by #2674

chalin:
Reopening until spec-side cleanup is merged.

PR #2667, opened 2023-05-07
Auto-update script: only match whole words, and other improvements

chalin:
- Prep for #2387
- When searching for version fields, the script now only matches:
  - At word boundaries (for the "variable" name)
  - An entire line
- Executes `git reset --hard origin/main` at the start of the script when run GH
- Adds support for running the script locally (macOS and Linux)
- When run locally, the script defaults to a dry run. Use `-f` to force a regular run.

/cc @trask


svrnm (approved):
lgtm, please merge on your needs @chalin

PR #2519, opened 2023-03-17
Don't use __name__ for tracers in python docs

cartermp:
Fixes https://github.com/open-telemetry/opentelemetry.io/issues/2517

Issue #2517, opened 2023-03-16
Python docs incorrectly use `__name__` for `instrumenting_module_name`


WesleyAC:
**What needs to be changed?**

The [docs for `trace.get_tracer`](https://opentelemetry-python.readthedocs.io/en/latest/api/trace.html#opentelemetry.trace.TracerProvider.get_tracer) state:

> The uniquely identifiable name for instrumentation scope, such as instrumentation library, package, module or class name. `__name__` may not be used as this can result in different tracer names if the tracers are in different files. It is better to use a fixed string that can be imported where needed and used consistently as the name of the tracer.

However, in many places in the docs, `__name__` is used for this.

**What is the name + path of the page that needs changed?**

https://opentelemetry.io/docs/instrumentation/python/cookbook/ and others

srikanthccv:
It is fine to use `__name__`. There are many OTEL contrib instrumentation that uses `__name__` https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-python-contrib+get_tracer%28__name__&type=code and examples/docs maintained on the core repo do the same https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-python%20get_tracer(__name__&type=code. The `MAY` clause is used to let the reader know it may not always be the correct to use `__name__` for the scope.

cartermp:
> as long as the user knows the scope differentiation

Hmm, that seems to me that maybe it shouldn't be what's documented then? Or it is documented, but called with this caveat?

srikanthccv:
> The uniquely identifiable name for instrumentation scope, such as instrumentation library, package, module or class name.

Does this part of the `get_tracer` documentation make it any clear?

WesleyAC:
Yeah, as a user I did a lot of reading docs and issues about this but still couldn't answer the question "what is this name actually used for" which is probably the most important think to know to be able to choose a useful one.

srikanthccv:
>What should people name a tracer, taking the behavior that python has in place into consideration

>We need to be opinionated here. Saying it can be anything depending on what works for you isn't helpful

Following is the OTEL spec description for the instrumentation scope. 

[...]

PR #2680, opened 2023-05-09
Revert "Don't use __name__ for tracers in python docs (#2519)"

srikanthccv:
See https://github.com/open-telemetry/opentelemetry.io/issues/2517

- There is a dedicated `service.name` resource attribute semantic convention for the service name. The tracer/meter name is used for the scope. It could be seen as a system -> subsystem -> module, and the user decides which is appropriate for them.
- tracer and meter wouldn't have different names within the same scope.

### DIFF
--- a/content/en/docs/demo/services/recommendation.md
+++ b/content/en/docs/demo/services/recommendation.md
@@ -30,7 +30,7 @@ endpoints, resource attributes, and service name are automatically set by the
 OpenTelemetry auto instrumentor based on environment variables.
 
 ```python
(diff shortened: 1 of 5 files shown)


svrnm (approved):
@srikanthccv apologies if the other PR was merged in too early, I was under the assumption that we had all the approvals. Since my python knowledge is limited here, we rely on your expertise. I am happy to merge this and revert those changes.

srikanthccv:
I don't mind the current changes.

PR #2689, opened 2023-05-10
Python getting started: use port 8080 consistently

pcasaretto:
1. Added surrounding double quotes to the metrics example.
2. Added port flag instructions to all flask run instances.


chalin (commented):
After rebasing, the change you marked at (1) is gone because a previous PR made that change. I'll let @cartermp et al. decide if they'd like the other changes.

svrnm (approved):

svrnm:
yeah, I think it makes sense to keep the port set to 8080 consistently.

PR #2681, opened 2023-05-09
feat(js-auto-instrumentation): Added Automatic Instrumentation Section for JS

samimusallam:
Added a JS Auto instrumentation section explaining how to set up and configure the module.

PR #2730, opened 2023-05-16
Updates to text-lint and cspell settings

svrnm:
Because of #2681 I updated some of the textlint and cspell configs and ran some semi-automatic clean ups, some more PRs to follow.


.vscode/cspell.json
```diff
@@ -14,6 +14,8 @@
     "hugo",
     "initializers",
     "Istio",
+    "Jaeger",
```
chalin, line 17: I suggest lowercase here and also adding a textlint rule for the proper capitalization. WDYT?

```suggestion
    "jaeger",
```

.vscode/cspell.json
```diff
@@ -14,6 +14,8 @@
...
+    "jaegertracing",
     "Javadoc",
     "Knative",
     "Kubecon",
```
chalin, line 21: Similar comment to the one above:

```suggestion
    "kubecon",
```

Then add "KubeCon" to the textlint word set.

.vscode/cspell.json
```diff
@@ -28,14 +30,17 @@
     "otlp",
     "OTLP",
```
chalin, line 31: "OTEP" above should be "otep", with an entry added to textlint. Ditto for OTLP:

```suggestion
    "otlp",
```

svrnm:
@chalin I updated the cspell & textlint configuration in a way that cspell has the job to take care of spelling and textlint takes care of proper capitalization. I also added some packages that help to exclude some false positives. Please take a look

package.json
```diff
@@ -67,18 +69,57 @@
...
   "textlint": {
+    "plugins": {
+      "@textlint/text": {
+        "extensions": [".yml"]
+      }
+    },
+    "filters": {
+      "comments": {
+        "disablingComment": "prettier-ignore-start",
+        "enablingComment": "prettier-ignore-end"
```
chalin, line 80: We generally prettier-ignore tabbed panes just because they confuse Prettier. Maybe we should still spell-check the content, and use text-lint specific disabling comments? Just a thought, I'm ok with this for now.

chalin (approved):
Thanks for working on this. See inline for question/comments.

package.json
svrnm: Textlint has exactly the same issue, the tabs confuse it because it does not understand that those are code blocks (while it understands \`\`\` just fine). My line of thinking now was instead of having two comments I just reuse the ones that are there already.

This btw also [explains the extra spaces](https://github.com/open-telemetry/opentelemetry.io/pull/2791#discussion_r1205297168), the plugin requires the lines to be surrounded by newlines

chalin: Ok, thanks for the explanation.

Yeah, I think that we need a kind of tab pane that allows use of standard markdown code fences (I've take note of this (IMHO) desirable feature).

cartermp (approved):

PR #2750, opened 2023-05-19
Upload new OTEL SDK diagram for No Collector page

hillarymc:
This is the latest diagram for the No Collector page. It is now compliant with the brand guidelines.


cartermp (approved):
Thanks!

PR #2755, opened 2023-05-22
Fix import

joostlek:
Hey!

I tried this in my own project and my IDE couldn't find `opentelemetry.SpanStatusCode`. My experience with typescript d.ts files isn't great, but what I saw is that `opentelemetry` in this context would only be `_default`.
afbeelding

Kind regards


svrnm:
thanks @joostlek , @open-telemetry/javascript-approvers please take a look if this is correct indeed.

pichlermarc (approved):
Yes, that's correct, thanks for fixing this. :slightly_smiling_face:

svrnm:
thank you @joostlek !

PR #2756, opened 2023-05-22
Registry Updates May'23

svrnm:
Registry updates, lots of PHP :-)

### DIFF
--- a/data/registry/collector-exporter-dataset.yml
+++ b/data/registry/collector-exporter-dataset.yml
@@ -1,0 +1,13 @@
+title: DataSet Exporter
+registryType: exporter
+isThirdParty: false
+language: collector
+tags:
+  - dataset
+  - exporter
+  - collector
+repo: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/datasetexporter
+license: Apache 2.0
+description: This exporter sends logs to DataSet.
+authors: OpenTelemetry Authors
+otVersion: latest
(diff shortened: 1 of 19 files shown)


martin-majlis-s1:
@svrnm : I am new to the process, so I was following this guideline - https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/CONTRIBUTING.md#adding-new-components - that I should register the component. 

Therefore I have created this PR for adding `Dataset Exporter` - https://github.com/open-telemetry/opentelemetry.io/pull/2759 - where I am adding it.

I would say that it should be marked as `ThirdParty: true` and the authors should `Dataset`.

svrnm:
> @svrnm : I am new to the process, so I was following this guideline - https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/CONTRIBUTING.md#adding-new-components - that I should register the component.

You did everything right, I just sometimes run this script to check if we missed any component and add them semi-automatically, I will remove dataset from this PR and leave it to your PR 

> I would say that it should be marked as `ThirdParty: true` and the authors should `Dataset`.

Neither the ThirdParty nor the authors are exposed right now outside of the github repo, so I have no strong feelings about them, however the contribution of that component to opentelemetry-collector-contrib makes it non-third-party and by license the authors are "The OpenTelemetry Authors" (which you are a part of by that), but to be honest I am not sure how consistent we are across entities in the repo by now.

cartermp (approved):

PR #2761, opened 2023-05-22
Remove concept page 'Data Collection'

svrnm:
I think that we should remove the following page:

https://opentelemetry.io/docs/concepts/data-collection/

It mostly holds redundant and outdated information:

- the collector is mentioned in the "Components" page within the concepts
- Deployment section is not in alignment anymore with [Deployment](https://opentelemetry.io/docs/collector/deployment/)
- Components is outdated because Connectors are missing, they are explained at other places as well (we might consider moving that to https://opentelemetry.io/docs/concepts/components/#collector)
- The repo section is not relevant for end users (and also incomplete since we have -releases)

@open-telemetry/collector-approvers PTAL as well

---

Preview: https://deploy-preview-2761--opentelemetry.netlify.app/docs/

Redirect test:

- https://deploy-preview-2761--opentelemetry.netlify.app/docs/concepts/data-collection/


chalin (approved):
+1

+1 to the redirect -- either as an alias in the new target file, or a redirect rule
(and the three broken links will need to be addressed)

cartermp:
Yeah, that's about what I'd expect.

svrnm:
@chalin @cartermp fixed the redirects, please take another look

content/en/docs/concepts/distributions.md
```diff
@@ -11,7 +11,7 @@ support multiple [signals](../signals). The reference implementation of
 OpenTelemetry is available as:
 
 - [Language-specific instrumentation libraries](../instrumenting)
-- [A Collector binary](../data-collection)
+- [A Collector binary](/docs/collector)
```
chalin, line 14: Use canonical path (with trailing `/`):

```suggestion
- [A Collector binary](/docs/collector/)
```

chalin (commented):
@svrnm - I'm wondering if it would make more sense to redirect `/docs/concepts/data-collection/` to `/docs/concepts/components/#collector` instead, that way it stays within the Concepts section.

dmitryax (approved):
LGTM

PR #2778, opened 2023-05-23
consistent image name

codeboten:
The docker registry example used `opentelemetry-collector` while the ghcr example uses `opentelemetry-collector-contrib`

This was causing an issue as the location where the configuration file must be mounted differs from one image to the other.


codeboten:
Pinging @tylerbenson as he ran into an issue due to this docs bug

tylerbenson (approved):

cartermp (approved):
woof

PR #2783, opened 2023-05-24
Fix import

joostlek:
Same as #2755. I think the javascript one also needs to be fixed, but I am not sure if Javascript supports the same way of importing like typescript does


cartermp (approved):
Thanks! Will merge since the other was merged (and I have run into this myself)

PR #2784, opened 2023-05-24
Improve consistency

joostlek:
In the Tracing part, opentelemetry is used instead of otel. This PR gives consistency in the naming


chalin (commented):
I'd maybe vote for `otel` throughout, but will defer to @svrnm and @cartermp

chalin:
Just to be clear, I don't have a strong preference either way -- but I do agree that consistency is good! So thanks for submitting this PR.

pichlermarc (approved):
> @open-telemetry/javascript-approvers PTAL, is there any recommendation on that?

Looks good, I don't think there's a preferred way. I myself use `import opentelemetry from "@opentelemetry/api";` :slightly_smiling_face:

chalin (approved):
👍

Issue #2642, opened 2023-04-28
Adjust to move of OTLP specs into opentelemetry-proto


tigrannajaryan:
We are [moving OTLP spec](https://github.com/open-telemetry/opentelemetry-specification/issues/3415) from https://opentelemetry.io/docs/reference/specification/protocol/otlp/ to https://github.com/open-telemetry/opentelemetry-proto/tree/main/specification/ (pending https://github.com/open-telemetry/opentelemetry-proto/pull/458).

We will need to update the documentation website appropriately (I am not sure how exactly it works today, so not sure what needs to change).

Note that only the following files are moving (not the entire directory):
```
specification/protocol/README.md
specification/protocol/design-goals.md
specification/protocol/otlp.md
specification/protocol/requirements.md
```

---

### Prep

- [x] #2676
- [x] #2677
  - [x] #2701
- [x] #2691
- [x] https://github.com/open-telemetry/opentelemetry-proto/pull/472
- [x] https://github.com/open-telemetry/opentelemetry-proto/pull/469

### Tasks

- [x] #2786
- [x] #2794
- [x] #2799
- [x] https://github.com/open-telemetry/opentelemetry-proto/pull/479
  - [ ] Add aliases to OTLP spec pages
- [x] #2839
- [x] #2846
- ~Run prettifier over OTLP spec pages~

tigrannajaryan:
@open-telemetry/docs-approvers can you please help with this?

svrnm:
@chalin can you give this a look? ty

chalin:
So what is it that you would like to see on the website @tigrannajaryan and @open-telemetry/specs-approvers? 

**Two options** come to mind and are presented below. Let me know which you prefer, or whether you had something else in mind.

### 1. Publish the OTLP spec separately under "Reference"

Something like this:

> <img width="1054" alt="Screen Shot 2023-05-08 at 08 40 18" src="https://user-images.githubusercontent.com/4140793/236826440-cf33d5f5-77d9-4d3b-9c28-cd766ec5b0e0.png">

### 2. Publish the `opentelemetry-proto` repo OTLP spec through it's current location in the OTel spec ...

... that is, https://opentelemetry.io/docs/reference/specification/protocol/otlp/, as an **override** to any pages that the OTel spec might provide.

While this option is minimally disruptive, it _might_ lead to confusion if/once #2666 lands and the spec version is visible as part of the page title and breadcrumbs. That is, we might end up with something like this:

[...]

PR #2786, opened 2023-05-24
Add OTLP to docs/specs

chalin:
- Contributes (possibly closes) #2642
- Adds the OTLP spec under `/docs/specs`

**Preview**: https://deploy-preview-2786--opentelemetry.netlify.app/docs/specs/otlp

/cc @tigrannajaryan

### DIFF
--- a/hugo.yaml
+++ b/hugo.yaml
@@ -182,8 +182,10 @@ module:
       target: content/docs/instrumentation/go
     - source: tmp/otel/specification
       target: content/docs/specs/otel
-    # - source: tmp/otlp/docs
-    #   target: content/docs/specs/otlp
+    - source: tmp/otlp/docs
+      target: content/docs/specs/otlp
+    - source: tmp/otlp/opentelemetry/proto
+      target: static/docs/specs/otlp/proto
     - source: tmp/community/mission-vision-values.md
       target: content/community/mission.md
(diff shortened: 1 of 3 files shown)


tigrannajaryan:
I think "Design Goals" and "Requirements" are unnecessary here. They contain text that is only of historical interest and is not part of the protocol spec.
We can probably keep just the "Specification" page and link it from the top-level "OTLP" link in the side navbar.
What do you think?

tigrannajaryan:
> > I think "Design Goals" and "Requirements" are unnecessary here. They contain text that is only of historical interest and is not part of the protocol spec. We can probably keep just the "Specification" page and link it from the top-level "OTLP" link in the side navbar. What do you think?
> 
> That makes sense to me, but it's your call @tigrannajaryan, and that of @open-telemetry/specs-approvers.

Let's go with that. Remove "Design Goals" and "Requirements", only keep "Specification".

chalin:
> Let's go with that. Remove "Design Goals" and "Requirements", only keep "Specification".

@tigrannajaryan - done. Only the OTLP spec page is not published. Also note that I've removed the proto files. Instead, links to proto files (and folders) point back to the repo.

tigrannajaryan (approved):
I think this is great.

cartermp (approved):

PR #2790, opened 2023-05-25
Textlint fixes for blog posts

svrnm:
Fixes based on capitalization rules

### DIFF
--- a/content/en/blog/2022/instrument-apache-httpd-server/index.md
+++ b/content/en/blog/2022/instrument-apache-httpd-server/index.md
@@ -56,7 +56,7 @@ port 4317.
 
 To send a request to Apache webServer you can either use curl from terminal
 (`curl localhost:9004`), or visit [localhost:9004][] in any browser. A default
-landing page saying "Testing 123..." for Apache Http Server on Centos will be
+landing page saying "Testing 123..." for Apache HTTPD Server on Centos will be
 displayed as below:
 
 ![Testing](testing.png)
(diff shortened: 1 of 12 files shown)


content/en/blog/2022/debug-otel-with-otel/index.md
```diff
@@ -205,35 +205,35 @@ Once again we ran `docker compose up`[^1] to bring up our sample app and we send
...
 In Jaeger we still see that the traces are disconnected. However, when we look
-into one of those traces, we can see the collected request headers from nginx to
+into one of those traces, we can see the collected request headers from NGINX to
 backend:
 
 ![A screenshot of the Jaeger UI showing that http.request.header.traceparent has multiple entries.](jaeger-with-request-headers.png)
 
 There it is! The trace headers (`baggage`, `traceparent`, `tracestate`) are send
-as multiple header fields: the nginx module added the value of each of those
+as multiple header fields: the module for NGINX added the value of each of those
```
chalin, line 214: The original "the NGINX module added ..." seemed ok to me, why change the word ordering (here and later in the doc)? To me "the NGINX module" can be interpreted as a shorthand for "the module for NGINX". Besides, that's how they write it in https://opentelemetry.io/blog/2022/instrument-nginx/, which is referred to at the top of this article. But if you feel that this is more correct, I'm ok with the change.

chalin (approved):
Great job!

Just one question about "NGINX module" vs. the alternative you propose (see inline). I'm ok either way.

content/en/blog/2022/debug-otel-with-otel/index.md
svrnm: let me revert that

PR #2804, opened 2023-05-29
Move insecure option under `tis`

HPaulson:
Moves `insecure` option on Agent Deploy docs under `tls`

re: https://github.com/open-telemetry/opentelemetry-collector/pull/4063/files


cartermp (approved):
Thanks, good catch (from quite some time ago)

PR #2805, opened 2023-05-30
Update exporters.md - Adding missing library to example instrumentation.js

tqi-raurora:
The example instrumentation.js had a missing library: PeriodicExportingMetricReader

When running the example as-is with: node --require ./instrumentation.js app.js

The user would get an error:
ReferenceError: PeriodicExportingMetricReader is not defined


cartermp (approved):
Oh dear, look at that -- was missing for JS! Thanks!

tqi-raurora:
Glad to help

Figured this could help other newcomers to Nodejs like me

Issue #1689, opened 2022-08-31
Add a section in Components about "agents" & what to use instead of "agent"?


svrnm:
This is a follow-up to @trask's question (#1661): 

Sub-projects like [opentelemetry-java-instrumentation](https://github.com/open-telemetry/opentelemetry-java-instrumentation), [opentelemetry-dotnet-instrumentation](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation) & [python's opentelemetry-distro](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/opentelemetry-distro) provide a solution that is described as _"layer that adds OpenTelemetry instrumentation to a service without modifying the source code for that service."_ in [OTEPS-0001](https://github.com/open-telemetry/oteps/blob/main/text/0001-telemetry-without-manual-instrumentation.md#motivation).

I would like to add a section to [Components](https://opentelemetry.io/docs/concepts/components/) that calls out the existence of those, what value they bring to the end-user ("zero source code modification.", auto instrumentation for all my libraries, packaging of SDK, exporters, resource detectors & other building blocks, extensibility & some), when to use them and when to go for a manual instrumentation, etc.

Coming from APM I would like to call them "Agents" but as stated in the OTEP _"that term is overloaded and ambiguous"_, so throughout our docs we try to avoid it so far. Im okay with that, but to add that section to the docs, I first need a canonical term that describes that layer properly. There are a few alternatives throughout the docs & code:

- [java](https://github.com/open-telemetry/opentelemetry-java-instrumentation): OpenTelemetry (Automatic) Instrumentation for `<Language>` (plus Javaagent of course)
- [.NET](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation) OpenTelemetry `<Language>` Automatic Instrumentation
- [Python](https://opentelemetry.io/docs/instrumentation/python/distro/) OpenTelemetry Distro for `<Language>` based on what is described here: [Distributions](https://opentelemetry.io/docs/concepts/distributions/)

So, how is that _"layer that adds OpenTelemetry instrumentation to a service without modifying the source code for that service."_  called?

cc @open-telemetry/docs-approvers

svrnm:
@martinjt have a look here :-)

reyang:
Here is my suggestion:
[...]

PR #2852, opened 2023-06-09
Standardize on manual instrumentation intro text

svrnm:
Doing this in one PR for all languages on purpose: the goal is to have a common intro text for the page on manual instrumentation. Key points are:

* API & SDK dependency for standalone applications
* API only for libraries

Open for debate: Add a sentence how it differentiates to Automatic Instrumentation (e.g. like ruby has it today), I personally don't think that's necessary on this particular page, because it's already clarified via the concept pages.


content/en/docs/instrumentation/erlang/manual.md
```diff
@@ -2,12 +2,24 @@
...
+description: Manual instrumentation for OpenTelemetry Erlang/Elixir
 ---
 
-Instrumentation is the act of adding observability code to your application.
-This can be done with direct calls to the OpenTelemetry API within your code or
-including a dependency which calls the API and hooks into your project, like a
-middleware for an HTTP server.
+[Manual instrumentation](https://opentelemetry.io/docs/concepts/instrumentation/manual/)
+is the process of adding observability code to your application yourself. If you
+are developing a standalone process or service, then you would take a dependency
```
tsloughter, line 10: I know its a different meaning but the use of "process" here is a little weird in the context of Erlang. Can we do just `standalone service` or `standalone program`?

content/en/docs/instrumentation/cpp/manual.md
```diff
@@ -2,10 +2,24 @@
 title: Manual Instrumentation
 linkTitle: Manual
 weight: 30
+description: Manual instrumentation for OpenTelemetry C++
 ---
 
-Manual instrumentation is the process of adding observability code to your
-application.
+[Manual instrumentation](https://opentelemetry.io/docs/concepts/instrumentation/manual/)
```
chalin, line 8:
```suggestion
[Manual instrumentation](/docs/concepts/instrumentation/manual/)
```

content/en/docs/instrumentation/erlang/manual.md
chalin: Right. I'd vote to just drop "process".

content/en/docs/instrumentation/cpp/manual.md
```diff
@@ -2,10 +2,24 @@
...
+is the process of adding observability code to your application yourself. If you
+are developing a standalone process or service, then you would take a dependency
+on the API and the SDK.
```
chalin, line 11: Start a new paragraph:

```suggestion
is the process of adding observability code to your application yourself.

If you
are developing a standalone process or service, then you would take a dependency
on the API and the SDK.
```

Also, I can't quite make sense of the wording of the last sentence.

content/en/docs/instrumentation/cpp/manual.md
```diff
@@ -2,10 +2,24 @@
...
+
+If you are developing a library, that wants to export telemetry data using
```
chalin, line 13:
```suggestion
If you are developing a library and want it to export telemetry data using
```

content/en/docs/instrumentation/cpp/manual.md
```diff
@@ -2,10 +2,24 @@
...
+OpenTelemetry, it MUST only depend on the API and should never configure or
```
chalin, line 14:
```suggestion
OpenTelemetry, it **must** depend only on the API and should never configure or
```

Kielek:
I am not familiar with docsy, but mayve there is a possibility to extract this common text in separate file, and inject it by some script during the page build?

It will be much easier to keep it in sync for all languages.

content/en/docs/instrumentation/cpp/manual.md
```diff
@@ -2,10 +2,24 @@
...
+depend on the OpenTelemetry SDK.
+
+This way libraries will obtain a real implementation of OpenTelemetry only if
+the user application is configured for it and otherwise will function as without
+added observability.
```
chalin, line 19: Combine the paragraphs since they are both dealing with libraries?

```suggestion
depend on the OpenTelemetry SDK. In this way, libraries will obtain a real implementation of OpenTelemetry only if
the user application is configured for it and otherwise will function as without
added observability.
```

I can't quite make sense of this last sentence either.

content/en/docs/instrumentation/cpp/manual.md
```diff
@@ -2,10 +2,24 @@
...
+
+For more details, check out the
+[Library Guidelines](https://opentelemetry.io/docs/concepts/instrumentation/libraries/).
```
chalin, line 22:
```suggestion
For more details, see
[Library Guidelines](/docs/concepts/instrumentation/libraries/).
```

chalin (changes requested):
- Let's aim to keep things DRY and have the shared intro text generated from a shortcode
- See inline for suggested edits to the added text.

svrnm:
On the shortcode, I was wondering something. @chalin you might have some opinion on that.

We already make use of shortcodes in the language-specific _index pages, e.g.:

```
{{% lang_instrumentation_index_head cpp /%}}
```

The good part about that is that I can inject variables for example the used language.

The docsy documentation also has a section on reused documentation and makes use of `readfile`:

https://www.docsy.dev/docs/adding-content/shortcodes/#reuse-documentation

```
{{% readfile "installation.md" %}}
```

What would be the way to go?

content/en/docs/instrumentation/cpp/manual.md
svrnm: > In this way, libraries will obtain a real implementation of OpenTelemetry only if the user application is configured for it

This was the original sentence I copied from one of the language SIG pages. I think the message is that the library making only use of the API works just as if the API was never used until someone adds the SDK to the application. It's really hard to express that in a short sentence. But we can drop it, since it is explained in more detail in the library guides.

content/en/docs/instrumentation/erlang/manual.md
svrnm: would application work?

ferd: Application also has a name that's used differently in Erlang. We have "applications" and "library applications" as the two types of 'collections of modules' sets of dependencies in the OTP framework (the former being stateful and the latter being stateless), and therefore a single "service" or end-user application can be made of multiple "OTP applications." 

These standards are confusing, but within the ecosystem they've been in use since the mid 1990s.

"standalone program" or "standalone service" are better words that do not clash within the Erlang/OTP ecosystem and that every user there will understand properly.

svrnm: it's service now :-)

content/en/docs/instrumentation/cpp/manual.md
svrnm: I reordered the original sentence, because most people come because the develop a standalone service, although that makes things harder to understand. Let me think about that for a little bit longer (or if anyone has good suggestions on the wording, please do :))

chalin (changes requested):
Very nice!

Note that the PR now includes regressions of the specs:

image

You'll want to remove those from the commit that included them.

layouts/shortcodes/docs/instrumentation/manual-intro.md
cartermp, line 1: Hmmm. I like where this is going, but I think we should clear things up a bit, not helped at all by opentelemetry's confusingly overloaded terminology around "API" and "SDK"

For application developers:

- They install their language's SDK
- They use the SDK to do setup, register dependencies, and critically, exporting the data
- They use the API to instrument
- It's...kind of an uninteresting detail that there's a distinction between API and SDK here, because the SDK they install already pulls in what they need

For library developers:

- They install only the appropriate "API" package for their library
- They instrument using the API
- They _do not_ configure export of any data

I think we need to be very clear about this distinction, which I don't think is clear enough in the current form.

cartermp (changes requested):

layouts/shortcodes/docs/instrumentation/manual-intro.md
svrnm: I agree with you, I tried to use as much existing wording as possible, but maybe it is time to rewrite. This intro should be short and refer to the concept page on manual instrumentation for more details, let me see how I can balance that.

content/en/docs/instrumentation/cpp/manual.md
svrnm: I rewrote the whole text, outdating this discussion, will mark it as resolved

layouts/shortcodes/docs/instrumentation/manual-intro.md
svrnm: I copied some text from the concept page on manual instrumentation, I hope that makes things clearer?

svrnm:
> Very nice!
> 
> Note that the PR now includes regressions of the specs:
> 
> image
> You'll want to remove those from the commit that included them.

fixed!

layouts/shortcodes/docs/instrumentation/manual-intro.md
```diff
@@ -1,15 +1,14 @@
 [Manual instrumentation](/docs/concepts/instrumentation/manual/)
-is the process of adding observability code to your application yourself. If you
-are developing a standalone service, then you would take a dependency
-on the API and the SDK.
+is the process of adding observability code to your application yourself.
```
chalin:
```suggestion
is the process of adding observability code to an application yourself.
```

layouts/shortcodes/docs/instrumentation/manual-intro.md
```diff
@@ -1,15 +1,14 @@
...
+To do so, you need to import OpenTelemetry to your code. If you're
```
chalin:
```suggestion
To do so, your app code needs to import OpenTelemetry. If you're
```

layouts/shortcodes/docs/instrumentation/manual-intro.md
```diff
@@ -1,15 +1,14 @@
...
+developing a library or some other component that is intended to be consumed by
```
chalin:
```suggestion
developing a library or component that will be part of
```

layouts/shortcodes/docs/instrumentation/manual-intro.md
```diff
@@ -1,15 +1,14 @@
...
+a runnable binary, then you would only take a dependency on the API, because
```
chalin:
```suggestion
a runnable binary, then your could would only depend on the API, because
```

layouts/shortcodes/docs/instrumentation/manual-intro.md
```diff
@@ -1,15 +1,14 @@
...
+unless the OpenTelemetry SDK, the instrumentation does nothing and does not
```
chalin:
```suggestion
without the OpenTelemetry SDK, the instrumentation code has no effect and so does not 
```

layouts/shortcodes/docs/instrumentation/manual-intro.md
```diff
@@ -1,15 +1,14 @@
...
+impact software performance. For more details, on instrumenting libraries read
+[these guidelines](/docs/concepts/instrumentation/libraries/).
```
chalin:
```suggestion
impact software performance. For more details on instrumenting libraries, see 
[Libraries](/docs/concepts/instrumentation/libraries/).
```

layouts/shortcodes/docs/instrumentation/manual-intro.md
```diff
@@ -1,15 +1,14 @@
...
 
-If you are developing a library and want it to export telemetry data using
-OpenTelemetry, it **must** depend only on the API and should never configure or
-depend on the OpenTelemetry SDK.
+If your artifact is a standalone service, then you would take a dependency on
```
chalin, line 10:
```suggestion
If your component is a standalone service, then it will depend on both
```

chalin (commented):
Copyedits, hopefully I understood the essence of what you are trying to communicate.

(Btw, why is it important to talk about dependencies?)

layouts/shortcodes/docs/instrumentation/manual-intro.md
```diff
@@ -0,0 +1,14 @@
+[Manual instrumentation](/docs/concepts/instrumentation/manual/) is the process
+of adding observability code to your application yourself. To do so, you need to
+import OpenTelemetry to your code. If you're developing a library or some other
+component that is intended to be consumed by a runnable binary, then you would
+only take a dependency on the API, because unless the OpenTelemetry SDK, the
```
cijothomas, line 5: > I, because unless the OpenTelemetry SDK

The sentense feels incomplete.

layouts/shortcodes/docs/instrumentation/manual-intro.md
```diff
@@ -0,0 +1,14 @@
...
+instrumentation does nothing and does not impact software performance. For more
+details, on instrumenting libraries read
+[these guidelines](/docs/concepts/instrumentation/libraries/).
+
+If your artifact is a standalone service, then you would take a dependency on
+the API and the SDK.
+
+For more information about the OpenTelemetry API and SDK, see the
+[specification](/docs/specs/otel/).
```
cartermp, line 14: I..ended up basically rewriting this. Couldn't help myself. Feel free to take it however you like, though.

```suggestion
[Manual instrumentation](/docs/concepts/instrumentation/manual/) is the act
of adding observability code to your application yourself.

If you're instrumenting an application, you need to use the OpenTelemetry SDK
for your language. You'll then use the SDK to initialize OpenTelemetry and the
API to instrument your code. This will emit telemetry from your app and any
library you installed that also comes with instrumentation.

If you're instrumenting a library, you must only install the OpenTelemetry API
package for your language. This means that your library will only emit telemetry
when an application that uses the OpenTelemetry SDK uses it. Otherwise, your
library will just behave as before. For more on instrumenting libraries, read
[these guidelines](/docs/concepts/instrumentation/libraries/).

For more information about the OpenTelemetry API and SDK, see the
[specification](/docs/specs/otel/).
```

svrnm: Looks like my brain is malfunctioning to write this down properly 😆 ... I take yours!

layouts/shortcodes/docs/instrumentation/manual-intro.md
svrnm: apologies, as said in my other comment it looks like I wrote this being not really working properly. Took @cartermp's proposal!

cartermp (approved):

layouts/shortcodes/docs/instrumentation/manual-intro.md
```diff
@@ -0,0 +1,16 @@
+[Manual instrumentation](/docs/concepts/instrumentation/manual/) is the act
+of adding observability code to your application yourself.
```
chalin, line 2:
```suggestion
of adding observability code to an application yourself.
```

layouts/shortcodes/docs/instrumentation/manual-intro.md
```diff
@@ -0,0 +1,16 @@
...
+
+If you're instrumenting an application, you need to use the OpenTelemetry SDK
+for your language. You'll then use the SDK to initialize OpenTelemetry and the
+API to instrument your code. This will emit telemetry from your app and any
```
chalin, line 6: I'm ok with your use of "app" here, but let's be consistent and replace "application" in the rest of this diff hunk by "app".

chalin, line 6:
```suggestion
API to instrument your code. This will emit telemetry from your app, and any
```

layouts/shortcodes/docs/instrumentation/manual-intro.md
```diff
@@ -0,0 +1,16 @@
...
+library you installed that also comes with instrumentation.
+
+If you're instrumenting a library, you must only install the OpenTelemetry API
+package for your language. This means that your library will only emit telemetry
+when an application that uses the OpenTelemetry SDK uses it. Otherwise, your
+library will just behave as before. For more on instrumenting libraries, read
+[these guidelines](/docs/concepts/instrumentation/libraries/).
```
chalin, line 13:
```suggestion
If you're instrumenting a library, only install the OpenTelemetry API package
for your language. Your library will not emit telemetry on its own. It will only
emit telemetry when it is part of an application that uses the OpenTelemetry
SDK. For more on instrumenting libraries, see
[Libraries](/docs/concepts/instrumentation/libraries/).
```

chalin (commented):
See inline comments

layouts/shortcodes/docs/instrumentation/manual-intro.md
```diff
@@ -0,0 +1,16 @@
+[Manual instrumentation](/docs/concepts/instrumentation/manual/) is the act
+of adding observability code to your application yourself.
+
+If you're instrumenting an application, you need to use the OpenTelemetry SDK
```
cijothomas, line 4: instrumenting part just requires API only.

layouts/shortcodes/docs/instrumentation/manual-intro.md
```diff
@@ -0,0 +1,16 @@
...
+for your language. You'll then use the SDK to initialize OpenTelemetry and the
+API to instrument your code. This will emit telemetry from your app and any
+library you installed that also comes with instrumentation.
+
+If you're instrumenting a library, you must only install the OpenTelemetry API
+package for your language. This means that your library will only emit telemetry
+when an application that uses the OpenTelemetry SDK uses it. Otherwise, your
+library will just behave as before. For more on instrumenting libraries, read
+[these guidelines](/docs/concepts/instrumentation/libraries/).
+
```
cijothomas, line 14: I'd also recommend including a sentence here about "manual instrumentation does not necessarily mean creating the span/metrics by yourself. One may leverage instrumentation libraries (add link to it)"..

layouts/shortcodes/docs/instrumentation/manual-intro.md
cartermp: True, but...it serves no purpose without an exporter, which you need the SDK for.

layouts/shortcodes/docs/instrumentation/manual-intro.md
cartermp: I would disagree with that. At the risk of making this a bikeshedding moment, I think instrumentation libraries are much more in the realm of automatic instrumentation than manual instrumentation. In many cases it's literally just one line of code and then you get a bunch of data.

cijothomas: We all have different opinions, but its already clarified here:
https://opentelemetry.io/docs/concepts/instrumentation/automatic/

cartermp: Well, the definition of manual instrumentation we have here is pretty clear: https://opentelemetry.io/docs/specs/otel/glossary/#manual-instrumentation

And given packages like node's autoinstrumentations-node package, I still stand by the statement that an instrumentation library is more close to automatic instrumentation than manual instrumentation.

svrnm: This is an [endless](https://github.com/open-telemetry/opentelemetry.io/issues/1689) [discussion](https://github.com/open-telemetry/opentelemetry-specification/issues/2866) and end-user's (aka readers of the documentation) should not be bothered with.

I agree with what @cijothomas said in essence: if an end user is doing manual instrumentation they will make use of 3rd party libraries and they don't want to instrument them themselves, so they either hope that this library is natively instrumented (yeah!) OR they need to make use of an instrumentation library.

If this is no "manual instrumentation" or not, does not really matter for the end user, what matters is that they can and should make use of those to get a better instrumentation experience.

So, I hope we can solve this by wording, e.g.: 

> if you **do** manually instrument your code, you don't have to create all telemetry by yourself: your dependencies either provide native instrumentation for you, or you can make use of instrumentation libraries. To learn more about instrumentation libraries, read [this]. 

My goal is to choose a wording that neither implies the one or the other stance. 

P.S.: If you like to continue the bikesheeding, I refer to the existing discussions on this topic I linked above and, for `autoinstrumentation-node` package additionally, see https://github.com/open-telemetry/opentelemetry-js/issues/3298 (although things changed because this package now actually contains the "auto instrumentation": https://opentelemetry.io/docs/instrumentation/js/automatic/

cartermp: I don't think that's a good paragraph to add/modify. In our own definition we make it very clear that manual instrumentation means using the API to instrument yourself.

The confusion around "automatic" that other maintainers have self-inflicted is just rearing its head again here, because we have components that do not require the use of the API to instrument, and yet they're somehow not considered "automatic instrumentation" or something close to that. It's just weird. You either get components that create data for you or you have to create that data yourself.

svrnm: > I don't think that's a good paragraph to add/modify. In our own definition we make it very clear that manual instrumentation means using the API to instrument yourself.

I actually think the opposite is true and adding this sentence is valuable to end-users, since this question comes up lately often like "can I do manual instrumentation but still use instrumentation library X for my dependency".

Again, that's why I suggested we phrase it in a way that does not suggest that Instrumentation Libraries are manual instrumentation, they can supplement it (if this is the right English word) or you can combine manual instrumentation and instrumentation libraries like you can combine manual instrumentation with the javaagent in java's case (and the same is true for the equivalent in most other languages (actively avoiding the A* word here))

If we don't find consensus here, let's create a follow-up PR to discuss that and try to get this one closed.

layouts/shortcodes/docs/instrumentation/manual-intro.md
svrnm: outdated

layouts/shortcodes/docs/instrumentation/manual-intro.md
svrnm: outdated

layouts/shortcodes/docs/instrumentation/manual-intro.md
cartermp: That bring us into the bikeshed about agents, though 🙂 

I won't die on this hill, for the sake of this PR, but I fear deeply that we're inching towards a path of "technically correct in the way the maintainers reason about things" rather than "what end-users typically care about". Perhaps I'm wrong and people typically see adding an instrumentation library as some form of manual instrumentation, in which case I would propose that we change that definition instead.

svrnm: > I won't die on this hill, for the sake of this PR,

Me neither, that's why I propose we merge this PR for now and create a follow-up one for the other discussion. If this requires a change of definition, then let's see if we re-vive that battle on the spec 😆

layouts/shortcodes/docs/instrumentation/manual-intro.md
svrnm: @reyang what part is not necessarily true? That it serves no purpose without an exporter or that you need the SDK, can you elaborate?

chalin (approved):
✨ 

LGTM, after the file-format issues are addressed.

mateuszrzeszutek (approved):

layouts/shortcodes/docs/instrumentation/manual-intro.md
svrnm: fixed

Kielek (approved):

cartermp:
🎉

PR #2854, opened 2023-06-09
Reorg text-sharing shortcodes before we add more

chalin:
- Suggested prep for #2852
- There are no changes to the generated site files
  ```console
  $ npm run build    
  ...
  $ (cd public && git diff -I '^<meta \w+="[^"]*(date|time)' -I 'Last modified' -- . ':(exclude)*.xml') # no output 
  $
  ```

/cc @svrnm


layouts/shortcodes/docs/instrumentation/lang-index-head.md
```diff
@@ -0,0 +1,35 @@
+{{ $data := index $.Site.Data.instrumentation.languages (.Get 0) }}
```
chalin, line 1: Maybe this should be renamed to `lang-index-intro.md`? WDYT @svrnm?

chalin: Yeah ... I prefer "intro". The term "head" makes me think that it includes content in `<head>`. Let me make the change.

layouts/shortcodes/docs/instrumentation/lang-index-intro.md
svrnm, line 1: maybe _index-intro or even _index/intro (and then later manual/intro, automatic/intro)? so `paganeme/section` would be the pattern (I assume that we will have more than one of these eventually per page)

svrnm:
thank you @chalin 

I agree that we should standardize this part first before proceeding with #2825.

ferd (approved):

arielvalentin (approved):
👍🏼 on behalf of @open-telemetry/ruby-approvers

layouts/shortcodes/docs/instrumentation/lang-index-intro.md
chalin: We just discussed this out-of-band with @svrnm, and we'll go for `_index-intro.md` for now. We'll refactor further at a later time as needed.

chalin: (Actually, I'm going to drop the leading `_`. It doesn't add much value to distinguish between `index` and `_index` in the case of the shortcode.)

svrnm:
lgtm @chalin, thanks!

chalin:
Given that there are no changes to the generated site files, and this is just a reorg, I'll merge this now.

Issue #2861, opened 2023-06-10
Need to clarify OTEL_EXPORTER_OTLP_TRACES_ENDPOINT format forOTLP Exporter Configuration


hu6360567:
I'm having trouble of setting otlp exporter endpoints by using envirnoment variable.

As referenced in https://github.com/open-telemetry/opentelemetry-java/issues/5517, java need to set with scheme, which fails in python auto-instrument.

We need to clarify the format of otlp gprc endpoints, since the default value comes with "http", buth example not.

PR #2863, opened 2023-06-11
Use https scheme ins grpc env var exporter examples

cartermp:
fixes https://github.com/open-telemetry/opentelemetry.io/issues/2861

### DIFF
--- a/content/en/docs/concepts/sdk-configuration/otlp-exporter-configuration.md
+++ b/content/en/docs/concepts/sdk-configuration/otlp-exporter-configuration.md
@@ -22,7 +22,7 @@ endpoint and want one environment variable to control the endpoint.
 
 **Example:**
 
-- gRPC: `export OTEL_EXPORTER_OTLP_ENDPOINT="my-api-endpoint:443"`
+- gRPC: `export OTEL_EXPORTER_OTLP_ENDPOINT="https://my-api-endpoint:443"`
 - HTTP: `export OTEL_EXPORTER_OTLP_ENDPOINT="http://my-api-endpoint/"`
 
 For OTLP/HTTP, exporters in the SDK construct signal-specific URLs when this
@@ -45,7 +45,7 @@ end with `v1/traces` if using OTLP/HTTP.
 
 **Example:**
 
-- gRPC: `export OTEL_EXPORTER_OTLP_TRACES_ENDPOINT="my-api-endpoint:443"`
+- gRPC: `export OTEL_EXPORTER_OTLP_TRACES_ENDPOINT="https://my-api-endpoint:443"`
 - HTTP:
   `export OTEL_EXPORTER_OTLP_TRACES_ENDPOINT="http://my-api-endpoint/v1/traces"`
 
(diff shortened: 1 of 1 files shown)


hu6360567:
If the endpoint always comes with the scheme, then what's the use of `OTEL_EXPORTER_OTLP_INSECURE` for?

svrnm:
> If the endpoint always comes with the scheme, then what's the use of `OTEL_EXPORTER_OTLP_INSECURE` for?

`insecure` allows self-signed keys

svrnm (approved):

hu6360567:
> > If the endpoint always comes with the scheme, then what's the use of `OTEL_EXPORTER_OTLP_INSECURE` for?
> 
> `insecure` allows self-signed keys

Is it described in any specification? 
Seems `OTEL_EXPORTER_OTLP_INSECURE` in [python](https://opentelemetry-python.readthedocs.io/en/latest/sdk/environment_variables.html#opentelemetry.sdk.environment_variables.OTEL_EXPORTER_OTLP_INSECURE) would affect scheme, rather than validating server certificate.

> The [OTEL_EXPORTER_OTLP_INSECURE](https://opentelemetry-python.readthedocs.io/en/latest/sdk/environment_variables.html#envvar-OTEL_EXPORTER_OTLP_INSECURE) represents whether to enable client transport security for gRPC requests. A scheme of https takes precedence over this configuration setting. Default: False

Will this PR cause a compatibility problem?

I'm fine with the format, as long as the value can be accepted accross the ecosystem

svrnm:
> > > If the endpoint always comes with the scheme, then what's the use of `OTEL_EXPORTER_OTLP_INSECURE` for?
> > 
> > 
> > `insecure` allows self-signed keys
> 
> Is it described in any specification? Seems `OTEL_EXPORTER_OTLP_INSECURE` in [python](https://opentelemetry-python.readthedocs.io/en/latest/sdk/environment_variables.html#opentelemetry.sdk.environment_variables.OTEL_EXPORTER_OTLP_INSECURE) would affect scheme, rather than validating server certificate.
> 
> > The [OTEL_EXPORTER_OTLP_INSECURE](https://opentelemetry-python.readthedocs.io/en/latest/sdk/environment_variables.html#envvar-OTEL_EXPORTER_OTLP_INSECURE) represents whether to enable client transport security for gRPC requests. A scheme of https takes precedence over this configuration setting. Default: False
> 
> Also, for signed keys, there should be a way to specify the key server used to verify the server identity, rather than just ignoring the certificate from server.
> 
> Will this PR cause a compatibility problem?
> 
> I'm fine with the format, as long as the value can be accepted accross the ecosystem.

May bad, I was wrong here!

From the spec [[1](https://github.com/open-telemetry/opentelemetry-specification/blob/f09624bb97e9be3be259733b93be507df18927bd/specification/protocol/exporter.md?plain=1#LL19C1-L20C1)]:
- **Insecure**: Whether to enable client transport security for the exporter's gRPC connection. This option only applies to OTLP/gRPC when an endpoint is provided without the `http` or `https` scheme - OTLP/HTTP always uses the scheme provided for the `endpoint`. Implementations MAY choose to not implement the `insecure` option if it is not required or supported by the underlying gRPC client implementation.

hu6360567:
It seems the python implementation can accept the export variable with shceme, but when it has conflict with our application when auto-instrument enabled, I'll inspect it later.

cartermp:
@hu6360567 do you have any issue open here? https://github.com/open-telemetry/opentelemetry-python-contrib

I would bet they're keen on fixing the issue.

PR #2867, opened 2023-06-12
Update discussion meeting dates

reese-lee:
This PR updates the End User Discussion Group meetings with the dates for the next three months.


chalin (approved):
👍

Issue #780, opened 2021-09-24
[CI/infra] Doc file names: use kebab-case not snake_case


chalin:
We need a consistent naming convention for doc file names. I suggest adopting [kebab-case] over [other cases], [snake_case] in particular.

In particular note that we have some **Getting started** pages in `getting_started.md` files and other in `getting-started.md` files. These should be consistently named.

- JS
  - [x] #841
  - [x] `content/en/docs/js/instrumentation_examples.md` - #1095
- PHP
  - [x] https://github.com/open-telemetry/opentelemetry-php/pull/578
  - [x] #1119
- Ruby
  - [x] https://github.com/open-telemetry/opentelemetry-ruby/pull/1120
  - [x] #1098
- [x] Demo: #2293
- [x] #2869
- [x] #3019
- [ ] All that might be left are the files in iconography.
- [ ] Another thing to consider might be enforcing lowercase letters too
- [ ] Spec -- probably won't want to changes these. Note that semconv are moving to a separate repo, and so the `semantic_conventions` directory names might be disappearing:
  ```console
  $ find content -name "*_*" -type f | perl -pe 's|/_index.md||' | grep -e _
  ...
  content/en/docs/reference/specification/metrics/semantic_conventions
  content/en/docs/reference/specification/trace/semantic_conventions
  content/en/docs/reference/specification/schemas/file_format_v1.0.0.md
  content/en/docs/reference/specification/logs/semantic_conventions
  content/en/docs/reference/specification/resource/semantic_conventions/deployment_environment.md
  content/en/docs/reference/specification/resource/semantic_conventions/cloud_provider/aws
  content/en/docs/reference/specification/resource/semantic_conventions
  ```

[other cases]: https://en.wikipedia.org/wiki/Letter_case
[kebab-case]: https://en.wikipedia.org/wiki/Letter_case#Kebab-case
[snake_case]: https://en.wikipedia.org/wiki/Letter_case#Snake_case

austinlparker:
kebabs sound good to me

himanshu007-creator:
Hi , I would like to work on this issue

chalin:
Come to think of it, this has dependencies on other issues I need to address in parallel. Thanks for the offer @himanshu007-creator, I'll let you know if/when you can contribute.

chalin:
@cartermp - when creating new files (e.g., https://github.com/open-telemetry/opentelemetry-ruby/pull/1060/files) please use kebab-case filenames.

[...]

PR #2869, opened 2023-06-13
Updating diagrams with new styles

hillarymc:
Updated the following diagrams to follow similar and cohesive styling

- otel general architecture diagram
- 2 otel baggage diagrams
- traces venn diagram 
- tail sampling process image
- waterfall trace image
- otel gateway lbs sdk v2 image

---

- Contributes to #780


static/img/otel_baggage_v2.svg
svrnm, line 1: Rename to otel_bagage_2.svg

svrnm (dismissed):
overall looks good to me, only nit is that you have some _v2 suffixes that I think we can remove.

content/en/docs/collector/deployment/gateway.md
```diff
@@ -28,7 +28,7 @@ same collector instance where the tail sampling policy is applied.
 
 Let's have a look at such a case where we are using the load-balancing exporter:
 
-![Gateway deployment with load-balancing exporter](../../img/gateway-lb-sdk.svg)
+![Gateway deployment with load-balancing exporter](../../img/otel_gateway_lb_sdk_v2.svg)
```
chalin, line 31: Per the following issue, please keep the kebab case in the filename:

- #780

Yeah, I know that the file names of all of the other files that you've updated the images for use snake case 🤷🏼‍♂️, but we're actually try to aim for kebab case as the standard.

(No need to change any of the other file names to kebab case in this PR, but if you do, that would be grand.)

chalin (approved):
Thanks for renaming the files @hillarymc, and for contributing the SVGs, of course!

LGTM, though I haven't cross-checked the diagrams themselves. I'll let @svrnm and @cartermp validate the diagram contents.

cartermp (approved):
Extremely into these updates. Thanks @hillarymc!

PR #2875, opened 2023-06-15
[demo/cartservice] Avoid warning - hiding variables

Kielek:
Corresponding demo PR https://github.com/open-telemetry/opentelemetry-demo/pull/943


svrnm (approved):
@Kielek let us know when the other PR is merged so we can merge this one (wondering if there is some way to automate that?)

svrnm:
> The demo requires to adjust documentation first (in PR template). I suppose it is sufficient check for .NET service (I do no see any upcoming adjustment in short term).
> 
> I think that we can merge this as is right now. It is pretty independent snippet.

cool then let me do that :-)

PR #2876, opened 2023-06-15
Update traces exporters in kubernetes-deployment.md

fabric73:
Had to add spanmetrics to traces/exporters, otherwise the otel collector pod crashed with this error:

"collector server run finished with error: invalid configuration: connectors::spanmetrics: must be used as both receiver and exporter but is not used as exporter"

### DIFF
--- a/content/en/docs/demo/kubernetes-deployment.md
+++ b/content/en/docs/demo/kubernetes-deployment.md
@@ -167,7 +167,7 @@ opentelemetry-collector:
         traces:
           receivers: [otlp]
           processors: [batch]
-          exporters: [otlphttp/example]
+          exporters: [spanmetrics, otlphttp/example]
 ```
 
 > **Note** When merging YAML values with Helm, objects are merged and arrays are


cartermp:
Thanks @fabric73  - can you sign the CLA? Then we can review.

cartermp (approved):

svrnm:
@open-telemetry/demo-approvers PTAL

julianocosta89 (commented):
In the main file [otelcol-config.yml](https://github.com/open-telemetry/opentelemetry-demo/blob/main/src/otelcollector/otelcol-config.yml#L39-L52) we have the following:

```yaml
service:
  pipelines:
    traces:
      receivers: [otlp]
      processors: [batch]
      exporters: [otlp, logging, spanmetrics]
    metrics:
      receivers: [otlp, spanmetrics]
      processors: [filter, batch]
      exporters: [prometheus, logging]
    logs:
      receivers: [otlp]
      processors: [batch]
      exporters: [logging]
```

What happens is that when using the `otelcol-config-extras.yml` the value at the pipelines are overwritten.
If you change `traces.exporters` or `metrics.receivers` the `spanmetrics` will have to be added to whatever new value is added, otherwise the error shared will be raised.

This PR is partially correct, but I'd prefer to explain the case, instead of just adding the `spanmetrics` value there.

Any suggestions on how to better do that?

Also, it is nice to highlight that if we use the suggested values,  we also lose the traces exporters to `otlp` and `logging`.
That's why I'd prefer to explain how to use 2 configs in a single OTel-Collector.

svrnm:
I think you should be able to fix that by using multiple pipelines, so update [otelcol-config.yml](https://github.com/open-telemetry/opentelemetry-demo/blob/main/src/otelcollector/otelcol-config.yml#L39-L52) to do something like:

```
service:
  pipelines:
    traces/default:
      receivers: [otlp]
      processors: [batch]
      exporters: [otlp, logging, spanmetrics]
    metrics/default:
      receivers: [otlp, spanmetrics]
      processors: [filter, batch]
      exporters: [prometheus, logging]
    logs/default:
      receivers: [otlp]
      processors: [batch]
      exporters: [logging]
```

and then if needed someone can have their own pipeline in the extras file

julianocosta89:
> I think you should be able to fix that by using multiple pipelines, so update [otelcol-config.yml](https://github.com/open-telemetry/opentelemetry-demo/blob/main/src/otelcollector/otelcol-config.yml#L39-L52) to do something like:
> 
> ```
> service:
>   pipelines:
>     traces/default:
>       receivers: [otlp]
>       processors: [batch]
>       exporters: [otlp, logging, spanmetrics]
>     metrics/default:
>       receivers: [otlp, spanmetrics]
>       processors: [filter, batch]
>       exporters: [prometheus, logging]
>     logs/default:
>       receivers: [otlp]
>       processors: [batch]
>       exporters: [logging]
> ```
> 
> and then if needed someone can have their own pipeline in the extras file

This sounds like a great approach!

svrnm:
> > I think you should be able to fix that by using multiple pipelines, so update [otelcol-config.yml](https://github.com/open-telemetry/opentelemetry-demo/blob/main/src/otelcollector/otelcol-config.yml#L39-L52) to do something like:
> > ```
> > service:
> >   pipelines:
> >     traces/default:
> >       receivers: [otlp]
> >       processors: [batch]
> >       exporters: [otlp, logging, spanmetrics]
> >     metrics/default:
> >       receivers: [otlp, spanmetrics]
> >       processors: [filter, batch]
> >       exporters: [prometheus, logging]
> >     logs/default:
> >       receivers: [otlp]
> >       processors: [batch]
> >       exporters: [logging]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and then if needed someone can have their own pipeline in the extras file
> 
> This sounds like a great approach!

@fabric73 , this is much more work than the initial PR but would you be willing to work on that? @julianocosta89 & I can provide you with guidance for things that are unclear

julianocosta89:
Discussed this case at the SIG Meeting and we agreed to update the docs, instead of defining a new pipeline.
I'll take care of that in a new PR.

Issue #2877, opened 2023-06-15
Move documentation to be user journey focused, not OTel concepts focused


markmandel:
Hi all,

To set some context, I work on (amongst other things) [Agones](https://agones.dev), which has used OpenCensus in Go for a number of years, with the primary use case being Metrics (we don't really do trace right now). We have the understanding that we'll need to eventually move to OTel if we want to have a long term supported observability library.

First I want to preface this by recognising all the awesome work you do for OTel, so apolgies up front for being critical, but please know it's with the intent of improvement. 

A while back (April) I looked at how to do the migration from Open Census, and I got _really_ stuck on the onboarding process (especially compared to what I was previously used to with Open Census documentation). I ended up chatting with a few people at Kubecon EU about it, and rightly so - they suggested I provide some feedback here! So here I am.

The TL;DR I found is that I found that Open Census's documentation is focused around _what I want to do_ (i.e. do I want to export metrics? Do I want to log something, do I want to implement tracing?), whereas it seems to me like OTel is focused around  _how does OTel work_ (what are the concepts, what is a collector, what is an exporter?), which for onboarding creates quite a large barrier to entry -- as I need to understand OTel before I can do anything, rather than learning how to do something, and using that as the vehicle to learn the underlying concepts.

I wrote this friction log back in April, so hopefully it's still relevant (looking at the docs, still seems to be the case), but this was the struggle I was finding, and eventually gave up on.

**Use case:** I want to create metrics in Go (language doesn't actually matter, but it's what i know best), and export it to Google Cloud Monitoring (Also doesn't quite matter here either), and I've never used OTel before.

* I go to https://opentelemetry.io/ and click on Docs (https://opentelemetry.io/docs/)
* I read that page, and look for metrics, I see nothing in the navbar, but I do see "metrics" on the docs landing page at the top ("...such as traces, metrics, logs"), I'll click "metrics" then!
* I get a section on "Reliability & Metrics" - cool. I know what metrics are, but I need to do a thing, and this isn't explaining that yet. Okay, I see another link in that section on "metrics", let's see where that goes.
* That link goes to exactly where I am. Why is it a link? Well that's frustrating. Okay, now I'm on my own to work out my own adventure.
* Next stop, I see on the navbar there's a "getting started" section. Awesome! I love Quickstarts, that will get me going on my journey! 
[...]

PR #2879, opened 2023-06-16
Add more details to metrics concept page

svrnm:
Based on the feedback from https://github.com/open-telemetry/opentelemetry.io/issues/2877, I tried to improve some of the metrics concept page:

* I gave it a structure that is closer to the one we have for traces already (i.e. calling out components like MeterProvider, Meter, etc.)
* Added a metrics support table that (a) provides a link to the languages, and (b) calls out metric stability across them, such that an end user is not going to be surprised if for a specific language they can not find metric instructions

If the latter proves valuable for metrics I can create the same for traces (and later for logs)


cartermp (approved):
This is a big improvement over the current...sparse...docs.

PR #2904, opened 2023-06-26
Add collector and language support to sampling concept page

svrnm:
This PR adds a section to the footer of the sampling concept page to point to relevant collector processors and to languages that have documentation on sampling (well-aware that for some languages just the docs are missing). This way readers can read on after going through the concepts, if this proves valuable we can use it as a pattern on other pages as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,6 +109,7 @@
           "Kubernetes",
           "MySQL",
           "NGINX",
+          "OpenTelemetry Collector",
           "OTel",
           "OTEP",
           "PHP",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -109,6 +109,7 @@
           "Kubernetes",
           "MySQL",
           "NGINX",
+          "OpenTelemetry Collector",
```
svrnm, line 112: This is just here to avoid a common mistake to write "OpenTelemetry collector"

cartermp (approved):

Issue #2721, opened 2023-05-15
Adjust to move Semantic Conventions out of OTel spec


jsuereth:
As part of [OTEP #227](https://github.com/open-telemetry/oteps/blob/main/text/0227-separate-semantic-conventions.md) we're moving the semantic conventions out of the Specification and into a new repository.

As part of this:

- We're leaving the Markdown files for semconv in the specification repository with links to the new location warnings.
- Semantic Conventions will be released with their own version number from this repository: https://github.com/open-telemetry/semantic-conventions

We may be re-architecting the structure of this repository, but we can negotiate how best to do so for the website to easily pull it in.

---

### Tasks

- [x] #2785
- [x] #2918
- [x] #2907
- [x] #2982
- [x] #2987
- [x] https://github.com/open-telemetry/semantic-conventions/pull/170
- [x] Ensure every section has a README
  - [x] https://github.com/open-telemetry/semantic-conventions/pull/171
- [x] https://github.com/open-telemetry/semantic-conventions/pull/175
- [x] https://github.com/open-telemetry/semantic-conventions/pull/177
- [x] #2984
- [x] #2998
- [x] #3008
  - [x] #3010
- [x] https://github.com/open-telemetry/semantic-conventions/pull/187
- [x] Ensure that links to OTel spec from semconv pages refer to website pages (`/docs/specs/otel`)
  - [x] Waiting on https://github.com/open-telemetry/semantic-conventions/pull/185
  - [x] #3012
- [x] #3033
- [x] Ensure that links into semconv from other specs refer to website pages:
  - [x] #3037
- [x] Add semconv version to landing page title - #3033
- [x] Add version update to autoupdate script - #3033
- [x] Fix links in semconv pages, and drop workaround in this repo: `|(/resource/faas\.md)#function-as-a-service|$1|;` - https://github.com/open-telemetry/semantic-conventions/pull/180
- [x] Ensure "View page source" and other links are working, esp. for index files (originally README.md files)
- [x] Address:
  - [x] https://github.com/open-telemetry/opamp-spec/issues/150 -- until then, this causes a build warning; the warning is skipped by an entry in `.warnings-skip-list.txt`
  - [x] Drop skip pattern from `.warnings-skip-list.txt` (after #3026 is merged)

Other to-dos from #2908:

- [ ] Add semantic conventions to ToC / index
- [ ] Ensure semantic-conventions -> specification links are accurately internalized to the website
[...]

PR #2918, opened 2023-06-27
Add semantic-conventions as a submodule and get schemas from there, with one exception

### LINKED ISSUE #2907: Schema files should now be served from the Semantic Conventions repository
Right now Hugo / OpenTelemetry.io is serving schema files from the local checkout of the specification.

See: https://github.com/open-telemetry/opentelemetry.io/blob/main/hugo.yaml#L194C14-L194C14

Instead, this should be pulled from this location: https://github.com/open-telemetry/semantic-conventions/tree/main/schemas

chalin:
- Contributes to #2721
  - Brings in https://github.com/open-telemetry/semantic-conventions as a `content-modules`
- Closes #2907
  - Also ensure that we temporarily get the 1.21.0 schema from the OTel spec repo (will track removal of this special case via 2721)

There are no changes in the generated site files:

```console
$ npm run build
...
$ (cd public && git diff) # no change
$
```

### DIFF
--- a/hugo.yaml
+++ b/hugo.yaml
@@ -191,8 +191,17 @@ module:
       target: content/community/roadmap.md
     - source: static
       target: static
+    - source: content-modules/semantic-conventions/schemas
+      target: static/schemas
+      # TODO: drop the lines marked below once
+      # https://github.com/open-telemetry/opentelemetry.io/issues/2721
+      # is fully resolved.
+      # [DROP-FROM-HERE
+      excludeFiles: [1.21.0]
     - source: content-modules/opentelemetry-specification/schemas
       target: static/schemas
+      includeFiles: [1.21.0]
+      # DROP-TO-HERE]
     - source: static/img
       target: static/img
     - source: content-modules/opentelemetry-specification/internal/img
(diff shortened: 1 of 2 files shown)


cartermp (approved):

hugo.yaml
```diff
@@ -191,8 +191,17 @@ module:
...
+    - source: content-modules/semantic-conventions/schemas
+      target: static/schemas
+      # TODO: drop the lines marked below once
+      # https://github.com/open-telemetry/opentelemetry.io/issues/2721
+      # is fully resolved.
+      # [DROP-FROM-HERE
+      excludeFiles: [1.21.0]
     - source: content-modules/opentelemetry-specification/schemas
       target: static/schemas
+      includeFiles: [1.21.0]
```
jsuereth, line 203: We should NOT be serving a `1.21.0` right now, from either location.

chalin: Ok, that's news to me. It has been served since the OTel spec 1.21.0 was released in May. Back then the website served whatever schemas are found under the OTel spec's `schemas` folder and no one from the specs team has informed us otherwise.

If you think that this is a problem, please open an issue over this repo stating that the schema should be removed and cc @tigrannajaryan and the rest of the OTel spec maintainers. Thanks!

chalin: Ah, ok, I just saw: https://github.com/open-telemetry/opentelemetry-specification/issues/3572.

I'll create a corresponding issue here, and a PR to remove the 1.21.0 schema.

chalin: Here's the PR:

- #2924

PR #2908, opened 2023-06-26
Fixes #2907 - Schema files are served from semantic-conventions.

jsuereth:
We'd like to move schema files to be served from the semantic-conventions repository.  This is the first step, next we'd like to add an OpenTelemetry-Bot action that will bump the content module every time semantic conventions cuts a release.

DO NOT MERGE:  This would make the `1.21.0` schema available immediately. 

This PR is for review, to make sure this is the right approach for publishing schema files into the website going forward.

## TODOS

- [X] Add semantic-conventions as subcontent module
- [X] Update auto-upgrade script to apply to semantic-conventions subcontent module
- [ ] Add semantic conventions to ToC / index
- [ ] Ensure semantic-conventions -> specification links are accurately internalized to the website
- [ ] Ensure specification -> semantic-conventions links are accurately internalized to the website.
- [ ] Remove `specification/semantic_conventions/*.md` from being pulled in OpenTelemetry specification.
- [ ] Figure out if supplementary guidelines belong on the website

Issue #137, opened 2020-03-12
Registry search results improvements


codeboten:
Searching the registry for https://opentelemetry.io/registry/?s=python+exporter returns results for all items containing python and all items containing exporter. I would expect it to return only python exporters.

codeboten:
This can now be accomplished by the following query: https://opentelemetry.io/registry/?s=&component=exporter&language=python

PR #2920, opened 2023-06-27
Add SemConv spec under docs/specs (preview)

chalin:
- Contributes to #2721
- Supersedes #2908
- Adds semconv spec to `docs/specs`
- Pages will only appear in deploy previews, not on the production server, since all spec pages are marked as `draft: true`
- We won't want to publish the semconv spec until the following restructuring is complete, IMHO:
  https://github.com/open-telemetry/semantic-conventions/issues/137

**Preview**: https://deploy-preview-2920--opentelemetry.netlify.app/docs/specs/semconv/

/cc @jsuereth @joaopgrassi

### DIFF
--- a/content/en/docs/specs/semconv/_index.md
+++ b/content/en/docs/specs/semconv/_index.md
@@ -1,0 +1,6 @@
+---
+title: OpenTelemetry Semantic Conventions
+linkTitle: SemConv
+cascade:
+  draft: true
+---
--- a/content/en/docs/specs/semconv/_index.md
+++ b/content/en/docs/specs/semconv/_index.md
@@ -1,6 +1,6 @@
 ---
 title: OpenTelemetry Semantic Conventions
-linkTitle: SemConv
+linkTitle: Semantic Conventions
 cascade:
   draft: true
 ---
--- a/hugo.yaml
+++ b/hugo.yaml
@@ -185,6 +185,14 @@ module:
       target: content/docs/specs/otlp/_index.md
     - source: tmp/otlp/docs/img
       target: content/docs/specs/otlp/img
+    - source: tmp/semconv/specification/logs
+      target: content/docs/specs/semconv/logs
+    - source: tmp/semconv/specification/metrics
+      target: content/docs/specs/semconv/metrics
+    - source: tmp/semconv/specification/resource
+      target: content/docs/specs/semconv/resource
+    - source: tmp/semconv/specification/trace
+      target: content/docs/specs/semconv/trace
     - source: tmp/community/mission-vision-values.md
       target: content/community/mission.md
     - source: tmp/community/roadmap.md
--- a/scripts/content-modules/cp-pages.sh
+++ b/scripts/content-modules/cp-pages.sh
@@ -65,3 +65,21 @@ FILES=$(find $DEST -name mission-vision-values.md -o -name roadmap.md)
 $SCRIPT_DIR/adjust-pages.pl $FILES
 
 echo "COMMUNITY pages: copied and processed"
+
+## Semantic Conventions
+
+SRC=content-modules/semantic-conventions/specification
+DEST=$DEST_BASE/semconv/specification
+
+rm -Rf $DEST
+mkdir -p $DEST
+cp -R $SRC/* $DEST/
+
+find $DEST/ -name "README.md" -exec sh -c 'f="{}"; mv -- "$f" "${f%README.md}_index.md"' \;
+
+# To exclude a file use, e.g.: -not -path '*/specification/_index.md'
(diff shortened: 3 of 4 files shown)


cartermp (approved):

chalin:
Thanks for the review @cartermp! Will wait for feedback from @jsuereth and @joaopgrassi.

cartermp:
Yep, figured as much

joaopgrassi (approved):
Looks good to me, given things will change after #137.

One question though: Do you think the structure defined here https://github.com/open-telemetry/semantic-conventions/issues/137#issuecomment-1609014622 is possible to reproduce on the website? Or: Does it even make sense to you?

content/en/docs/specs/semconv/_index.md
```diff
@@ -0,0 +1,6 @@
+---
+title: OpenTelemetry Semantic Conventions
+linkTitle: SemConv
```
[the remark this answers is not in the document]
jsuereth: +1

hugo.yaml
```diff
@@ -185,6 +185,14 @@ module:
       target: content/docs/specs/otlp/_index.md
     - source: tmp/otlp/docs/img
       target: content/docs/specs/otlp/img
+    - source: tmp/semconv/specification/logs
```
jsuereth, line 188: We're going through a refactoring of the semantic conventions markdown right now.

Can we just mount `tmp/semconv/specification`?  There will be a top-level readme file there shortly and I'm happy to give it all the hugo content you need.

We do *not* want to have logs/metrics/trace specific locations going forward.

chalin:
> Do you think the structure defined here [open-telemetry/semantic-conventions#137 (comment)](https://github.com/open-telemetry/semantic-conventions/issues/137#issuecomment-1609014622) is possible to reproduce on the website?

@joaopgrassi - yes, it will actually be easier to bring in. This is just a preview of what it would look like under the current organization.

> We're going through a refactoring of the semantic conventions markdown right now.
> 
> Can we just mount `tmp/semconv/specification`? There will be a top-level readme file there shortly and I'm happy to give it all the hugo content you need.
> 
> We do _not_ want to have logs/metrics/trace specific locations going forward.

@jsuereth - right, this PR is just in preparation, while https://github.com/open-telemetry/semantic-conventions/issues/137 is being addressed.

I'd you agree (and approve this PR @jsuereth), I'd like to merge this PR As I mentioned earlier, the semconv spec will _not_ show up in production. As work progresses on in the semconv repo, I can submit followup PRs.

hugo.yaml
chalin: I've answered further below.

chalin:
@jsuereth @joaopgrassi - btw, regarding the `linkTitle`, which only shows up in the sidenav: the reason I chose `SemConv` as a side-nav title is so we can avoid word-wrap (as you can see below), and because it is more consistent with the other specs. If we use the full name "Semantic Conventions", plus the version, it's a bit too wide IMHO:

> image

It's a detail, not very important, but I wanted to at least explain why so that you can make a more informed choice of what you'd like the `linkTitle` to be.

joaopgrassi:
> btw, regarding the linkTitle, which only shows up in the sidenav: the reason I chose SemConv as a side-nav title is so we can avoid word-wrap (as you can see below), and because it is more consistent with the other specs. If we use the full name "Semantic Conventions", plus the version, it's a bit too wide IMHO:

I see. I think the word wrap isn't great, but I think I prefer that over the "cript" SemConv one. But even in my 1080p monitor there was no wrap, so should be OK? I guess if it comes to that, we can do some css magic :D

joaopgrassi:
Also: We will be listing under "Semantic Conventions" just the "latest version" right? We will not something like

``` 
Semantic Conventions
- 1.22.0
- 1.23.0
...
```

chalin:
@joaopgrassi - correct, only the latest officially released version (at least, that is what we're doing with the OTel spec, and we can do the same here).

chalin:
@jsuereth - so, is this preview PR good to go?

jsuereth (approved):
Under the caveat that we'll change the import going forward for new layout, yeah, let's keep making steady progress.

I do have a few question for follow up PRs:

- How do we get left-hand outline like the spec has for sections?
- Can we change all links to specification in semconv to point at the otel.io copy of the specification?
- Can we change all links to semconv from the specification to point at the otel.io copy of the semconv?



For the last two I was working on the perl script to make this happen.

chalin:
@jsuereth - I've answered in https://github.com/open-telemetry/opentelemetry.io/issues/2721#issuecomment-1611598495.

PR #2921, opened 2023-06-27
Add Node version constraint to package.json & update contrib notes

chalin:
- Updates `CONTRIBUTING.md`: we don't use symlinks anymore; clarifies where content comes from.
- `package.json`:
  - Adds `engines.node` entry to require active LTS version (currently 18.x)
  - Moves dependencies entries to be before `textlint` config etc.


svrnm (approved):
lgtm

PR #2932, opened 2023-06-29
Updates for cspell and textlint

svrnm:
I will provide a set of PRs that will help to clean up all the cspell "missing words", some typos and some language improvements. Those updates to the cspell and textlint rules are covering some common words used across the docs


chalin (approved):
Nice, thanks!

Eventually, could we move the textlint config out of `package.json`? WDYT?

svrnm:
> Nice, thanks!
> 
> Eventually, could we move the textlint config out of `package.json`? WDYT?

Yes, I initially put it there since other configs also lived there, but with it growing bigger we should have it out.

PR #2937, opened 2023-06-29
cSpell fixes for instrumentation pages: cpp, ruby, swift

svrnm:
I batched the changes for Ruby, Swift & C++ because they only have 1 or 2 pages that require updates (cspell ignore lists, one typo)


chalin (approved):
Maybe `darwin` and `wifi` could be made global, but this LGTM as is.

svrnm:
> Maybe `darwin` and `wifi` could be made global, but this LGTM as is.

Agreed, but I'll do this as a follow up eventually. Maybe I write a quick script that extracts all ignore blocks and we can see which terms are used often, so we can decide which ones we want to have global

PR #2940, opened 2023-06-29
cSpell fixes for instrumentation pages: java

svrnm:
cSpell fixes & ignore list updates


mateuszrzeszutek (approved):

chalin (approved):

content/en/docs/instrumentation/java/automatic/agent-config.md
```diff
@@ -4,7 +4,12 @@ linkTitle: Configuration
 weight: 10
 spelling:
   cSpell:ignore akka autoconfiguration Autoconfiguration Dotel HSET javaagent
-  cSpell:ignore serverlessapis Servlet servlet
+  cSpell:ignore serverlessapis Servlet servlet Customizer classloaders logback
```
chalin, line 7: Do we actually want both Servlet and servlet in this list? Only "servlet" should be enough, right?

svrnm:
```suggestion
  cSpell:ignore serverlessapis servlet Customizer classloaders logback
```

content/en/docs/instrumentation/java/automatic/agent-config.md
```diff
@@ -4,7 +4,12 @@ linkTitle: Configuration
 weight: 10
 spelling:
   cSpell:ignore akka autoconfiguration Autoconfiguration Dotel HSET javaagent
-  cSpell:ignore serverlessapis Servlet servlet
+  cSpell:ignore serverlessapis servlet Customizer classloaders logback
```
breedx-splk, line 7: I notice start-cap on `Customizer` but I suppose case is probably not important.

svrnm: it's like "if upper case, the case counts, if lowercase it's both upper & lower case", so yes we are good here :)

PR #2941, opened 2023-06-29
cSpell fixes for instrumentation pages: go

svrnm:
cSpell fixes & ignore list updates


content/en/docs/instrumentation/go/resources.md
```diff
@@ -44,7 +45,7 @@ resources := resource.New(context.Background(),
     resource.WithOS(), // This option configures a set of Detectors that discover OS information
     resource.WithContainer(), // This option configures a set of Detectors that discover container information
     resource.WithHost(), // This option configures a set of Detectors that discover host information
-    resource.WithDetectors(thirdparty.Detector{}), // Bring your own external Detector implementation
+    resource.WithDetectors(thirdParty.Detector{}), // Bring your own external Detector implementation
```
pellared, line 48:
```suggestion
    resource.WithDetectors(thirdparty.Detector{}), // Bring your own external Detector implementation
```

`thirdparty` is a name of the Go package name. The idiomatic way is to have them lowercase. See https://go.dev/blog/package-names:

> Good package names are short and clear. They are lower case, with no under_scores or mixedCaps.

svrnm: My assumption was that this is a place holder for any kind of third party package, isn't it?

pellared: True. But even a placeholder should use idiomatic naming to look good 😉

content/en/docs/instrumentation/go/getting-started.md
```diff
@@ -1,6 +1,8 @@
 ---
 title: Getting Started
 weight: 10
+spelling: cSpell:ignore fibonacci struct fscanf chan println strconv stdouttrace
```
chalin, line 4: There is at least one place in this file where Fibonacci is misspelled -- it should be capitalized, e.g.:

> // Fibonacci returns the n-th fibonacci number.

should be

> // Fibonacci returns the n-th Fibonacci number.

chalin: In that sense, probably `fibonacci` (all in lowercase) should be removed from the `cSpell:ignore` list.

chalin (changes requested):
At least one more fix

content/en/docs/instrumentation/go/getting-started.md
svrnm: fixed

chalin (approved):
✨

PR #2957, opened 2023-06-30
Update cspell and textlint config

svrnm:
Move textlint config out of package.json

Add more words to cspell list, a lot of names of authors that have published multiple times on the blog.

@chalin PTAL


chalin (approved):
LGTM

As for blog author names, as a general rule, we should have an ignore rule in the blog post itself. Otherwise over time we'll end up adding every single blog author name to the global list. My 2cents, though I don't feel strongly about it.

As for author names, IMHO, we should leave the ignore rule in the blog post.

svrnm:
> LGTM
> 
> As for blog author names, as a general rule, we should have an ignore rule in the blog post itself. Otherwise over time we'll end up adding every single blog author name to the global list. My 2cents, though I don't feel strongly about it.
> 
> As for author names, IMHO, we should leave the ignore rule in the blog post.

I don't want to have all authors in that list, I only picked names of people that show up multiple times because the have created multiple blog posts already, or are mentioned within the text e.g. as candidates or elected members of the GC or speakers at a conference. But if the overall tendency is that we should not do that I can revert the changes.

chalin:
@svrnm - your call.

svrnm:
as discussed yesterday during the SIG call with @chalin, I will merge this in: we can go backwards if we want anyways.

I will also write down some updates to the contributors guidelines that updates to that cspell file should not be made for individuals that write a blog post once but for names that came up (and will come up) again and again

PR #2958, opened 2023-06-30
cSpell fixes for blog posts

svrnm:
More cspell ignore lists, typo fixes, etc.


svrnm:
depends on #2957

content/en/blog/2022/collector-builder-sample/index.md
```diff
@@ -4,6 +4,9 @@ linkTitle: Collector builder with GCP
 date: 2022-10-17
 author: >-
   [Mike Dame](https://github.com/damemi) (Google)
+spelling: cSpell:ignore customizability otlpreceiver loggingexporter rollouts
+spelling: cSpell:ignore configmap otlpexporter batchprocessor zipkinexporter 
+spelling: cSpell:ignore gomod
```
chalin, line 9: Oh, I hadn't noticed this before! There should be at most one `spelling` front-matter field. Instead we can do something like the following:

```suggestion
spelling: |
  cSpell:ignore customizability otlpreceiver loggingexporter rollouts
  cSpell:ignore configmap otlpexporter batchprocessor zipkinexporter 
  cSpell:ignore gomod
```

In fact, I thought that I had found a way to get Prettier to leave the local dictionary words all on one line (which is what cSpell requires.

chalin (approved):

content/en/blog/2022/collector-builder-sample/index.md
chalin: If this builds as is, I'm ok if you leave it as repeated front-matter keys. We can consider cleaning that up later.

svrnm: prettier just stumbled upon this and created some really strange version of that, but let me double check that

chalin: This is what I did before: include `# prettier-ignore` just before the (single) `spelling:` entry, like this: https://github.com/open-telemetry/opentelemetry.io/blob/d61b9fde23c8e84ee5b1b5aca45aa2bbe9780e59/content/en/ecosystem/vendors.md?plain=1#L1-L8

svrnm: @chalin, I'll merge it in as is, since we have *many* documents right now that have the multiline spelling cSpell:ignore pattern I am using here, we can have a follow up discussion which format we prefer to have

PR #2975, opened 2023-07-05
add collector benchmark page

codeboten:
This is based on the work that @mwear started in the collector repository. Rather than publishing a github page from the collector repo, I suggest this should live in the opentelemetry.io repository and be displayed on the website as well.

For reviewers, is including style/html in the markdown ok? Or should this be moved somewhere else?

---

**Preview**: https://deploy-preview-2975--opentelemetry.netlify.app/docs/collector/benchmarks/

### DIFF
--- a/content/en/docs/collector/benchmarks.md
+++ b/content/en/docs/collector/benchmarks.md
@@ -2,14 +2,8 @@
 title: Benchmarks
 weight: 99
 ---
-  <style>
 
-    header {
-      width: 100%;
-      display: flex;
-      flex-direction: row;
-      background-color: rgb(79, 98, 173);
-    }
+  <style>
 
     main {
       margin: 8px;
@@ -45,10 +39,6 @@ weight: 99
       align-items: center;
     }
 
-    .header-label {
-      margin-right: 4px;
-    }
-
     .benchmark-set {
       margin: 8px 0;
       width: 100%;
@@ -76,15 +66,6 @@ weight: 99
       max-width: 1000px;
     }
 
-    .logo svg {
-      height: 48px;
-      margin-left: 30px;
-    }
-
-    .header-item {
-      flex: 1;
-    }
-
     div.container {
       max-width: 1012px;
       margin-right: auto;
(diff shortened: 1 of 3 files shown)


content/en/docs/collector/benchmarks.md
```diff
@@ -0,0 +1,306 @@
...
+---
+  <style>
+
+    header {
+      width: 100%;
+      display: flex;
+      flex-direction: row;
+      background-color: rgb(79, 98, 173);
+    }
+
```
mwear, line 13: We removed the header from the html, we can probably delete this and rely on the header css for opentelemetry.io

```suggestion
```

content/en/docs/collector/benchmarks.md
```diff
@@ -0,0 +1,306 @@
...
+    .logo svg {
+      height: 48px;
+      margin-left: 30px;
+    }
+
+    .header-item {
+      flex: 1;
+    }
+
```
mwear, line 87: These too

```suggestion
```

mwear (commented):
I think we can delete some CSS, but generally this looks good to me.

austinlparker:
cc @chalin -- thoughts on this? some wonkiness with breakpoints due to max-width looks like.

austinlparker:
Couple of other notes -

- Would be great to have an index/TOC here, it's a _lot_ of charts.
- Would be great to have explanations/descriptions of the benchmarks
- Is there a way we can consolidate this to top-k (10?) critical benchmarks for web and put the rest in the repo?

chalin:
@austinlparker - once your concerns are addressed (which seem reasonable to me -- e.g. top-x), I can certainly look at integrating the page. Actually, it could always land with embedded CSS etc. and be refactored after.

jpkrohling:
I came here for the same comments @austinlparker has already made :-) An intro section setting the context (is it about the contrib image? Core? both?) and a short description of the scenarios would be required.

codeboten:
I've added a paragraph at the top, PTAL

mwear (approved):

cartermp (approved):

Issue #2491, opened 2023-03-13
Ensure that `fuse.js`-generated registry entries also render markdown in descriptions


chalin:
- Markdown is being rendered in descriptions in the initial full list of entries
- The registry.json.json layout needs to be updated to markdownify descriptions.

PR #2977, opened 2023-07-06
Apply textlint rules to registry

svrnm:
This PR cleans up all textlint errors in `data/` by fixing registry entries and adding some more rules to the textlintrc

### DIFF
--- a/.textlintrc.json
+++ b/.textlintrc.json
@@ -17,8 +17,7 @@
         "github.com/open-telemetry",
         "#opentelemetry",
         "/^(?:language|repo): .*$/m",
-        "/tags:([\\n\\r\\s]*-[\\s\\S]*?[\\n\\r]+)*/m",
-        "Instrumentation for net/http"
+        "/tags:([\\n\\r\\s]*-[\\s\\S]*?[\\n\\r]+)*/m"
       ]
     }
   },
--- a/.textlintrc.json
+++ b/.textlintrc.json
@@ -17,8 +17,7 @@
         "github.com/open-telemetry",
         "#opentelemetry",
         "/  - .*/",
-        "/(?:language|repo): .*/",
-        "net/http"
+        "^/(?:language|repo): .*/",
       ]
     }
   },
--- a/.textlintrc.json
+++ b/.textlintrc.json
@@ -15,7 +15,10 @@
         "/<http://.*>/",
         "medium.com/opentelemetry",
         "github.com/open-telemetry",
-        "#opentelemetry"
+        "#opentelemetry",
+        "/  - .*/",
+        "/(?:language|repo): .*/",
+        "net/http"
       ]
     }
   },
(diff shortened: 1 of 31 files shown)


.textlintrc.json
```diff
@@ -15,7 +15,10 @@
         "/<http://.*>/",
         "medium.com/opentelemetry",
         "github.com/open-telemetry",
-        "#opentelemetry"
+        "#opentelemetry",
+        "/  - .*/",
+        "/(?:language|repo): .*/",
+        "net/http"
```
chalin, line 21: Is this entry still necessary given that you've put occurrences of "net/http" inside code fences, like this:

    `net/http`

chalin (commented):
Again, very nice. Thanks for doing this.

For my questions and suggestions, see inline comments.

.textlintrc.json
svrnm: There is at least one instance were net/http is in the title of a registry entry, where we do not apply `markdownify`, that's why I put it here. I need to raise an issue with the registry because of that, but we can keep it out for now

.textlintrc.json
```diff
@@ -15,7 +15,10 @@
         "/<http://.*>/",
         "medium.com/opentelemetry",
         "github.com/open-telemetry",
-        "#opentelemetry"
+        "#opentelemetry",
+        "/  - .*/",
```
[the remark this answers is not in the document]
svrnm: The idea was to rule out tags in the registry, but you are right this will probably overshoot...

data/registry/tools-python-sentry.yml
```diff
@@ -14,7 +14,6 @@ license: MIT
 description:
   The Sentry OpenTelemetry Python integration provides a processor and
   propagator to send data to Sentry and to associate traces/spans to Sentry
-  errors. For configuration details, see [OpenTelemetry
-  Support](https://docs.sentry.io/platforms/python/performance/instrumentation/opentelemetry/).
```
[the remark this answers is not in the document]
svrnm: because the markdownify is broken if I search, will raise an issue on that.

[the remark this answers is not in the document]
chalin: It's a known issue (#2491), but has been marked low priority. Feel free to bump up the priority. I'd vote to keep the link in the description (it at least gives us a test case for #2491 :)).

.textlintrc.json
chalin: Right, I'd still apply code formatting, and drop this entry. But, in the end, it's your call.

svrnm:
```suggestion
```

data/registry/tools-python-sentry.yml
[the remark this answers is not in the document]
svrnm: reverted

.textlintrc.json
```diff
@@ -15,7 +15,10 @@
         "/<http://.*>/",
         "medium.com/opentelemetry",
         "github.com/open-telemetry",
-        "#opentelemetry"
+        "#opentelemetry",
+        "/^(?:language|repo): .*$/m",
+        "/tags:([\\n\\r\\s]*-[\\s\\S]*?[\\n\\r]+)*/m",
```
chalin, line 20: Nice!

.textlintrc.json
```diff
@@ -15,7 +15,10 @@
...
+        "Instrumentation for net/http"
```
chalin, line 21: I'd vote to add code ticks to:

> title: splunkhttp -- Instrumentation for net/http

so that it reads

> title: splunkhttp -- Instrumentation for `net/http`

instead of adding this one-off rule.

I can address title markdown at the same time as I tackle the markdown in descriptions. WDYT?

svrnm: done

chalin (approved):
Splendid! ✨

PR #2978, opened 2023-07-06
Enable website telemetry in production only

chalin:


chalin:
> What if I want to improve that analytics? Is there a way to turn it on if required? Is there a flag I can set my dev to "prod"?

Yes, if you'd like to turn it on when you serve locally use:

```console
$ npm run _serve:hugo -- -e production
```

svrnm (approved):

svrnm:
cool, then we should be good here :-)

PR #2980, opened 2023-07-06
Update pr-actions.yml: fix fix:format

svrnm:


chalin (approved):
Ah! Thanks for the fix ✨

PR #2983, opened 2023-07-07
php updates

brettmc:
* use guzzle http client to satisfy simplified http client/message requirements
* update pickle installation instructions to work now that we've moved the extension source
into a subdirectory
* disable metrics by default, since it often trips up users trying out tracing


pdelewski (approved):

svrnm:
>     * disable metrics by default, since it often trips up users trying out tracing

maybe put in a short note on this and that people can turn them on (set the value to otlp) as needed

brettmc:
> maybe put in a short note on this and that people can turn them on (set the value to otlp) as needed

On a second reading, I decided that the disabling metrics thing is adequately covered in https://opentelemetry.io/docs/instrumentation/php/sdk/#autoloading so I've removed it from this PR.

svrnm (approved):

PR #2982, opened 2023-07-06
Update semantic-conventions

chalin:
- Contributes to #2721
- Updates semantic-conventions under `/docs/specs/semconv` using the new organization brought in by https://github.com/open-telemetry/semantic-conventions/pull/166

> ~**IMPORTANT NOTE**: if this PR is merged, then the semantic-conventions will be **published to the production server**. Are y'all ok with that @jsuereth @joaopgrassi @open-telemetry/specs-semconv-approvers?~

**Preview**: https://deploy-preview-2982--opentelemetry.netlify.app/docs/specs/semconv/

PR #2984, opened 2023-07-07
Semconv: adjust doc-page titles

chalin:
- Followup to #2982
- Adjusts the titles of semconv pages to match https://github.com/open-telemetry/semantic-conventions/pull/175
- Note: semconv pages are still in preview-only (draft) mode

/cc @jsuereth @joaopgrassi


chalin:
@jsuereth @joaopgrassi - since the semconv pages are still only visible in Netlify previews, I'm going to merge this now because I have followup changes to make to the scripts. Feedback is still welcome -- I'll address it, if any, in a subsequent PR.

PR #2989, opened 2023-07-07
Introduce version variables for erlang docs

svrnm:
Similar to what we have for Java & Collector already this pushes the versions used in erlang docs to the page front matter. This is only an intermediate step, in a further update we should have those taken from `data/`


chalin (changes requested):
Nice, but let's keep this DRY.. The version params should exist in a single file: `content/en/docs/instrumentation/erlang/_index.md`. All files in the section will have access to the params via the `param` shortcode.

svrnm:
> Nice, but let's keep this DRY.. The version params should exist in a single file: `content/en/docs/instrumentation/erlang/_index.md`. All files in the section will have access to the params via the `param` shortcode.

cool, I was wondering the same, I guess java is doing that and I somehow missed it

svrnm:
@chalin fixed

ferd (approved):

cartermp (approved):

cartermp:
Alright I did a fuckywucky with the refcache somehow in another PR, so the red in this PR is a false negative. I'm just going to merge it and then fix my mess up

PR #3007, opened 2023-07-10
CI: cache dependencies for check-format GH action

chalin:
I'm not 100% sure that this is properly setup, but it does reduced the action execution time by more than 50%.

The action does contain the following warning, so I'm not sure that `npm install` is actually being called or not:

> npm WARN exec The following package was not found and will be installed: prettier@3.0.0

/cc @trask


chalin:
@trask - does the change in this PR seem reasonable to you?

chalin:
Works for me. Merging.

PR #3008, opened 2023-07-10
Ensure that semconv pages are preview only, not in production + add watermark

chalin:
@jsuereth @joaopgrassi - my apologies, but #2984 inadvertently caused the semconv pages to be published to the production server (confusion arose as I switched to using the name-title adjustment script and forgot that I had dropped the `draft: true` front matter field).

This PR:

- Adds a DRAFT ~(was: PREVIEW-ONLY)~ wartermark for the semconv draft pages
- Drops the semconv pages from the production server

Given that the semconv pages are pretty much in their final form (given that the reorg is done), maybe we can keep the pages in production, but with the watermark. Or even just forget about the watermark too? Your call, let me know.

**Preview**: https://deploy-preview-3008--opentelemetry.netlify.app/docs/specs/semconv/

**Screenshot**:

> image

### DIFF
--- a/assets/scss/_styles_project.scss
+++ b/assets/scss/_styles_project.scss
@@ -263,7 +263,7 @@ body.td-page--draft .td-content {
   position: relative;
 
   &::after {
-    content: 'PREVIEW-ONLY';
+    content: 'Draft';
     position: absolute;
     top: 80px;
     left: 50%;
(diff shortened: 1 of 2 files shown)


chalin:
Btw, I can also change the watermark text to be DRAFT.

svrnm (approved):

chalin:
I've switched to the DRAFT watermark.

chalin:
@jsuereth - Going to merge this now before the semconv pages get too much traffic -- the pages weren't meant to be published (yet) in the first place. Apologies for the churn!

Issue #2814, opened 2023-06-02
CI: consider GH action to surface build warnings and other secondary checks


chalin:
Context: https://github.com/open-telemetry/opentelemetry.io/pull/2812#issuecomment-1573800658

- [x] Surface these checks as a GH action - #3026
  - Apparently we can no longer have an action status be "neutral"

Extra:

- Add warning for links with the text "here"

chalin:
I'm going to consider this closed by #3026. Adding other checks, as suggested in the opening comment, can be done in the scope of another PR or issue.

PR #3026, opened 2023-07-12
CI: check-links: enable caching and check for warnings

chalin:
- Contributes to #2814
- Enable simple node-based caching
- Separates concerns. We now have separate jobs for:
  - Building the site and running the link checker
  - Checking for required refcache updates
  - Checking for build issues (including warnings about use of external links when a local path should be used)
- I've disabled the failure of the BUILD WARNINGS job until #2721 is fully resolved

https://github.com/open-telemetry/opentelemetry.io/actions/workflows/check-links.yml used to take almost 3 minutes, it is now down to ~1 min 40 sec~. Update: it's now 2 min 15 sec on a rerun; I couldn't get it to work using a partial install. Not as much an improvement as I would have hoped, but it is an improvement. I could probably trim `devDependencies`, but this will do for now.

### DIFF
--- a/.github/workflows/check-links.yml
+++ b/.github/workflows/check-links.yml
@@ -23,7 +23,7 @@ jobs:
 
       - name: Minimal install and prep
         run: |
-          npm install npm-run-all postcss-cli
+          npm install npm-run-all postcss-cli postcss autoprefixer
 
       - run: npm run build | tee tmp/build-log.txt
       - run: npm run _check:links
--- a/.github/workflows/check-links.yml
+++ b/.github/workflows/check-links.yml
@@ -10,13 +10,19 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - name:
-          Create NPM cache-hash input file and reduced-dependencies package.json
+      - name: Create NPM cache-hash input file
         run: |
           mkdir -p tmp
-          jq '{devDependencies, dependencies, engines, gitHubActionCacheCounter}' package.json > tmp/package-ci.json
-          jq 'del(.devDependencies[] | select(test("spell|gulp|netlify|prettier|require|text")))' package.json > tmp/package-min.json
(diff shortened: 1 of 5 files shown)


svrnm (approved):

svrnm:
> [open-telemetry/opentelemetry.io/actions/workflows/check-links.yml](https://github.com/open-telemetry/opentelemetry.io/actions/workflows/check-links.yml) used to take almost 3 minutes, it is now down to ~1 min 40 sec~. Update: it's now 2 min 15 sec on a rerun; I couldn't get it to work using a partial install. Not as much an improvement as I would have hoped, but it is an improvement. I could probably trim `devDependencies`, but this will do for now.

every minute counts :-) it looks like checking the links itself is taking ~50s, another 50s go in `npm install` (`added 2236 packages, and audited 2237 packages in 53s`)

svrnm:
@chalin just FYI, I ran [this tool](https://www.npmjs.com/package/howfat) today to see which packages have the most dependencies. The biggest contributor are netlify-cli (1336), textlint & deps (696) gulp (482), cspell (183), so if you want to tune the job in a way that it only installs the packages needed, you might get some time back.

chalin:
I tried that already, see https://github.com/open-telemetry/opentelemetry.io/pull/3026/commits/6c02f16bfac48c6f38d01875950160b5b31483fe, but it didn't save us any time as far as I could tell. If you think that it might still be helpful to try, I can roll back to that commit. WDYT?

chalin:
Ok, I'm going to try the reduced dependencies list again.

chalin:
Using a reduced-dependencies build worked! The problem I faced before was that my `jq` query wasn't working. The slowest action is now slightly below 2 min.

chalin:
Seeded refcache change and the action successfully detected it.

PR #3029, opened 2023-07-13
Fix typo in agent-config.md

dgcom:
Correct environment variable name when converting from configuration property.


cartermp (approved):
Thanks, good catch!

PR #3033, opened 2023-07-14
Update semantic-conventions to 1.21.0, publish all schemas, drop draft status, enable auto-update

chalin:
- Contributes to #2721
- Updates semconv submodule to 1.21.0
- Enabled the publication of schema for 1.21.0
- Removes `draft` status from semconv pages -- **SEMCONV PAGES WILL APPEAR ON THE PRODUCTION SERVER**
- Enabled auto-update of the semconv version
- Note that the OTel spec is at 1.23.0, so links into the OTel spec at 1.22.0 appear as external links; they do not link to the corresponding OTel website spec page

**Preview**: https://deploy-preview-3033--opentelemetry.netlify.app/docs/specs/semconv

/cc @jsuereth @joaopgrassi


joaopgrassi:
Does https://github.com/open-telemetry/semantic-conventions/issues/194#event-9817060867 impact in anything after we release this version to production? I guess the URLs right, but that can be fixed with the aliases?

joaopgrassi (approved):

svrnm (approved):

chalin:
@joaopgrassi - we can do it in either order. Fixing the names first is a bit easier in that we don't need to add aliases.

If you and @jsuereth prefer, I can leave the senconv pages in draft status. Let me know.

jsuereth (approved):

jsuereth:
I think publishing as-is is fine.  We can continue to refine/update pages as we go.

PR #3034, opened 2023-07-14
Update demo architecture

julianocosta89:
Loadgenerator now sends requests through FrontEndProxy. 
And OTelCollector is using OTLPExporter instead of JaegerExporter (which was deprecated) for traces.


mviitane (approved):
LGTM. Good to keep these updated. You could also fix one typo in row 134:
"Open Telemetry" -> "OpenTelemetry"

julianocosta89:
Thx @mviitane, done!

chalin (approved):
Thanks!

PR #3045, opened 2023-07-15
Simplify Java onboarding instructions

punya:
Use a Gradle plugin to manage the Java agent and the environment variables. This cuts down on manual steps.

### DIFF
--- a/.vscode/cspell.json
+++ b/.vscode/cspell.json
@@ -95,6 +95,7 @@
     "Rexed",
     "roadmap",
     "runtimes",
+    "ryandens",
     "Rynn",
     "semconv",
     "Severin",
--- a/content/en/docs/instrumentation/java/getting-started.md
+++ b/content/en/docs/instrumentation/java/getting-started.md
@@ -40,6 +40,8 @@ content:
 ```kotlin
 plugins {
   id("java")
+  id("application")
+  id("com.ryandens.javaagent-application") version "0.4.2"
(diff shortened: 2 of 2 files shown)


mateuszrzeszutek (changes requested):
Hey @punya ,

While I understand that using a gradle plugin might simplify things for you application, this is supposed to be just an example of how to use the javaagent. We don't really want to have extra gradle logic (some of our users don't use gradle at all), and we want the whole thing to be as simple (!= short) as possible. I believe that using standard tools like environment variables and `java -jar` is easier to get for first time users (especially if they use maven for builds).

svrnm:
Thanks for raising this PR @punya and thank you for your comment @mateuszrzeszutek: I agree that while the gradle plugin (further) simplifies the getting started, it makes things complicated for people not using gradle. I think it is important for people to know about "-javaagent" and how to use it. 

Nevertheless, I think the gradle plugin is good to know about. Maybe we can add a note somewhere that this is possible? Maybe in https://opentelemetry.io/docs/instrumentation/java/automatic/?

punya:
@mateuszrzeszutek and @svrnm, thanks for your feedback. I can see how the use of a "magic" Gradle plugin would make this sample less approachable, and harder to adapt to Maven. If I have time, I'll start a separate PR to mention the agent in https://opentelemetry.io/docs/instrumentation/java/automatic/.

PR #3047, opened 2023-07-17
Remove OTLP label from arrows

### LINKED ISSUE #2890: Overview: define OTLP on first use?
Refering to diagram on main page https://opentelemetry.io/docs/

Short form OTel is explained in text and used in diagram, BUT
OTLP (guessing OpenTelemetryProtocol) is only used in diagram but not introduced for the reader.

kDot:
Removes OTLP labels from arrows.

Fixes #2890


svrnm (commented):
this would work for me.

Another reason for removing those OTLP arrows is that not all of this data i necessarily sent via OTLP

cartermp (approved):
Thanks! And agree, it's not always OTLP that goes in.

PR #3061, opened 2023-07-22
Add /fix:refcache action

cartermp:
Maybe, just maybe, this will let us just update the refcache in PR comments. Let's find out with https://www.gooddeals.com/

### DIFF
--- a/.github/workflows/pr-actions.yml
+++ b/.github/workflows/pr-actions.yml
@@ -70,3 +70,59 @@ jobs:
           fi
         env:
           GH_TOKEN: ${{ secrets.OPENTELEMETRYBOT_GITHUB_TOKEN }}
+
+  fix-refcache:
+    name: /fix:refcache
+    runs-on: ubuntu-latest
+
+    if: |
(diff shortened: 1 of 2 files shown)


cartermp:
God I hate computers sometimes

chalin (approved):
LGTM

I'm not sure that we can test an action like this in the PR that adds it.
Maybe just remove the seeded error and commit. We can test after that. WDYT?

JamieDanielson:
@cartermp looks like this worked 🎉  (sorry if this is old news but I thought it was kinda cool anyway) Check out over here https://github.com/open-telemetry/opentelemetry.io/pull/3361#issuecomment-1755916347

svrnm:
> @cartermp looks like this worked 🎉  (sorry if this is old news but I thought it was kinda cool anyway) Check out over here https://github.com/open-telemetry/opentelemetry.io/pull/3361#issuecomment-1755916347

I haven't used that action for a while assuming it's still broken so 🎉

PR #3062, opened 2023-07-22
update refcache again

cartermp:
Why? Well, a PR I merged was 100% green. However, it wasn't -- github actions was just taking its sweet time deciding to even queue up a job. And so the refcache job hadn't even been run yet.


cartermp:
using my all-powerful admin privileges to un-red the repo since I redded it over the weekend while one of the other maintainers is on PTO

PR #3064, opened 2023-07-24
Add hints for textlint workflow

svrnm:
This PR introduces annotations for textlint workflow runs, I added mistakes to two files to visualize the change (those need to be fixed before merging)

Screenshot 2023-07-24 at 17 43 40


cartermp (approved):
dank

chalin (approved):
Ooooh, very convenient! 

Approved, once seeded errors are removed :)

Also, one caveat: it requires any contributor who wishes to run this check locally to have `jq` locally installed.

(Maybe a Gulp task might be better? That could be addressed later.)

svrnm:
> Also, one caveat: it requires any contributor who wishes to run this check locally to have `jq` locally installed.
> 
> (Maybe a Gulp task might be better? That could be addressed later.)

Outside of improving that script I think contributors should not need to run this locally (there is still the npm check:text), but if a gulp task helps to standardize things I am good with it later.

PR #3072, opened 2023-07-25
[CI] fix:refcache - ensure all packages are installed

chalin:
This is a quick fix; we can optimize later by using "reduced-dependencies package.json" like in check-links.yml.

### DIFF
--- a/.github/workflows/pr-actions.yml
+++ b/.github/workflows/pr-actions.yml
@@ -108,8 +108,9 @@ jobs:
           cache: npm
           cache-dependency-path: tmp/package-ci.json
 
+      - run: npm install
       - run: |
-          npm run test
+          npm run check:links
           git status
           git branch -v
 


.github/workflows/pr-actions.yml
```diff
@@ -108,8 +108,9 @@ jobs:
           cache: npm
           cache-dependency-path: tmp/package-ci.json
 
+      - run: npm install
       - run: |
-          npm run test
+          npm run check:links
```
cartermp, line 113: hmmm, doesn't this then _not_ update the refcache?

chalin: This change means that we're only checking links, not doing other checks (which, I suppose we could still do). Unless we're only checking for internal links, the link checker will rewrite the refcache once it's done checking. Notice the `refcache.json` manipulation at the end of the `check:links` run:

```console
$ npm run check:links
...
> build
> npm run _build
...
tmp/bin/htmltest --log-level 1
htmltest started at 12:04:22 on public
========================================================================
✔✔✔ passed in 32.994806527s
tested 489 documents

cp tmp/.htmltest/refcache.json static/
npm run _prettier:any -- --write static/refcache.json
```

cartermp: Mmm, gotcha, excellent!

cartermp (approved):

PR #3075, opened 2023-07-25
update go metric status

alvarowolfx:
Accordingly to release v1.16.0 of Open Telemetry Go, `metrics` is now stable. Updating the docs to reflect that.

Reference
* https://github.com/open-telemetry/opentelemetry-go/releases/tag/v1.16.0
  > This release contains the first stable release of the OpenTelemetry Go [metric AP](https://pkg.go.dev/go.opentelemetry.io/otel/metric) 

---

**Preview**: https://deploy-preview-3075--opentelemetry.netlify.app/docs/instrumentation/go/#status-and-releases

### DIFF
--- a/data/instrumentation.yaml
+++ b/data/instrumentation.yaml
@@ -25,7 +25,7 @@ languages:
     urlName: go
     status:
       traces: stable
-      metrics: beta
+      metrics: stable
       logs: not yet implemented
   java:
     name: Java
--- a/data/instrumentation.yaml
+++ b/data/instrumentation.yaml
@@ -25,7 +25,7 @@ languages:
     urlName: go
     status:
       traces: stable
-      metrics: stable
+      metrics: mixed (beta SDK & stable API)
       logs: not yet implemented
   java:
     name: Java


chalin (approved):
LGTM

(@ow, 1.16.0 was released in May!)

MrAlias (changes requested):
The metric API has a stable release, but the SDK does not. That needs to be reflected in the status or it should not be changed until the SDK is also released as stable.

data/instrumentation.yaml
```diff
@@ -25,7 +25,7 @@ languages:
     urlName: go
     status:
       traces: stable
-      metrics: beta
+      metrics: stable
```
cartermp, line 28:
```suggestion
      metrics: mixed (beta SDK & stable API)
```

cartermp (changes requested):
Recommend changing as per @MrAlias's suggestion. I agree that we can't claim the status to be stable until the SDK is stable (since most developers install the SDK, not just the API).

data/instrumentation.yaml
chalin: Hmm, I'm not sure how that will render. Let's commit the suggestion and see.

alvarowolfx: Screenshot 2023-07-25 at 16 55 29

I ran it locally here and it looks like this. Not sure if is going to be clear enough.

chalin:
Ok, this is the end result:

image

The shortcode muddles the capitalization. (I can fix that separately.)

Do we have a final vote on this @MrAlias and @cartermp? Leave it as is, or go with the above?

cartermp (approved):

MrAlias (approved):

cartermp:
Thanks @alvarowolfx!

chalin:
> The shortcode muddles the capitalization. (I can fix that separately.)

I'll address this via:

- #3077

PR #3076, opened 2023-07-25
Fix broken link to CNCF logos

mtwo:
Old link generates a 404

### DIFF
--- a/hugo.yaml
+++ b/hugo.yaml
@@ -97,7 +97,7 @@ params:
         icon: fab fa-stack-overflow
         desc: Practical questions and curated answers
       - name: CNCF branding
-        url: https://cncf-branding.netlify.app/projects/opentelemetry/
+        url: https://github.com/cncf/artwork/tree/master/projects/opentelemetry
         icon: fas fa-image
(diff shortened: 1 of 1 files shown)


chalin (changes requested):
Thanks for reporting the 404 @mtwo.

There were a few things wrong with the original link entry. Firstly, the link wasn't to "CNCF branding" (if it were, the proper link would be https://www.cncf.io/brand-guidelines); the link was to OTel logos. In that case, the updated URL you provided is correct, but the `name` field should be changed.

The `links.user` array appears under the [Community](https://opentelemetry.io/community/) page's **Learn and Connect** section. I don't think that it makes sense to have logos in that list.

I'd suggest that we move the logos reference into the [Marketing Guidelines](https://opentelemetry.io/community/marketing-guidelines/).

Thoughts @mtwo @svrnm @cartermp @austinlparker ?

chalin (approved):
Ok, let's merge this since it fixes a 404.

I'll carry the discussion to a separate issue, and submit a proposal when I have the time.

Thanks again for the fix @mtwo!

chalin:
Will track via:

- #3081

PR #3096, opened 2023-07-28
Update app.js code in NGINX BLOG

880831ian:
- Solve the problem that app.js under the front-end data folder cannot be executed normally
- Adjust docker-compose version


chalin (commented):
Contrary to regular doc pages, we won't generally update old blog posts like this. Blog posts are dated, and it's understood that they'll age and may no longer be accurate or useful.

That being said, I'll let the other maintainers chime in.

cartermp:
I think this is fine to bring in. We don't have nginx docs and I don't see us writing those any time soon, so I'd rather take this in than not.

cartermp (approved):

880831ian:
I was also looking for NGINX articles and just came across this blog, it helped me understand opentelemetry very useful, thanks @cartermp @chalin

PR #3097, opened 2023-07-28
[CI] Adjust spell-check action's path to omit registry entries

chalin:
- Adjusts `streetsidesoftware/cspell-action` file path so registry data entries are ignored (at least I hope that it'll work now :)
- Adjusts markdownlint option: `link-fragments`: false -- the linter doesn't understand anchor redefinition markdown syntax, so we'll let the link checker handle fragment checking
- Adjusts NPM scripts


svrnm (approved):

.github/workflows/check-spelling.yml
```diff
@@ -16,7 +16,7 @@ jobs:
             content/**/*.md
             layouts/**/*.md
             data/**/*
-            !data/registry
+            !data/registry/**/*
```
chalin, line 19: I'm not convinced that this is actually needed, b/c we're not seeing any errors being reported in this PR, only in the JS PR. But it shouldn't hurt.

Issue #3083, opened 2023-07-26
CI improvements - 2023/07


chalin:
Misc. CI improvements:

- [x] #3072
- [x] #3084
- [x] Spelling:
  - [x] #3095
  - [x] #3093
  - [x] #3116
- [x] #3097
  - [x] #3099 -- oops, made that last edit too quickly!
- [x] #3101
- [x] #3103 && #3110
- [x] Site build info:
  - [x] #3111 
  - [x] #3113
- [x] #3115
- ➡️  Make npm-run-all optional or replace it with a simple script or task, since we only use it to run scripts sequentially

PR #3116, opened 2023-07-31
[CI] fix:dict - ensure workflow fails if files have changed

chalin:
- Contributes to #3083
- Adds a check after running `fix:dict` to ensure that no files have been updated; if files have changed, the workflow fails and gives instructions
- ~**NOTE**: the action will fail now because some files don't have their dictionaries normalized~

### DIFF
--- a/.github/workflows/check-spelling.yml
+++ b/.github/workflows/check-spelling.yml
@@ -25,3 +25,13 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - run: npm run fix:dict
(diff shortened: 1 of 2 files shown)


chalin:
PR reports a failure as expected, see https://github.com/open-telemetry/opentelemetry.io/actions/runs/5720037680/job/15499213451?pr=3116:

> image

chalin:
I'll fix the page now.

cartermp (approved):

PR #3120, opened 2023-08-01
[CI] textlint workflow: use npm script for same coverage

### LINKED ISSUE #3119: [CI] Textlint fails locally but not in GH action
Here's the output of when I run the linter locally:

```console
> check:text
> npm run _check:text


> _check:text
> npx textlint content data


/Users/chalin/git/lf/open-telemetry/opentelemetry.io/content/en/docs/instrumentation/php/automatic.md
  102:22  ✓ error  Incorrect usage of the term: “php”, use “PHP” instead  terminology

✖ 1 problem (1 error, 0 warnings)
✓ 1 fixable problem.
Try to run: $ textlint --fix [file]

ERROR: "check:text" exited with 1.
ERROR: "check" exited with 1.
```

Maybe the script that generates errors for GH gobbles up the failure sometimes? /cc @svrnm

chalin:
- Closes #3119
- Uses `npm run _check:text` rather than raw call to textlint, so that we get the same files checked as when we run `check:text`
- ~**NOTE**: if all goes well, this PR's textlint action will report an error.~
- Adds ignore directives to a PHP file that textlint reports an issue on

### DIFF
--- a/.github/workflows/scripts/textlint.sh
+++ b/.github/workflows/scripts/textlint.sh
@@ -2,16 +2,16 @@
 
 # Run textlint and translate the JSON output into a format that provides file annotations for github PRs
 # https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-error-message
-ERRORS=$(npx textlint -f json content/en/*.md | jq -r  '
-    .[] | 
-    ( .filePath | sub(env.PWD + "/"; "")  ) as $fp | 
-    .messages[] | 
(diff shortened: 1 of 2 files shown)


chalin:
Yes, it does now report the error, see https://github.com/open-telemetry/opentelemetry.io/actions/runs/5721423252/job/15502989504?pr=3120:

> image

I'll commit a fix to that file now.

svrnm (approved):

PR #3121, opened 2023-08-01
Rename .NET Core to .NET

Kielek:
Rename .NET Core to .NET
.NET Core is legacy name and was renamed to .NET in 5.0 release


chalin (commented):
In that case, also change the URL to be canonical -- the existing URL redirects. The Framework URL my need to be changed too?

Kielek:
> In that case, also change the URL to be canonical -- the existing URL redirects. The Framework URL my need to be changed too?

Link updated in 6640eb4b52c0d7effe9de986a88afe105ec26fd7
It still redirect to language specific version, but I would keep it as it is.

Change for .NET Framework are not needed.

chalin (approved):
✨

PR #3125, opened 2023-08-02
Add a warning to browser getting started

svrnm:
We have the same warning now for [Manual instrumentation - Browser](https://opentelemetry.io/docs/instrumentation/js/manual/#browser), I think the [getting started](https://opentelemetry.io/docs/instrumentation/js/getting-started/browser/) should have it as well, because at least I get ~weekly questions around OTel for browser/mobile/etc.

### DIFF
--- a/content/en/docs/instrumentation/js/getting-started/browser.md
+++ b/content/en/docs/instrumentation/js/getting-started/browser.md
@@ -5,10 +5,7 @@ weight: 20
 ---
 
 {{% alert title="Warning" color="warning" %}}
-
-Browser instrumentation **experimental**. If you are interested in helping out, see [Contributing](https://github.com/open-telemetry/opentelemetry-js#contributing).
-
-{{% /alert %}}
+{{% _param notes.browser-instrumentation %}} {{% /alert %}}
 
 While this guide uses the example application presented below, the steps to
 instrument your own application should be similar.
--- a/content/en/docs/instrumentation/js/getting-started/browser.md
+++ b/content/en/docs/instrumentation/js/getting-started/browser.md
@@ -6,12 +6,7 @@ weight: 20
 
 {{% alert title="Warning" color="warning" %}}
 
-Instrumentation for the browser is experimental and this is mostly unspecified:
-this may break in the future. If you want to help changing that, you can join
-the Client Instrumentation SIG on CNCF Slack channel
-[#otel-client-side-telemetry](https://cloud-native.slack.com/archives/C0239SYARD2)
-or by attending
-[their meetings](https://docs.google.com/document/d/16Vsdh-DM72AfMg_FIt9yT9ExEWF4A_vRbQ3jRNBe09w/edit)
+Browser instrumentation **experimental**. If you are interested in helping out, see [Contributing](https://github.com/open-telemetry/opentelemetry-js#contributing).
 
 {{% /alert %}}
 
--- a/content/en/docs/instrumentation/js/getting-started/browser.md
+++ b/content/en/docs/instrumentation/js/getting-started/browser.md
@@ -4,6 +4,17 @@ aliases: [/docs/js/getting_started/browser]
 weight: 20
 ---
 
+{{% alert title="Warning" color="warning" %}}
+
+Instrumentation for the browser is experimental and this is mostly unspecified:
+this may break in the future. If you want to help changing that, you can join
+the Client Instrumentation SIG on CNCF Slack channel
+[#otel-client-side-telemetry](https://cloud-native.slack.com/archives/C0239SYARD2)
+or by attending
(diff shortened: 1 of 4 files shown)


chalin (commented):
Oh, sorry, I missed that warning in the manual page earlier, but now it makes me want to ask: Why is the [OTel JS status](https://opentelemetry.io/docs/instrumentation/js/#status-and-releases) not reflecting this "experimental" status?

svrnm:
> Oh, sorry, I missed that warning in the manual page earlier, but now it makes me want to ask: Why is the [OTel JS status](https://opentelemetry.io/docs/instrumentation/js/#status-and-releases) not reflecting this "experimental" status?

Because browser and nodejs are different here. Good point, maybe we should also add some clarity on that page.

pichlermarc:
> > Oh, sorry, I missed that warning in the manual page earlier, but now it makes me want to ask: Why is the [OTel JS status](https://opentelemetry.io/docs/instrumentation/js/#status-and-releases) not reflecting this "experimental" status?
> 
> Because browser and nodejs are different here. Good point, maybe we should also add some clarity on that page.

I fully agree. We should also do it in the OTel JS repo's README, as it could be clearer from that one too. I opened an issue on the JS repo to tackle that: https://github.com/open-telemetry/opentelemetry-js/issues/4035

content/en/docs/instrumentation/js/getting-started/browser.md
```diff
@@ -4,6 +4,17 @@ aliases: [/docs/js/getting_started/browser]
...
 ---
 
+{{% alert title="Warning" color="warning" %}}
+
+Instrumentation for the browser is experimental and this is mostly unspecified:
+this may break in the future. If you want to help changing that, you can join
+the Client Instrumentation SIG on CNCF Slack channel
+[#otel-client-side-telemetry](https://cloud-native.slack.com/archives/C0239SYARD2)
+or by attending
+[their meetings](https://docs.google.com/document/d/16Vsdh-DM72AfMg_FIt9yT9ExEWF4A_vRbQ3jRNBe09w/edit)
```
chalin, line 14: I think that this warning prose should be greatly simplified, especially if the warning is going to be repeated on several pages. For example:

```suggestion
Browser instrumentation **experimental**. If you are interested in helping out, see [Contributing](https://github.com/open-telemetry/opentelemetry-js#contributing).
```

Of course having such a warning means we need to remember to pull it out once the status changes. It also sets a precedent for all other languages. My 2 cents.

chalin (changes requested):
Ok, I'd vote to clarify the status on the JS landing page _first_.

If we do add a warning, it should be short -- especially if you're wanting to repeat the warning across several pages (maybe we need a shortcode to help keep things DRY)?

See inline comment.

svrnm:
@chalin I added the note to the hugo.yaml similar to the one we have for docker compose. I use the same wording now on three places: _index, getting-started and manual.

I also rephrased it once again to the following and decided *against* updating the status table of OTel JS:

> Client instrumentation for the browser is **experimental** and mostly **unspecified**. If you are interested in helping out, get in touch with the [Client Instrumentation SIG](https://docs.google.com/document/d/16Vsdh-DM72AfMg_FIt9yT9ExEWF4A_vRbQ3jRNBe09w/edit)

Some rational for that: Upading the status table would be confusing, since the API & SDK for JavaScript in general are following the specification for traces, metrics, logs, etc. What we are looking at here is **different** from that as it is for Client-side instrumentation (or real user monitoring) which is (a) not solely a problem of the JS community and (b) will likely lead to an additional set of API/SDK specs or make use of existing ones in a certain way. 

I took the wording "mostly unspecified" from the [Roadmap - P2: Client Instrumentation (RUM)]:

>  OpenTelemetry JS has technically supported capturing spans from web browsers since its first releases, however this behavior was mostly unspecified

chalin (approved):
Well done with the refactoring. LGTM ✨

cartermp (approved):

hugo.yaml
```diff
@@ -171,6 +171,9 @@ params:
     docker-compose-v2: |
       `docker-compose` is deprecated. For details, see
       [Migrate to Compose V2](https://docs.docker.com/compose/migrate/).
+    browser-instrumentation: |
+      Client instrumentation for the browser is **experimental** and mostly **unspecified**. If you are interested in 
+      helping out, get in touch with the [Client Instrumentation SIG](https://docs.google.com/document/d/16Vsdh-DM72AfMg_FIt9yT9ExEWF4A_vRbQ3jRNBe09w/edit)
```
chalin, line 176: Oops, missed that: this sentence should end with a period.

svrnm: let's have a follow up PR then for the period.

chalin: See #3134.

PR #3130, opened 2023-08-03
Get rid of bugfix textlint rule and update .NET and eBPF

cartermp:
Not a fan of that bugfix rule, heh

### DIFF
--- a/.textlintrc.json
+++ b/.textlintrc.json
@@ -86,7 +86,7 @@
         [".net", ".NET"],
         ["3rd[- ]party", "third-party"],
         ["cpp", "C++"],
-        ["dotnet", ".NET"],
+        ["(\\.|dot)net", ".NET"],
         ["ebpf", "eBPF"],
         ["[^\\.]json", "JSON"],
         ["[^\\.]html", "HTML"],


.textlintrc.json
```diff
@@ -83,10 +83,12 @@
...
         "Zend",
         "Zipkin",
+        [".net", ".NET"],
         ["3rd[- ]party", "third-party"],
-        ["bug[- ]fix(es)?", "bugfix$1"],
         ["cpp", "C++"],
-        ["dotnet", ".NET"],
+        ["[Dd]ot[Nn]et", ".NET"],
+        ["EBPF", ".eBPF"],
+        ["eb[Pp][Ff]", ".eBPF"],
```
chalin, line 91: I suspect that the `.` was a copy-paste error. Also, maybe we can combine the two rules:

```suggestion
        ["[eE][bB][Pp][Ff]", "eBPF"],
```

svrnm, line 91: As far as I understand the "documentation" for textlint-rule-terminology those regular expressions are case insensitive: https://github.com/sapegin/textlint-rule-terminology (see "        // RegExp (case-insensitive) → replacement" in the sample config

You should get the same result by just putting "eBPF" in the list.

.textlintrc.json
```diff
@@ -83,10 +83,12 @@
         "YouTube",
         "Zend",
         "Zipkin",
+        [".net", ".NET"],
         ["3rd[- ]party", "third-party"],
-        ["bug[- ]fix(es)?", "bugfix$1"],
```
svrnm, line 87: Do you want to add the reverse rule then?

.textlintrc.json
chalin: Oh, that's even better.

chalin: I've adapted the suggestion above based on @svrnm's comment about the regex being case insensitive.

.textlintrc.json
chalin: Yeah, we should.

cartermp: Eh, I don't know if we should add the reverse rule. Both are fine. I don't feel strongly about this though.

.textlintrc.json
```diff
@@ -83,10 +83,11 @@
...
         ["cpp", "C++"],
-        ["dotnet", ".NET"],
+        ["[Dd]ot[Nn]et", ".NET"],
```
chalin, line 89:
```suggestion
        ["dotnet", ".NET"],
```

.textlintrc.json
chalin: I think that we should opt for consistency, but we can add the reverse rule in a followup PR.

chalin (approved):
Other than the fix suggested inline for "dotnet", LGTM

chalin:
Actually even the "dotnet" fix can go in the followup PR.

.textlintrc.json
```diff
@@ -83,10 +83,11 @@
         "YouTube",
         "Zend",
         "Zipkin",
+        [".net", ".NET"],
```
chalin, line 86: Wait, this rule is wrong(it should use `\\.`) but in any case, it should be merged with the original "dotnet" rule.

chalin (changes requested):

.textlintrc.json
```diff
@@ -83,10 +83,11 @@
...
         ["3rd[- ]party", "third-party"],
-        ["bug[- ]fix(es)?", "bugfix$1"],
         ["cpp", "C++"],
         ["dotnet", ".NET"],
```
chalin, line 89:
```suggestion
        ["(\\.|dot)net", ".NET"],
```

chalin (approved):
Ok, let's get this in.

Issue #3138, opened 2023-08-06
Opentelemery Collector doc crand import alias needs to be mentioned on "check your work" or import section.


bagmeg:
**What needs to be changed?** Describe the update that is required.
From "Building a Trace receiver" tutorial, found that NewSpanID() function is using crand.Reader and it is not mentioned neither inside code (import) nor "check your work" section. I think it is confusing and people will have to make guess which package to use. For myself I used crypto/rand and it worked but not sure if document was meant to use this package.
So I think this should be mentioned somewhere on document.
```go
func NewSpanID() pcommon.SpanID {
	var rngSeed int64
	_ = binary.Read(crand.Reader, binary.LittleEndian, &rngSeed)
	randSource := rand.New(rand.NewSource(rngSeed))
       ...
}
```

**What is the name + path of the page that needs changed?** The relative path
and page title where you found a problem.
content/en/docs/collector/trace-receiver.md

svrnm:
thanks for raising this issue @bagmeg, do you want to provide a PR for fixing this yourself?

bagmeg:
@svrnm sure I can make PR for this. Can you assign it to me?

svrnm:
@bagmeg [We have a policy](https://github.com/open-telemetry/opentelemetry.io/blob/main/CONTRIBUTING.md#found-a-problem) that we do not assign issues, just go ahead and raise your PR and point out that it fixes this particular issue

svrnm:
fixed via #3140

PR #3140, opened 2023-08-08
Add crand explain

bagmeg:
Added an explanation that crand is an alias for the crypto/rand package. 
This pull request is for issue #3138 

**Preview**: https://deploy-preview-3140--opentelemetry.netlify.app/docs/collector/trace-receiver/


content/en/docs/collector/trace-receiver.md
```diff
@@ -1918,6 +1918,11 @@ func NewSpanID() pcommon.SpanID {
 }
 ```
 
+> #### Check your work
+>
+> - Importing crypto/rand package as crand because name is duplicated with
+>   math/rand.
```
svrnm, line 1924:
```suggestion
> - Imported the `crypto/rand` package as `crand` to avoid name conflict with 
>   math/rand.
```

svrnm:
thank you @bagmeg! Please check the CLA that needs to be signed by you, so we can accept your contribution

content/en/docs/collector/trace-receiver.md
bagmeg: Applied suggestion. Thank you

content/en/docs/collector/trace-receiver.md
```diff
@@ -1918,6 +1918,11 @@ func NewSpanID() pcommon.SpanID {
 }
 ```
 
+> #### Check your work
+>
+> - Imported the `crypto/rand` package as `crand` to avoid name conflict with
+>   math/rand.
+
```
chalin, line 1925: It is my understanding that "check your work" sections highlight additions or changes that were mentioned previously.

In this case, there is no sample code that illustrates the import.

I'd suggest that we instead add the required import to the code block that ends at line 1919, and then drop this "Check your work" call out.

chalin (changes requested):
I'd vote to illustrate the import rather than add a "check your work" callout. See inline comment for details.

content/en/docs/collector/trace-receiver.md
svrnm: yes, let's have that partial import

svrnm: @chalin please take another look

chalin (changes requested):
Oh wait, this needs to be changed to use the new `{{% alert %}}` syntax, that is being used as of:

- #3204

cartermp (approved):

Issue #1661, opened 2022-08-25
Is it OpenTelemetry Java Agent or OpenTelemetry Javaagent?


trask:
"Java Agent" is the established term in the Java community, but I have seen many people outside of the Java community confused by this, b/c when they see "Agent", they think of out-of-process "Agent" (i.e. OpenTelemetry Collector/Agent). [btw, this confusion was [mentioned way back in the very first OTEP](https://github.com/open-telemetry/oteps/blob/main/text/0001-telemetry-without-manual-instrumentation.md#motivation)]

I've sort of been trying to make "OpenTelemetry Javaagent" happen to help with this confusion, but I'd like to see what others think, if this is a big enough problem to deviate from the official Java terminology. While the JVM arg is `-javaagent:...` and so I don't think "Javaagent" is a huge stretch, I believe all official Java docs refer to these as "Java Agents", and not "Javaagents".

cc @open-telemetry/java-instrumentation-approvers

svrnm:
I remember that we had that discussion with another PR a few months back as well.

While it makes things much longer (and harder to read?) what about calling it "Java Auto Instrumentation Agent"? 

Independent of what we use as a term, it might be got to have an explicit statement on that potential confusion in the documentation.

trask:
> what about calling it "Java Auto Instrumentation Agent"?

do you know if this works for other languages? (e.g. .NET Auto Instrumentation Agent)

if so that's definitely a nice +

cartermp:
Heh, to make matters more confusing, .NET autoinstrumentation is explicit _not_ an agent, but a set of runtime diagnostics hook...

svrnm:
> do you know if this works for other languages? (e.g. .NET Auto Instrumentation Agent)

If you come from the APM world, yes, that works. Same for auto instrumentation for any other language. Reading through [OTEP-1](https://github.com/open-telemetry/oteps/blob/main/text/0001-telemetry-without-manual-instrumentation.md#motivation) I see that this already has some mentioning of that topic:

> One way to navigate situations like these is with a software layer that adds OpenTelemetry instrumentation to a service without modifying the source code for that service. **(In the conventional APM world, these software layers are often called “agents”, though that term is overloaded and ambiguous so we try avoid it in this document.)**

[...]

PR #3144, opened 2023-08-08
Rework java manual instrumentation

svrnm:
@open-telemetry/java-approvers @open-telemetry/docs-approvers 

[As mentioned](https://github.com/open-telemetry/opentelemetry-java/pull/5665#issuecomment-1667487031) this is a try to improve the manual instrumentation docs for Java. I used the updated version for [JavaScript](https://opentelemetry.io/docs/instrumentation/js/manual/) as a blueprint and I put the auto-configuration at the top.

There is still some work that needs to be done, but I would appreciate some early feedback. Thanks!

Preview: 
https://deploy-preview-3144--opentelemetry.netlify.app/docs/instrumentation/java/manual/

Additional pages that will be introduced:
https://deploy-preview-3144--opentelemetry.netlify.app/docs/instrumentation/java/exporters/
https://deploy-preview-3144--opentelemetry.netlify.app/docs/instrumentation/java/libraries/ (tbd)


content/en/docs/instrumentation/java/exporters.md
```diff
@@ -0,0 +1,109 @@
...
+
+For [manual instrumentation](/docs/instrumentation/java/manual), you will find
+some introductions below on how to set up backends and the matching exporters.
+
+## OTLP endpoint
+
+To send trace data to a OTLP endpoint (like the [collector](/docs/collector) or
+Jaeger) you'll want to use `opentelemetry-exporter-otlp`:
+
+<!-- prettier-ignore-start -->
```
chalin, line 24: Just curious why you added `prettier-ignore` directives? They shouldn't be needed anymore.

svrnm: probably a copy&paste error or I forgot to remove them

content/en/docs/instrumentation/java/manual.md
```diff
@@ -7,31 +7,198 @@ aliases:
...
 
-## Setup
+{{% alert title="Note" color="info" %}}
 
-The first step is to get a handle to an instance of the `OpenTelemetry`
-interface.
+On this page you will learn how you can add traces, metrics and logs to your
+code _manually_. But, you are not limited to only use one kind of
+instrumentation: use
+[automatic instrumentation](/docs/instrumentation/java/automatic/) to get
```
breedx-splk, line 23: Not sure of exactly phrasing, but it would be nice to have something here that also points out the use case of dev/operators who don't have access to the source code or otherwise can't or don't want to modify source. Maybe something like:

"If you cannot modify the source code for your project, you should skip manual instrumentation and just use automatic instrumentation." 

Feel free to improve on that. :)

content/en/docs/instrumentation/java/manual.md
```diff
@@ -7,31 +7,198 @@ aliases:
...
+dependencies for the OpenTelemetry API.
+
+<!-- prettier-ignore-start -->
+
+{{< tabpane text=true >}}
+{{% tab Gradle %}}
+
+```kotlin
+dependencies {
+    implementation 'io.opentelemetry:opentelemetry-api:{{% param javaVersion %}}'
```
breedx-splk, line 196: This could be slightly confusing to inexperienced users who might replace their ENTIRE dependencies block. I think it should be additive since it builds on the previous block.
```suggestion
    implementation("org.springframework.boot:spring-boot-starter-web")
    implementation 'io.opentelemetry:opentelemetry-api:{{% param javaVersion %}}'
```

breedx-splk: This might also be true of other gradle dependency blocks below.

svrnm: good call out, I reduced the list to make it more obvious what is new, but maybe [this](https://gohugo.io/content-management/syntax-highlighting/#highlighting-in-code-fences) can help

breedx-splk: Yeah I'd be totally happy with that. Thanks @svrnm

content/en/docs/instrumentation/java/exporters.md
```diff
@@ -0,0 +1,109 @@
...
+title: Exporters
+weight: 50
+---
+
+In order to visualize and analyze your traces, you will need to export them to a
+backend such as [Jaeger](https://www.jaegertracing.io/) or
+[Zipkin](https://zipkin.io/). OpenTelemetry Java provides exporters for some
+common open source backends.
+
+If you use the Javaagent for
```
cartermp, line 11: Should we refer to it as Javaagent or "Java agent"? I've seen this both ways, so unclear what's preferred

content/en/docs/instrumentation/java/manual.md
```diff
@@ -7,31 +7,198 @@ aliases:
...
+```text
+[5,6,5,3,6,1,2,5,4,4,2,4]
+```
+
+## Manual instrumentation setup
+
+For both library and app instrumentation, the first step is to install the
+dependencies for the OpenTelemetry API.
+
+<!-- prettier-ignore-start -->
```
cartermp, line 189: Same deal here, I assume we can get rid of these

content/en/docs/instrumentation/java/manual.md
```diff
@@ -55,93 +271,277 @@ telemetry source-identifying info.
...
+For debugging and local development purposes, use the `logging` exporter. After
+you have finished setting up manual instrumentation, provide an appropriate
+exporter library in the classpath of the application to
+[export the app's telemetry data](/docs/instrumentation/java/exporters/) to one
+or more telemetry backends.
+
+The SDK auto-configuration has to be initialized from your code in order to
+allow the module to go through the provided environment variables (or system
+properties) and set up the `OpenTelemetry` instance by using the builders
+internally.
```
cartermp, line 363: I think it's important to emphasize _where_ in the code as well. The example shows where, but explaining more generally where you'd do it matters.

cartermp (approved):
Overall lgtm

content/en/docs/instrumentation/java/exporters.md
```diff
@@ -0,0 +1,109 @@
...
+    </dependencies>
+</project>
+```
+
+{{< /tab >}}
+{{< /tabpane>}}
+
+Next, configure the exporter to point at an OTLP endpoint.
+
+If you use [SDK auto-configuration](/docs/instrumentation/java/manual/#automatic-configuration)
```
jack-berg, line 54: Are you going to break out a page dedicated to auto configuration?

content/en/docs/instrumentation/java/exporters.md
```diff
@@ -0,0 +1,109 @@
...
+
+```java
+import io.opentelemetry.exporter.otlp.metrics.OtlpGrpcMetricExporter;
+import io.opentelemetry.exporter.otlp.trace.OtlpGrpcSpanExporter;
+import io.opentelemetry.exporter.otlp.logs.OtlpGrpcLogRecordExporter;
+import io.opentelemetry.sdk.metrics.export.PeriodicMetricReader;
+import io.opentelemetry.sdk.trace.export.BatchSpanProcessor;
+import io.opentelemetry.sdk.logs.export.BatchLogRecordProcessor;
+
+        SdkTracerProvider sdkTracerProvider = SdkTracerProvider.builder()
```
jack-berg, line 74: Is the intent for this to be psuedocode or a blub that folks can copy and paste into their app? If a copy / pastable blurb, maybe consider wrapping this in a main function:
```
  public static void main(String[] args) {
    // Initialize opentelemetry early in application lifecycle
   SdkTracerProvider sdkTracerProvider = ...

   // ... continue with application startup
  }
```

content/en/docs/instrumentation/java/exporters.md
```diff
@@ -0,0 +1,109 @@
...
+  -p 6831:6831/udp \
+  -p 6832:6832/udp \
+  -p 5778:5778 \
+  -p 16686:16686 \
+  -p 4317:4317 \
+  -p 4318:4318 \
+  -p 14250:14250 \
+  -p 14268:14268 \
+  -p 14269:14269 \
+  -p 9411:9411 \
```
jack-berg, line 107: There's a lot of ports here.. surely they're not all needed?

content/en/docs/instrumentation/java/manual.md
```diff
@@ -46,7 +213,56 @@ telemetry source-identifying info.
...
+If you instrument a Java app, install the dependencies for the OpenTelemetry
+SDK.
+
+{{< tabpane text=true >}} {{% tab Gradle %}}
+
+```kotlin
+dependencies {
+    implementation 'io.opentelemetry:opentelemetry-api:{{% param javaVersion %}}'
+    implementation 'io.opentelemetry:opentelemetry-sdk:{{% param javaVersion %}}'
+    implementation 'io.opentelemetry:opentelemetry-sdk-metrics:{{% param javaVersion %}}'
```
jack-berg, line 244: This dependency is not needed - its packaged with `opentelemetry-sdk`.

content/en/docs/instrumentation/java/manual.md
jack-berg: As early as possible in the application lifecycle.

content/en/docs/instrumentation/java/manual.md
```diff
@@ -55,93 +271,277 @@ telemetry source-identifying info.
...
+  public static void main(String[] args) {
+    SpringApplication app = new SpringApplication(DiceApplication.class);
+    app.setBannerMode(Banner.Mode.OFF);
+    app.run(args);
+  }
+
+  @Bean
+  public OpenTelemetry openTelemetry() {
+    Resource resource = Resource.getDefault()
+        .merge(Resource.create(Attributes.of(ResourceAttributes.SERVICE_NAME, "dice-server", ResourceAttributes.SERVICE_VERSION, "0.1.0")));
```
jack-berg, line 465: Minor, but I would do `Resource.getDefault().toBuilder().put(SERVICE_NAME, "dice-server").put(SERVICE_VERSION, "0.1.0").build()` instead.

content/en/docs/instrumentation/java/exporters.md
svrnm: [Time to resolve this issue](https://github.com/open-telemetry/opentelemetry.io/issues/1661)

content/en/docs/instrumentation/java/manual.md
svrnm: fair point, it's missing some "as early as possible" or "ideally in your main method" etc.

content/en/docs/instrumentation/java/manual.md
```diff
@@ -55,93 +271,277 @@ telemetry source-identifying info.
...
+
+To verify your code, build and run the app:
+
+```sh
+gradle assemble
+env \
+OTEL_SERVICE_NAME=dice-server \
+OTEL_TRACES_EXPORTER=logging \
+OTEL_METRICS_EXPORTER=logging \
+OTEL_LOGS_EXPORTER=logging \
```
jack-berg, line 409: Nothing in this setup is bridging logs into opentelemetry. Maybe consider using one of the appenders as described in [log appenders](https://opentelemetry.io/docs/instrumentation/java/manual/#log-appenders)?

content/en/docs/instrumentation/java/manual.md
```diff
@@ -157,38 +557,158 @@ instrumenting][instrumentation library] the [instrumented library] or
...
+  public Dice(int min, int max, OpenTelemetry openTelemetry) {
+    this.min = min;
+    this.max = max;
+    this.tracer = openTelemetry.getTracer(Dice.class.getName(), "0.1.0");
+  }
+
+  public Dice(int min, int max) {
+    this.min = min;
+    this.max = max;
+    this.tracer = OpenTelemetry.noop().getTracer(Dice.class.getName(), "0.1.0");
```
jack-berg, line 631: Replace with: `this(min, max, OpenTelemetry.noop())`

content/en/docs/instrumentation/java/resources.md
```diff
@@ -0,0 +1,57 @@
...
+indicate latency in your system, you can narrow it down to a specific container,
+pod, or Kubernetes deployment.
+
+If you use the Javaagent for
+[automatic instrumentation](/docs/instrumentation/java/automatic) you can learn
+how to setup resource detection following the [Agent Configuration
+Guide]/docs/instrumentation/java/automatic/agent-config)
+
+For manual instrumentation, you will find some introductions on how to set up
+resource detection below.
```
jack-berg, line 24: There's a whole suite of resource providers that detect / add resource attributes based on environment: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/resources/library/README.md

These can be used with or without autoconfigure, and are built into the otel java agent.

content/en/docs/instrumentation/java/exporters.md
```diff
@@ -0,0 +1,109 @@
...
+{{< /tab >}}
+{{< /tabpane>}}
+
+Next, configure the exporter to point at an OTLP endpoint.
+
+If you use [SDK auto-configuration](/docs/instrumentation/java/manual/#automatic-configuration)
+all you need to do is update your environment variables:
+
+```shell
+env OTEL_TRACES_EXPORTER=otlp OTEL_METRICS_EXPORTER=otlp OTEL_LOGS_EXPORTER=otlp java -jar ./build/libs/java-simple.jar
```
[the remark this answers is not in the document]
svrnm: Ah, that's a great point!

content/en/docs/instrumentation/java/manual.md
svrnm: Oddly, if I do not include `-metrics` the setup with auto-configurations throws me an error, maybe thats a bug?

```
***************************
APPLICATION FAILED TO START
***************************

Description:

An attempt was made to call a method that does not exist. The attempt was made from the following location:

    io.opentelemetry.sdk.autoconfigure.MeterProviderConfiguration.configureMeterProvider(MeterProviderConfiguration.java:47)

The following method did not exist:

    'io.opentelemetry.sdk.metrics.internal.exemplar.ExemplarFilter io.opentelemetry.sdk.metrics.internal.exemplar.ExemplarFilter.traceBased()'

The calling method's class, io.opentelemetry.sdk.autoconfigure.MeterProviderConfiguration, was loaded from the following location:

    jar:file:/Users/neumanns/Projects/manual-java/java-simple/build/libs/java-simple.jar!/BOOT-INF/lib/opentelemetry-sdk-extension-autoconfigure-1.28.0.jar!/io/opentelemetry/sdk/autoconfigure/MeterProviderConfiguration.class

The called method's class, io.opentelemetry.sdk.metrics.internal.exemplar.ExemplarFilter, is available from the following locations:

    jar:file:/Users/neumanns/Projects/manual-java/java-simple/build/libs/java-simple.jar!/BOOT-INF/lib/opentelemetry-sdk-metrics-1.19.0.jar!/io/opentelemetry/sdk/metrics/internal/exemplar/ExemplarFilter.class

The called method's class hierarchy was loaded from the following locations:

    io.opentelemetry.sdk.metrics.internal.exemplar.ExemplarFilter: jar:file:/Users/neumanns/Projects/manual-java/java-simple/build/libs/java-simple.jar!/BOOT-INF/lib/opentelemetry-sdk-metrics-1.19.0.jar!/


Action:

Correct the classpath of your application so that it contains compatible versions of the classes io.opentelemetry.sdk.autoconfigure.MeterProviderConfiguration and io.opentelemetry.sdk.metrics.internal.exemplar.ExemplarFilter
```

content/en/docs/instrumentation/java/manual.md
svrnm: At this stage of the documentation no telemetry will be emitted, later I will update the log appenders section (and metrics) to also mention that demo application, I just started with tracing for now

content/en/docs/instrumentation/java/exporters.md
```diff
@@ -0,0 +1,109 @@
...
+how to setup exporters following the
+[Agent Configuration Guide](/docs/instrumentation/java/automatic/agent-config)
+
+For [manual instrumentation](/docs/instrumentation/java/manual), you will find
+some introductions below on how to set up backends and the matching exporters.
+
+## OTLP endpoint
+
+To send trace data to a OTLP endpoint (like the [collector](/docs/collector) or
+Jaeger) you'll want to use `opentelemetry-exporter-otlp`:
```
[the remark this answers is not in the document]
svrnm: Do you want this to be included in this PR or in a follow-up update?

content/en/docs/instrumentation/java/exporters.md
svrnm: Eventually if needed, yes, but maybe not in this PR.

content/en/docs/instrumentation/java/exporters.md
svrnm: I just copied that over from the JS one, indeed with the setup here most ports are not required

content/en/docs/instrumentation/java/manual.md
svrnm: ah, my Java is rusty ;-)

I am wondering, would it make sense here to replace `OpenTelemetry.noop()` with `GlobalOpenTelemetry`. If I understand things correctly, this would add another layer of fallback, because if a global object is registered it can be used, otherwise it is a noop as well.

content/en/docs/instrumentation/java/exporters.md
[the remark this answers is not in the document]
jack-berg: I'm happy to do in a followup. Just wanted to make the comment because its been a topic of conversation lately and I don't want it to get lost in the shuffle.

content/en/docs/instrumentation/java/manual.md
jack-berg: Oh this is dependency collision issue with spring boot dependency management. I've come across this before. Just did a quick search and found this message I send about this:
>I’ve just discovered something rather annoying upgrading to spring boot 3. If you use the [io.spring.dependency-management](https://docs.spring.io/dependency-management-plugin/docs/current/reference/html/), it brings in dependencies in a way that conflicts and overrides our opentelemetry-bom. I.e. I was using opentelemetry-bom:1.22.0 and yet opentelemetry-api and other artifacts were resolving to 1.19.0. After some investigation I determined that spring boot has added opentelemetry artifacts at version 1.19.0 to its set of managed dependencies: https://docs.spring.io/spring-boot/docs/3.0.1/reference/htmlsingle/#appendix.dependency-versions
I got around it by bailing on the io.spring.dependency-management plugin and instead just depending on the spring boot BOM like this: [https://github.com/newrelic/newrelic-opentelemetry-examples/pull/258/commits/46571423558a46a49229617f5872c1135f8c9490#diff-21bc7265[…]7eb2c226b4b6b5](https://github.com/newrelic/newrelic-opentelemetry-examples/pull/258/commits/46571423558a46a49229617f5872c1135f8c9490#diff-21bc7265f74c0805ef365c54fd9e2696b565e660a8f7237ecd7eb2c226b4b6b5)

Notice how the error message is referencing `opentelemetry-sdk-metrics-1.19.0`? That's the sign that the opentelemetry-bom is being overridden.

content/en/docs/instrumentation/java/manual.md
jack-berg: Potentially... It definitely provides some small increased chance that a valid OpenTelemetry instance is used. Not sure how to weigh that against the fact that we advise against using GlobalOpenTelemetry.

content/en/docs/instrumentation/java/exporters.md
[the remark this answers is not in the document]
svrnm: Ok, let's keep this comment unresolved now and before merging this PR I'll create a follow-up issue

content/en/docs/instrumentation/java/manual.md
svrnm: Thanks, need to think about that one. On the one hand I like additional layers of fallback, but since these are the official docs it is very likely that people copy from them into their libraries and then a pattern emerges that GlobalOpenTelemetry is used because it is easy, so if we want people to _not_ use it, we should propagate the noop one

content/en/docs/instrumentation/java/manual.md
svrnm: I added a sentence, please take a look

content/en/docs/instrumentation/java/manual.md
svrnm: updated and I used `hl_lines` to highlight the lines of new dependencies and code changes overall

content/en/docs/instrumentation/java/exporters.md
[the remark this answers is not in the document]
svrnm: removed the OTEL_*_EXPORTERS, added the endpoint and added a note that `otlp` is the default value

content/en/docs/instrumentation/java/exporters.md
svrnm: I copied the example app code from `manual.md` and put highlights for the changed lines

content/en/docs/instrumentation/java/exporters.md
svrnm: fixed

content/en/docs/instrumentation/java/manual.md
svrnm: let me take a look ...

content/en/docs/instrumentation/java/manual.md
svrnm: done

content/en/docs/instrumentation/java/manual.md
svrnm: fixed, I went with the noop. There is a note about GlobalOpenTelemetry below, shall we keep that one?

content/en/docs/instrumentation/java/manual.md
breedx-splk: good improvement, thanks!

content/en/docs/instrumentation/java/resources.md
jack-berg: Add a dependency on the artifact, and use the public APIs to merge various environment based resources with your resource. E.g. to include the [OsResource](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/849e0d5229a1085a80de3867b7a013cf26aaa8d6/instrumentation/resources/library/src/main/java/io/opentelemetry/instrumentation/resources/OsResource.java#L24):
```
Resource resource = Resource.getDefault()
  .merge(OsResource.get())
  .merge(Resource.create(Attributes.builder()
    .put(SERVICE_NAME, "dice-service")
    // ..
    .builder()));
```

content/en/docs/instrumentation/java/manual.md
jack-berg: You can keep it, but may want to use some the language we use in the [GlobalOpenTelemetry javadoc](https://github.com/open-telemetry/opentelemetry-java/blob/ebf96dc5de85676a4d2d29e8ea0be7f49b3ed6b0/api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java#L34-L37) so that we're consistently warning against using it.

content/en/docs/instrumentation/java/resources.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Adding resources in code
+
+Custom resources can be configured in your code like the following:
+
+```java
+    Resource resource = Resource.getDefault()
+        .merge(Resource.create(Attributes.builder()
```
mateuszrzeszutek, line 32: Should this be just `Attributes`? Since you're creating a resource in lines 40-41

content/en/docs/instrumentation/java/exporters.md
```diff
@@ -0,0 +1,146 @@
...
+public class DiceApplication {
+  public static void main(String[] args) {
+    SpringApplication app = new SpringApplication(DiceApplication.class);
+    app.setBannerMode(Banner.Mode.OFF);
+    app.run(args);
+  }
+
+  @Bean
+  public OpenTelemetry openTelemetry() {
+    Resource resource = Resource.getDefault().toBuilder().put(SERVICE_NAME, "dice-server").put(SERVICE_VERSION, "0.1.0").build()
```
mateuszrzeszutek, line 106:
```suggestion
    Resource resource = Resource.getDefault().toBuilder().put(SERVICE_NAME, "dice-server").put(SERVICE_VERSION, "0.1.0").build();
```

content/en/docs/instrumentation/java/manual.md
svrnm: @jack-berg , @mateuszrzeszutek you are the experts here, so I'll remove/add/change it to whatever works for you.

Note that the manual instrumentation guide is right now not talking a lot about combining the agent with manual instrumentation. We might need to do another round on thinking about how to incorporate that best (maybe it requires it's own section somewhere in the "Setup" section?)

content/en/docs/instrumentation/java/resources.md
svrnm: Ok, that's something we have not described at all ([see my other comment as well](https://github.com/open-telemetry/opentelemetry.io/pull/3144#discussion_r1317066398)). If I understandit correctly you can use those resource providers only if you use the agent? The agent already comes with the SDK and everything included, so I guess the dependency file will look very different then?

content/en/docs/instrumentation/java/manual.md
```diff
@@ -55,93 +269,281 @@ telemetry source-identifying info.
...
+    SdkLoggerProvider sdkLoggerProvider = SdkLoggerProvider.builder()
+        .addLogRecordProcessor(BatchLogRecordProcessor.builder(SystemOutLogRecordExporter.create()).build())
+        .setResource(resource)
+        .build();
+
+    OpenTelemetry openTelemetry = OpenTelemetrySdk.builder()
+        .setTracerProvider(sdkTracerProvider)
+        .setMeterProvider(sdkMeterProvider)
+        .setLoggerProvider(sdkLoggerProvider)
+        .setPropagators(ContextPropagators.create(W3CTraceContextPropagator.getInstance()))
```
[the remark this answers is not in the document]
svrnm: Happy to have it

content/en/docs/instrumentation/java/manual.md
mateuszrzeszutek: > We might need to do another round on thinking about how to incorporate that best (maybe it requires it's own section somewhere in the "Setup" section?)

I think even a small section that focuses just on that would be a good idea 👍 
Definitely can be a separate PR though

content/en/docs/instrumentation/java/resources.md
mateuszrzeszutek: > If I understandit correctly you can use those resource providers only if you use the agent? The agent already comes with the SDK and everything included, so I guess the dependency file will look very different then?

No, it's a "normal" library instrumentation module: https://mvnrepository.com/artifact/io.opentelemetry.instrumentation/opentelemetry-resources/1.29.0-alpha
It is included by default with the agent, but you can also use it if you configure the SDK manually or via autoconfigure.

content/en/docs/instrumentation/java/manual.md
```diff
@@ -55,93 +269,281 @@ telemetry source-identifying info.
...
-    implementation 'io.opentelemetry:opentelemetry-semconv:{{% param javaVersion %}}-alpha'
+    implementation("org.springframework.boot:spring-boot-starter-web");
+    implementation("io.opentelemetry:opentelemetry-api:{{% param javaVersion %}}");
+    implementation("io.opentelemetry:opentelemetry-sdk:{{% param javaVersion %}}");
+    implementation("io.opentelemetry:opentelemetry-sdk-metrics:{{% param javaVersion %}}");
+    implementation("io.opentelemetry:opentelemetry-exporter-logging:{{% param javaVersion %}}");
+    implementation("io.opentelemetry:opentelemetry-semconv:{{% param javaVersion %}}-alpha");
+    implementation("opentelemetry-sdk-extension-autoconfigure:{{% param javaVersion %}}");
+    implementation("opentelemetry-sdk-extension-autoconfigure:{{% param javaVersion %}}");
+    implementation("opentelemetry-sdk-extension-autoconfigure-spi:{{% param javaVersion %}}");
```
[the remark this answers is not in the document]
svrnm: Following previous comments I keep the list as is and grow it step-by-step but highlight the new lines, here's how this looks like in hugo:

Screenshot 2023-09-06 at 14 57 45

content/en/docs/instrumentation/java/manual.md
```diff
@@ -157,38 +559,158 @@ instrumenting][instrumentation library] the [instrumented library] or
...
+    // Make the span the current span
+    try (Scope scope = span.makeCurrent()) {
+      int result = this.getRandomNumber(1, 6);
+      if (player.isPresent()) {
+        logger.info("{} is rolling the dice: {}", player.get(), result);
+      } else {
+        logger.info("Anonymous player is rolling the dice: {}", result);
+      }
+      return Integer.toString(result);
+    } finally {
```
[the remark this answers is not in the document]
svrnm: `catch` for what exceptions?

content/en/docs/instrumentation/java/resources.md
svrnm: Ah, okay, then maybe I misunderstood that, let me double check on that :)

content/en/docs/instrumentation/java/resources.md
svrnm: let me double check that, could be a copy & paste issue

content/en/docs/instrumentation/java/manual.md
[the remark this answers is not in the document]
mateuszrzeszutek: The highlight looks great, but the duplicate `opentelemetry-sdk-extension-autoconfigure` line on the first glance looks like a mistake to me; perhaps the previous code sample shouldn't include the autoconfigure module then? I.e. if the previous section is about general SDK usage we can skip the autoconfigure module there, and in this module (specifically about using autoconfigure) we include it.

content/en/docs/instrumentation/java/manual.md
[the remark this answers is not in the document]
svrnm: Ah, cool, that should be there indeed!

content/en/docs/instrumentation/java/manual.md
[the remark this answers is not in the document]
svrnm: ah... somehow my eye missed that one!

content/en/docs/instrumentation/java/resources.md
svrnm: I added a section, see below

content/en/docs/instrumentation/java/manual.md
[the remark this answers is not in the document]
svrnm: done, ptal

content/en/docs/instrumentation/java/manual.md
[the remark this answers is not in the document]
svrnm: added

content/en/docs/instrumentation/java/resources.md
svrnm: it was a copy&past error indeed

content/en/docs/instrumentation/java/manual.md
```diff
@@ -486,6 +481,7 @@ public class DiceApplication {
         .setMeterProvider(sdkMeterProvider)
         .setLoggerProvider(sdkLoggerProvider)
         .setPropagators(ContextPropagators.create(W3CTraceContextPropagator.getInstance()))
+        .setPropagators(ContextPropagators.create(W3CBaggagePropagator.getInstance()))
```
mateuszrzeszutek, line 484: Unfortunately this will simply override the previous setting; you have to do it the following way:
```suggestion
        .setPropagators(ContextPropagators.create(TextMapPropagator.composite(W3CTraceContextPropagator.getInstance(), W3CBaggagePropagator.getInstance())))
```

content/en/docs/instrumentation/java/manual.md
```diff
@@ -672,15 +668,24 @@ import io.opentelemetry.context.Scope;
 
     // Make the span the current span
     try (Scope scope = span.makeCurrent()) {
-      int result = this.getRandomNumber(1, 6);
+
+      if (!rolls.isPresent()) {
+        throw new ResponseStatusException(HttpStatus.BAD_REQUEST, "Missing rolls parameter", null);
+      }
```
mateuszrzeszutek, line 674: Nice 👍

content/en/docs/instrumentation/java/resources.md
```diff
@@ -23,6 +23,63 @@ how to setup resource detection following the
 For manual instrumentation, you will find some introductions on how to set up
 resource detection below.
 
+## Detecting resources from common environments
+
+You can use `ResourceProvider`s for filling in attributes related to common
+environments, like docs/specs/otel/resource/semantic_conventions/container/,
```
mateuszrzeszutek, line 29: Missing link?
```suggestion
environments, like [Container](/docs/specs/otel/resource/semantic_conventions/container/),
```

content/en/docs/instrumentation/java/resources.md
```diff
@@ -23,6 +23,63 @@ how to setup resource detection following the
...
+import io.opentelemetry.instrumentation.resources.ProcessRuntimeResource;
+
+...
+    Resource resource = Resource.getDefault()
+    .merge(ContainerResource.get())
+    .merge(HostResource.get())
+    .merge(OsResource.get())
+    .merge(ProcessResource.get())
+    .merge(ProcessRuntimeResource.get())
+    .merge(Resource.create(Attributes.builder()
```
mateuszrzeszutek, line 76: Nit: indentation 
```suggestion
      .merge(ContainerResource.get())
      .merge(HostResource.get())
      .merge(OsResource.get())
      .merge(ProcessResource.get())
      .merge(ProcessRuntimeResource.get())
      .merge(Resource.create(Attributes.builder()
```

content/en/docs/instrumentation/java/manual.md
```diff
@@ -69,79 +287,259 @@ telemetry source-identifying info.
...
+        .addLogRecordProcessor(BatchLogRecordProcessor.builder(SystemOutLogRecordExporter.create()).build())
+        .setResource(resource)
+        .build();
+
+    OpenTelemetry openTelemetry = OpenTelemetrySdk.builder()
+        .setTracerProvider(sdkTracerProvider)
+        .setMeterProvider(sdkMeterProvider)
+        .setLoggerProvider(sdkLoggerProvider)
+        .setPropagators(ContextPropagators.create(W3CTraceContextPropagator.getInstance()))
+        .setPropagators(ContextPropagators.create(W3CBaggagePropagator.getInstance()))
```
svrnm, line 484:
```suggestion
       .setPropagators(ContextPropagators.create(TextMapPropagator.composite(W3CTraceContextPropagator.getInstance(), W3CBaggagePropagator.getInstance())))
```

content/en/docs/instrumentation/java/manual.md
```diff
@@ -69,79 +287,258 @@ telemetry source-identifying info.
...
+    SdkLoggerProvider sdkLoggerProvider = SdkLoggerProvider.builder()
+        .addLogRecordProcessor(BatchLogRecordProcessor.builder(SystemOutLogRecordExporter.create()).build())
+        .setResource(resource)
+        .build();
+
+    OpenTelemetry openTelemetry = OpenTelemetrySdk.builder()
+        .setTracerProvider(sdkTracerProvider)
+        .setMeterProvider(sdkMeterProvider)
+        .setLoggerProvider(sdkLoggerProvider)
+       .setPropagators(ContextPropagators.create(TextMapPropagator.composite(W3CTraceContextPropagator.getInstance(), W3CBaggagePropagator.getInstance())))
```
svrnm, line 483:
```suggestion
        .setPropagators(ContextPropagators.create(TextMapPropagator.composite(W3CTraceContextPropagator.getInstance(), W3CBaggagePropagator.getInstance())))
```

content/en/docs/instrumentation/java/exporters.md
```diff
@@ -0,0 +1,141 @@
...
+To try out the exporters quickly, you can run Jaeger with OTLP enabled in a
+docker container:
+
+```shell
+docker run -d --name jaeger \
+  -e COLLECTOR_OTLP_ENABLED=true \
+  -p 16686:16686 \
+  -p 4317:4317 \
+  -p 4318:4318 \
+  jaegertracing/all-in-one:latest
```
jack-berg, line 140: Jaeger isn't going to accept the metrics or logs that the SDK is configured to send to it.

svrnm: That's true, this is actually a bigger issue since we have the same setup for other languages as well. 

As a short fix let me change language to call out that this will only show you traces.

Additionally I'll raise an issue to discuss how we want to setup those Exporter pages to have something for all signals.

mateuszrzeszutek (approved):

content/en/docs/instrumentation/java/exporters.md
svrnm: [Not perfect, but accurate for now](https://github.com/open-telemetry/opentelemetry.io/pull/3144/commits/839ab5e3172202f6525347cdcbbf211cf7cae0cb#diff-b4c7cabc3b96c8b722a2bbbf1c2f26c2b27b2c3f86cba61a49be91ca06c034b3R131)

svrnm:
Applied additional changes to solve conflict with #3239

content/en/docs/instrumentation/java/manual.md
```diff
@@ -311,7 +315,7 @@ dependencies {
     implementation("io.opentelemetry:opentelemetry-sdk:{{% param javaVersion %}}");
     implementation("io.opentelemetry:opentelemetry-sdk-metrics:{{% param javaVersion %}}");
     implementation("io.opentelemetry:opentelemetry-exporter-logging:{{% param javaVersion %}}");
-    implementation("io.opentelemetry:opentelemetry-semconv:{{% param javaVersion %}}-alpha");
+    implementation 'io.opentelemetry.semconv:opentelemetry-semconv:{{% param semconvJavaVersion %}}-alpha'
```
mateuszrzeszutek, line 318: Single quotes won't work with Kotlin build files:
```suggestion
    implementation("io.opentelemetry.semconv:opentelemetry-semconv:{{% param semconvJavaVersion %}}-alpha")
```

svrnm: Thx again @mateuszrzeszutek !

content/en/docs/instrumentation/java/exporters.md
svrnm: follow-up issue #3266

content/en/docs/instrumentation/java/manual.md
svrnm: Follow-up issue #3270

content/en/docs/instrumentation/java/exporters.md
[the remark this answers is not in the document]
svrnm: follow up issue: #3271

PR #3148, opened 2023-08-09
Fix Erlang/Elixir example for linking spans

PhillippOhlandt:
I noticed that the Elixir example does not work while going through the guide. 

I am not sure if the changes for the Erlang example are 100% correct as I am not that experienced with plain Erlang.

### DIFF
--- a/content/en/docs/instrumentation/erlang/manual.md
+++ b/content/en/docs/instrumentation/erlang/manual.md
@@ -209,7 +209,7 @@ created.
 {{< tabpane text=true langEqualsHeader=true >}} {{% tab Erlang %}}
 
 ```erlang
-Parent = otel_tracer.current_span_ctx(),
+Parent = ?current_span_ctx,
 proc_lib:spawn_link(fun() ->
                         %% a new process has a new context so the span created
                         %% by the following `with_span` will have no parent
--- a/content/en/docs/instrumentation/erlang/manual.md
+++ b/content/en/docs/instrumentation/erlang/manual.md
@@ -209,7 +209,7 @@ created.
 {{< tabpane text=true langEqualsHeader=true >}} {{% tab Erlang %}}
 
 ```erlang
-Parent = ?current_span_ctx,
+Parent = otel_tracer.current_span_ctx(),
 proc_lib:spawn_link(fun() ->
                         %% a new process has a new context so the span created
                         %% by the following `with_span` will have no parent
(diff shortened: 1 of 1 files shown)


content/en/docs/instrumentation/erlang/manual.md
```diff
@@ -209,7 +209,7 @@ created.
 {{< tabpane text=true langEqualsHeader=true >}} {{% tab Erlang %}}
 
 ```erlang
-Parent = ?current_span_ctx,
+Parent = otel_tracer.current_span_ctx(),
```
ferd, line 212: I believe the macro should work. Maybe @tsloughter can double-check, but I know if we're making this explicit, the following works in my code in general:

```suggestion
Parent = otel_tracer:current_span_ctx(otel_ctx:get_current()),
```

bryannaegele: It's defined in the tracer module.

https://github.com/open-telemetry/opentelemetry-erlang/blob/main/apps/opentelemetry_api/src/otel_tracer.erl#L34

PhillippOhlandt: Yes, that's where I found it. The erlang examples in the docs sometimes reference an explicit module and sometimes they don't so my first assumption was that the functions without explicit module (like the example here) are meant to reference the `opentelemetry` module. And since the function here does not exist in the `opentelemetry` module but in `otel_tracer` I thought it's a mistake or simply an outdated example. But maybe I am also missing an erlang related convention here.

ferd: The call started with a question mark (`?`) which points to a macro. Macros are traditionally defined in `include/` directories and these can be found at https://github.com/open-telemetry/opentelemetry-erlang/blob/b409a5fa8e3e004e7d0946e0c741da71ce518709/apps/opentelemetry_api/include/otel_tracer.hrl#L4-L5

PhillippOhlandt: @ferd Ohh, that explains a lot. I did not know that macros in erlang are global, unlike in elixir where they are module scoped and need to be explicitly imported. I will revert the erlang example later then. Thanks a lot for clarifying this!

PhillippOhlandt: Reverted.

tsloughter: The confusion is also that ` otel_tracer.current_span_ctx(),` and similar functions just shouldn't exist. It would be better if the only way to use the pdict implicit context was with the macros and the calls to the module `otel_tracer` all required passing a context variable.

Probably we should add a deprecation warning on all those functions since we can't actually remove them until a major release (not happening anytime soon at all).

ferd (approved):

chalin (approved):
Thanks @PhillippOhlandt, @ferd et al.

Issue #2554, opened 2023-03-29
CI: refcache update policy and cron-job proposal


chalin:
@svrnm @cartermp @austinlparker: Severin brought up the topic of "burden on PR contributors" of having to update the refcache file. I'd like to propose the following.

- If a PR introduces "lots" of new external links, then the PR must/should include an updated refcache file.
- For other PRs, feel free to merge the PR whether the refcache is fresh or not -- freshness is checked via a separate GH action so it is possible to merge w/o the action passing.

We can then have, say, a cron job that periodically update the website's refcache file. WDYT?

### Tasks

Main:
- [x] Create update scripts #2960

Setup a cron job to run, say, once every 4 weeks and attempt to refresh, say, 500 entries:
- [x] #7954
- [x] #7958
- [x] #7963
- [x] #7980
- [x] #7982
  - [x] Handel case when there are no changes to commit, e.g., https://github.com/open-telemetry/opentelemetry.io/actions/runs/18189265987/job/51780417292
- [x] #7989
- [x] #7991
- [x] Try to use Puppeteer as a fallback link checker when htmltest returns 4XX statuses:
  - [x] #8014 
  - [x] #8015 
  - [x] #8016
  - [x] #8017
  - [x] #8021
  - [x] #8028
- Try validating NPM package URLs using the CLI `npm list` or `npm view`
  - [x] #8033
  - [x] #8034
- Handle crates.io special case:
  - [x] #8038
  - [x] #8040
- [x] #8048 
- [x] #8065
- [x] #8104
- [x] #8145
- [x] #8170
- [x] #8342

Cleanup:
  - [x] #7965
  - [x] #8055

cartermp:
I like it a lot.

chalin:
I'd also like to propose the following tweak to the build script: it _should_ fail (it doesn't currently), if there's an invalid link or if a new external link results in a 404 (independent of the refcache freshness). How does that sound?

chalin:
@trask - the semconv daily build job is working great! I'm wondering if you'd be willing to help with this issue by writing a similar script that creates or commits to a "Refcache refresh" PR on a daily basis. I think that I'm getting a sense of how to automate this process, and now have the base scripts ready to support that process. If you aren't available to write such a script, that's ok, but I figured that I'd ask. (Otherwise, I'll try my hand at it when I next have the time.)

PR #3156, opened 2023-08-11
Registry - fix link for .NET AWS Instrumentation package

Kielek:
Based on https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1275

### DIFF
--- a/data/registry/instrumentation-dotnet-instrumentation-aws.yml
+++ b/data/registry/instrumentation-dotnet-instrumentation-aws.yml
@@ -6,7 +6,7 @@ tags:
   - instrumentation-aws
   - instrumentation
   - dotnet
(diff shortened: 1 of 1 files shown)


Kielek:
Links checker is not verifying links in registry-yaml files.

I think it can be handled as a separate issue.

chalin (approved):
👍🏻

chalin:
> Links checker is not verifying links in registry-yaml files

The issue is that the link check is currently **ignoring** links into github: https://github.com/open-telemetry/opentelemetry.io/blob/e52373fdf7b058a76395576055adee9a53c7936e/.htmltest.yml#L33-L34

The reason I've done that is because our refcache has grown considerably, and because of that the link-checking GH action over 2 min. I have a task open to address this -- e.g., #2554.

PR #3162, opened 2023-08-11
Fix links in Rust getting-started

chalin:
- Links to pages that are redirects need to be absolute, otherwise the link checker will complain.


cartermp (approved):

chalin:
I'm not sure why these weren't reported by GH. The check passed even though there was an error. I'm investigating.

PR #3168, opened 2023-08-14
Improve erlang/elixir config examples

PhillippOhlandt:
Adding the OTP app name to the config examples so that they can simply be copy/pasted and don't cause any confusion.


tsloughter (approved):
Looks good :+1:

chalin (approved):
Thanks @PhillippOhlandt and @tsloughter. RSLGTM

PR #3208, opened 2023-08-29
Update Lightstep name on vendors page

martinkuba:
Lightstep is now part of ServiceNow, and the official name is "ServiceNow Cloud Observability".


cartermp (approved):
😢

PR #3223, opened 2023-09-04
Fix incorrect reference of AttributeKey.stringKey

EddieEldridge:
## Context

The example code in the [Java Manual metrics documentation](https://opentelemetry.io/docs/instrumentation/java/manual/#metrics) simply doesn't compile.

It doesn't compile because of the incorrect reference to `stringKey`, a non-existent function.

This PR should fix that as `AttributeKey.stringKey` is used above and in many other places in the same doc.

```java
  // Right, stringKey comes from AttributeKey
  Attributes attributes_right = Attributes.of(AttributeKey.stringKey(message), "SomeWork");        
  
  // Wrong, stringKey is not defined
  Attributes attributes_wrong = Attributes.of(stringKey(message), "SomeWork");
```

### DIFF
--- a/content/en/docs/instrumentation/java/manual.md
+++ b/content/en/docs/instrumentation/java/manual.md
@@ -610,7 +610,7 @@ LongCounter counter = meter
       .build();
 
 // It is recommended that the API user keep a reference to Attributes they will record against
-Attributes attributes = Attributes.of(stringKey("Key"), "SomeWork");
+Attributes attributes = Attributes.of(AttributeKey.stringKey("Key"), "SomeWork");
 
 // Record data
 counter.add(123, attributes);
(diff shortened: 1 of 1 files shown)


cartermp (approved):
Ahh, good catch, thanks. I was curious, and I think this may be an old holdover from 2+ year old docs where this might have been compilable code as a part of some pre-release. Or it was just a doc bug that's existed forever, heh.

laurit:
> Ahh, good catch, thanks. I was curious, and I think this may be an old holdover from 2+ year old docs where this might have been compilable code as a part of some pre-release. Or it was just a doc bug that's existed forever, heh.

It is just a missing `import static io.opentelemetry.api.common.AttributeKey.stringKey;` As the same doc already uses `AttributeKey.longKey` and `AttributeKey.stringKey` without a static import I think this change is good.

laurit (approved):

PR #3236, opened 2023-09-08
Fix copy paste bug in JS metrics instrumentation

pellared:


pichlermarc (approved):
good catch, thanks! :slightly_smiling_face:

svrnm (approved):

PR #3239, opened 2023-09-08
Update docs to point to new semconv artifact

jack-berg:


cartermp:
@jack-berg yeah, it should be pretty simple to update. I think the main concern is further changes that require code updates? One of the reasons why we have the existing ones auto-update is because the large majority of the time it's just the version number that has to update. If that's the case for semconv artifacts for Java, then I'd be in favor of automating it.

content/en/docs/instrumentation/java/manual.md
```diff
@@ -1103,5 +1108,6 @@ io.opentelemetry.sdk.trace.export.BatchSpanProcessor = io.opentelemetry.extensio
 [parentbased]:
   https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/ParentBasedSampler.java
 [releases]: https://github.com/open-telemetry/opentelemetry-java#releases
+[semantic-conventions-java]: https://github.com/open-telemetry/semantic-conventions-java/releases
```
cartermp, line 1111: For this you'll need to run `npm run test-and-fix` to update the refcache.

jack-berg:
@cartermp code changes are possible since the artifact is still alpha, but this was previously the case as well. For the most part, each release will be just a version bump. I'll make sure to followup in the rare event that a version bump coincides with a code change.

mateuszrzeszutek (approved):

cartermp:
Sounds good @jack-berg -- looks like a PR I merged introduced a merge conflict. Once that's done and you can update the refcache with `npm run-and-fix` then this should be good to go.

mateuszrzeszutek (approved):

svrnm (approved):

PR #3241, opened 2023-09-08
Update refcache action to actually run the fix

cartermp:
It only runs the check - the update to the refcache isn't done.


theletterf (approved):

.github/workflows/pr-actions.yml
```diff
@@ -104,6 +104,7 @@ jobs:
       - run: npm install --omit=optional
       - run: |
           npm run check:links
+          npm run test-and-fix
```
pellared, line 107: Is the previous step needed then?

svrnm: `test-and-fix` does much more than fixing the refcache (e.g. it also runs `fix:dict` and `fix:filenames`)

If I remember correctly `check:links` *should* update the refcache files except there is any issue going on.

cartermp: It doesn't appear to be updating anything, but `test-and-fix` does.

svrnm (approved):

.github/workflows/pr-actions.yml
svrnm: that's odd ... but let's try it out then, I approved the PR

cartermp:
Huh. I tried on a another branch and indeed the original check worked? Maybe github actions are just so slow that I never noticed.

PR #3242, opened 2023-09-09
Update Hugo to 0.118.2

ydFu:
The latest version has been released, follow the upstream update version information.
Hugo [Releases v0.118.2](https://github.com/gohugoio/hugo/releases) .

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,7 +85,7 @@
     "autoprefixer": "^10.4.15",
     "cspell": "^7.0.0",
     "gulp": "^4.0.2",
-    "hugo-extended": "0.117.0",
+    "hugo-extended": "0.118.2",
     "markdownlint": "^0.30.0",
     "postcss-cli": "^10.1.0",
     "prettier": "^3.0.1",


svrnm:
@ydFu, thanks for raising this PR

Is there a particular reason why you think we should upgrade to that version? We normally do upgrades to new versions of the underlying software on the cadence that is needed and doable by the maintainers, so if there is no compelling reason to upgrade, we will not introduce this chance

ydFu:
> @ydFu, thanks for raising this PR
> 
> Is there a particular reason why you think we should upgrade to that version? We normally do upgrades to new versions of the underlying software on the cadence that is needed and doable by the maintainers, so if there is no compelling reason to upgrade, we will not introduce this chance

@svrnm ,Thank you for the reminder. I noticed that Hugo has the following changes:

v0.118.0
- Hugo 0.118.0 now builds with Go 1.21. This version also includes:
- Proper CJK support in Markdown.
- A revamped implementation of hugo new site and hugo new theme. See details.

v0.118.1
- This release only fixes a GLIBC_xxx not found issue with the Linux AMD64 binary when running on older Linux versions, which is the situation when building on Netlify/Vercel, etc.

v0.118.2
- Release: Revert to building with Go 1.20.

That's why I raised the version change request. However, I respect that 
community normally upgrade to new versions of the underlying software on the cadence that is needed and doable by the maintainers. Therefore, I'm closing this change request.

Issue #3157, opened 2023-08-11
Fix sidenav height and padding by updating Docsy


chalin:
Context:

- Update Docsy to bring in https://github.com/google/docsy/pull/1661 once it lands
- Remove: https://github.com/open-telemetry/opentelemetry.io/blob/ad9ebd53dfa98900871e5e1700c43fca8cce7d14/assets/scss/_styles_project.scss#L106-L108

chalin:
Fixed by:

- [forthcoming PR]

Screenshots:

| Before | After |
|--------|--------|
| <img width="887" alt="Screen Shot 2023-11-08 at 12 38 02" src="https://github.com/open-telemetry/opentelemetry.io/assets/4140793/5a80be72-cf5c-4857-8543-daa7ecebb341"> | <img width="894" alt="image" src="https://github.com/open-telemetry/opentelemetry.io/assets/4140793/a719d547-e2e5-409a-9c04-88f6c7434a58">

PR #3243, opened 2023-09-09
Remove foldable nav

pegasas:
try to resolve https://github.com/open-telemetry/opentelemetry.io/issues/3157

### DIFF
--- a/assets/scss/_styles_project.scss
+++ b/assets/scss/_styles_project.scss
@@ -103,10 +103,6 @@
   }
 }
 
-.foldable-nav {
-  padding-top: 0.5rem;
-} // TODO: propagate this fix to Docsy. Also class name should be prefixed with td
-
 // Adjust the spacing of page-meta and page-TOC (https://github.com/open-telemetry/opentelemetry.io/pull/354)
 // TODO: upstream
 .td-toc #TableOfContents {


svrnm:
thanks for raising this @pegasas 

@chalin can you take a look, thanks.

assets/scss/_styles_project.scss
```diff
@@ -103,10 +103,6 @@
   }
 }
 
-.foldable-nav {
```
theletterf, line 106: I think you should also remove the class from whatever HTML element has it attached?

theletterf (commented):
Not sure if this project pulls the Docsy theme as a submodule. If that's the case, how do we go about branching out from upstream?

assets/scss/_styles_project.scss
pegasas: Sure.
I checked the repo.
It seems there is not any HTML element in this repo.

svrnm:
> Not sure if this project pulls the Docsy theme as a submodule. If that's the case, how do we go about branching out from upstream?

@chalin has all the answers for that :-) Please be patient with an answer here, I don't think this issue here is super urgent right now.

chalin (changes requested):
The necessary Docsy fix mentioned in the original issue hasn't officially landed in a Docsy release yet, so this PR won't work. 

A new Docsy release is planned for "real soon now" :).

Thanks for the effort @pegasas, but I'm going to close this since there are more changes for me to do as I address #3157.

Issue #3247, opened 2023-09-11
.Net documentation: Manual Instrumentation - create new root activities - example unlcear.


technimad-splunk:
It is unclear where `source` comes from in the code example about [setting up a new root](https://opentelemetry.io/docs/instrumentation/net/manual/#creating-new-root-activities). 

Should source in the example given be replaced by MyActivitySource?

```
public static void DoWork()
{
    var previous = Activity.Current;
    Activity.Current = null;

-    var newRoot = source.StartActivity("NewRoot");
+    var newRoot = MyActivitySource.StartActivity("NewRoot");

    // Re-set the previous Current Activity so the trace isn't messed up
    Activity.Current = previous;
}
```
If not, can we explain what the `source` variable holds?

svrnm:
@open-telemetry/dotnet-approvers please take a look

cijothomas:
Look like that particular doc page is broken.

In the 1st snippet `var serviceName = "MyServiceName";` is done, presumable to use for OTel Resource. But the next snippet `var serviceName = "MyCompany.MyProduct.MyService";` looks like trying to define `ActivitySource` name... But the 3rd snippet `public static readonly ActivitySource MyActivitySource = new(TelemetryConstants.ServiceName);` is defining `ActivitySource` with a name from `TelemetryConstants` class, which is not shown before... There is also usage of `DemoSource` but that is not defined either.

I think the following should get what you are looking for 
`var newRoot = MyActivitySource.StartActivity("NewRoot");`

(The name of `ActivitySource` used to create Span/Activity must be added to the `TracerProvider` via `AddSource()` calls. This is the key thing.)

Please check if this helps. The doc fixes might take some time.

cartermp:
Yes, for this specifically it should just use `MyActivitySource`

cartermp:
Fixed in https://github.com/open-telemetry/opentelemetry.io/pull/3250

PR #3250, opened 2023-09-11
Update .NET manual instrumentation root activity creation

cartermp:
Use `MyActivitySource` instead of `source`, the latter of which was missed as the wrong thing at some point

fixes https://github.com/open-telemetry/opentelemetry.io/issues/3247

### DIFF
--- a/content/en/docs/instrumentation/net/manual.md
+++ b/content/en/docs/instrumentation/net/manual.md
@@ -260,7 +260,7 @@ public static void DoWork()
     var previous = Activity.Current;
     Activity.Current = null;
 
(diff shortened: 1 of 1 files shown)


cijothomas (approved):
LGTM.

Also there are few other issues in this doc as noted [here](https://github.com/open-telemetry/opentelemetry.io/issues/3247#issuecomment-1714034923). For another time!

cartermp:
Yeah, agreed. We are seeing about standardizing each manual instrumentation doc and this can fit in there maybe? I'll file a separate tracking issue.

PR #3255, opened 2023-09-11
Rename javaagent to java agent in prose

cartermp:
See https://github.com/open-telemetry/opentelemetry.io/issues/1661#event-10336576662

I didn't touch blogs because they're more of a historical artifact.

Config stuff is all the same.

### DIFF
--- a/content/en/docs/instrumentation/java/automatic/agent-config.md
+++ b/content/en/docs/instrumentation/java/automatic/agent-config.md
@@ -78,22 +78,22 @@ extension.
 
 {{% /config_option %}}
 
-### Javaagent logging output
+### Java agent logging output
 
 The agent's logging output can be configured by setting the following property:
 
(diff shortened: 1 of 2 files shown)


content/en/docs/instrumentation/java/automatic/agent-config.md
```diff
@@ -78,22 +78,22 @@ extension.
...
 
-### Javaagent logging output
+### Java agent logging output
 
 The agent's logging output can be configured by setting the following property:
 
 {{% config_option name="otel.javaagent.logging" %}}
 
-The javaagent logging mode. The following 3 modes are supported:
+The java agent logging mode. The following 3 modes are supported:
```
mateuszrzeszutek, line 87: Should we use the uppercase Java?

breedx-splk (approved):
Assuming we get `Java` (capitalized) looks good otherwise.

PR #3279, opened 2023-09-15
update references from logging exporter to debug exporter

codeboten:
The logging exporter is being deprecated in favour of the debug exporter. This changes references to the logging exporter to use the new debug exporter instead.

Note: this will remain in draft until v0.86.0 is released which will be the first release to include the debug exporter.


theletterf (approved):
LGTM

svrnm:
Should we add a note somewhere that this change is happening with 0.86, so ppl using collectors before that need `logging` and afterwards `debug`? We could have it similar to the `docker-compose` vs `docker compose` note that we have a centralised footnote (see https://github.com/open-telemetry/opentelemetry.io/blob/main/hugo.yaml#L170) and then add it at the places where the change is applied. After a certain period (3 months, 6 months, 12 months, etc.) we remove it once again

codeboten:
> Should we add a note somewhere that this change is happening with 0.86, so ppl using collectors before that need logging and afterwards debug?

The collector logs a warning message for any users that continue using the logging exporter. The plan is to keep that component in a deprecated state for at least one year to give users time to adjust. I don't know if an additional message in the documentation is necessary.

djaglowski (approved):

content-modules/opentelemetry-specification
svrnm, line 1: make sure that your PR does not touch any submodules

jpkrohling:
> Should we add a note somewhere that this change is happening with 0.86, so ppl using collectors before that need logging 

I think it's important to add something like this. Quite a few users do attempt to use new instructions with older versions of the collector for a reason or another. Having it explicit in the docs that the debug exporter exists starting from 0.86.0 will be a good starting point for users who have problems using it with older versions.

svrnm:
> > Should we add a note somewhere that this change is happening with 0.86, so ppl using collectors before that need logging
> 
> I think it's important to add something like this. Quite a few users do attempt to use new instructions with older versions of the collector for a reason or another. Having it explicit in the docs that the debug exporter exists starting from 0.86.0 will be a good starting point for users who have problems using it with older versions.

Apologies, I didn't think about the other way round, you are completely right! Let's raise an issue for that

jaronoff97 (approved):

content-modules/opentelemetry-specification
codeboten: i definitely did not intentionally modify a submodule :) will revert this part of the change 👍🏻

codeboten: reverted

cartermp (approved):

PR #3318, opened 2023-09-27
Refine Go: Sampling

pellared:
1. Add hyperlinks to pkg.go.dev
2. Describe what is the default sampling


pellared:
CC @theletterf

content/en/docs/instrumentation/go/sampling.md
```diff
@@ -1,36 +1,43 @@
...
 a system. The exact sampler you should use depends on your specific needs, but
 in general you should make a decision at the start of a trace, and allow the
 sampling decision to propagate to other services.
 
-A sampler needs to be set on the tracer provider when its configured, as
-follows:
+A [`Sampler`](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/trace#Sampler) can
+to be set on the tracer provider when register it using the
+[`WithSampler`](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/trace#WithSampler)
+option, as follows:
```
theletterf, line 14:
```suggestion
When you registar a [`Sampler`](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/trace#Sampler) 
you can set it to the tracer provider using the
[`WithSampler`](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/trace#WithSampler)
option as follows:
```

content/en/docs/instrumentation/go/sampling.md
```diff
@@ -1,36 +1,43 @@
...
-`AlwaysSample` and `NeverSample` are fairly self-explanatory. Always means that
-every trace will be sampled, the converse holds as true for Never. When you're
-getting started, or in a development environment, you'll almost always want to
-use `AlwaysSample`.
+[`AlwaysSample`](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/trace#AlwaysSample)
+and
+[`NeverSample`](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/trace#NeverSample)
+are fairly self-explanatory. Always means that every trace will be sampled, the
+converse holds as true for Never. When you're getting started, or in a
+development environment, you'll almost always want to use `AlwaysSample`.
```
theletterf, line 27:
```suggestion
are self-explanatory values. `AlwaysSample` means that every trace is
sampled, while `NeverSample` means that no trace is sampled. When
you're getting started, or in a development environment, use
`AlwaysSample`.
```

theletterf: Are there production recommended values? When one would use `NeverSample`?

content/en/docs/instrumentation/go/sampling.md
```diff
@@ -1,36 +1,43 @@
...
-- `ParentBased`, which behaves differently based on the incoming sampling
-  decision. In general, this will sample spans that have parents that were
-  sampled, and will not sample spans whose parents were _not_ sampled.
+- [`TraceIDRatioBased`](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/trace#TraceIDRatioBased),
+  which will sample a fraction of traces, based on the fraction given to the
+  sampler. Thus, if you set this to .5, half of traces will be sampled.
+- [`ParentBased`](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/trace#ParentBased),
+  which behaves differently based on the incoming sampling decision. In general,
+  this will sample spans that have parents that were sampled, and will not
+  sample spans whose parents were _not_ sampled.
```
theletterf, line 37:
```suggestion
- [`TraceIDRatioBased`](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/trace#TraceIDRatioBased),
  which samples a fraction of traces, based on the fraction given to the
  sampler. If you set .5, half of all the traces are sampled.
- [`ParentBased`](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/trace#ParentBased),
  which behaves differently based on the incoming sampling decision. `ParentBased`
  samples spans that have parents that were sampled, and doesn't sample spans
  whose parents were not sampled.
```

content/en/docs/instrumentation/go/sampling.md
```diff
@@ -1,36 +1,43 @@
...
 
-When you're in production, you should consider using the `TraceIDRatioBased`
-sampler with the `ParentBased` sampler.
+By default, the tracer provider uses a `ParentBased` sampler with the
+`AlwaysSample`.
```
theletterf, line 40:
```suggestion
`AlwaysSample` setting.
```

content/en/docs/instrumentation/go/sampling.md
```diff
@@ -1,36 +1,43 @@
...
+
+When you're in production, you should consider using the `ParentBased` sampler
```
theletterf, line 42:
```suggestion
When in a production environment, consider using the `ParentBased` sampler
```

theletterf (changes requested):
Thanks! Some changes requested.

content/en/docs/instrumentation/go/sampling.md
pellared: E.g. one someone would like to have it configurable.

content/en/docs/instrumentation/go/sampling.md
pellared: I do not understand this change.

If you want to set a sampler then you have to do it via `WithSampler`

I see that my sentence was a total mess. Changing to

```suggestion
A [`Sampler`](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/trace#Sampler) can
be set on the tracer provider using the
[`WithSampler`](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/trace#WithSampler)
option, as follows:
```

content/en/docs/instrumentation/go/sampling.md
theletterf: Could you add a sentence on when would `NeverSample` be useful?

content/en/docs/instrumentation/go/sampling.md
pellared:
```suggestion
`AlwaysSample` sampler.
```

theletterf (approved):

content/en/docs/instrumentation/go/sampling.md
```diff
@@ -1,36 +1,43 @@
...
-  will be sampled.
-- `ParentBased`, which behaves differently based on the incoming sampling
-  decision. In general, this will sample spans that have parents that were
-  sampled, and will not sample spans whose parents were _not_ sampled.
+- [`TraceIDRatioBased`](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/trace#TraceIDRatioBased),
+  which samples a fraction of traces, based on the fraction given to the
+  sampler. If you set .5, half of all the traces are sampled.
+- [`ParentBased`](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/trace#ParentBased),
+  which behaves differently based on the incoming sampling decision.
+  `ParentBased` samples spans that have parents that were sampled, and doesn't
```
dashpole, line 36: nit: can we be consistent about whether the "trace" or the "span" is sampled?  Generally, I would actually prefer "spans" being sampled, but we appear to use "trace is sampled" more often.

content/en/docs/instrumentation/go/sampling.md
```diff
@@ -1,36 +1,43 @@
...
-- `TraceIDRatioBased`, which will sample a fraction of traces, based on the
-  fraction given to the sampler. Thus, if you set this to .5, half of traces
-  will be sampled.
-- `ParentBased`, which behaves differently based on the incoming sampling
-  decision. In general, this will sample spans that have parents that were
-  sampled, and will not sample spans whose parents were _not_ sampled.
+- [`TraceIDRatioBased`](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/trace#TraceIDRatioBased),
+  which samples a fraction of traces, based on the fraction given to the
+  sampler. If you set .5, half of all the traces are sampled.
+- [`ParentBased`](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/trace#ParentBased),
```
dashpole, line 34: Should ParentBased include something about delegating to another sampler?

content/en/docs/instrumentation/go/sampling.md
pellared: https://github.com/open-telemetry/opentelemetry.io/pull/3318/commits/7d941581f6bbfcb3d8247ffac3c2453f96343a01

content/en/docs/instrumentation/go/sampling.md
pellared: PTAL 5847bd3c

content/en/docs/instrumentation/go/sampling.md
```diff
@@ -1,36 +1,45 @@
...
 weight: 80
-cSpell:ignore: sdktrace
 ---
 
-Sampling is a process that restricts the amount of traces that are generated by
-a system. The exact sampler you should use depends on your specific needs, but
-in general you should make a decision at the start of a trace, and allow the
+Sampling is a process that restricts the amount of spans that are generated by a
+system. The exact sampler you should use depends on your specific needs, but in
+general you should make a decision at the start of a span, and allow the
```
dashpole, line 8: "start of a trace" was correct here. Alternative: "first span in a trace"

content/en/docs/instrumentation/go/sampling.md
```diff
@@ -1,36 +1,45 @@
 ---
 title: Sampling
 weight: 80
-cSpell:ignore: sdktrace
 ---
 
-Sampling is a process that restricts the amount of traces that are generated by
-a system. The exact sampler you should use depends on your specific needs, but
-in general you should make a decision at the start of a trace, and allow the
+Sampling is a process that restricts the amount of spans that are generated by a
```
cartermp, line 6: Recommend linking to the concept page, `/docs/concepts/sampling/`

content/en/docs/instrumentation/go/sampling.md
```diff
@@ -1,36 +1,45 @@
...
-`AlwaysSample` and `NeverSample` are fairly self-explanatory. Always means that
-every trace will be sampled, the converse holds as true for Never. When you're
-getting started, or in a development environment, you'll almost always want to
-use `AlwaysSample`.
+[`AlwaysSample`](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/trace#AlwaysSample)
+and
+[`NeverSample`](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/trace#NeverSample)
+are self-explanatory values. `AlwaysSample` means that every span is sampled,
+while `NeverSample` means that no span is sampled. When you're getting started,
+or in a development environment, use `AlwaysSample`.
```
cartermp, line 27: Since this is the default, I think it's worth calling out that you don't need to specify the sampler unless you're already configuring a head sampler and want to change it for development.

content/en/docs/instrumentation/go/sampling.md
```diff
@@ -1,36 +1,45 @@
...
 Other samplers include:
 
-- `TraceIDRatioBased`, which will sample a fraction of traces, based on the
-  fraction given to the sampler. Thus, if you set this to .5, half of traces
-  will be sampled.
-- `ParentBased`, which behaves differently based on the incoming sampling
-  decision. In general, this will sample spans that have parents that were
-  sampled, and will not sample spans whose parents were _not_ sampled.
+- [`TraceIDRatioBased`](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/trace#TraceIDRatioBased),
+  which samples a fraction of spans, based on the fraction given to the sampler.
```
cartermp, line 32: Mmmm, I think we still do want to say traces, or at least clarify what's happening here. Yes, it will sample spans, but it's really about sampling traces at  given rate at the head. For example, if you emit 100 traces each request and specify a rate of 0.5, you won't end up with 50 missing spans in each trace, but you will end up with 50% of the total span volume.

content/en/docs/instrumentation/go/sampling.md
```diff
@@ -1,36 +1,45 @@
...
+  samples spans that have parents that were sampled, and doesn't sample spans
+  whose parents were not sampled.
 
-When you're in production, you should consider using the `TraceIDRatioBased`
-sampler with the `ParentBased` sampler.
+By default, the tracer provider uses a `ParentBased` sampler with the
+`AlwaysSample` sampler.
+
+When in a production environment, consider using the `ParentBased` sampler with
+the `TraceIDRatioBased` sampler.
```
cartermp, line 45: I'm thinking we may want to revise this, since it's a little complicated.

If you combine both, then yes, it will typically behave how you like. However, if your application is called by an upstream one you don't necessarily control, then it can be advantageous to omit the `ParentBased` composite sampler. The result is that you can sometimes end up with traces that have missing root spans, but the rest of your traces will be coherent and at the volume you care about. We've run into this several times with customers who do head sampling and it unfortunately means they have to pick one or the other. We've tried to get them to switch to tail-based sampling but they often can't.

I'm thinking of rolling that guidance out across all the languages, but if you agree, would you want to incorporate here, or handle orthogonally?

cartermp (changes requested):
Some minor comments, but otherwise this looks great.

I think we should also document setting the environment variable in a top-level section in this guide. That can be especially helpful for development since you can just use the default, but then have production do something different.

content/en/docs/instrumentation/go/sampling.md
pellared: The default is different and described at the end.

content/en/docs/instrumentation/go/sampling.md
pellared: > dif you emit 100 traces each request and specify a rate of 0.5, you won't end up with 50 missing spans in each trace

I am not sure if you are correct. I think this will be correct if you combine `TraceIDRatioBased` and `ParentBased`.

pellared: Also see https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#traceidratiobased

content/en/docs/instrumentation/go/sampling.md
pellared: I am not sure if I follow. The `ParentBased` is configurable regarding local vs remote parent spans. See the available options https://pkg.go.dev/go.opentelemetry.io/otel/sdk/trace#ParentBasedSamplerOption

> I'm thinking of rolling that guidance out across all the languages, but if you agree, would you want to incorporate here, or handle orthogonally?

We can try incorporating it here. If we will have problems doing so then we can handle it orthogonally.

pellared:
@cartermp Are your comments a blocker? My original intent of this PR was just make a little improvement (editorial changes and describe what is the default sampling). We can make an issue to address your comments and you assign me there.

content/en/docs/instrumentation/go/sampling.md
pellared: https://github.com/open-telemetry/opentelemetry.io/pull/3318/commits/dc2a8087756660335c32a696436bbe6836a466b5

pellared:
> I think we should also document setting the environment variable in a top-level section in this guide.

I created https://github.com/open-telemetry/opentelemetry.io/issues/3325 with necessary references.

pellared:
@cartermp can please you take a look again?

content/en/docs/instrumentation/go/sampling.md
cartermp: Okay, sounds good. We can tackle separately.

cartermp (approved):

cartermp:
Thanks!

PR #3334, opened 2023-09-29
Update adopters.yaml

unix4you2:


theletterf (approved):

theletterf:
@unix4you2 Can you update with changes from `main`?

data/ecosystem/adopters.yaml
```diff
@@ -99,3 +99,9 @@
   components: [Ruby]
   reference: ''
   contact: ''
+- name: Practico Framework
+  url: https://practico.org/
+  components: [Collector, PHP]
+  reference: https://www.practico.org/documentaci%C3%B3n
```
svrnm, line 105: @unix4you2 happy to have Practico Framework listed, but for me it's hard to identify in your docs where OpenTelemetry is mentioned to be adopted? Also if this is an "adoption" in the sense of otel being integrated into practico this belongs to "integrations.yaml" (those terms can easily be confused, we still have to figure out a better way to distinguish these.)

theletterf: I couldn't find references to OTel in the source either... https://github.com/search?q=repo%3Aunix4you2%2Fpractico%20opentelemetry&type=code

theletterf (changes requested):
Please address @svrnm comment.

data/ecosystem/adopters.yaml
unix4you2: Hi Severin,

Practico is a RAD and low code application designer for PHP so many of its features, final code and application components are dinamically generated from the XML definitions (even the config files for the project). 
https://github.com/unix4you2/practico/tree/master/xml.   Visually you can't see code because all the components are XML that will be displayed in each execution.

So please tell me if we should mention it in another place for people that are interested in use it with opentelemetry.  We could try to be it more visible but maybe need to expose some code out of our XML.

Best regards!

svrnm: @unix4you2 there is no need to expose code, what we are looking for is a clear pointer in your documentation that states that Practico (a) supports OpenTelemetry and (b) how end users can set it up (even if that is done by any low code mechanisms). As @theletterf said there is no findable reference for OpenTelemetry on your website, and without that we can not link you.

cartermp:
Closing due to inactivity

Issue #2829, opened 2023-06-07
Common weights/structure for language-specific pages


svrnm:
As discussed during the SIG meeting 05/22/23 we want to have a common structure for all language navigations:

  - 10 Getting Started
  - 20 Automatic
    - 10 Configuration
    - 20 Extensions
    - ...
  - 30 Manual
  - 40 Libraries
  - 50 Exporters
  - 60 (Context) Propagation
  - 70 Resources
  - 80 Sampling
  - 100-200 *language specific docs*
  - 210 API (external link)
  - 220 Examples (external link)

I added the weights after the discussion to make it consistent through that

this also is going to fix #2431 

TODO:
- [x] #2831
- [x] #2830
- [x] #2832
- [x] Go:
  - [x] https://github.com/open-telemetry/opentelemetry-go/pull/4211
  - [x] #2849
- [x] #2833
- [x] #2834
- [x] #2835
- [x] #2836
- [x] #2837
- [x] Rust (not applicable)
- [x] #2838

svrnm:
> My only suggestion would be to start language-specific pages at 100 instead of 90. Otherwise, great initiative, thanks!

good call, let me fix some of my PRs then ;-)

svrnm:
@chalin , what do we need to do to get the latest updates for the go pages, since https://github.com/open-telemetry/opentelemetry-go/pull/4211 got merged?

chalin:
@svrnm - refresh the Go submodule by running this command:

```sh
npm run update:submodule -- content-modules/opentelemetry-go
```

and submit the update as a PR.

chalin:
Closed by #2849

PR #3342, opened 2023-10-03
Add page for JS benchmarks

martinkuba:
The Javascript SDK now has microbenchmark tests that are automatically run on every push to main. The results are currently published at https://open-telemetry.github.io/opentelemetry-js/benchmarks/.

This PR adds a page here with the same results. This is similar to what has already been done for the Collector load tests.
https://opentelemetry.io/docs/collector/benchmarks/

### DIFF
--- a/content/en/docs/instrumentation/js/benchmarks.md
+++ b/content/en/docs/instrumentation/js/benchmarks.md
@@ -1,0 +1,255 @@
+---
+title: Benchmarks
+weight: 99
+---
+
+ <style>
+
+    main {
+      margin: 8px;
+      width: 100%;
+      display: flex;
+      flex-direction: column;
+    }
+
+    button {
+      color: #fff;
+      background-color: #3298dc;
+      border-color: transparent;
+      cursor: pointer;
+      text-align: center;
+    }
+
+    button:hover {
+      background-color: #2793da;
+      flex: none;
+    }
+
+    .spacer {
+      flex: auto;
+    }
+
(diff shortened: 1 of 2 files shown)


cartersocha:
@chalin could you take a look at the errors please?

cartermp:
In this case you can look at the logs to see what failed, an `npm run fix` should take care of it.

martinkuba:
@cartermp It looks like the spelling is failing on words that are part of the JS code. Is there a way to bypass that?

cartermp:
Hrm. That should have fixed the spelling thing. Ugh, sorry.

In the yaml metadata at the top of the file, you can add:

```
cSpell:ignore: rrggbbaa Elems
```

content/en/docs/instrumentation/js/benchmarks.md
```diff
@@ -0,0 +1,257 @@
+---
+title: Benchmarks
+weight: 99
```
svrnm, line 3: Can you do a weight above 100? (see https://github.com/open-telemetry/opentelemetry.io/issues/2829), serverless has 100 right now, so either swap it or pick 101 or 110

svrnm (approved):
lgtm, only a minor change request.

Since this is the second benchmark page we should consider moving the CSS to a central place for that (maybe that also works for the JS piece?), not necessarily a part of this PR, but let's have a follow up issue

svrnm (approved):

Issue #2623, opened 2023-04-24
[Meta] Make the "Roll the dice" application the default sample application


svrnm:
The [Python Getting Started](https://opentelemetry.io/docs/instrumentation/python/getting-started/) has a very simple but great sample application, where instead of just giving back a "hello world", it rolls a virtual dice and returns the value of that. I think we should use that as a sample for ALL of our getting started guides, because later in the manual instrumentation it can be used to add some custom stuff easily:

* Custom span attribute for the number rolled
* Some custom metrics for the numbers rolled individually & totally
* We can add logs (see java example) easily, which for example is not given with the grpc sample

I created 2 updated version for Node.JS & Java:

https://github.com/open-telemetry/opentelemetry.io/pull/2624
https://github.com/open-telemetry/opentelemetry.io/pull/2625

For Java the additional advantage is that we can show logs, which is not available for the gRPC sample used 

Background:
This is inspired by the work we did for the talk we did during KubeCon EU (see the sample app [here](https://github.com/pavolloffay/kubecon-eu-2023-opentelemetry-kubernetes-tutorial/tree/main/app)) and I wanted to give some of that back to the docs. We might even consider to build on top of that in the Operator Docs, but that's another discussion)

-- 

Tracking:

- [x] C++
- [x] .NET
- [x] [Erlang/Elixir](https://github.com/open-telemetry/opentelemetry.io/pull/2685)
- [x] Go
- [x] [Java](https://github.com/open-telemetry/opentelemetry.io/pull/2625)
- [x] [JavaScript](https://github.com/open-telemetry/opentelemetry.io/pull/2624)
- [x] [PHP](#2758)
- [x] Python
- [x] [Ruby](https://github.com/open-telemetry/opentelemetry.io/pull/2655)
- [x] [Rust](https://github.com/open-telemetry/opentelemetry.io/pull/3123)
- [x] Swift

austinlparker:
sgtm. ++ consistency

cartermp:
Yep let's do that

tsloughter:
+1 I already implemented this and planned to introduce it into the Erlang/Elixir docs soon. Had been meaning to check if there was desire to make it the default used by all languages :)

tsloughter:
Including a client would be nice. It may be too much for the docs but to have in the examples for each language would at least be useful.

svrnm:
> Including a client would be nice. It may be too much for the docs but to have in the examples for each language would at least be useful.

[...]

PR #3344, opened 2023-10-04
Add basic getting started for swift

svrnm:
contributes to #2623

I have not a lot of experience with Swift, so I hope what I wrote makes sense to a proper swift developer. The purpose of this is not to be all neat & tidy, but quick and easy to follow, such that people can continue their journey from there.

### DIFF
--- a/content/en/docs/instrumentation/swift/getting-started.md
+++ b/content/en/docs/instrumentation/swift/getting-started.md
@@ -1,0 +1,207 @@
+---
+title: Getting Started
+description: Get telemetry for your app in less than 5 minutes!
+cSpell:ignore: rolldice
+weight: 10
+---
+
+This page will show you how to get started with OpenTelemetry in Swift.
+
+You will learn how you can instrument a simple application, in such a way that
+[traces][] are emitted to the console.
+
+## Prerequisites
+
+Ensure that you have the following installed locally:
+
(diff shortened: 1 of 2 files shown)


vvydier (approved):
I tried the instructions and was able to get the Webserver and the otel StdoutExporter set up in a few minutes. Its a good addition to the docs

svrnm:
> I tried the instructions and was able to get the Webserver and the otel StdoutExporter set up in a few minutes. Its a good addition to the docs

Thank you for taking a look @vvydier

cartermp (approved):
So I think this is good, but my understanding is that Swift is primarily used for iOS apps instead of server apps. I don't want to block on this because of that, but I think we'd be remiss not to have an iOS getting started scenario.

But that's a whole new kind of getting started app.

svrnm:
> So I think this is good, but my understanding is that Swift is primarily used for iOS apps instead of server apps. I don't want to block on this because of that, but I think we'd be remiss not to have an iOS getting started scenario.
> 
> But that's a whole new kind of getting started app.

Based on what I see with Vapor and some other frameworks, server side swift is a thing on it's own. I think what we eventually should have is 2 Getting Started docs like we have it in JavaScript (Node.JS & Browser, here: Server & Client ?)

PR #3365, opened 2023-10-11
Add concept page for resources

svrnm:
Adds a concept page explaining what resources are, what they are for and on a highlevel how they can be used.

It also adds a common intro text on resources that can now be used across languages (+special treatment for erlang:-) )

I also applied some weight changes to make them closer to what we have defined for language pages (see https://github.com/open-telemetry/opentelemetry.io/issues/2829)


content/en/docs/concepts/components.md
```diff
@@ -88,15 +88,18 @@ For more information, see
 
 ### Resource Detectors
 
-A resource represents the entity producing telemetry as resource attributes. For
-example, a process producing telemetry that is running in a container on
-Kubernetes has a Pod name, a namespace, and possibly a deployment name. All
-three of these attributes can be included in the resource.
+A [resource](/docs/concepts/resources/) represents the entity producing
```
cartermp, line 91: hrm. This page doesn't exist?
image

svrnm: oh... it looks like I missed to add that page to my PR, odd...

svrnm:
```
$ git add content/en/docs/concepts/resources/_index.md
The following paths are ignored by one of your .gitignore files:
content/en/docs/concepts/resources
hint: Use -f if you really want to add them.
hint: Turn this message off by running
hint: "git config advice.addIgnoredFile false"
$ head .gitignore
# Hugo-generated assets
.hugo_build.lock
/public
resources/
```

content/en/docs/instrumentation/erlang/resources.md
```diff
@@ -9,16 +9,7 @@ cSpell:ignore: behaviour
...
-producing telemetry as attributes. For example, an OTP Release producing
-telemetry that is running in a container on Kubernetes has an OTP Release name,
-a Pod name, a namespace, and possibly a deployment name. All four of these
-attributes can be included in the resource.
-
-In your observability backend, you can use resource information to better
-investigate interesting behavior. For example, if your trace or metrics data
-indicate latency in your system, you can narrow it down to a specific container,
-pod, or Kubernetes deployment.
+{{% docs/instrumentation/resources-intro erlang %}}
```
ferd, line 12: As an Erlang/Elixir Reviewer here, I guess this is okay, but it certainly is weird that the Erlang-specific term ("an OTP Release") ends up living in the generic-side of the documentation rather than being passed as another optional term—I notice that Go and Java docs don't need to pass the programming language in there either and there are no problems with that.

If you were to truly extract this bit to share it elsewhere, I'd argue you want to keep the language-specific terms on the caller side, and the OTel-general terms on the template's side.

content/en/docs/concepts/components.md
cartermp: oh shit, well that's fine. Does it really need to be a subfolder though? I think just a markdown file for now is good enough

content/en/docs/instrumentation/erlang/resources.md
svrnm: fair point, it felt odd doing it the way I did it here, so let me fix that.

content/en/docs/concepts/components.md
svrnm: There's an image, that's why I need a folder, I updated the `.gitignore` to only apply to top-level resources/ folder, hope that's not going to be a problem in the future

ferd (approved):

MrAlias (approved):

content/en/docs/concepts/components.md
svrnm: fixed, PTAL

content/en/docs/concepts/resources/_index.md
```diff
@@ -0,0 +1,58 @@
...
+
+If you use [Jaeger](https://www.jaegertracing.io/) as your observability
+backend, resource attributes are grouped under the **Process** tab:
+
+![A screenshot from Jaeger showing an example output of resource attributes associated to a trace](screenshot-jaeger-resources.png)
+
+In practice a resource will be associated with the `TraceProvider` or
+`MetricProvider` when they are created. The association can not be changed
+later. When associated, all metrics and spans produced from a `Tracer` or
+`Meter` from the provider will have the resource associated with them.
```
cartermp, line 18:
```suggestion
A resource is added to the `TraceProvider` or `MetricProvider`
when they are created during initialization. This association can
not be changed later. After a resource is added, all spans and
metrics produced from a `Tracer` or `Meter` from the provider
will have the resource associated with them.
```

content/en/docs/concepts/resources/_index.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Semantic Attributes with SDK-provided Default Value
+
+There are attributes which must be provided by the OpenTelemetry SDK. One of
+them is the `service.name`, which represents the logical name of the service.
+The SDK will provide a default value for this attribute, but it is recommended
+to set it explicitly, either in code or via populating the environment variable
+`OTEL_SERVICE_NAME`
```
cartermp, line 26:
```suggestion
There are attributes provided by the OpenTelemetry SDK. One of
them is the `service.name`, which represents the logical name of the service.
By default, SDKs will assign the value `unknown_service` for this value, so
it is recommended to set it explicitly, either in code or via setting the
environment variable `OTEL_SERVICE_NAME`.
```

content/en/docs/concepts/resources/_index.md
```diff
@@ -0,0 +1,58 @@
...
+
+Additionally, the SDK will also provide a resource with attributes
+`telemetry.sdk.name`, `telemetry.sdk.language` and `telemetry.sdk.version` to
+identify itself.
```
cartermp, line 30:
```suggestion
Additionally, the SDK will also provides the following resource attributes to
identify itself: `telemetry.sdk.name`, `telemetry.sdk.language` and
`telemetry.sdk.version`.
```

cartermp (changes requested):
Overall looks good, but some suggestions in the main concept doc

svrnm:
> Overall looks good, but some suggestions in the main concept doc

thanks, suggestions applied

Kielek (approved):

hanyuancheung (approved):

cartermp (approved):
LGTM, just need to appease the CI robots now

PR #3366, opened 2023-10-11
Add more subsections for Go Getting Started

cartermp:
This breaks things up a bit so the getting-started guide is a little more amenable to people coming in with an existing app, looking to take things up piecemeal.

### DIFF
--- a/content/en/docs/instrumentation/go/getting-started.md
+++ b/content/en/docs/instrumentation/go/getting-started.md
@@ -26,7 +26,7 @@ complete list of libraries for supported frameworks, see the
 
 For more elaborate examples, see [examples](/docs/instrumentation/go/examples/).
 
-## Installation
+### Setup
 
 To begin, set up a `go.mod` in a new directory:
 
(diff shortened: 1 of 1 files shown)


content/en/docs/instrumentation/go/getting-started.md
```diff
@@ -79,14 +79,44 @@ func rolldice(w http.ResponseWriter, r *http.Request) {
...
+  "go.opentelemetry.io/otel/sdk/trace" \
+  "go.opentelemetry.io/otel/semconv/v1.21.0" \
+  "go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
+```
+
+This installs OpenTelemetry SDK components and `net/http` instrumentation.
+
+If you're instrumenting a different library for network requests, you'll need to
+install the appropriate instrumentation library. See
+[libraries](/docs/instrumentation/go/libraries/) for more info.
```
pellared, line 114: I am not sure if we need to write it.

`go mod tidy` installs missing Go modules.

I think it is good to minimize the steps if they are not needed.

```suggestion
```

cartermp: Mmm, I think it's better to install up-front. Otherwise highly-used editors like go/vscode won't work well
image

pellared (approved):

pellared:
@cartermp Thanks a lot ❤️

dashpole (approved):

cartermp:
@svrnm would like to get your view here. The overall structure is the same, but I wonder if this kind of subsectioning should also apply to other languages? We kind of do that with Node. I'm also fine with this just being done on a case-by-case basis.

svrnm (approved):

svrnm:
> @svrnm would like to get your view here. The overall structure is the same, but I wonder if this kind of subsectioning should also apply to other languages? We kind of do that with Node. I'm also fine with this just being done on a case-by-case basis.

Yes, I think we should apply that to the other languages as well (as far as possible). Today we do not have a good mechanism to keep those pages in sync and I don't think we are ready yet to turn all of that in some template pages and then only keep the language specific stuff in place. So +1 for doing this one-by-one as of today.

PR #3374, opened 2023-10-13
(chore) GHA auto-update-versions workflow -- restrict forks

### LINKED ISSUE #3373: Consider not running `auto-update-versions.yml` on forks if possible
See here: https://github.com/cartermp/opentelemetry.io/actions

That's a lotta cpu cycles!

Thanks @puckpuck for the pointer

puckpuck:
Fixes #3373 

Restricts the `auto-update-versions.yml` GitHub Action workflow from running on forks. This workflow is scheduled to run hourly filling Action logs on forks with failed attempts.


.github/workflows/auto-update-versions.yml
```diff
@@ -9,6 +9,8 @@ on:
 jobs:
   auto-update-versions:
     name: Auto-update versions
+    # restrict forks from running this workflow
+    if: github.repository == 'open-telemetry/opentelemetry.io'
```
cartermp, line 13: my lack of yaml-foo -- should the following blocks of code also be indented, or no?

cartermp (approved):

.github/workflows/auto-update-versions.yml
puckpuck: job execution conditions are at the `jobs.<job-id>.if` level so this is proper

https://docs.github.com/en/actions/using-jobs/using-conditions-to-control-job-execution

PR #3292, opened 2023-09-19
Registry Updates September 2023, Part 2: cspell fixes coming

svrnm:
This PR includes multiple changes towards the registry:

* most changes are for including `cspell:ignore` in files where it is needed, this will allow us to enable spell checking on the registry
* remove the go components (like #3291)
* remove other deprecated components (jaeger exporters)
* add new components

PR #3376, opened 2023-10-13
Add missing cspell:ignore to registry entries

svrnm:


chalin (approved):
Yep, was hitting these locally.

chalin:
@svrnm - is this the followup to #3292, or is there more?

PR #3378, opened 2023-10-14
Update vendors.yaml to add ClickHouse

tbragin:
Hi! Adding ClickHouse to OpenTelemetry vendor list, per the vendor docs instructions.

Link to the documentation that details how your offering consumes OpenTelemetry natively via OTLP. 
https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/clickhouseexporter 
https://clickhouse.com/blog/storing-log-data-in-clickhouse-fluent-bit-vector-open-telemetry 
https://clickhouse.com/blog/storing-traces-and-spans-open-telemetry-in-clickhouse 
https://clickhouse.com/use-cases/logging-and-metrics

Link to your distribution, if applicable
N/A

Link that proves that your offering is open source, if applicable. An open source distribution does not qualify your offering to be marked “open source”. 
https://github.com/ClickHouse/ClickHouse/blob/master/LICENSE

GitHub handle or email address as a point of contact so that we can reach out in case we have questions 
@tbragin


data/ecosystem/vendors.yaml
```diff
@@ -321,3 +321,10 @@
   contact: 'info@greptime.com'
   oss: true
   commercial: true
+name: ClickHouse
+  distribution: false
+  nativeOTLP: true
+  url: 'https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/clickhouseexporter'
+  contact: ''
```
svrnm, line 328:
```suggestion
  contact: 'https://github.com/tbragin'
```

data/ecosystem/vendors.yaml
```diff
@@ -321,3 +321,10 @@
   contact: 'info@greptime.com'
   oss: true
   commercial: true
+name: ClickHouse
+  distribution: false
+  nativeOTLP: true
+  url: 'https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/clickhouseexporter'
```
svrnm, line 327: That doesn't fit, if this would be a support for native OTLP you would not need a vendor specific exporter

svrnm (changes requested):

data/ecosystem/vendors.yaml
tbragin: I must have misinterpreted what "native OTLP" means. If it means that we support OTLP protocol in ClickHouse database itself -- then yes, it is not the case, we do require an exporter. Then, we should change it to false.

svrnm: then set it to false please

tbragin: Yes, just wanted to triple check my understanding. I set "nativeOTLP: false".

svrnm (approved):

PR #3380, opened 2023-10-16
Update check-spelling.yml

svrnm:
https://github.com/open-telemetry/opentelemetry.io/pull/3292 allows us to also apply cspell checks on the registry files.

See https://github.com/open-telemetry/opentelemetry.io/pull/3376 for newly introduced cspell issues.


chalin (approved):
Works for me if this is what we want.

PR #3388, opened 2023-10-16
Update to spec 1.26.0 and semconv 1.22.0 at the SAME time

chalin:
- Closes #3387 by superseding it
- Closes #3386 by superseding it
- Updates both submodules at the same time since they are mutually dependent on each other. I suspect that this will probably always be the case but we should confirm with the spec and semconv folks.

**Preview**:

- https://deploy-preview-3388--opentelemetry.netlify.app/docs/specs/
- https://deploy-preview-3388--opentelemetry.netlify.app/schemas/1.22.0/

/cc @svrnm

### DIFF
--- a/.htmltest.yml
+++ b/.htmltest.yml
@@ -11,6 +11,10 @@ IgnoreURLs: # list of regexs of paths or URLs to be ignored
   - ^/community/end-user/feedback-survey/$
   - ^(/docs/migration/)?opencensus/$
 
+  # TODO(@chalin) remove once the website pages have caught up with latest semconv changes:
+  - ^/docs/specs/otel/\w+/semantic_conventions
+  - \#instrument-advice
+
   - ^https://deploy-preview-\d+--opentelemetry.netlify.app/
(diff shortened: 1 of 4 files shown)


chalin:
An OpAMP semconv link is invalid, and is breaking the build (because we have build-time link checking for cross-spec links -- something to be revisited soon). So, atm, we're waiting on:

- https://github.com/open-telemetry/opamp-spec/pull/171

svrnm (approved):
lgtm, I'll approve this now so you can merge it when ever the dependencies are fixed

tigrannajaryan:
> So, atm, we're waiting on:
> 
> * [[editorial] specification: fix semconv link opamp-spec#171](https://github.com/open-telemetry/opamp-spec/pull/171)

Merged.

reyang (approved):

chalin:
So that we can get schema 1.22.0 out, I've disabled some link checks etc. I'll reenable them via:

- #3392

Issue #3393, opened 2023-10-16
Trace, metric, and resource semantic conventions are 404'ing


heidmotron:
While the root for semantic conventions appears to be up:
https://opentelemetry.io/docs/concepts/semantic-conventions/

Trace, metric, and resource semantic conventions now 404:
https://opentelemetry.io/docs/specs/otel/trace/semantic_conventions/
https://opentelemetry.io/docs/specs/otel/metrics/semantic_conventions/
https://opentelemetry.io/docs/specs/otel/resource/semantic_conventions/

svrnm:
Thanks for raising this issue @heidmotron.

The semantic conventions have been moved from the otel spec into their own repository from where we source them, let me fix that. Here are the URLs you are looking for:

https://opentelemetry.io/docs/specs/semconv/general/trace/
https://opentelemetry.io/docs/specs/semconv/general/metrics/
https://opentelemetry.io/docs/specs/semconv/resource/

chalin:
This is a known and temporary issue:

- #3392

I'm starting work on this now.

Issue #3392, opened 2023-10-16
Post OTel-spec and semconv update: address and revert disabled checks


chalin:
- Followup to #3388
- Revert temporary changes:
  - Link checking: ignored local paths to semconv and missing hash
  - Build warning

PR #3395, opened 2023-10-17
Fix broken semantic convention links

svrnm:
addresses #3393, but not yet a final fix: @chalin can we have aliases for pages in the semantic convention such that the old URLs are pointing to the new pages?

### DIFF
--- a/content/en/docs/concepts/semantic-conventions.md
+++ b/content/en/docs/concepts/semantic-conventions.md
@@ -4,13 +4,14 @@ description: Common names for different kinds of operations and data.
 weight: 30
 ---
 
-OpenTelemetry defines Semantic Conventions (sometimes called Semantic
-Attributes) that specify common names for different kinds of operations and
-data. The benefit to using Semantic Conventions is in following a common naming
-scheme that can be standardized across a codebase, libraries, and platforms.
+OpenTelemetry defines [Semantic Conventions](/docs/specs/semconv/general/trace/)
+(sometimes called Semantic Attributes) that specify common names for different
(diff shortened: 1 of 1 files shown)


chalin:
Yes, that is part of a clean fix for #3392. I'd rather address all link fixes together. I'm starting work on that now.

content/en/docs/concepts/semantic-conventions.md
```diff
@@ -4,13 +4,14 @@ description: Common names for different kinds of operations and data.
 weight: 30
 ---
 
-OpenTelemetry defines Semantic Conventions (sometimes called Semantic
-Attributes) that specify common names for different kinds of operations and
-data. The benefit to using Semantic Conventions is in following a common naming
-scheme that can be standardized across a codebase, libraries, and platforms.
+OpenTelemetry defines [Semantic Conventions](/docs/specs/semconv/general/trace/)
```
chalin, line 7:
```suggestion
OpenTelemetry defines [Semantic Conventions](/docs/specs/semconv/)
```

chalin:
Fixing all the broken links is going to be tricky (and a bit lengthy since each needs to be analyzed on a case-by-case basis), and I'm not sure that whether we'll be able to use aliases in all cases. So maybe let's move ahead with this partial fix for now as I work on the others. WDYT @svrnm?

chalin:
Or we can wait for my PR :)

svrnm:
> Or we can wait for my PR :)

Since you have the best overview on all the moving pieces right now, I'll leave it to you to either merge this PR or group it into one you have

PR #3403, opened 2023-10-17
Update kubernetes-deployment.md

aiddroid:
code and docs mismatch, so many same errors in the docs, please don't link to the "main" branch from the docs.

### DIFF
--- a/content/en/docs/demo/kubernetes-deployment.md
+++ b/content/en/docs/demo/kubernetes-deployment.md
@@ -61,7 +61,7 @@ To expose the frontendproxy service use the following command (replace
 `my-otel-demo` with your Helm chart release name accordingly):
 
 ```shell
-kubectl port-forward svc/my-otel-demo-frontendproxy 8080:8080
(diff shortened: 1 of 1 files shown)


julianocosta89 (changes requested):
I don't understand this change.

We are stating in the previous paragraph the following:
> To expose the frontendproxy service use the following command (replace
`my-otel-demo` with your Helm chart release name accordingly)

puckpuck:
I think the issue is the service name prefix when installing via helm gets set to `my-otel-demo` but when installing via kubectl apply it will be `opentelemetry-demo`.

We either need to update how we generate the K8s manifest in the demo, or redo this section in the docs.

svrnm:
> I think the issue is the service name prefix when installing via helm gets set to `my-otel-demo` but when installing via kubectl apply it will be `opentelemetry-demo`.
> 
> We either need to update how we generate the K8s manifest in the demo, or redo this section in the docs.

If the docs need to be redone, let's have a separate issue for that 👍

PR #3430, opened 2023-10-24
chore: add missing ':'s in registry cSpell ignore lists

svrnm:
Not mandatory, but this creates consistency with other `cSpell:ignore:` lists we have

### DIFF
--- a/data/registry/collector-exporter-apiclarity.yml
+++ b/data/registry/collector-exporter-apiclarity.yml
@@ -1,4 +1,4 @@
-# cSpell:ignore apiclarity
+# cSpell:ignore: apiclarity
 title: APIClarity HTTP Exporter
 registryType: exporter
 isThirdParty: true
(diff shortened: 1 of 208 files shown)


theletterf (approved):
If we haven't yet, could we document this in a contribution guide together with other particularities of our docs? If that doc doesn't exist, I can create an issue and work on it.

cartermp (approved):

simi:
I have not seen `ignore:` in docs, shouldn't it be opposite? Remove `:` in cSpell ignore lists? Or is there any reason to keep those in there?

https://cspell.org/configuration/document-settings/

chalin (changes requested):
Catching up on PRs. @simi is right, we don't need a `:`, ... unless we're **encoding** the ignore rule as part of the **Hugo front-matter**. In such cases, we need the colon because `cSpell:ignore` is being used as a YAML key name.

I'd suggest not making these changes.

chalin:
With @cartermp's thumbs up, I'm going to close this for now. @svrnm - feel free to reopen if you'd like to explore this further.

PR #3441, opened 2023-10-26
Fix missing package name in Gradle dependencies for auto-configuration

AlexanderWert:
There's a typo in the documentation regarding the Gradle dependency definition for auto-configuration.


trask (approved):
thx!

svrnm (approved):

PR #3449, opened 2023-10-27
Update vendors.yaml to add TingYun

ale0512:


data/ecosystem/vendors.yaml
```diff
@@ -335,3 +335,10 @@
   contact: 'https://github.com/tbragin'
   oss: true
   commercial: true
+- name: TingYun
+  distribution: true
```
svrnm, line 339: please provide evidence that there is a distribution of OpenTelemetry by TingYun (e.g. a link to a repository or within the documentation), if none is provided, set this to false

ale0512: Thanks for pointing out my mistake

ale0512:
@svrnm Could you please advise on how to update refcache.json, Thanks.

PR #3469, opened 2023-11-01
List available schemas in 404 response to `/schemas`

### LINKED ISSUE #3256: Is there an index for all available schema?
I am writing code that uses these schema and would like to create automation that checks all published schema are supported. To do this I need to be able to dynamically get a list of all published schema.

Ideally, I would be able to make a request to the opentelemetry.io website and get an index of all published schema and their URLs.

Does this already exist, or is it something that would need to be added?

### LINKED ISSUE #3416: Ensure that schema files are served with content-type application/yaml
/cc @open-telemetry/specs-semconv-maintainers

### LINKED ISSUE #3468: `/schemas` is the top 404
Clearly folks are interested in listing the available schemas, because `/schemas` (or `/schemas/`) is the #1 legitimate 404. I think that there should be a schemas index page, and will submit a proposal soon.

@tigrannajaryan had already argued against an index page, see https://github.com/open-telemetry/opentelemetry.io/issues/1144#issuecomment-1041845740. I'd like to open this up for discussion and see if we can find a win-win solution.

Related:

- #3256

/cc @open-telemetry/specs-semconv-maintainers @jsuereth @svrnm @MrAlias

chalin:
- Fixes #3468 by adding an index page, if semconv and spec folks agree to this feature. 
  Note that the list is autogenerated from `content-modules/semantic-conventions/schemas`
- ~Fixes #3256~ moved to #3475
- ~Fixes #3416~ moved to #3474

---

**Preview**: https://deploy-preview-3469--opentelemetry.netlify.app/schemas/

### Screenshot (mobile)

> image

### DIFF
--- a/content/en/schemas.md
+++ b/content/en/schemas.md
@@ -1,0 +1,17 @@
+---
+title: OpenTelemetry Schemas
+linkTitle: Schemas
+---
+
+{{% blocks/section color="white" %}}
+
+## {{% param title %}}
+
+Available schemas are listed below. To learn more about schemas, see [Telemetry
+Schemas][].
+
+{{% schemas %}}
+
+{{% /blocks/section %}}
+
+[Telemetry Schemas]: /docs/specs/otel/schemas/#opentelemetry-schema
--- a/layouts/index.redirects
+++ b/layouts/index.redirects
@@ -48,6 +48,7 @@
 {{ $latestSchemaFile := index $schemaFiles 0 -}}
 
 /schemas/latest  /schemas/{{ $latestSchemaFile.Name }}
+/schemas /schemas/ 404!
 
 {{/*
   Social-media image redirects. As mentioned in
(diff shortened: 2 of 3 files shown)


tigrannajaryan:
I think it is a good idea to point `/schemas/latest` to the latest published schema.

I don't think we should publish a human-readable list at `/schemas/`. My previous argument still holds: if we want `/schemas/` to contain anything it should be a machine-readable format defined in Otel spec. Since we don't have such format defined I think we should not publish anything at the index yet.

chalin:
Thanks for your response @tigrannajaryan - see https://github.com/open-telemetry/opentelemetry.io/issues/3468#issuecomment-1789781267 for a clarification of the current situation (`/schemas/` _already_ returns human readable HTML) and two suggested ways for us to have our cake and eat it too 🎂 😄.

chalin:
Converting to draft until we reach a consensus at #3468.

In the meantime, I'm testing the option of returning a 404 status for the schema index page, while still providing the list of schemas in human readable form.

chalin:
Ok, I'm going to split this up so as to not hold back the individual fixes. Here's the first:

- #3474

chalin:
Here's the other feature factored out:

- #3475

I'll repurpose this PR to be only about the index page.

chalin:
Closing in favor of:

- #3537

PR #3474, opened 2023-11-02
Schema content-type fix

### LINKED ISSUE #3416: Ensure that schema files are served with content-type application/yaml

chalin:
- Fixes #3416
- (Was formerly a part of #3469)
- **Content-type** fix confirmed:
  ```console
  $ curl -sI https://deploy-preview-3474--opentelemetry.netlify.app/schemas/1.20.0 | grep type
  content-type: application/yaml
  ```
- **Download tests**: 
  - https://deploy-preview-3474--opentelemetry.netlify.app/schemas/1.20.0

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -14,3 +14,8 @@ force = true
 from = "https://blog.opentelemetry.io/*"
 to = "https://opentelemetry.io/blog/:splat"
 force = true
+
+[[headers]]
+  for = "/schemas/:version"
+  [headers.values]
+    content-type = "application/x-yaml"
--- a/netlify.toml
+++ b/netlify.toml
@@ -18,4 +18,4 @@ force = true
 [[headers]]
   for = "/schemas/:version"
   [headers.values]
-    content-type = "application/x-yaml"
+    content-type = "application/yaml"


netlify.toml
```diff
@@ -13,4 +13,9 @@ force = true
...
 from = "https://blog.opentelemetry.io/*"
 to = "https://opentelemetry.io/blog/:splat"
-force = true
\ No newline at end of file
+force = true
+
+[[headers]]
+  for = "/schemas/:version"
+  [headers.values]
+    content-type = "application/x-yaml"
```
tigrannajaryan, line 21: I think `application/yaml` is the the latest accepted media type.

From https://www.ietf.org/archive/id/draft-ietf-httpapi-yaml-mediatypes-00.html#name-media-type-application-yaml

>Deprecated alias names for this type: application/x-yaml

tigrannajaryan (approved):
LGTM, with just a comment on the media type string to use.

netlify.toml
chalin: I've found a more recent (and non-expired) draft: https://www.ietf.org/archive/id/draft-ietf-httpapi-yaml-mediatypes-10.html, dated 2023/08/30. This is still a draft, but I'll switch to `application/yaml` as you suggest. Thanks for catching that!

Issue #3458, opened 2023-10-31
CI/infra improvements - 2023/11


chalin:
- [x] #3492
- [x] #3494
- [x] #3495
- [x] #3491
- [x] #3496
- [x] #3499
- [x] Cleanup: #3506
- [x] #3501
- [x] #3503
- [x] #3490
  - [x] #3504
- [x] #3502
- [x] #3519, which supersedes the following:
  - [x] #3522
- [x] #3540
- [x] #3572
- [x] #3563
- [x] #3611
- [x] #3615
- [x] #3623

---

For later?

- #3411 
- #3131
- Carried over from #3218:
  - #3117
  - #3183
  - Consider saving the generated website using the artifact action, and passing that along to the link checker

Other possible contenders for this month:

- #780

Previous / next:

- #3218
- #3738

chalin:
All done. Following through with:

- #3738

PR #3500, opened 2023-11-05
[CI] Convert textlintrc file to YAML, consolidate and clean up config

### LINKED ISSUE #3499: [CI] Textlint: convert rc file to YAML so we can add comments etc
Some of the patterns in the textlintrc file would warrant some explanation. It's easier to use YAML in such a case, since it supports comments. Textlint supports YAML, among other [formats for its config file](https://textlint.github.io/docs/configuring.html).

I'll wait for the currently pending PRs over textlint config to be merged before making this change -- **waiting on the merge of**:

- #3497
- #3496
- #3491

chalin:
- Fixes #3499
- Contributes to #3458
- Consolidates and cleans up some entries, dropping unnecessary ones
- Submits a few fixed files, from issues previously missed by the linter
- **Notes** about the YAML syntax:
  - In general, strings don't require quotes, so we avoid them, unless we need to avoid having part of the string be interpreted as YAML
  - Unlike in JSON strings, we don't need escapes like `\\s`, _unless_ the YAML string is in double quotes.


svrnm (approved):
nice!

PR #3517, opened 2023-11-08
Update table in Observability primer

svrnm:
With HTTP semantic conventions stable I updated the observability primer with a set of example keys&values from there such that we don't have to make sure they are up-to-date anymore.

Also some OTel -> OpenTelemetry, because why not?

### DIFF
--- a/content/en/docs/concepts/observability-primer.md
+++ b/content/en/docs/concepts/observability-primer.md
@@ -2,7 +2,7 @@
 title: Observability Primer
 description: Core observability concepts.
 weight: 9
-cSpell:ignore:  webshop
+cSpell:ignore: webshop
 ---
 
 ## What is Observability?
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/observability-primer.md
```diff
@@ -2,7 +2,7 @@
 title: Observability Primer
 description: Core observability concepts.
 weight: 9
-cSpell:ignore: KHTML
+cSpell:ignore: KHTML webshop
```
chalin, line 5: Hmm. KHTML isn't used in the page anymore AFAICT.
`webshop` shouldn't be necessary since it only appears in code. Can you double check @svrnm?

chalin (commented):
Nice! See inline comment, thx.

cartermp (approved):

content/en/docs/concepts/observability-primer.md
svrnm: fixed, `webshop` is necessary because it's not in a code block but inline code which seems to be checked

Hhffuhvg:
Fork

PR #3527, opened 2023-11-11
Fix example JSON in the docs

jdreesen:
I'm very new to this whole topic and just started reading the docs, but I found an inconsistency in the JSON examples.

I am not quite sure if this is the correct structure, but this is how it is done above in the examples, and the text also reads as if the fields `trace_id` and `span_id` should be nested in `context`.

### DIFF
--- a/content/en/docs/concepts/signals/traces.md
+++ b/content/en/docs/concepts/signals/traces.md
@@ -198,10 +198,12 @@ Sample span:
 
 ```json
 {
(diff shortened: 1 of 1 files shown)


svrnm:
Thank you for raising this issue @jdreesen! You are right it's like that in the other examples. This brings also to my attention that the span attributes need to be updated, but we can have that in another PR.

svrnm (approved):

PR #3529, opened 2023-11-12
Fix typo in port number for OTLP exporter

otienoanyango:
Just a small typo in port number as 413* instead if 431*


chalin (approved):
Thanks!

Issue #3546, opened 2023-11-15
Blog Template instructions incorrect


davidgs:
The page https://github.com/open-telemetry/opentelemetry.io/blob/main/archetypes/blog.md says: 

> [Top-level heading](https://github.com/open-telemetry/opentelemetry.io/issues/new#top-level-heading)
> 
> Top-level headings start at level 2, as shown above.

But there is no top-level heading shown above, so the instruction is unclear, at best.

davidgs:
Blog template page also has the following issues: 

<img width="1059" alt="2023-11-15_08-16-37" src="https://github.com/open-telemetry/opentelemetry.io/assets/2071898/ba730ab8-45fa-41d1-907f-75b61f579de0">

1. That error message is confusing on the template page. 
2. For anyone that is not already familiar with Hugo and the mustache-templates, the line 
`{{ replaceRE "[-_]" " " .Name | title }}` is confusing. Are they to replace that line with the title? 
3. `linkTitle: ADD A SHORT TITLE HERE # Mandatory, make sure that your short title.` is an incomplete sentence. Make sure that the short title is what? 
4. `author: >- ` do we replace this `>-` with the author's name? Are those characters required? 
5. Including the `TODO: ` in instructions lines leaves it ambiguous as to if the submitter is to replace these? Or if it's a TODO item for the maintainers.

svrnm:
> But there is no top-level heading shown above, so the instruction is unclear, at best.

Above that line is `## Top-level heading` as top-level heading example. All it wants to say is don't use `# Top-level heading` in your blog post since the `title:` is going to be h1

> That error message is confusing on the template page.

There is nothing we can do against that, because this file is not intended to be viewed on GitHub, it should be used at outlined in the README.md: `npx hugo new content/en/blog/2023/short-name-for-post.md` will make a copy of that file for you

> {{ replaceRE "[-_]" " " .Name | title }}

See my comment above, the `npx hugo` will take care of that

> Make sure that the short title is what?

Make sure that your title is short, let me fix that.

> `author: >-` do we replace this `>-` with the author's name? Are those characters required?

No, the authors are listed one by one as shown in the lines below. We very often have blog posts with multiple authors, that's why we use this structure.

> Including the TODO: in instructions lines leaves it ambiguous as to if the submitter is to replace these? 

This is for the submitter. If you run `npx huo new...` this will remain part of your template and leave it to you to fix that

svrnm:
[...]

PR #3548, opened 2023-11-15
Update blog.md with clearer instructions

davidgs:
Clear up some minor confusion with the blog.md


davidgs:
Addresses some of issue #3546

davidgs:
I am also working on getting the EasyCLA problems fixed and have an open JIRA ticket for that.

archetypes/blog.md
```diff
@@ -17,7 +17,7 @@ With contributions from secondary-author-name-1, ..., and secondary-author-n.
 
 ## Top-level heading
 
-Top-level headings start at **level 2**, as shown above.
+Top-level headings start at **level 2**. Your post should not include `# headings` but should instead only use `## headings`.
```
svrnm, line 20:
```suggestion
Top-level headings start at **level 2**. This means, that your post should not include `# headings` for top-level headings but `## headings` instead.
```

The sentence _can_ lead to the assumption that we only accept level 2 headings, but of course they can use level3, level4, etc if they want to.

archetypes/blog.md
```diff
@@ -33,7 +33,7 @@ directory. Put the images into the same directory.
 If you have an image stored at `content/en/{{ .File.Dir }}imagename.png`, you
 can reference them like the following:
 
-![Provide a good image description for improved accessibility](imagename.png)
+`![Provide a good image description for improved accessibility](imagename.png)`
```
svrnm, line 36:
```suggestion
![Provide a good image description for improved accessibility](imagename.png)
```

I disagree with this change, although this will show up as a broken image, the intend is that if someone runs the hugo new command they see how an image will be used.

archetypes/blog.md
```diff
@@ -44,3 +44,7 @@ npm run format
 ```
 
 Happy writing!
+
+**Note:** You can safely ignore the warning at the top of this page.
```
svrnm, line 48:
```suggestion
**Note:** If you view this page with the GitHub file viewer, you can safely ignore the `Error in user YAML` at the top of this page.
```

archetypes/blog.md
```diff
@@ -44,3 +44,7 @@ npm run format
...
+
+<img width="1024" alt="2023-11-15_10-32-38" src="https://github.com/davidgs/opentelemetry.io/assets/2071898/d7da3f54-64df-4120-b4f7-861134eb71d9">
```
svrnm, line 50:
```suggestion
```

svrnm (changes requested):
thanks for taking the time to update this page, I provided a few comments.

archetypes/blog.md
davidgs: Maybe have both, so that users who may be unfamiliar with images in Markdown can see what it would look like?

svrnm: > Maybe have both, so that users who may be unfamiliar with images in Markdown can see what it would look like?

Not sure if I follow. Users can see how it looks like by taking a look at the file & the preview? We can add a dummy image here to _unbreak_ it and make it clearer by that

davidgs:
Thanks for all the attention to this! Much appreciated.

cartermp (approved):

PR #3555, opened 2023-11-16
docs: Update .net framework docs with metrics info

cedstrom:
Took me quite a while to find this information buried on Microsoft's website, so explicitly calling out the package needed for metrics support in legacy .net.

### DIFF
--- a/content/en/docs/instrumentation/net/netframework.md
+++ b/content/en/docs/instrumentation/net/netframework.md
@@ -22,6 +22,8 @@ First, install the following NuGet packages:
 - [OpenTelemetry.Instrumentation.AspNet](https://www.nuget.org/packages/OpenTelemetry.Instrumentation.AspNet)
 - [OpenTelemetry.Exporter.Console](https://www.nuget.org/packages/OpenTelemetry.Exporter.Console)
 
+If you plan to use metrics, you need to add a reference to the ```System.Diagnostics.DiagnosticsSource``` package as well to get access to the backported System.Diagnostics.Metrics namespace.
+
 Next, modify your `Web.Config` file to add a required HttpModule:
 
 ```xml


content/en/docs/instrumentation/net/netframework.md
```diff
@@ -22,6 +22,8 @@ First, install the following NuGet packages:
 - [OpenTelemetry.Instrumentation.AspNet](https://www.nuget.org/packages/OpenTelemetry.Instrumentation.AspNet)
 - [OpenTelemetry.Exporter.Console](https://www.nuget.org/packages/OpenTelemetry.Exporter.Console)
 
+If you plan to use metrics, you need to add a reference to the ```System.Diagnostics.DiagnosticsSource``` package as well to get access to the backported System.Diagnostics.Metrics namespace.
```
cijothomas, line 25: This package is needed irrespective of Metrics are used or not. But also, this package is automatically brought by the other packages, so not required to be installed explicitly.

cijothomas (changes requested):
See https://github.com/open-telemetry/opentelemetry.io/pull/3555/files#r1396175225

svrnm:
Thanks @cedstrom , @open-telemetry/dotnet-approvers please take a look!

cartermp (changes requested):
In addition to @cijothomas comment, to add metrics support here we'd need one of two things:

1. Explicitly adding the meterprovider in the code sample, similar to how the tracerprovider is initialized (preferred, with a code comment)
2. Adding a second code sample that shows how to add the meterprovider

Otherwise, just referencing `System.Diagnostics.DiagnosticsSource` won't activate ASP.NET metrics.

Existing example in the contrib package is here: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main/src/OpenTelemetry.Instrumentation.AspNet#metrics

PR #3355, opened 2023-10-06
Update exporters documentation for JS

svrnm:
Major rework of the [Exporters](https://opentelemetry.io/docs/instrumentation/js/exporters/) page for JavaScript. 

The goal is once again to create a blueprint that then can be reused for other languages.

PR #3558, opened 2023-11-17
Rework the java exporters page

svrnm:
Based on https://github.com/open-telemetry/opentelemetry.io/pull/3355 I updated the java exporters documentation to have a similar structure.


breedx-splk (approved):
I gave it a read-through. Looks good from here. 👍🏻

content/en/docs/instrumentation/java/exporters.md
```diff
@@ -17,13 +17,61 @@ how to setup exporters following the
...
+[configure the collector](/docs/collector/configuration) to send your telemetry
+to your observability backend.
+
+### Dependencies {#otlp-dependencies}
 
-There are multiple OTLP options available, each catering to different use cases.
-For most users, the default artifact will suffice and be the most simple:
+If you want to send telemetry data to an OTLP endpoint (like the
+[OpenTelemetry Collector](#collector-setup), [Jaeger](#jaeger) or
+[Prometheus](#prometheus)), there are multiple OTLP options available, each
```
jack-berg, line 72: Prometheus is a reference in here because it added experimental support for OTLP ingest, right?

content/en/docs/instrumentation/java/exporters.md
```diff
@@ -171,14 +219,240 @@ public class DiceApplication {
...
+use exporters writing telemetry data to the console (stdout).
+
+If you followed the
+[Getting Started](/docs/instrumentation/java/getting-started/) or
+[Manual Instrumentation](/docs/instrumentation/java/manual/) guides, you already
+have the console exporter installed.
+
+The `LoggingSpanExporter`, the `LoggingMetricExporter` and the
+`SystemOutLogRecordExporter` are included in the
+`opentelemetry-exporter-logging` artifact.
```
jack-berg, line 234: Might be good to mirror the comment above about how to specify the logging exporter when using autoconfigure:

> Note, that in the case of exporting via OTLP you do not need to set
`OTEL_TRACES_EXPORTER`, `OTEL_METRICS_EXPORTER` and `OTEL_LOGS_EXPORTER` since
`otlp` is their default value

The value which selects the logging exports is `logging`, or `logging-otlp` for a version which logs the OTLP json encoding of the data (though `logging-otlp` requires the `opentelemetry-exporter-logging-otlp` artifact).

content/en/docs/instrumentation/java/exporters.md
```diff
@@ -171,14 +219,240 @@ public class DiceApplication {
...
   jaegertracing/all-in-one:latest
 ```
+
+Now following the instruction to setup the [OTLP exporters](#otlp-dependencies).
+
+## Prometheus
+
+To send your metric data to [Prometheus](https://prometheus.io/), you can either
+[enable Prometheus' OTLP Receiver](https://prometheus.io/docs/prometheus/latest/feature_flags/#otlp-receiver)
+and use the [OTLP exporter](#otlp) or you can use the `PrometheusExporter`.
```
jack-berg, line 258: The camel case of "PrometheusExporter" suggests its the name of a class in java. We call it the "prometheus exporter", but the class name is `PrometheusHttpServer`. Not sure how you want to reflect that in this document.

jack-berg (commented):
Just left a few comments but looks pretty good. 👍

cartermp (approved):
LGTM pending feedback addressed from @jack-berg

content/en/docs/instrumentation/java/exporters.md
svrnm: We have Prometheus across most other languages as well (had them for a while, not only by me introducing them via these PRs). I think it's more around them being CNCF as well/being a standard metrics endpoint for many

content/en/docs/instrumentation/java/exporters.md
svrnm:
```suggestion
and use the [OTLP exporter](#otlp) or you can use the [`PrometheusHttpServer`](https://javadoc.io/doc/io.opentelemetry/opentelemetry-exporter-prometheus/latest/io/opentelemetry/exporter/prometheus/PrometheusHttpServer.html), a `MetricReader`, that starts an HTTP server that will collect metrics and serialize to Prometheus text format on request.
```

Is it OK to just replace it like this? I used the wording from the API docs.

content/en/docs/instrumentation/java/exporters.md
svrnm:
~~~suggestion
`opentelemetry-exporter-logging` artifact.

If you use
[SDK auto-configuration](/docs/instrumentation/java/manual/#automatic-configuration)
all you need to do is update your environment variables:
```shell
env OTEL_TRACES_EXPORTER=logging OTEL_METRICS_EXPORTER=logging OTEL_LOGS_EXPORTER=logging  java -jar ./build/libs/java-simple.jar
```
~~~

Like this? The `logging-otlp` is an interesting corner case, I am wondering if it should live here OR be better used in the OTLP section since it is helpful there for debugging?

content/en/docs/instrumentation/java/exporters.md
jack-berg: Sorry for the delay. Yeah that looks good.

content/en/docs/instrumentation/java/exporters.md
jack-berg: 👍

> The logging-otlp is an interesting corner case, I am wondering if it should live here OR be better used in the OTLP section since it is helpful there for debugging?

Yeah probably worth omitting logging-otlp for now.

jack-berg (approved):
Looks good with the recommendations from @svrnm.

content/en/docs/instrumentation/java/exporters.md
```diff
@@ -171,14 +219,247 @@ public class DiceApplication {
...
+{{< /tab >}} {{< /tabpane>}}
+
+Update your OpenTelemetry configuration to use the exporter and to send data to
+your Zipkin backend:
+
+```java
+import io.opentelemetry.exporter.zipkin.ZipkinSpanExporter;
+
+SdkTracerProvider sdkTracerProvider = SdkTracerProvider.builder()
+        .addSpanProcessor(BatchSpanProcessor.builder(OZipkinSpanExporter.builder().setEndpoint("http://localhost:9411/api/v2/spans").build()).build())
```
trask, line 411:
```suggestion
        .addSpanProcessor(BatchSpanProcessor.builder(ZipkinSpanExporter.builder().setEndpoint("http://localhost:9411/api/v2/spans").build()).build())
```

trask (approved):
we generally recommend using the autoconfigure API, do you think that could work on this page? (i'm also totally good with this PR as-is and we can always explore converting it to autoconfigure API later)

PR #3622, opened 2023-11-30
fix permission issue with fix:refcache PR action

svrnm:
Fixes the fix:refcache action from failing.


chalin (approved):
Oh, such a simple fix! Thx! ✨

PR #3623, opened 2023-11-30
[CI] Speed up link checking by testing files concurrently

chalin:
Since we're using an external link refcahce, I think that it is safe to run checks concurrently. On my local machines this makes the checks run 60% faster. On GH, it seems to just speed it up by 40%, but still, that's pretty good.


theletterf (commented):
I'm a bit worried about the fact that `TestFilesConcurrently` is still experimental (https://github.com/wjdp/htmltest/blob/master/README.md?plain=1#L168). Sites like DataDog's docs have it switched off. See also: https://github.com/wjdp/htmltest/wiki/Concurrency

chalin:
@theletterf - right, it is experimental. If we didn't have an external link refcache, I wouldn't enable this feature, for the reasons mention in that wiki page you referenced (thanks for that!). But otherwise, I'm not worried, and feel that it's worth a try. We can always drop it if it causes issues. My 2cents ✨

theletterf (approved):
Sounds good! Let's try this out.

PR #3624, opened 2023-11-30
Fix incorrect hyperlink in docs

ekquasar:
There is no demo with manual instrumentation for Python, the link in the demo index points to automatic instrumentation.


svrnm (approved):

theletterf (approved):

cartermp:
@ekquasar unfortunately it looks like your npm format commit was done with an email that isn't covered by the CLA, so we can't merge this. Could you either also sign the CLA from that email, or rebase?

puckpuck:
There is manual instrumentation in the Recommendation service for the demo. You can see a span being created and started [here](https://github.com/open-telemetry/opentelemetry-demo/blob/main/src/recommendationservice/recommendation_server.py#L66). This is also mentioned in the documentation for the demo [here](https://opentelemetry.io/docs/demo/services/recommendation/#create-new-spans)

We should not be doing the changes in this PR.

svrnm:
> > > There is manual instrumentation in the Recommendation service for the demo. You can see a span being created and started [here](https://github.com/open-telemetry/opentelemetry-demo/blob/main/src/recommendationservice/recommendation_server.py?rgh-link-date=2023-12-07T21%3A12%3A42Z#L66). This is also mentioned in the documentation for the demo [here](https://opentelemetry.io/docs/demo/services/recommendation/#create-new-spans)
> > > We should not be doing the changes in this PR.
> > 
> > 
> > Thanks.
> > To clarify that: from the docs it seems both is used (manual & automatic), can/should this be reflected in that table?
> 
> The current docs site does reflect both today. This is the same as the ad service which is Java-based using an auto-instrumentation agent, and manual instrumentation in code.

True, I missed that somehow

svrnm:
Will close this PR for now. If you disagree @ekquasar please follow-up

PR #3629, opened 2023-12-01
Add script to scan cspell ignore lists

svrnm:
@chalin update to you mentioning recently that we should add some terms (like ott) to our global ignore list. 

The script in this PR can help with that, it will collect all ignore lists and then print out the words with the most files having them in their ignore list, here are all with >5:

```
   6 darwin
   6 debugexporter
   6 farfetch
   6 loggingexporter
   6 loglevel
   6 millis
   6 otlpreceiver
   6 struct
   7 autoconfigure
   7 kubelet
   7 myapp
   8 sdktrace
  11 distro
  13 rolldice
```

Note, that I am not saying that we should add all of them, but the output of the script can help to identify those we should take into consideration.

### DIFF
--- a/scripts/cspell-ignore-scanner
+++ b/scripts/cspell-ignore-scanner
@@ -1,0 +1,8 @@
+#!/bin/bash
+#
+# This script scans directories content, layouts, data for files with "cSpell:ignore" in them and then prints how often a certain word has been added to an ignore list
+# This helps to add new words to the .vscode/cspell.json file
+# 
+
+DIRECTOR=${1:-.}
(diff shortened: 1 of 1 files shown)


cartermp (approved):

chalin (approved):
Cool! 👍🏻

scripts/cspell-ignore-scanner
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+#
+# This script scans directories content, layouts, data for files with "cSpell:ignore" in them and then prints how often a certain word has been added to an ignore list
+# This helps to add new words to the .vscode/cspell.json file
+# 
+
+DIRECTOR=${1:-.}
```
chalin, line 7:
```suggestion
DIRECTORY=${1:-.}
```

chalin: Oh, don't you actually want `DIRECTORY` to be empty by default?

PR #3630, opened 2023-12-01
Registry Updates December 2023

svrnm:
Irregular monthly update of registry entries that need to be added or removed.


static/refcache.json
```diff
@@ -2071,6 +2071,10 @@
     "StatusCode": 200,
     "LastSeen": "2023-06-30T08:34:11.155334-04:00"
   },
+  "https://github.com/astaxie/beego%29": {
+    "StatusCode": 404,
+    "LastSeen": "2023-12-01T16:02:49.666565+01:00"
+  },
```
chalin, line 2077: There are a bunch of URLs that result in 404s in the refcache -- btw, all of these URLs seem to end with %29 (`)`). All 404 entries should be removed.

Actually, I think that we should fail link checking when there are such entries.

chalin: > Actually, I think that we should fail link checking when there are such entries.

See:

- #3640

svrnm: Ah... they shouldn't be there, thanks!

chalin (approved):
👍🏻

PR #3637, opened 2023-12-02
Update vendors.yaml to add VictoriaMetrics

hagen1778:
Hi! Adding VictoriaMetrics to OpenTelemetry vendor list, per the vendor docs instructions.

**Link to the documentation that details how your offering consumes OpenTelemetry natively via OTLP.**
VictoriaMetrics supports [metrics format for ingestion](https://docs.victoriametrics.com/#sending-data-via-opentelemetry).

**Link to your distribution, if applicable**
N/A

**Link that proves that your offering is open source, if applicable. An open source distribution does not qualify your offering to be marked “open source”.**
https://github.com/VictoriaMetrics/VictoriaMetrics/blob/master/LICENSE

**GitHub handle or email address as a point of contact so that we can reach out in case we have questions**
@hagen1778

### DIFF
--- a/data/ecosystem/vendors.yaml
+++ b/data/ecosystem/vendors.yaml
@@ -351,7 +351,7 @@
   commercial: true
 - name: VictoriaMetrics
   distribution: false
-  nativeOTLP: false
+  nativeOTLP: true
   url: 'https://github.com/VictoriaMetrics/VictoriaMetrics/tree/master/docs#sending-data-via-opentelemetry'
   contact: 'https://github.com/hagen1778'
   oss: true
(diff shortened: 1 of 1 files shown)


data/ecosystem/vendors.yaml
```diff
@@ -349,3 +349,10 @@
   contact: 'otel@tingyun.com'
   oss: false
   commercial: true
+- name: VictoriaMetrics
+  distribution: false
+  nativeOTLP: false
```
cartermp, line 354: Hmm. Your docs would indicate you support direct OTLP ingest of metrics. In that case should this be`true`?

theletterf (approved):

PR #3643, opened 2023-12-04
Add SRI hash to fuse.js import

theletterf:
This fixes https://github.com/open-telemetry/opentelemetry.io/security/code-scanning/1

I used https://www.srihash.org/ to calculate the SRI hash.


chalin (approved):
Works for me if it makes the error go away. Thanks @theletterf!

On a side note, I'd like to get rid of fuse, but that's another story :).

PR #3646, opened 2023-12-05
Update use of metadata in glossary.md

flands:
Metadata and its definition are used inconsistently in the glossary. Make it consistent and clarify its use.


content/en/docs/concepts/glossary.md
```diff
@@ -118,9 +117,8 @@ pull-based.
 
 ### **Field**
 
-Name/value pairs added to [`Log Records`](#log-record) (similar to
-[`Attributes`](#attribute) for [`Spans`](#span) and [`Labels`](#label) for
-[`Metrics`](#metric)). See [field spec][field].
+Term used specifically for [`Log Records`](#log-record). See
+[`Attribute`](#attribute) and the [field spec][field].
```
theletterf, line 121: Attribute then sends folks directly to Metadata. Perhaps we should direct them there right away?

content/en/docs/concepts/glossary.md
```diff
@@ -36,9 +36,8 @@ alerting and tooling for finding the root cause.
 
 ### **Attribute**
 
-A key-value pair. Used across telemetry signals - e.g. in [`Traces`](#trace) to
-attach data to a [`Span`](#span), or in [`Metrics`](#metric). See [attribute
-spec][attribute].
+See [`Metadata`](#metadata). Used across [`Signals`](#signal) and [`Resources`]
```
theletterf, line 39: Hmm, I think it might still be worth it to describe what an attribute looks like, and the fact that some vendors call them "tags" still. WDYT?

content/en/docs/concepts/glossary.md
```diff
@@ -152,7 +150,7 @@ Short for [JavaScript Object Notation][json].
 
 ### **Label**
 
-See [Attribute](#attribute).
+Term used specifically for [`Metrics`](#metric). See [`Attribute`](#attribute).
```
theletterf, line 153:
```suggestion
Term used for [`Metrics`](#metric). See [`Attribute`](#attribute).
```

Same thing about Attribute and Metadata.

content/en/docs/concepts/glossary.md
```diff
@@ -178,9 +176,10 @@ describes what happened, where it happened, etc. See [more][log record].
 
 ### **Metadata**
 
-A name/value pair added to telemetry data. OpenTelemetry calls this
-[`Attributes`](#attribute) on [`Spans`](#span), [`Labels`](#label) on
-[`Metrics`](#metric) and [`Fields`](#field) on [`Logs`](#log).
+A name/value pair (e.g. `foo="bar"`) added to telemetry data. OpenTelemetry
+generally calls this [`Attributes`](#attribute) though [`Metrics`](#metric) have
+[`Dimensions`](#dimension) and [`Labels`](#label) while [`Logs`](#log) have
+[`Fields`](#field).
```
theletterf, line 182:
```suggestion
A key-value pair, for example `foo="bar"`, added to telemetry data. OpenTelemetry
calls these pairs [`Attributes`](#attribute). In addition, [`Metrics`](#metric) have
[`Dimensions`](#dimension) and [`Labels`](#label), while [`Logs`](#log) have
[`Fields`](#field).
```

theletterf (changes requested):

mhausenblas (changes requested):
Thanks and LGTM, essentially the same points that @theletterf already highlighted from my side.

content/en/docs/concepts/glossary.md
flands: I'm trying to avoid saying key-value pair everywhere otherwise, it doesn't seem to make sense to have a definition for or use the word metadata - thoughts? When the problems I see is that attributes is the standard here and neither is dimension, labels, or fields. At the end of the day, all of this is metadata.

content/en/docs/concepts/glossary.md
flands: I went back and forth here - attribute appears to the word OTel wants to use to mean metadata, but it's not used everywhere. So option is to link everything to attribute, which just links to metadata or direct to metadata. I'm going switching direct to metadata and will update this.

content/en/docs/concepts/glossary.md
```diff
@@ -118,9 +117,8 @@ pull-based.
 
 ### **Field**
 
-Name/value pairs added to [`Log Records`](#log-record) (similar to
-[`Attributes`](#attribute) for [`Spans`](#span) and [`Labels`](#label) for
-[`Metrics`](#metric)). See [field spec][field].
+Term used specifically for [`Log Records`](#log-record). See
```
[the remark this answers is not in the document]
flands: That's an interesting point - technically, all fields are key-value pairs, but the value could be another nested key-value pair. In addition, fields support the notion of resource and attribute. What's also interesting is that you have fields such as SeverityText or TraceId, which can technically be used as metadata that may or may not be included in attributes or resources.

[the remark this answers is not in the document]
flands: OK, attempted to address this feedback - feel free to reopen

content/en/docs/concepts/glossary.md
theletterf: Could we describe it as a <something> that describes data and might take the form of a KV pair? Maybe with an example?

flands: OK, done!

theletterf (approved):

cartermp (approved):

PR #3649, opened 2023-12-05
All mentions of Twitter now mention X

nmummau:
#3647

PR #3653, opened 2023-12-06
Drop obsolete hugo config param

chalin:
FontAwesome is a derived dependency, pulled in via Docsy. There should be no mention of FA versions in this repo. I think that the `font_awesome_version` is a vestige from long long ago.

Removing that like doesn't affect generated site files:

```console
$ (cd public && git diff -I 'var buildDate')
$
```


chalin:
For context, see https://github.com/open-telemetry/opentelemetry.io/pull/3649/files#r1416482560.

svrnm (approved):

PR #3659, opened 2023-12-08
Add CVE page

martinkuba:
This is adding a page with a list of published security vulnerabilities across all repositories, similar to [Kubernetes docs](https://kubernetes.io/docs/reference/issues-security/official-cve-feed/).

### DIFF
--- a/content/en/docs/cve.md
+++ b/content/en/docs/cve.md
@@ -32,12 +32,7 @@ is refreshed daily.
       var url = 'https://raw.githubusercontent.com/open-telemetry/sig-security/data-source/published_output.json';
       return fetch(url)
         .then(function(response) {
-          return response.text();
-        })
-        .then(data => {
-          data = data.replace(/}\s*{/g, '},{');
-          var json = JSON.parse('[' + data + ']');
-          return json
+          return response.json();
         });
     }
 
--- a/content/en/docs/cve.md
+++ b/content/en/docs/cve.md
@@ -1,0 +1,78 @@
+---
+title: Common Vulnerabilities and Exposures
+weight: 102
+---
+
+This is a list of reported CVEs across all repositories in the OpenTelemetry org
+on GitHub. The raw data is stored in
+[this GitHub repository](https://github.com/open-telemetry/sig-security), and it
+is refreshed daily.
+
+  <table id="cve-table">
+    <thead>
+      <tr>
+        <th>CVE ID</th>
+        <th>Issue Summary</th>
+        <th>Severity</th>
+        <th>Repository</th>
+      </tr>
+    </thead>
+    <tbody>
+    </tbody>
+  </table>
+
+<script id="main-script">
+  'use strict';
+  (function() {
+    function fetchAndRender() {
+      fetchData()
+        .then(renderTable);
+    }
+
+    function fetchData() {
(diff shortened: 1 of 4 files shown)


content/en/docs/cve.md
```diff
@@ -0,0 +1,78 @@
+---
+title: Common Vulnerabilities and Exposures
+weight: 102
+---
+
+This is a list of reported CVEs across all repositories in the OpenTelemetry org
+on GitHub. The raw data is stored in
+[this GitHub repository](https://github.com/open-telemetry/sig-security), and it
+is refreshed daily.
```
svrnm, line 9:
```suggestion
This is a list of reported Common Vulnerabilities and Exposures (CVEs) across all repositories in the [OpenTelemetry organisation on GitHub](https://github.com/open-telemetry/). The raw data is stored in
the [open-telemetry/sig-security](https://github.com/open-telemetry/sig-security) repository, and it
is refreshed daily.
```

content/en/docs/cve.md
```diff
@@ -0,0 +1,78 @@
...
+        <th>Issue Summary</th>
+        <th>Severity</th>
+        <th>Repository</th>
+      </tr>
+    </thead>
+    <tbody>
+    </tbody>
+  </table>
+
+<script id="main-script">
```
svrnm, line 24: I guess, the purpose of this script is to make sure that the most up-to-date data is always live on the website? 

There is an alternative way: hugo supports the ability to read from JSON files (https://gohugo.io/functions/data/getjson/), so we could replace the javascript with a template function and the reload with a re-build of the website triggered on a new CVE being published? The **advantage** would be that we don't need javascript, which feels a little bit more accessible and also makes it easier for crawlers to collect this page (helping the CVEs to be findable)

We can start with the JS, but I would like to get @chalin's perspective when he's back

svrnm (dismissed):
Looking forward to have this, I provided some initial feedback

content/en/docs/cve.md
chalin: Yes, let's do build-time fetch of the data if y'all agree!

chalin (commented):
I agree with both points that @svrnm raised: let's move this into the newly proposed `/security` section, and do data-fetch at build time.

content/en/docs/cve.md
```diff
@@ -0,0 +1,72 @@
...
+      <tr>
+        <th>CVE ID</th>
+        <th>Issue Summary</th>
+        <th>Severity</th>
+        <th>Repository</th>
+      </tr>
+    </thead>
+    <tbody>
+    </tbody>
+  </table>
```
chalin, line 21: Nit: unindent the HTML.

chalin:
Otherwise, great job! LGTM

content/en/docs/cve.md
martinkuba: How often is the site rebuilt? If we do this, should I remove the text about the list being updated daily?

chalin: As @svrnm suggested, let's go with the dynamic update for now. We can revisit the page in 2024 :).

theletterf (approved):

cartermp (approved):

PR #3661, opened 2023-12-08
improve dotnet getting started instructions

andrew-hillier:
- on the initial instructions, added a function for RollDice() that was referenced, but not included.
- when adding the Open Telemetry code, instructions to replace certain lines of code, including the required using statements.

### DIFF
--- a/content/en/docs/instrumentation/net/getting-started.md
+++ b/content/en/docs/instrumentation/net/getting-started.md
@@ -46,6 +46,8 @@ code:
 ```csharp
 using System.Globalization;
 
+using Microsoft.AspNetCore.Mvc;
+
 var builder = WebApplication.CreateBuilder(args);
 var app = builder.Build();
 
@@ -63,6 +65,12 @@ string HandleRollDice([FromServices]ILogger<Program> logger, string? player)
     }
 
     return result.ToString(CultureInfo.InvariantCulture);
+}
+
+int RollDice()
+{
+    var rng = new Random();
+    return rng.Next(1, 6);
 }
 
 app.MapGet("/rolldice/{player?}", HandleRollDice);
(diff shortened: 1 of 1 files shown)


cartermp:
@andrew-hillier can you sign the CLA? Then we can review

cartermp (approved):
Thank you, I think some of this got lost in a transition from a previous version of this doc. Thanks for correcting it!

andrew-hillier:
> Thank you, I think some of this got lost in a transition from a previous version of this doc. Thanks for correcting it!

No problem, happy to contribute! 😀

cijothomas (approved):

content/en/docs/instrumentation/net/getting-started.md
```diff
@@ -65,6 +67,12 @@ string HandleRollDice([FromServices]ILogger<Program> logger, string? player)
     return result.ToString(CultureInfo.InvariantCulture);
 }
 
+int RollDice()
+{
+    var rng = new Random();
+    return rng.Next(1, 6);
```
reyang, line 73: Curious, is this the right way of using Random?

cartermp: Technically no, but that's more for when you've got a tight loop, since then it will be "less random" due to initializing each iteration. But I think per-request is fine for demonstrative purposes.

pjanotti: While this is certainly fine for a demo why not use the same as in the auto-instrumentation docs?
https://github.com/open-telemetry/opentelemetry.io/blob/210e9c6c0d0279a7292e31bfe6fb54032ae44b87/content/en/docs/instrumentation/net/automatic/getting-started.md?plain=1#L54-L57

cartermp: mmm, yes, I had forgotten about the new .NET API. Let's go with that. @andrew-hillier can you make this change?

reyang (approved):

PR #3678, opened 2023-12-13
Fix code license, add NOTICE

svrnm:
The LICENSE-CODE file has a copy&paste error, fixed by this. I checked with other projects and the meaning of the license and the `Copyright [yyyy] [name of copyright owner]` should **not** be replaced by a project name.

I also added a NOTICE file as suggested by the apache license, see [section 4.4](https://www.apache.org/licenses/LICENSE-2.0.html#redistribution) for details. Although our repo does not have a lot of content, I think it still helps to improve potential attributions.

### DIFF
--- a/LICENSE-CODE
+++ b/LICENSE-CODE
@@ -187,9 +187,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright 2020 The KEDA Authors.
-
-   and others that have contributed code to the public domain.
+   Copyright [yyyy] [name of copyright owner]
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


theletterf (approved):
Out of curiosity: Is this something the Legal folks of CNCF should/could review?

svrnm:
> Out of curiosity: Is this something the Legal folks of CNCF should/could review?

the change of the LICENSE file: no, the NOTICE file: maybe?

svrnm:
I removed the NOTICE for now, because it may require more discussion

cartermp (approved):

chalin:
Oh my, thanks for noticing that copy-paste error. There was back and forth some time ago in terms of which file(s) to use, I think that that's how it slipped 🤷🏼‍♂️ 

IMHO the LICENSE-CODE file needs further changes, the copyright notice is for the code in this repo, not the code in the documentation AFAIK. I'll open a separate issue to track this.

chalin:
See:

- #3742

PR #3680, opened 2023-12-13
Add 'Building' section to collector docs

svrnm:
We discussed this a few times already that we should group some of the "Building..." and add a new section for them. Here it is. I thought about including [Building a custom collector](https://opentelemetry.io/docs/collector/custom-collector/) or not, and decided against inclusion: I think it is a very important topic that needs to be at a higher level, and it is  also different to the other as it is not something that requires you to write code. That's why this section is now called "Building custom components"

I also dropped/added some qualifiers:
* All mention of "custom" as this is now given by the top level
* Renamed "trace receiver" to only "receiver", as this is just an example but could be used for other components as well (I actually used it as an entry point to write a processor)
* Renamed "custom authenticator" to "authenticator extension" to highlight that this is about extensions as well and _might_ be helpful for other extensions.

### DIFF
--- a/content/en/docs/collector/building/authenticator-extension.md
+++ b/content/en/docs/collector/building/authenticator-extension.md
@@ -1,6 +1,7 @@
 ---
-title: Building a custom authenticator
-weight: 30
+title: Building an authenticator extension
+weight: 40
+aliases: [/docs/collector/custom-auth/]
 cSpell:ignore: configauth oidc
 ---
 
--- a/content/en/docs/collector/building/_index.md
+++ b/content/en/docs/collector/building/_index.md
@@ -1,0 +1,11 @@
+---
+title: Building custom components
+description: Instructions on how to build your own collector components
+weight: 90
+---
+
+The OpenTelemetry Collector can not only be extended by existing components, but
+also by custom components, that you develop and build on your own. Here you will
+find instructions on how to build some of those components. For additional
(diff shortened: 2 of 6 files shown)


content/en/docs/collector/building/authenticator-extension.md
```diff
@@ -1,6 +1,7 @@
 ---
-title: Building a custom authenticator
-weight: 30
+title: Building an authenticator extension
```
jpkrohling, line 2: An authenticator is an extension. In fact, I think the previous title still resonates better with me.

jpkrohling (approved):
Despite still preferring the previous title, this LGTM.

content/en/docs/collector/building/authenticator-extension.md
svrnm:
```suggestion
title: Building an authenticator
```

svrnm: works for me as well, do you want the "custom" still be there as well?

jpkrohling: I have a slight preference for the version with "custom", but they are both fine.

svrnm: I will leave "custom" out more because it makes the navbar links a little shorter

PR #3687, opened 2023-12-15
Move context propagation into dedicated page

svrnm:
This PR moves the concept of context propagation out of traces into a dedicated page. I tried to fix some of the wording to make it clear that this concept is not exclusive to tracing/for all signals, but overall I tried to keep the text untouched as far as possible, as this may require a follow up PR to cleanup (so this is a "perfect is the enemy of good" PR)


cartermp (approved):
I love it.

PR #3701, opened 2023-12-18
php: fix namespace of Globals

brettmc:
Globals was moved just prior to going GA, but the documentation was not updated accordingly.


svrnm (approved):

svrnm:
thank you @brettmc

PR #3703, opened 2023-12-21
Clarify advice on using global TracerProvider

### LINKED ISSUE #3692: Advice to use GlobalOpenTelemetry is misleading
The library instrumentation page [says](https://github.com/open-telemetry/opentelemetry.io/blob/main/content/en/docs/concepts/instrumentation/libraries.md#getting-a-tracer) that "libraries should obtain tracer from global TracerProvider by default". It goes on to say that "its useful for libraries to have an API that allows applications to pass instances of TracerProvider explicitly".

This isn't necessarily wrong, but the advice is flipped in java and the code snippet (along with [another that shortly follows](https://github.com/open-telemetry/opentelemetry.io/blob/main/content/en/docs/concepts/instrumentation/libraries.md#public-apis)) use `opentelemetry-java` as an example. See [language in the java manual instrumentation page](https://github.com/open-telemetry/opentelemetry.io/blame/main/content/en/docs/instrumentation/java/manual.md#L641-L651) which directly contradicts this advice:

> if you are writing library instrumentation, it is strongly
recommended that you provide your users the ability to inject an instance of
`OpenTelemetry` into your instrumentation code. If this is not possible for some
reason, you can fall back to using an instance from the `GlobalOpenTelemetry`
class

Options:
- Clarify that the preference for global versus providing instances TracerProvider may vary by language
- Use an example language that isn't java, and maybe add a footnote that at least in java the behavior is flipped

jack-berg:
Resolves #3692.


cartermp:
@jack-berg if you run `npm run fix:format` then this should be good to go

svrnm (approved):

svrnm:
Thanks @jack-berg !

jack-berg:
Nice trick @svrnm! I'll have to remember that next time..

svrnm:
> Nice trick @svrnm! I'll have to remember that next time..

You mean the fix:format? Ye, it's finally showing some reliability 😆 makes live easier this way (we also have fix:refcache but afaik it's broken again)

PR #3705, opened 2023-12-22
Expand otel collector docs

hughesjj:
I've had this kinda stuff sitting around for a while.  It's a big chunk, so even though it's incomplete I figure I should at least start soliciting advice, sharing direction, or factoring out smaller bits to merge upstream.


svrnm:
> I've had this kinda stuff sitting around for a while. It's a big chunk, so even though it's incomplete I figure I should at least start soliciting advice, sharing direction, or factoring out smaller bits to merge upstream.

Thanks for sharing your work @hughesjj. As you said it's a big chunk to review, so I have to postpone the review for after the break

TODO.md
```diff
@@ -0,0 +1,15 @@
+# Review
```
theletterf, line 1: This is meant to be removed once the PR is ready, I guess.

theletterf (commented):
Hi there! Thanks for this!

Could you explain a bit more the scope and purpose of the changes?

PR #3707, opened 2023-12-22
Fix path issue

svrnm:


austinlparker (approved):

layouts/partials/hooks/body-end.html
```diff
@@ -1,4 +1,2 @@
-{{ if hugo.IsProduction -}}
-  {{ partial "script.html" (dict "src" "js/tracing.js") -}}
-{{ end -}}
+{{ partial "script.html" (dict "src" "js/tracing.js") -}}
```
svrnm, line 1:
```suggestion
{{ if hugo.IsProduction -}}
  {{ partial "script.html" (dict "src" "js/tracing.js") -}}
{{ end -}}
```

chalin:
Thanks for noticing the build error @svrnm and/or @austinlparker. Strange, I can't see what changed that might have affected the production build. Anyhow, I'm glad that this is resolved.

chalin:
Followups:

- #3736
- #3737

PR #3715, opened 2023-12-28
Use correct link for values.yaml

jeff-d:


cartermp:
Thanks @jeff-d but I think this one came first, so I'll close your PR

PR #3719, opened 2024-01-03
Fix typo in automatic.md

Jeldo:
Correct the endpoint value that differs from the description(`http://otel-collector.opentelemetry.svc.cluster.local:4317`) below.

### DIFF
--- a/content/en/docs/kubernetes/operator/automatic.md
+++ b/content/en/docs/kubernetes/operator/automatic.md
@@ -656,7 +656,7 @@ For example:
 ```yaml
 spec:
   exporter:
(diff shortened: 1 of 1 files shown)


cartermp (approved):

jaronoff97:
It looks like the rest of the docs reference demo-collector, should we use that here instead?

Jeldo:
> It looks like the rest of the docs reference demo-collector, should we use that here instead?

@jaronoff97 It would be helpful to understand if we use the same names throughout the document.

jaronoff97 (approved):

svrnm (approved):

PR #3723, opened 2024-01-04
Update Python automatic instrumentation directions

flands:
OTel Python automatic instrumentation doesn't support Flask or Werkzeug 3.0+. This will be addressed in [this PR](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2013) As a result, the automatic instrumentation example doesn't work. This has been an issue since Oct and reported in multiple issues across multiple repositories. The getting started automatic instrumentation does work as it's pinned to versions less than 3.0.

Update the documentation for the automatic instrumentation so it work. Addresses: https://github.com/open-telemetry/opentelemetry-python/issues/3543 Also switch to venv over virtualenv to be consistent with getting started documentation and to remove an additional dependency.

Once the above PR is merged, both the getting started and automatic instrumentation example documentation should be updated to remove the pinning.


svrnm (approved):
@open-telemetry/python-approvers please take a look

srikanthccv (approved):
Thanks

PR #3730, opened 2024-01-08
Registry Refresher: UI changes, search updates & more

svrnm:
Over the holidays I picked an update for the registry as a side project, this PR has all the results. We might consider to cherry-pick a few changes, but I wanted to also try out a few experimental things.

Preview: https://deploy-preview-3730--opentelemetry.netlify.app/ecosystem/registry/

Note that the PR touches 700+ files because I applied a lot of transformations on the registry entries, so the most relevant changes are to 

* assets/js/registrySearch.js
* assets/scss/_registry.scss
* scripts/update-registry-versions.sh
* layouts/shortcodes/ecosystem/registry/search-form.html
* layouts/partials/ecosystem/registry/quickinstall/*.md
* layouts/partials/ecosystem/registry/entry.html
* layouts/ecosystem/registry.json.json
* templates/registry-entry.yml

What I changed:
* The look and feels of the entries, leveraging a lot of CSS that comes with Bootstrap 
* Update authors to also include links, emails, multiple authors, etc
* Allow multiple links (repo, docs, package, etc.)
* Show package versions + update script for common registries (npm, nuget, hex, etc.)
* Quick Install for known registries + collector
* Badges like "new" "native" "included with the otel demo", "is first party" or "deprecated"
* Replaced fuse with minisearch, because latest versions of fuse provides strange/incorrect search results and the repo activity is fairly stalled, while minisearch seems to be more active
* Add auto suggestion via minisearch


theletterf (approved):
Nice work!

Wouldn't it be more "educational" to leave optional fields such as "thirdParty" even if they're false? That could give contributors a better idea of what they'd fill out.

svrnm:
> Nice work!
> 
> Wouldn't it be more "educational" to leave optional fields such as "thirdParty" even if they're false? That could give contributors a better idea of what they'd fill out.

My concern with "thirdParty" always was, that it was not 100% clear what we mean by that: 
* It was often used to mean "this was not created by the OpenTelemetry community", but that's identified by the "authors" attribute
* I think it was meant as "this instrumentation library is not created by the same party as the instrumented library", but rarely used as that.

I just realised that I introduced "isFirstParty" with the same ambiguity, we might need a better word to describe that (or have proper documentation around it)

svrnm:
@open-telemetry/php-approvers the request for your approval was triggered by https://github.com/open-telemetry/opentelemetry.io/pull/3730/commits/433289539a0bcc00de8dc340210164cf44aff1a0, since this PR adds `packagist` to the list of known words

theletterf:
How about `isVendor`?

svrnm:
> How about `isVendor`?

without context it also is not really clear what this means... 🤔 

`isFirstPartyInstrumentation` or `isThirdPartyInstrumentation` maybe to be more explicit on what we want to express? Because this flag should also only apply on registry entries of type `instrumentation`

Note, that I also introduced `isNative` in a case where not an instrumentation library is provided, but otel is integrated into the app/library directly (like Next.js is doing it)

data/registry/collector-exporter-alibaba-cloud-log-service.yml
```diff
@@ -1,14 +1,20 @@
+# cSpell:ignore alibabacloudlogserviceexporter
+# cspell:ignore alibabacloudlogserviceexporter
```
chalin, line 2: Remove repeated line:

```suggestion
# cSpell:ignore alibabacloudlogserviceexporter
```

chalin (changes requested):
Very nice improvements! 

Drive-by comment: card content overflows on mobile / narrow displays, e.g.

image

I'll take a closer as soon as I can, or maybe we can discuss it at the next comms meeting.

chalin:
> Over the holidays I picked an update for the registry as a side project

Wow, can see that you had some fun! ✨ 😄

svrnm:
> Drive-by comment: card content overflows on mobile / narrow displays, e.g.

I think I have fixed that now. I used "flex" but now changed it to rows/cols by bootstrap with breakpoints set accordingly

svrnm:
@open-telemetry/docs-approvers please take a look. Although this is a big change, I am not worried about merging it, since the current registry is not really used that heavily and this might give it some more utility. wdyt?

chalin (approved):
(RS)LGTM! I haven't look at the JS and CSS code in details (don't have the bandwidth atm), but overall this seems like a very nice improvement 👍🏻

static/refcache.json
```diff
@@ -5587,6 +6515,9 @@
     "StatusCode": 206,
     "LastSeen": "2023-06-29T18:39:49.775529-04:00"
   },
+  "https://rubygems.org/gems/opentelemetry-instrumentation-redis": {
+    "StatusCode": 200,
+    "LastSeen": "2024-01-08T12:17:25.018558+01:00"
   "https://rubygems.org/gems/opentelemetry-instrumentation-all": {
```
chalin, line 6521: These lines are misformatted.

svrnm: hopefully fixed in https://github.com/open-telemetry/opentelemetry.io/pull/3730/commits/dbef3153d066cff1514140411fb6639173d9699f

svrnm:
> (RS)LGTM! I haven't look at the JS and CSS code in details (don't have the bandwidth atm), but overall this seems like a very nice improvement 👍🏻

There is probably a lot of space for improvement (and also some stuff I didn't remove yet, which is not required anymore), but I would review this in a follow up PR then

PR #3735, opened 2024-01-08
Add dependabot config for GitHub actions

svrnm:
Via https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot


chalin (approved):
Looks reasonable to me

PR #3752, opened 2024-01-10
Adding Immersive Fusion

dankoverride:


data/ecosystem/vendors.yaml
```diff
@@ -132,6 +132,13 @@
   contact: 'support@hyperdx.io'
   oss: true
   commercial: true
+- name: Immersive Fusion
+  distribution: true
```
svrnm, line 136: can you provide a link where we can understand/see the distribution you have?

dankoverride: That was oversight. I've submitted a fix. Thank you for catching.

svrnm (approved):

PR #3757, opened 2024-01-11
Update opentelemetry-specification version to 1.29.0

svrnm:
Replacement for #3755 (see #3756 as well)


chalin (approved):
I removed the check since, as the TODO message said "drop the following warning once the checks are enabled in the spec repos" and the check is now in the spec repo.

All checks are green now (and I don't know why it didn't fail before -- looking at the logs it seems that is _did_ fail, but maybe just not noticed before the merge?).

chalin:
All green, merging!

PR #3769, opened 2024-01-12
Add new blog post for Distributed tracing in event driven architecture: 2024

TamimiGitHub:
I added a new blog post under 2024 that talks about Distributed Tracing in Event Driven Architecture Systems. 

Any reviews of further opinions on this is greatly appreciated! Thanks!

CC: @chalin


TamimiGitHub:
Thanks @svrnm for the feedback. I attempted to rewrite it to include a focus on open source components like the solace otel receiver in the collector and solace APIs. I have also referenced otel open source tools like the agent as well. This re-write of the blog had more of an architecture approach discussing the issues that could be faced with distributed tracing in an event driven architecture in general while taking a tool as an example, in this case Solace. 

The previously declined PR had a solace focus to it while this one was more generic leaving the audience with thoughts about EDA systems... Do you have any recommendations on what should be changed in the blog to make it more generic? I took a similar approach to the [Apache Kafka](https://opentelemetry.io/blog/2022/instrument-kafka-clients/) blog back in 2022.

Any feedback and consideration is appreciated!

svrnm:
Solace PubSub+ Event Broker is commercial, so we can not accept it as an example tool in such a blog post. If the blog post focus is "issues that could be faced with distributed tracing in an event driven architecture in general", then use an OSS tool as an example.

TamimiGitHub:
Thanks for the input. The issue is there isint an OSS event broker that supports span generation from within the broker, so presenting a solution with an open source technology will not fit in this. I can present the problem in the blog post but no solution without mentioning an enterprise approach. 
Any thoughts on this? Thanks!

svrnm:
> Thanks for the input. The issue is there isint an OSS event broker that supports span generation from within the broker, so presenting a solution with an open source technology will not fit in this. I can present the problem in the blog post but no solution without mentioning an enterprise approach. Any thoughts on this? Thanks!

Sorry, but I am not a subject matter expert on this topic, what is the "enterprise approach" in this context which is not possible to be presented without an OSS solution?

But, to reiterate that point: we can and will not accept a blog post that promotes a commercial offering as a solution. If you think your blog post is impossible to get to some points without that, then unfortunately we can not accept the blog post here.

TamimiGitHub:
No worries, I completely understand. The Enterprise approach covered in this article is the ability to generate spans from _within_ the broker in the overall distributed system. So for example spans like what happens when a message goes to a particular queue, dead letter queue, rejections due to broker specific issues, hops between multiple brokers in an event mesh, and the time the message takes to traverse inside the broker. 

OSS solution approach telemetry with message broker as a "black box", in other words telemetry data show messages going in and out of the broker with no details on what's going on inside the broker. This requires further broker specific customizations and a dedicated receiver on the collector. I did write another blog post about this on [DZone](https://dzone.com/articles/datadog-solace-tracing) (tho that one was more solace specific so I didnt submit it to the otel blog). 

However I do understand the guideline around refusing to promote an enterprise solution, which makes sense for a site that promotes only OSS. I'll think of another way to address this gap in the industry without mentioning a solution..

Thanks again for looking into this.

svrnm:
oss message brokers and middleware are a black box until they are not: eventually some of them will have opentelemetry built-in as well, as this is what OpenTelemetry is all about. A good way for you to address this gap in the industry is collaborating with the community to close it, e.g. by providing a helping hand in the Semantic Conventions Messaging WG. They meet every Thursday at 8:00PT, and you can get in touch via CNCF slack at [#otel-messaging](https://cloud-native.slack.com/archives/C02Q4AAHDSA)

I will close this PR now

Issue #2557, opened 2023-03-30
Document how to change the span status


pellared:
**Desired feature or idea**: What would you like to see happen?

Document how people can change the span status using the [Collector's Span Processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/spanprocessor).

This could be documented here: https://opentelemetry.io/docs/collector/transforming-telemetry/

**Additional context**: 

I have seen frequently questions how people can change "false negative" error span status for 404 HTTP Client calls or NOT_FOUND gRPC calls.

svrnm:
@alejandraquintero018 are you still interested in helping with this?

Gaurang-Patel:
Hi @alejandraquintero018 , @svrnm  I would like to work on this issue, should I document using a span processor or transform processor ?

svrnm:
Thank you @Gaurang-Patel, update https://opentelemetry.io/docs/collector/transforming-telemetry/ with instructions for the SpanProcessor. We should have a dedicated section on the TransformProcessor eventually as well, so that could be a follow up thing.

PR #3772, opened 2024-01-14
collector-span:  added documentation for changing span status in collector

Gaurang-Patel:
Fixes: https://github.com/open-telemetry/opentelemetry.io/issues/2557

---

**Preview**: https://deploy-preview-3772--opentelemetry.netlify.app/docs/collector/transforming-telemetry/


content/en/docs/collector/transforming-telemetry.md
```diff
@@ -169,6 +169,39 @@ processors:
...
+
+The following setting is required:
+
+- `code:` Represents span status. One of the following values `Unset`, `Error`,
+  `Ok`.
+- `description:` setting is allowed only for code `Error`.
+
+Example:
+
+```yaml
```
cartermp, line 189: I would prefer a more realistic example like this:

```
span/set_status_ok:
  include:
    attributes:
      - key: http.status_code
        value: 400
  status:
    code: "Ok"
```

The example you have below unconditionally sets all spans to an error which is not a scenario most people will look for. They will instead likely look for either a specific span or a condition in which to change the status.

content/en/docs/collector/transforming-telemetry.md
```diff
@@ -169,6 +169,43 @@ processors:
...
 ```
 
+## Set status for span
+
+**Processor**:
+[span processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/spanprocessor)
+
+The
+[span processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/spanprocessor)
+can be used to change span status.
```
svrnm, line 179:
```suggestion
To change the [status of a span](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#set-status), the
[span processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/spanprocessor) can be used.
```

Not a native speaker, but I had the feeling that the other way round it implies that this is all the span processor those, so by putting the thing you want to accomplish (change the span status) first, it's clearer what this is about.

content/en/docs/collector/transforming-telemetry.md
```diff
@@ -169,6 +169,43 @@ processors:
...
+
+The following setting is required:
```
svrnm, line 181:
```suggestion
The following settings are available:
```

svrnm (changes requested):

content/en/docs/collector/transforming-telemetry.md
```diff
@@ -169,6 +169,43 @@ processors:
...
+- `description:` setting is allowed only for code `Error`.
+
+Example:
+
+```yaml
+# Set status allows to set specific status for a given span. Possible values are
+# Ok, Error and Unset as per
+# https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#set-status
+# The description field allows to set a human-readable message for errors.
+#Example below sets status to success only when attribute http.status_code is equal to 400
```
chalin, line 194: Fix spacing. And you might want to line-wrap.

```suggestion
# Example below sets status to success only when attribute http.status_code is equal to 400
```

content/en/docs/collector/transforming-telemetry.md
```diff
@@ -169,6 +169,43 @@ processors:
...
+- `code:` Represents span status. One of the following values `Unset`, `Error`,
+  `Ok`.
+- `description:` setting is allowed only for code `Error`.
+
+Example:
+
+```yaml
+# Set status allows to set specific status for a given span. Possible values are
+# Ok, Error and Unset as per
+# https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#set-status
```
chalin, line 192: All of this descriptive prose should be part of the main text, outside of the YAML.
@cartermp et al., WDYT?

Also, for the link use https://opentelemetry.io/docs/specs/otel/trace/api/#set-status if you embed it as a comment, otherwise use `/docs/specs/otel/trace/api/#set-status`

```suggestion
# Ok, Error and Unset as per https://opentelemetry.io/docs/specs/otel/trace/api/#set-status
```

chalin (commented):
Drive-by comments

content/en/docs/collector/transforming-telemetry.md
cartermp: Eh, I kind of agree, but it's also helpful because if someone just copy-pastes this content, they will see a helpful comment in their config.

theletterf: I tend to like having comments in the YAML config snippets, as it's often where the eyes of our readers will go first. But I also like having the same content in the docs, to have some redundancy.

content/en/docs/collector/transforming-telemetry.md
theletterf: Let's get rid of passive voice altogether as per our style recommendations:

```suggestion
You can use the
[span processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/spanprocessor)
to change span status.
```

content/en/docs/collector/transforming-telemetry.md
```diff
@@ -169,6 +169,43 @@ processors:
...
+      - key: http.status_code
+        value: 400
+  status:
+    code: 'Ok'
+```
+
+The span processor also allows to modify the span name based on its attributes
+or extract span attributes from the span name. Refer to
+[config.yaml](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/spanprocessor/testdata/config.yaml)
+for detailed examples on using the processor.
```
theletterf, line 207:
```suggestion
You can also use the span processor also to modify the
span name based on its attributes or extract span attributes
from the span name. See the
[config.yaml](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/spanprocessor/testdata/config.yaml)
file for detailed examples on how to use the span processor.
```

svrnm:
@Gaurang-Patel please take a look at the feedback & suggestions, let us know if there is anything where we can help you!

content/en/docs/collector/transforming-telemetry.md
Gaurang-Patel: hi @cartermp , @theletterf  do we need to duplicate the content outside the YAML block?

content/en/docs/collector/transforming-telemetry.md
```diff
@@ -169,6 +169,43 @@ processors:
   k8sattributes/default:
 ```
 
+## Set status for span
```
chalin, line 172: Let's be consistent with the other sections:

```suggestion
## Setting a span status 
```

content/en/docs/collector/transforming-telemetry.md
chalin: None of the other YAML excerpts have such comments. I'd vote to put it all in the prose, and after, if necessary, we can see which snippet might make sense to include as a YAML comment. WDYT @cartermp and @theletterf?

content/en/docs/collector/transforming-telemetry.md
```diff
@@ -169,6 +169,43 @@ processors:
...
+[span processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/spanprocessor)
+to change span status.
+
+The following settings are available:
+
+- `code:` Represents span status. One of the following values `Unset`, `Error`,
+  `Ok`.
+- `description:` setting is allowed only for code `Error`.
+
+Example:
```
chalin, line 187:
```suggestion
The following example sets the span status to `Ok` when the `http.status_code` attribute is 400:
```

content/en/docs/collector/transforming-telemetry.md
```diff
@@ -169,6 +169,43 @@ processors:
...
+
+```yaml
+# Set status allows to set specific status for a given span. Possible values are
+# Ok, Error and Unset as per
+# https://opentelemetry.io/docs/specs/otel/trace/api/#set-status
+# The description field allows to set a human-readable message for errors.
+# Example below sets status to success only when attribute http.status_code is equal to 400
```
chalin, line 194:
```suggestion
# The description field allows to set a human-readable message for errors.
```

content/en/docs/collector/transforming-telemetry.md
```diff
@@ -169,6 +169,43 @@ processors:
...
+span/set_status_ok:
+  include:
+    attributes:
+      - key: http.status_code
+        value: 400
+  status:
+    code: 'Ok'
```
chalin, line 201:
```suggestion
    code: Ok
```

chalin (commented):
Another pass of copyedits.

content/en/docs/collector/transforming-telemetry.md
```diff
@@ -169,6 +169,43 @@ processors:
...
+## Set status for span
+
+**Processor**:
+[span processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/spanprocessor)
+
+You can use the
+[span processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/spanprocessor)
+to change span status.
+
+The following settings are available:
```
chalin, line 181:
```suggestion
Use the span processor to set a span's status. The following settings are available:
```

content/en/docs/collector/transforming-telemetry.md
```diff
@@ -169,6 +169,43 @@ processors:
...
+
+- `code:` Represents span status. One of the following values `Unset`, `Error`,
+  `Ok`.
```
chalin, line 184:
```suggestion
- `code`: represents span status, one of:
  - `Ok`
  - `Unset`
  - `Error`
```

content/en/docs/collector/transforming-telemetry.md
```diff
@@ -169,6 +169,43 @@ processors:
...
+- `description:` setting is allowed only for code `Error`.
```
chalin, line 185:
```suggestion
- `description`: status-change description, only allowed when `code` is `Error`
```

content/en/docs/collector/transforming-telemetry.md
chalin: Oh, maybe the section title should actually be 

> ## Setting span attributes

content/en/docs/collector/transforming-telemetry.md
```diff
@@ -169,6 +169,43 @@ processors:
...
+      - key: http.status_code
+        value: 400
+  status:
+    code: 'Ok'
+```
+
+You can also use the span processor also to modify the span name based on its
+attributes or extract span attributes from the span name. See the
+[config.yaml](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/spanprocessor/testdata/config.yaml)
+file for detailed examples on how to use the span processor.
```
chalin, line 207:
```suggestion
You can also use the span processor to modify the span name based on its
attributes or extract span attributes from the span name. For examples, see `spanprocessor` test data 
[config.yaml](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/spanprocessor/testdata/config.yaml)
file.
```

cartermp (approved):

svrnm (approved):

svrnm:
@open-telemetry/collector-approvers PTAL

mx-psi (commented):
@evan-bradley @TylerHelmuth do we want to use the span processor for this? Or do we want to encourage people to use the transform processor?

TylerHelmuth (changes requested):
@mx-psi is correct, we should not be pushing the spanprocessor.  Once I have some free time I should really deprecate that component; I don't think there is anything left that it does that the transformprocessor does not.

The transformprocessor config to change a span status in the collector is:

```yaml
transform:
  error_mode: ignore
  trace_statements:
    - context: span
      statements:
        - set(status.code, STATUS_CODE_OK) where attributes["http.request.status_code"] == 200
```

(I used the new stable http semantic convention value for status code)

content/en/docs/collector/transforming-telemetry.md
```diff
@@ -169,6 +169,40 @@ processors:
...
+span/set_status_ok:
+  include:
+    attributes:
+      - key: http.status_code
+        value: 400
+  status:
+    code: Ok
+```
+
+You can also use the span processor to modify the span name based on its
```
svrnm, line 200:
```suggestion
You can also use the transform processor to modify the span name based on its
```

content/en/docs/collector/transforming-telemetry.md
```diff
@@ -169,6 +169,40 @@ processors:
...
+attributes or extract span attributes from the span name. For examples, see
+`span processor` test data
```
svrnm, line 202:
```suggestion
`transform processor` test data
```

content/en/docs/collector/transforming-telemetry.md
```diff
@@ -169,6 +169,40 @@ processors:
...
+[config.yaml](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/spanprocessor/testdata/config.yaml)
```
svrnm, line 203:
```suggestion
[config.yaml](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/transformprocessor/testdata/config.yaml)
```

svrnm (changes requested):
@Gaurang-Patel @TylerHelmuth if we want to use the transform instead of the span processor here are the changes that need to be applied. Overall the docs works with both.

svrnm:
@mx-psi @TylerHelmuth please take another look!

mx-psi (approved):

svrnm (approved):

content/en/docs/collector/transforming-telemetry.md
```diff
@@ -169,6 +169,31 @@ processors:
   k8sattributes/default:
 ```
 
+## Setting a span status
+
+**Processor**:
+[transform processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/transformprocessor)
+
+Use the transform processor to set a span's status. The following example sets
+the span status to `Ok` when the `http.status_code` attribute is 400:
```
evan-bradley, line 178:
```suggestion
the span status to `Ok` when the `http.request.status_code` attribute is 200:
```
I understand it's just an example, but would using an actual HTTP success status code make more sense? It jumped out to me that we're setting the span status to `Ok` when we see a failure status code.

Also, we need to make sure the attribute here lines up with the example.

content/en/docs/collector/transforming-telemetry.md
```diff
@@ -169,6 +169,31 @@ processors:
...
+
+```yaml
+# The description field allows to set a human-readable message for errors.
+transform:
+  error_mode: ignore
+  trace_statements:
+    - context: span
+      statements:
+        - set(status.code, STATUS_CODE_OK) where
+          attributes["http.request.status_code"] == 400
```
evan-bradley, line 188:
```suggestion
        - set(status.code, STATUS_CODE_OK) where attributes["http.request.status_code"] == 200
```
Not super important, and it will work like this, but OTTL statements are generally written on a single line. I realize it's a bit long, but is it hard to read in the rendered version if it's like this?

Also, same as another comment, I think we should use a success code or try to explain a scenario where we want failure statuses to be set with an `Ok` span status.

content/en/docs/collector/transforming-telemetry.md
```diff
@@ -169,6 +169,31 @@ processors:
...
+## Setting a span status
+
+**Processor**:
+[transform processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/transformprocessor)
+
+Use the transform processor to set a span's status. The following example sets
+the span status to `Ok` when the `http.status_code` attribute is 400:
+
+```yaml
+# The description field allows to set a human-readable message for errors.
```
evan-bradley, line 181:
```suggestion
```
We're not modifying the description field here, should we include this comment?

content/en/docs/collector/transforming-telemetry.md
```diff
@@ -169,6 +169,31 @@ processors:
...
+      statements:
+        - set(status.code, STATUS_CODE_OK) where
+          attributes["http.request.status_code"] == 400
+```
+
+You can also use the transform processor to modify the span name based on its
+attributes or extract span attributes from the span name. For examples, see
+`transform processor` test data
+[config.yaml](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/transformprocessor/testdata/config.yaml)
+file.
```
evan-bradley, line 195:
```suggestion
attributes or extract span attributes from the span name. For examples, see an example
[config file](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/9b28f76c02c18f7479d10e4b6a95a21467fd85d6/processor/transformprocessor/testdata/config.yaml)
file for the transform processor.
```
Mostly just minor rewording, feel free to change this if you think it could be improved.

Also, I think we should pin this link to an exact commit. As it stands, if this file is moved in the contrib repo on the main branch, the link will break. I found a commit that should work, but another commit would work as well.

evan-bradley (commented):
Thanks for adding this. I think this mostly looks good, just a few notes.

content/en/docs/collector/transforming-telemetry.md
svrnm: This is intentional, the 400 has probably set `Error` as status code in the SDK code, and then in the collector we want to revert that, explicitly stating that those errors are OK for us.

I agree with the missing `request`

content/en/docs/collector/transforming-telemetry.md
svrnm: I have no concerns with a long code, will eave that to @Gaurang-Patel , @TylerHelmuth and @mx-psi , but see my comment above why 400

content/en/docs/collector/transforming-telemetry.md
svrnm: Ideally we point people to the most recent version of that file. @chalin if that file gets moved, are we getting informed about that with a 404? I don't think that we are doing that right now.

content/en/docs/collector/transforming-telemetry.md
evan-bradley: >This is intentional, the 400 has probably set `Error` as status code in the SDK code, and then in the collector we want to revert that, explicitly stating that those errors are OK for us.

Got it, thanks for the explanation. Could we include a short description of our goal here, then? My first impression was that the span was starting with an `Unknown` status code or similar and we wanted to populate it with the correct information.

content/en/docs/collector/transforming-telemetry.md
chalin: We aren't currently checking and then caching links into GH repos. Do we want to start doing that now?

svrnm: I think especially for links to other otel resources this would be great to have, I am not that worried about outside sources (yet), if this makes things easier in any way

content/en/docs/collector/transforming-telemetry.md
TylerHelmuth: Agree with a single line.  Although not a requirement of OTTL, it is idiomatic.

content/en/docs/collector/transforming-telemetry.md
svrnm: Yes, it would make sense to be a little bit clearer here.

content/en/docs/collector/transforming-telemetry.md
Gaurang-Patel: I wrote it in single line only but actually it failed formating test , any way we can exclude this line from the test if we want to go with single line ?

content/en/docs/collector/transforming-telemetry.md
chalin: I created the following issue to track the additional checking:

- #3994

content/en/docs/collector/transforming-telemetry.md
Gaurang-Patel: thanks @svrnm .

content/en/docs/collector/transforming-telemetry.md
svrnm:
```suggestion
the span status to `Ok` when the `http.request.status_code` attribute is 200:
```

@Gaurang-Patel please apply this change

content/en/docs/collector/transforming-telemetry.md
svrnm: @Gaurang-Patel please apply this change

content/en/docs/collector/transforming-telemetry.md
Gaurang-Patel: Hi @svrnm , it will be like 
```the span status to `Ok` when the `http.request.status_code` attribute is 400:```
 correct? as in example the condition is on 400.

svrnm: exactly! All that needs to be added is that "request"

TylerHelmuth (approved):

evan-bradley (approved):

PR #3774, opened 2024-01-15
added lumigo distro for python, node and java to the registry

developersteve:
I am adding lumigo Python, Java and Node.JS language instrumentations (distributions) to the registry.

Note: There didn't appear to be a `distribution` registry type yet, so I've listed them as instrumentations. Is `distribution` something that is going to be added as a registry type (or can it?)

### DIFF
--- a/data/registry/instrumentation-java-lumigo.yml
+++ b/data/registry/instrumentation-java-lumigo.yml
@@ -1,0 +1,14 @@
+title: Lumigo Java Distribution
+registryType: instrumentation
+isThirdParty: true
+language: java
+tags:
+  - java
+  - instrumentation
+  - lumigo
+  - distribution
+repo: https://github.com/lumigo-io/opentelemetry-java-distro
+license: Apache 2.0
+description: The lumigo Java OpenTelemetry distribution
+authors: lumigo, lumigo authors
+otVersion: latest
--- a/data/registry/instrumentation-js-node-lumigo.yml
+++ b/data/registry/instrumentation-js-node-lumigo.yml
@@ -1,0 +1,15 @@
+title: Lumigo Node.js Distribution
+registryType: instrumentation
+isThirdParty: true
+language: js
+tags:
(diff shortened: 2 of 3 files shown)


svrnm:
Just as an heads-up, we have an ongoing PR to refresh the whole registry, which also changes some of the file formats, so depending on the process of that we might need to figure out a merge order: https://github.com/open-telemetry/opentelemetry.io/pull/3730/

> Note: There didn't appear to be a `distribution` registry type yet, so I've listed them as instrumentations. Is `distribution` something that is going to be added as a registry type (or can it?)

As of today honeycomb seems to be the only distribution that we have in our registry (see https://opentelemetry.io/ecosystem/registry/?s=distribution&component=&language=). We might consider having it, but I am not sure if just adding a registry type `distribution` is doing the job. Let's keep `instrumentation` for now.

theletterf:
@svrnm Interesting. Shall we perhaps think of listing distros in a separate place / category? Or would that perhaps be too vendor-focused?

svrnm:
> @svrnm Interesting. Shall we perhaps think of listing distros in a separate place / category? Or would that perhaps be too vendor-focused?

I am not 100% sure, I see two solutions:
* We incorporate them into https://opentelemetry.io/ecosystem/vendors/
* We add them to the registry & at the same time think about adding all the vendors to the registry as well (and by that replace the /vendors/ page with a restricted search on the registry)

theletterf:
I'll open an issue on this.

svrnm:
@developersteve please take a look at https://github.com/open-telemetry/opentelemetry.io/pull/3854, @theletterf can add the lumigo repo in that initial PR for you. Will close this one for now.

theletterf:
Added in https://github.com/open-telemetry/opentelemetry.io/pull/3854/commits/f8e92e6eb3d16b87286f1ba3e2a530d3db7e42f3

PR #3775, opened 2024-01-15
Remove outdated templates

svrnm:
Those files are outdated and haven't been used in a long time, let's simply remove them


theletterf (approved):
I believe we no longer mention them anywhere, do we?

We do link the archetype from the blog contrib guide: https://opentelemetry.io/docs/contributing/blog/

svrnm:
> I believe we no longer mention them anywhere, do we?

No we don't and those pages have changed significantly. If we ever add a new language my suggestion is to copy from another language and work from there.

PR #3821, opened 2024-01-19
Registry updates Jan 2024

svrnm:
Adding some files, updating some versions, fixing scripts to work with the changed registry entry format.


theletterf:
Hmm, seems like the linter and spellcheckers are complaining...

chalin (approved):
RSLGTM

Issue #3229, opened 2023-09-06
[Meta] Standardize manual instrumentation docs with sample app


svrnm:
Since we are very close to have standardized the "Getting Started" docs for all languages (see https://github.com/open-telemetry/opentelemetry.io/issues/2623), we can do the same now for other pages as well. Manual Instrumentation is the next important one, so I create a ticket similar to #2623.

Each page should have a structure like the following:

- Introduction (standardized already)
- Advanced "roll the dice" example app
- Setup (language specific, load dependencies, etc.), initialize the SDK
- Traces
- Metrics
- Logs

See the JavaScript documentation for reference:

https://opentelemetry.io/docs/instrumentation/js/manual/

Tracking:

- C++
  - [ ] Example app
  - [ ] Setup
  - [ ] Traces
  - [ ] Metrics
  - [ ] Logs
- .NET
  - [ ] Example app
  - [ ] Setup
  - [ ] Traces
  - [ ] Metrics
  - [ ] Logs
- Erlang/Elixir
  - [ ] Example app
  - [ ] Setup
  - [ ] Traces
  - [ ] Metrics
  - [ ] Logs
- Go
  - [ ] Example app
  - [ ] Setup
  - [ ] Traces
  - [ ] Metrics
  - [ ] Logs
- Java (WIP: https://github.com/open-telemetry/opentelemetry.io/pull/3144)
  - [ ] Example app
  - [ ] Setup
  - [ ] Traces
  - [ ] Metrics
  - [ ] Logs
- JavaScript
  - [x] Example app
  - [x] Setup
  - [x] Traces
  - [ ] Metrics
  - [ ] Logs
- PHP
  - [ ] Example app
  - [ ] Setup
  - [ ] Traces
  - [ ] Metrics
  - [ ] Logs
- Python
  - [ ] Example app
  - [ ] Setup
  - [ ] Traces
  - [ ] Metrics
  - [ ] Logs
- Ruby
  - [ ] Example app
  - [ ] Setup
  - [ ] Traces
  - [ ] Metrics
  - [ ] Logs
- Rust
  - [ ] Example app
  - [ ] Setup
  - [ ] Traces
  - [ ] Metrics
  - [ ] Logs
- Swift
  - [ ] Example app
  - [ ] Setup
  - [ ] Traces
  - [ ] Metrics
  - [ ] Logs

theletterf:
I'd like to help with this.

IAMebonyhope:
[...]

PR #3834, opened 2024-01-20
Update .NET manual instrumentation docs with example app

IAMebonyhope:
This is to solve the .Net aspect of this issue #3229 
Completed the Example App, Setup and Traces section.

I can add metrics and logs in another PR.


svrnm:
thanks for kicking this off @IAMebonyhope ! 

Here's the preview: https://deploy-preview-3834--opentelemetry.netlify.app/docs/languages/net/instrumentation/

cc @open-telemetry/docs-approvers , @open-telemetry/dotnet-approvers

content/en/docs/languages/net/instrumentation.md
```diff
@@ -22,6 +36,218 @@ are covered here as well as the `System.Diagnostics` API.
...
+```
+
+## Manual instrumentation setup
+
+### Dependencies
+
+Install the [OpenTelemetry API and SDK NuGet packages](https://www.nuget.org/profiles/OpenTelemetry):
+
+```sh
+dotnet add package OpenTelemetry
```
cijothomas, line 173: `OpenTelemetry` is not required as it is brought automatically.

content/en/docs/languages/net/instrumentation.md
```diff
@@ -22,6 +36,218 @@ are covered here as well as the `System.Diagnostics` API.
...
+
+```text
+[5,6,5,3,6,1,2,5,4,4,2,4]
+```
+
+## Manual instrumentation setup
+
+### Dependencies
+
+Install the [OpenTelemetry API and SDK NuGet packages](https://www.nuget.org/profiles/OpenTelemetry):
```
cijothomas, line 170: https://www.nuget.org/profiles/OpenTelemetry -- this link is not really helpful here. Please point to the specific packages.

content/en/docs/languages/net/instrumentation.md
```diff
@@ -22,6 +36,218 @@ are covered here as well as the `System.Diagnostics` API.
...
+dotnet add package OpenTelemetry.Extensions.Hosting
+```
+
+### Initialize the SDK
+
+{{% alert title="Note" color="info" %}} If you’re instrumenting a library,
+**skip this step**. {{% /alert %}}
+
+Before any other module in your application is loaded, you must initialize the SDK. 
+If you fail to initialize the SDK or initialize it too late, no-op
```
cijothomas, line 184: This may not be true.. If the initialization is late, then at that point onwards instrumentations work.

content/en/docs/languages/net/instrumentation.md
```diff
@@ -22,6 +36,218 @@ are covered here as well as the `System.Diagnostics` API.
...
+```sh
+dotnet add package OpenTelemetry
+dotnet add package OpenTelemetry.Exporter.Console
+dotnet add package OpenTelemetry.Extensions.Hosting
+```
+
+### Initialize the SDK
+
+{{% alert title="Note" color="info" %}} If you’re instrumenting a library,
+**skip this step**. {{% /alert %}}
```
cijothomas, line 181: not clear what is "this step" referring to. Lets point to the next concrete step instead.

content/en/docs/languages/net/instrumentation.md
```diff
@@ -22,6 +36,218 @@ are covered here as well as the `System.Diagnostics` API.
...
+information, see [Resources](/docs/languages/net/resources/).
+
+To verify your code, build and run the app:
+
+```sh
+dotnet build
+dotnet run
+```
+
+This basic setup has no effect on your app yet. You need to add code for
```
cijothomas, line 248: "no effect" is not true for Logs. Asp.Net Core components could be logging and they are now sent via OpenTelemetry.

content/en/docs/languages/net/instrumentation.md
```diff
@@ -7,6 +7,20 @@ description: Instrumentation for OpenTelemetry .NET
...
+
+On this page you will learn how you can add traces, metrics and logs to your
+code _manually_. But, you are not limited to only use one kind of
+instrumentation: use [automatic instrumentation](/docs/languages/net/automatic/)
+to get started and then enrich your code with manual instrumentation as needed.
+
+Also, for libraries your code depends on, you don't have to write
+instrumentation code yourself, since they might come with OpenTelemetry built-in
+_natively_ or you can make use of
+[instrumentation libraries](/docs/languages/net/libraries/).
```
theletterf, line 20:
```suggestion
On this page you will learn how you can add traces, metrics and logs to your
code manually. You are not limited to using one kind of instrumentation:
you can also use [automatic instrumentation](/docs/languages/net/automatic/)
to get started and then enrich your code with manual instrumentation as needed.

Also, for libraries your code depends on, you don't have to write
instrumentation code yourself, since they might come with OpenTelemetry
or you can make use of
[instrumentation libraries](/docs/languages/net/libraries/).
```

@svrnm Not sure if the include has the same wording everywhere, but it shouldn't use italics for emphasis.

content/en/docs/languages/net/instrumentation.md
```diff
@@ -22,6 +36,218 @@ are covered here as well as the `System.Diagnostics` API.
...
+
+- [.NET SDK](https://dotnet.microsoft.com/download/dotnet) 6+
+
+### Create and launch an HTTP Server
+
+To highlight the difference between instrumenting a _library_ and a standalone
+_app_, split out the dice rolling into a _library file_, which then will be
+imported as a dependency by the _app file_.
+
+Create the _library file_ named `Dice.cs` and add the following code to it:
```
theletterf, line 58: Please remove all the italics/emphasis.

content/en/docs/languages/net/instrumentation.md
```diff
@@ -22,6 +36,218 @@ are covered here as well as the `System.Diagnostics` API.
...
+                logger.LogInformation("{player} is rolling the dice: {result}", player, result);
+            }
+
+            return result;
+        }
+    }
+}
+```
+
+Replace the program.cs content with the following code:
```
theletterf, line 138:
```suggestion
Replace the content of the program.cs file with the following code:
```

content/en/docs/languages/net/instrumentation.md
theletterf: Also, no bold styling, please.

content/en/docs/languages/net/instrumentation.md
```diff
@@ -22,6 +36,218 @@ are covered here as well as the `System.Diagnostics` API.
...
+### Initialize the SDK
+
+{{% alert title="Note" color="info" %}} If you’re instrumenting a library,
+**skip this step**. {{% /alert %}}
+
+Before any other module in your application is loaded, you must initialize the SDK. 
+If you fail to initialize the SDK or initialize it too late, no-op
+implementations will be provided to any library that acquires a tracer or meter from the API.
+
+To intialize the sdks, replace the program.cs content with the following code:
```
theletterf, line 187:
```suggestion
To initialize the SKDs, replace the content of the program.cs file with the
following code: 
```

content/en/docs/languages/net/instrumentation.md
```diff
@@ -22,6 +36,218 @@ are covered here as well as the `System.Diagnostics` API.
...
+
+app.Run();
+
+```
+
+For debugging and local development purposes, the following example exports
+telemetry to the console. After you have finished setting up manual
+instrumentation, you need to configure an appropriate exporter to
+[export the app's telemetry data](/docs/languages/net/exporters/) to one or more
+telemetry backends.
```
theletterf, line 231:
```suggestion
telemetry back ends.
```

content/en/docs/languages/net/instrumentation.md
```diff
@@ -22,6 +36,218 @@ are covered here as well as the `System.Diagnostics` API.
...
+
+The example also sets up the mandatory SDK default attribute `service.name`,
+which holds the logical name of the service, and the optional (but highly
+encouraged!) attribute `service.version`, which holds the version of the service
+API or implementation.
+
```
theletterf, line 237:
```suggestion
The example also sets up the mandatory SDK default attribute `service.name`,
which holds the logical name of the service, and the optional, but highly
encouraged, attribute `service.version`, which holds the version of the service
API or implementation.
```

theletterf (changes requested):
Thanks! Some language and formatting suggestions.

content/en/docs/languages/net/instrumentation.md
IAMebonyhope: Yes, it is the same wording everywhere (currently js and java) I just changed the references for .Net where necessary.

content/en/docs/languages/net/instrumentation.md
```diff
@@ -7,6 +7,20 @@ description: Instrumentation for OpenTelemetry .NET
...
 
+{{% alert title="Note" color="info" %}}
+
+On this page you will learn how you can add traces, metrics and logs to your
+code manually. You are not limited to using one kind of instrumentation:
+you can also use [automatic instrumentation](/docs/languages/net/automatic/)
+to get started and then enrich your code with manual instrumentation as needed.
+
+Also, for libraries your code depends on, you don't have to write
+instrumentation code yourself, since they might come with OpenTelemetry
```
reyang, line 18: Minor, I find "come with OpenTelemetry" a bit confusing, consider:

"since they might be already instrumented or there are instrumentation libraries for them"

content/en/docs/languages/net/instrumentation.md
```diff
@@ -22,361 +36,583 @@ are covered here as well as the `System.Diagnostics` API.
...
 ```csharp
-using OpenTelemetry;
-using OpenTelemetry.Resources;
-using OpenTelemetry.Trace;
+/*DiceController.cs*/
+using Microsoft.AspNetCore.Mvc;
+using System.Net;
 
-// ...
+namespace otel
```
reyang, line 102: Do we need a namespace here? (what value does it provide to the user)
If we need a namespace, consider following the .NET convention and use upper camel case.

content/en/docs/languages/net/instrumentation.md
```diff
@@ -7,6 +7,20 @@ description: Instrumentation for OpenTelemetry .NET
 
 {{% docs/languages/manual-intro %}}
 
+{{% alert title="Note" color="info" %}}
+
+On this page you will learn how you can add traces, metrics and logs to your
+code manually. You are not limited to using one kind of instrumentation:
+you can also use [automatic instrumentation](/docs/languages/net/automatic/)
+to get started and then enrich your code with manual instrumentation as needed.
```
cijothomas, line 15: confirm with auto-instrumentation owners, that it is feasible to have auto-instrumentation and enrichment via manual at the same time.

content/en/docs/languages/net/instrumentation.md
```diff
@@ -22,361 +36,581 @@ are covered here as well as the `System.Diagnostics` API.
...
+app.MapControllers();
+
+app.Run();
+```
+
+To ensure that it is working, run the application with the following command and
+open <http://localhost:8080/rolldice?rolls=12> in your web browser:
+
+```sh
+dotnet build
```
cijothomas, line 157: nit : build is not required explicitly as run will do it automatically.

content/en/docs/languages/net/instrumentation.md
```diff
@@ -22,361 +36,581 @@ are covered here as well as the `System.Diagnostics` API.
...
+          serviceVersion: serviceVersion))
+    .AddConsoleExporter()
+    .Build();
+
+var meterProvider = Sdk.CreateMeterProviderBuilder()
+    .AddMeter(serviceName)
+    .AddConsoleExporter()
+    .Build();
+
+//...
```
cijothomas, line 266: logs omitted intentionally?

content/en/docs/languages/net/instrumentation.md
```diff
@@ -22,361 +36,581 @@ are covered here as well as the `System.Diagnostics` API.
...
+    int result;
+    var childActivity = MyActivitySource.StartActivity("rollOnce");
+
+    try
+    {
+        result = Random.Shared.Next(min, max + 1);
+    }
+    catch (Exception ex)
+    {
+        childActivity?.SetStatus(ActivityStatusCode.Error, ex.Message);
```
cijothomas, line 602: is it recommended to set status description as ex.Message?

content/en/docs/languages/net/instrumentation.md
```diff
@@ -22,361 +36,581 @@ are covered here as well as the `System.Diagnostics` API.
...
 
 ```sh
-dotnet add package OpenTelemetry
-dotnet add package OpenTelemetry.Extensions.Hosting
 dotnet add package OpenTelemetry.Exporter.Console
+dotnet add package OpenTelemetry.Extensions.Hosting
 ```
 
-Then you can install the Instrumentation package
+For ASP.NET Core-based applications, also install the AspNetCore instrumentation package
```
cijothomas, line 183: this is not used in the rest of the code... missed to add?

content/en/docs/languages/net/instrumentation.md
Kielek: See https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/main/docs/manual-instrumentation.md

content/en/docs/languages/net/instrumentation.md
Kielek: consider https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-10.0/file-scoped-namespaces

It reduces nesting.

content/en/docs/languages/net/instrumentation.md
```diff
@@ -22,361 +36,581 @@ are covered here as well as the `System.Diagnostics` API.
...
+
+            return result;
+        }
+    }
+}
+```
 
-var serviceName = "MyServiceName";
-var serviceVersion = "1.0.0";
+Replace the content of the program.cs file with the following code:
```
Kielek, line 139:
```suggestion
Replace the content of the Program.cs file with the following code:
```

svrnm:
All, thanks for providing code reviews! @IAMebonyhope please take a look at the feedback, let us know if you have any questions around any comment provided. Also, do not worry too much about the not successful jobs, we can fix them later.

content/en/docs/languages/net/instrumentation.md
IAMebonyhope: Yes, I forgot to include it in the "Initialize the SDK" section.

I initially thought it wasnt needed but I noticed the code couldnt generate traces until I included the package.

content/en/docs/languages/net/instrumentation.md
IAMebonyhope: I missed it. 

Thanks for pointing it out 🙂

content/en/docs/languages/net/instrumentation.md
IAMebonyhope: ohh no. 
It is probably not even a good idea 🙂.

IAMebonyhope:
Thanks everyone for the review.
I pushed new iterations incorporating the feedback.

svrnm:
thanks @IAMebonyhope! @open-telemetry/dotnet-approvers please take another look!

content/en/docs/languages/net/instrumentation.md
```diff
@@ -22,39 +35,219 @@ are covered here as well as the `System.Diagnostics` API.
...
+
+## Manual instrumentation setup
+
+### Dependencies
+
+Install the following OpenTelemetry NuGet packages:
+
+[OpenTelemetry.Exporter.Console](https://www.nuget.org/packages/OpenTelemetry.Exporter.Console)
+
+[OpenTelemetry.Extensions.Hosting](https://www.nuget.org/packages/OpenTelemetry.Extensions.Hosting)
```
svrnm, line 171: @cijothomas similar to the PR on exporters, I assume also here the Extensions.Hosting is only for ASP.NET core?

content/en/docs/languages/net/instrumentation.md
```diff
@@ -22,39 +35,219 @@ are covered here as well as the `System.Diagnostics` API.
...
+[OpenTelemetry.Instrumentation.AspNetCore](https://www.nuget.org/packages/OpenTelemetry.Instrumentation.AspNetCore)
+
+```sh
+dotnet add package OpenTelemetry.Instrumentation.AspNetCore
+```
+
+
+### Initialize the SDK
+
+{{% alert title="Note" color="info" %}} If you’re instrumenting a library,  you dont need to initialize the SDK. {{% /alert %}}
```
svrnm, line 190:
```suggestion
{{% alert title="Note" color="info" %}} If you’re instrumenting a library, you dont need to initialize the SDK. {{% /alert %}}
```

content/en/docs/languages/net/instrumentation.md
```diff
@@ -63,320 +256,362 @@ using var tracerProvider = Sdk.CreateTracerProviderBuilder()
...
-using OpenTelemetry;
-using OpenTelemetry.Resources;
-using OpenTelemetry.Trace;
+### Initialize Tracing
 
-// Define some important constants and the activity source.
-// These can come from a config file, constants file, etc.
-var serviceName = "MyCompany.MyProduct.MyService";
-var serviceVersion = "1.0.0";
+{{% alert title="Note" color="info" %}} If you’re instrumenting a library, you dont need to initialize a TraceProvider. {{% /alert %}}
```
pellared, line 301:
```suggestion
{{% alert title="Note" color="info" %}} If you’re instrumenting a library, you don't need to initialize a TraceProvider. {{% /alert %}}
```

content/en/docs/languages/net/instrumentation.md
```diff
@@ -22,39 +35,219 @@ are covered here as well as the `System.Diagnostics` API.
...
+
+You don't have to use the example app: if you want to instrument your own app or
+library, follow the instructions here to adapt the process to your own code.
+
+### Dependencies {#example-app-dependencies}
+
+- [.NET SDK](https://dotnet.microsoft.com/download/dotnet) 6+
+
+### Create and launch an HTTP Server
+
```
svrnm, line 52:
```suggestion
### Create and launch an HTTP Server

To begin, set up an environment in a new directory called `dotnet-otel-example`. Within that directory, execute following command:

```shell
dotnet new web
```
```

File: content/en/docs/languages/net/instrumentation.md [hunk 24]
```diff
@@ -22,39 +35,219 @@ are covered here as well as the `System.Diagnostics` API.
...
+builder.Services.AddControllers();
+
+var app = builder.Build();
+
+app.MapControllers();
+
+app.Run();
+```
+
```
svrnm, line 149:
~~~suggestion

In the `Properties` subdirectory, replace the content of `launchSettings.json` with the following:

```json
{
  "$schema": "http://json.schemastore.org/launchsettings.json",
  "profiles": {
    "http": {
      "commandName": "Project",
      "dotnetRunMessages": true,
      "launchBrowser": true,
      "applicationUrl": "http://localhost:8080",
      "environmentVariables": {
        "ASPNETCORE_ENVIRONMENT": "Development"
      }
    }
  }
}
```

~~~

content/en/docs/languages/net/instrumentation.md
```diff
@@ -22,39 +35,219 @@ are covered here as well as the `System.Diagnostics` API.
...
+
+### Initialize the SDK
+
+{{% alert title="Note" color="info" %}} If you’re instrumenting a library, you dont need to initialize the SDK. {{% /alert %}}
+
+It is important to configure an instance of the OpenTelemetry SDK as early as possible in your application.
+
+To initialize the OpenTelemetry SDK for an ASP.NET Core app like in the case of the example app, 
+replace the content of the program.cs file with the following code:
```
svrnm, line 195:
```suggestion
update the content of the `Program.cs` file with the following code: 
```

content/en/docs/languages/net/instrumentation.md
```diff
@@ -22,39 +35,219 @@ are covered here as well as the `System.Diagnostics` API.
...
+replace the content of the program.cs file with the following code: 
+
+```csharp
+using OpenTelemetry.Logs;
+using OpenTelemetry.Metrics;
+using OpenTelemetry.Resources;
+using OpenTelemetry.Trace;
+
+// Ideally, you will want this name to come from a config file, constants file, etc.
+var serviceName = "Dice.*";
```
svrnm, line 204:
```suggestion
var serviceName = "dice-server";
```

content/en/docs/languages/net/instrumentation.md
```diff
@@ -63,320 +256,362 @@ using var tracerProvider = Sdk.CreateTracerProviderBuilder()
...
 }
+
 ```
+Note, that it’s required to `Stop()` the activity, otherwise it will not be exported.
+
+If you followed the instructions using the [example app](#example-app) up to
+this point, you can copy the code above in your library file `Dice.cs`. You
+should now be able to see activities/spans emitted from your app.
+
+Start your app as follows, and then send it requests by visiting http://localhost:8080/rolldice?rolls=12 with your browser or curl.
```
svrnm, line 418:
```suggestion
Start your app as follows, and then send it requests by visiting <http://localhost:8080/rolldice?rolls=12> with your browser or curl.
```

svrnm (changes requested):
overall lgtm, some change requests around functionality and formatting

content/en/docs/languages/net/instrumentation.md
IAMebonyhope: @svrnm if we change this "dice-server" then the name to create the ActivitySource for both the server and the lib will have to be "dice-server" as well.

svrnm: let me double check on that, might be that .NET behaves here differently compared to other languages

svrnm: Ok... so if I understand it correctle then, the ActivitySource name should be the instrumentation scope name, not the service name, so they should be different.

svrnm:
@open-telemetry/dotnet-approvers please take another look, would be great if we can merge this PR eventually!

content/en/docs/languages/net/instrumentation.md
svrnm: @open-telemetry/dotnet-approvers please take a look, I am no expert on that matter here, might be my misunderstanding

content/en/docs/languages/net/instrumentation.md
martinjt: Hosting is for anything that can take BackgroundServices and Dependency Injection, so you can use that with console apps.

content/en/docs/languages/net/instrumentation.md
martinjt: ActivitySource == Tracer
Activity == Span

Since it's normally 1 tracer per app, I'd say `dice-server` is fine?

content/en/docs/languages/net/instrumentation.md
```diff
@@ -63,320 +279,362 @@ using var tracerProvider = Sdk.CreateTracerProviderBuilder()
...
     }
 
-    // 'child1' and 'child2' both share 'parent' as a parent, but are independent
-    // from one another
+    activity?.Stop();
+    return results;
 }
+
+```
+Note, that it’s required to `Stop()` the activity, otherwise it will not be exported.
```
martinjt, line 435: I'm not sure I'd put this, it's more accepted practice to us a using statement that will automatically dispose of the activity and therefore send it.

content/en/docs/languages/net/instrumentation.md
```diff
@@ -63,320 +279,362 @@ using var tracerProvider = Sdk.CreateTracerProviderBuilder()
...
+using System.Net;
 
-```csharp
-public static void ParentOperation()
+public class DiceController : ControllerBase
 {
-    using var parentActivity = MyActivitySource.StartActivity("ParentActivity");
+    private ILogger<DiceController> logger;
+    
+    public static readonly ActivitySource MyActivitySource = new("Dice.Server", "1.0.0");
```
martinjt, line 376: I'm not sure this sort of approach is what we should be promoting. The ActivitySource should be shared at an assembly level, not at a controller/class level.

IAMebonyhope: Hello @martinjt I am not totally clear on what you mean here.
Do you mean the ActivitySource should be initialized in Program.cs and then passed in here?

svrnm:
@open-telemetry/dotnet-approvers ptal

content/en/docs/languages/net/instrumentation.md
cijothomas: It'd nice to avoid creating new and new ActivitySource per controller/class, and instead use a single one throughout. Unless there is a reason to separate it out, like enabling/disabling telemetry per ActivitySource. For this doc, it is best to use a single ActivitySource.

https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/examples/AspNetCore Shows one way of doing this in an asp.net core application.

cijothomas:
> @open-telemetry/dotnet-approvers ptal

Given there a lot of comments, perhaps its best to reduce scope and split to multiple, smaller PRs, to unblock progress.

Also, is there a view to see the rendered content somehow? Such a view will help us better review relatively big PRs like this, as opposed to just git diffs!

content/en/docs/languages/net/instrumentation.md
```diff
@@ -63,320 +279,362 @@ using var tracerProvider = Sdk.CreateTracerProviderBuilder()
     .AddConsoleExporter()
     .Build();
 
-// ...
-```
-
-This is also where you can configure instrumentation libraries.
+var meterProvider = Sdk.CreateMeterProviderBuilder()
```
cijothomas, line 282: since we are removing "using" keyword from here, we need to explicitly dispose() at the end of the application, else metrics won't be flushed to console! (Similar concern for logs/traces too, though logs/traces will still show up in console even without dispose, but we should show the right thing that works across all exporters.)
in short - please add explicit .Dispose() call on meterProvider, tracerProvider, loggerFactory at the end of the application.

content/en/docs/languages/net/instrumentation.md
```diff
@@ -63,320 +279,362 @@ using var tracerProvider = Sdk.CreateTracerProviderBuilder()
...
-public static void DoWork()
+public List<int> rollTheDice(int rolls)
 {
-    using var parent = MyActivitySource.StartActivity("parent");
-
-    using (var child1 = DemoSource.StartActivity("child1"))
-    {
-        // Do some work that 'child1' tracks
-    }
+    var activity = MyActivitySource.StartActivity("rollTheDice");
```
cijothomas, line 422: I'd also suggest to add a note here saying something like:
The example here is creating a new Span for demonstration purposes only. It is recommended to create Spans, only if it is doing something worth measuring independently. Too many spans makes it harder to visualize in tools like Jaeger.

https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/examples/AspNetCore/Controllers/WeatherForecastController.cs#L45-L52 - These wording maybe borrowed.

svrnm:
> > @open-telemetry/dotnet-approvers ptal
> 
> Given there a lot of comments, perhaps its best to reduce scope and split to multiple, smaller PRs, to unblock progress.

The core goal of this PR is to get the .NET instrumentation docs consistent with other languages, that's why it may appear very big in scope, but if you see ways to split it up to accelerate the progress, I am open for that. @IAMebonyhope wdyt?

> Also, is there a view to see the rendered content somehow? Such a view will help us better review relatively big PRs like this, as opposed to just git diffs!

https://deploy-preview-3834--opentelemetry.netlify.app/ -- this is generated for every PR, you can find that link by scrolling in the Checks View and click on Details for "netlify/opentelemetry/deploy-preview"

IAMebonyhope:
> > > @open-telemetry/dotnet-approvers ptal
> > 
> > 
> > Given there a lot of comments, perhaps its best to reduce scope and split to multiple, smaller PRs, to unblock progress.
> 
> The core goal of this PR is to get the .NET instrumentation docs consistent with other languages, that's why it may appear very big in scope, but if you see ways to split it up to accelerate the progress, I am open for that. @IAMebonyhope wdyt?
> I am okay with splitting as well, but I am not sure what parts I can split out that wont make it confusing to the readers. I already reduced the scope by focusing on tracing alone.

> > Also, is there a view to see the rendered content somehow? Such a view will help us better review relatively big PRs like this, as opposed to just git diffs!
> 
> https://deploy-preview-3834--opentelemetry.netlify.app/ -- this is generated for every PR, you can find that link by scrolling in the Checks View and click on Details for "netlify/opentelemetry/deploy-preview"

content/en/docs/languages/net/instrumentation.md
```diff
@@ -63,320 +281,360 @@ using var tracerProvider = Sdk.CreateTracerProviderBuilder()
...
+First we instantiate the activitySource in the `program.cs`:
 
-```csharp
-using var myActivity = MyActivitySource.StartActivity("SayHello");
+```csharp { hl_lines=["2-3"]}
 
-// do work that 'myActivity' will now track
-```
+// Register the ActivitySource as a singleton to have a single instance of activitySource across the application
+builder.Services.AddSingleton<ActivitySource>(new ActivitySource(serviceName, serviceVersion));
```
cijothomas, line 376: this should not be done (`ActivitySource` is not a class defined/owned by this app, and should not be added to DI without wrapping inside a custom class on your own.)

See example from here : https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/examples/AspNetCore/Program.cs

IAMebonyhope: I ll update to include a new instrumentation class to wrap the ActivitySource

content/en/docs/languages/net/instrumentation.md
```diff
@@ -22,39 +35,244 @@ are covered here as well as the `System.Diagnostics` API.
...
+    {
+        this.logger = logger;
+    }
+
+    [HttpGet("/rolldice")]
+    public List<int> RollDice(string player, int? rolls)
+    {
+        if(!rolls.HasValue)
+        {
+            logger.LogError("Missing rolls parameter");
```
reyang, line 121: Avoid using extension methods from LoggerExtension, use [compile-time logging source generation](https://docs.microsoft.com/dotnet/core/extensions/logger-message-generator) pattern to achieve the best performance.

Check https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/docs/logs#use-logger

content/en/docs/languages/net/instrumentation.md
```diff
@@ -22,39 +35,244 @@ are covered here as well as the `System.Diagnostics` API.
...
+
+```
+
+If initializing the OpenTelemetry SDK for a console app,
+add the following code at the beginning of your program, during any important startup operations.
+
+```csharp
+using OpenTelemetry.Resources;
+using OpenTelemetry.Trace;
+using OpenTelemetry.Logs;
```
reyang, line 268: Alphabetic ordering

content/en/docs/languages/net/instrumentation.md
martinjt: While in general I agree, I don't think that's a widespread enough methodology at the moment, and would take away from what this example is trying to achieve.

Using `ILogger` in the way this example does will be familiar to enough people that it won't cause them to learn something new.

content/en/docs/languages/net/instrumentation.md
```diff
@@ -22,39 +35,244 @@ are covered here as well as the `System.Diagnostics` API.
...
+builder.Services.AddControllers();
+
+var app = builder.Build();
+
+app.MapControllers();
+
+app.Run();
+```
+
+In the `Properties` subdirectory, replace the content of `launchSettings.json` with the following:
```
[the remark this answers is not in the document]
martinjt: I agree that we shouldn't include this as it's purpose is only to ensure that we get a consistent port number. I don't think either is best practice for what it's worth Reiley, and I don't think that tone helps at all.

content/en/docs/languages/net/instrumentation.md
```diff
@@ -22,39 +35,244 @@ are covered here as well as the `System.Diagnostics` API.
...
 
 ```csharp
-using OpenTelemetry;
+using OpenTelemetry.Logs;
+using OpenTelemetry.Metrics;
 using OpenTelemetry.Resources;
 using OpenTelemetry.Trace;
 
-// ...
+// Ideally, you will want this name to come from a config file, constants file, etc.
```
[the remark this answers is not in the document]
martinjt: Because environment variables are the preferred way of configuring Otel, and constants would be a better way to achieve this if it's in code.

That said, I'm not sure commenting the code like this in the getting-started example in the Markdown is worthwhile as it will cause the reader to think about too much.

martinjt (commented):
It's definitely looking good @IAMebonyhope! Added some comments, feel free to tag me if you have more!

content/en/docs/languages/net/instrumentation.md
```diff
@@ -22,39 +35,244 @@ are covered here as well as the `System.Diagnostics` API.
...
+var builder = WebApplication.CreateBuilder(args);
+
+builder.Services.AddOpenTelemetry()
+    .ConfigureResource(resource => resource.AddService(
+        serviceName: serviceName,
+        serviceVersion: serviceVersion))
+    .WithTracing(tracing => tracing
+        .AddSource(serviceName)
+        .AddAspNetCoreInstrumentation()
+        .AddConsoleExporter())
```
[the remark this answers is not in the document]
martinjt: I don't think that's needed for the getting started docs. We have a standard across all docs that Hope is trying to mirror, not specific to .NET, so I think it's fine to not say that.

content/en/docs/languages/net/instrumentation.md
```diff
@@ -63,320 +281,388 @@ using var tracerProvider = Sdk.CreateTracerProviderBuilder()
...
-You can instantiate several `ActivitySource`s if that suits your scenario,
-although it is generally sufficient to just have one defined per service.
-
-### Creating Activities
+It’s generally recommended to define `ActivitySource` once per app/service that is been instrumented, but you can instantiate several `ActivitySource`s if that suits your scenario
 
-To create an [`Activity`](/docs/concepts/signals/traces/#spans), give it a name
-and create it from your
-[`ActivitySource`](/docs/concepts/signals/traces/#tracer).
+In the case of the example app, we will create a new file `Instrumentation.cs` as a custom type to hold reference for the ActivitySource.
```
cijothomas, line 369: it looks like there is already a `public static class Telemetry` shown above. That must be removed if going with `instrumentation.cs`?

content/en/docs/languages/net/instrumentation.md
```diff
@@ -63,320 +281,388 @@ using var tracerProvider = Sdk.CreateTracerProviderBuilder()
...
+using System.Diagnostics;
 
-    using (var child2 = DemoSource.StartActivity("child2"))
+public class Dice
+{
+    public ActivitySource activitySource;
+    private int min;
+    private int max;
+
+    public Dice(int min, int max, ActivitySource activitySource)
```
cijothomas, line 453: not able to see who is injecting ActivitySource to this constructor? Or is that instantiated manually? Can you add the code also?

content/en/docs/languages/net/instrumentation.md
```diff
@@ -63,320 +281,388 @@ using var tracerProvider = Sdk.CreateTracerProviderBuilder()
...
-`using` block is explicitly defined, rather than scoped to `DoWork` itself like
-`parentActivity`.
 
-### Creating independent Activities
+```
 
-The previous examples showed how to create Activities that follow a nested
-hierarchy. In some cases, you'll want to create independent Activities that are
-siblings of the same root rather than being nested.
+The same activitySource is also passed to the library file `Dice.cs`:
```
cijothomas, line 441: I think the code to do this is missing? You need to add something like new Dice(activitySource,min,max).

svrnm:
@IAMebonyhope anything you need to move this forward?

IAMebonyhope:
> @IAMebonyhope anything you need to move this forward?

@svrnm I'll work on resolving the comments this week

content/en/docs/languages/net/instrumentation.md
IAMebonyhope: @reyang could we address the use of OpenTelemetry .NET Log practices in the Logs section, which will be added in another PR?

content/en/docs/languages/net/instrumentation.md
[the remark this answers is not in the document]
IAMebonyhope: This was copied from the https://opentelemetry.io/docs/languages/net/getting-started/#example-application 
As this PR is building on top of that...

content/en/docs/languages/net/instrumentation.md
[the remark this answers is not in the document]
IAMebonyhope: We also included this note in line 305/306 below

"For debugging and local development purposes, the example exports
telemetry to the console. After you have finished setting up manual
instrumentation, you need to configure an appropriate exporter to
[export the app's telemetry data](/docs/languages/net/exporters/) to one or more
telemetry backends."

content/en/docs/languages/net/instrumentation.md
```diff
@@ -22,39 +35,244 @@ are covered here as well as the `System.Diagnostics` API.
...
+```
+
+In the `Properties` subdirectory, replace the content of `launchSettings.json` with the following:
+
+```json
+{
+  "$schema": "http://json.schemastore.org/launchsettings.json",
+  "profiles": {
+    "http": {
+      "commandName": "Project",
```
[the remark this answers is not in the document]
IAMebonyhope: Same comment as above

content/en/docs/languages/net/instrumentation.md
IAMebonyhope: Added it above

IAMebonyhope:
Pushed a new iteration and responded to comments @svrnm

@cijothomas @reyang please can you take another look?

content/en/docs/languages/net/instrumentation.md
```diff
@@ -22,361 +35,643 @@ are covered here as well as the `System.Diagnostics` API.
...
-            serviceVersion: serviceVersion))
-      .AddAspNetCoreInstrumentation()
-      .AddConsoleExporter();
-  });
-```
+    .ConfigureResource(resource => resource.AddService(
+        serviceName: serviceName,
+        serviceVersion: serviceVersion))
+    .WithTracing(tracing => tracing
+        .AddSource(serviceName)
```
cijothomas, line 240: AddSource should be using the name of the `ActivitySource`. In this example, it just happens that both service name and activitysource name are the same, but i'd recommend to explicitly make them same by calling something like AddSource(Instrumentation.ActivitySource.Name)

content/en/docs/languages/net/instrumentation.md
```diff
@@ -22,361 +35,643 @@ are covered here as well as the `System.Diagnostics` API.
...
-In the preceding example, `childActivity` is ended because the scope of the
-`using` block is explicitly defined, rather than scoped to `DoWork` itself like
-`parentActivity`.
+Then we will update the `Program.cs` to add the Instrument object as a dependency injection:
+
+```csharp 
+//...
+
+// Register the ActivitySource as a singleton in the DI container.
+// Create a service to expose the ActivitySource Instruments.
```
cijothomas, line 389:
```suggestion
// Create a service to expose the ActivitySource.
```

Unless you plan to add metric instruments as well to Instrumentation.cs?

content/en/docs/languages/net/instrumentation.md
```diff
@@ -22,361 +35,643 @@ are covered here as well as the `System.Diagnostics` API.
...
-We recommend that all Tag names are defined in constants rather than defined
-inline as this provides both consistency and also discoverability.
+After a while, you should see the spans printed in the console by the
+`ConsoleExporter`, something like this:
+
+```json
+Activity.TraceId:            841d70616c883db82b4ae4e11c728636
+Activity.SpanId:             9edfe4d69b0d6d8b
+Activity.TraceFlags:         Recorded
+Activity.ParentSpanId:       39fcd105cf958377
```
cijothomas, line 516: given this is the child Activity, would it better to show the parent activity also here.

content/en/docs/languages/net/instrumentation.md
```diff
@@ -22,361 +35,643 @@ are covered here as well as the `System.Diagnostics` API.
...
+      result = Random.Shared.Next(min, max + 1);
+        
+      return result;
+    }
+}
+
+```
 
-// ...
+When you view the spans in a trace visualization tool, `rollOnce` childActivity will be
```
cijothomas, line 553: for demonstrating the idea of child spans, this is very good! I'd also want to keep adequate warning here for users to avoid too much activity, as that'll expensive, but more importantly - makes visualization harder.

content/en/docs/languages/net/instrumentation.md
```diff
@@ -22,361 +35,643 @@ are covered here as well as the `System.Diagnostics` API.
...
+tracked as a nested operation under `rollTheDice` activity.
 
-myActivity?.AddEvent(new("Gonna try it!"));
+### Get the current Activity
 
-// ...
+Sometimes it’s helpful to do something with the current/active span at a particular point in program execution.
```
cijothomas, line 558:
```suggestion
Sometimes it’s helpful to do something with the current/active Activity/Span at a particular point in program execution.
```

content/en/docs/languages/net/instrumentation.md
```diff
@@ -22,361 +35,643 @@ are covered here as well as the `System.Diagnostics` API.
...
+      result = Random.Shared.Next(min, max + 1);
+      childActivity?.SetTag("dicelib.rolled", result);
+        
+      return result;
+    } 
+}
+```
 
-// ...
+### Create Activities with events
```
cijothomas, line 585:
```suggestion
### Add Events to Activities
```

cijothomas (approved):
Thanks. Left some more comments, but okay to address them as follow ups.

content/en/docs/languages/net/instrumentation.md
```diff
@@ -22,361 +35,643 @@ are covered here as well as the `System.Diagnostics` API.
...
+about manual instrumentation.
 
-There are two main ways to initialize [tracing](/docs/concepts/signals/traces/),
-depending on whether you're using a console app or something that's ASP.NET
-Core-based.
+You don't have to use the example app: if you want to instrument your own app or
+library, follow the instructions here to adapt the process to your own code.
 
-#### Console app
+### Dependencies {#example-app-dependencies}
```
reyang, line 47: Maybe use "Prerequisites"

content/en/docs/languages/net/instrumentation.md
IAMebonyhope: Removed that line and updated the above comment.

content/en/docs/languages/net/instrumentation.md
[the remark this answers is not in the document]
svrnm: Let's follow up on this in a separate discussion

content/en/docs/languages/net/instrumentation.md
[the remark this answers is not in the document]
svrnm: resolving this conversation now as well to merge in this PR, we can have a follow up discussion on that.

svrnm (approved):

svrnm:
Thanks @IAMebonyhope for all the effort put into this! I merged the PR and this should be live in a minute. I will also create a few follow up issues on the remaining non-blocking feedback

PR #3838, opened 2024-01-22
[Go] Bump to newer semconv

pellared:
Required because of https://github.com/open-telemetry/opentelemetry-go/pull/4754 (part of https://github.com/open-telemetry/opentelemetry-go/releases/tag/v1.22.0).

CC @open-telemetry/go-approvers


dmathieu (approved):

pellared:
@open-telemetry/go-approvers

Do you think that here

https://github.com/open-telemetry/opentelemetry.io/blob/6e3bf5839719f5e01dda8d90e8cdbc55f1f5c3e0/content/en/docs/languages/go/getting-started.md#L198

it would be better to use `resource.NewSchemaless` instead of `resource.NewWithAttributes` to mitigate problem that the users would have when bumping OTel Go?

Related to https://github.com/open-telemetry/opentelemetry-go/issues/2341

content/en/docs/languages/go/getting-started.md
```diff
@@ -104,7 +104,7 @@ go get "go.opentelemetry.io/otel" \
   "go.opentelemetry.io/otel/sdk/metric" \
   "go.opentelemetry.io/otel/sdk/resource" \
   "go.opentelemetry.io/otel/sdk/trace" \
-  "go.opentelemetry.io/otel/semconv/v1.21.0" \
+  "go.opentelemetry.io/otel/semconv/v1.24.0" \
```
svrnm, line 107: non blocking comment, @open-telemetry/go-approvers: we can implement an automatic version updating similar to what java, collector, etc have:

* https://github.com/open-telemetry/opentelemetry.io/blob/main/.github/workflows/auto-update-versions.yml
* https://github.com/open-telemetry/opentelemetry.io/blob/6e3bf5839719f5e01dda8d90e8cdbc55f1f5c3e0/content/en/docs/languages/java/_index.md?plain=1#L7

pellared: This could be nice. However, take notice that the version is not equal version of OTel Go. It may not be easy to create such script.

svrnm: Also for java we have different versions for SDK, semconv, etc. So we might be able to replicate that.

pellared: I also created https://github.com/open-telemetry/opentelemetry-go/pull/4846 so that at least the example would not break in runtime if the code uses an older version of semconv.

cartermp (approved):

cartermp:
@pellared feel free to merge when you feel you've gotten enough feedback from other folks on the Go SIG for your question. Otherwise, I'll plan to merge by end of week.

pellared:
> You’re not [authorized](https://docs.github.com/articles/about-protected-branches/) to merge this pull request.

@cartermp Can you merge it? This one is a trivial fix. It just propagates the changes: https://github.com/open-telemetry/opentelemetry-go/blob/main/example/dice/otel.go#L29 

[I will propose a solution that does not require bumping semconv package.](https://github.com/open-telemetry/opentelemetry-go/pull/4846#issuecomment-1904656729). This one would require more approvals from the Go SIG.

cartermp:
Sounds good.

PR #3844, opened 2024-01-22
Registry: put 'offical' languages first in dropdown

svrnm:
In the registry, when you click on the "languages" drop down you get a mix of languages otel official supports (the ones we have SIGs around) and some languages that have 3rd party support. This PR makes sure that the "official" languages are put first.


cartermp (approved):
Collector isn't a language (okay, technically OTTL is), but I get why it's put in there since we don't have a way to split based on collector and SDK.

PR #3847, opened 2024-01-23
Replace or add loopback address in Collector OTLP examples

### LINKED ISSUE #3839: Explicitly set endpoint host to 0.0.0.0 in all Collector examples
**What needs to be changed?** 

Per the poll at https://github.com/open-telemetry/opentelemetry-collector/issues/8510#issuecomment-1737598294, the Collector will eventually change the default endpoint host in all servers from `0.0.0.0` to `localhost`. Before this happens, I would like to update all examples in opentelemetry.io to explicitly set the endpoint to `0.0.0.0` to minimize disruption to end users as much as possible.

The most common issue is on the OTLP receiver examples, where we need to change examples like:

```yaml
receivers:
  otlp:
    protocols:
      grpc:
```

to this:

```yaml
receivers:
  otlp:
    protocols:
      grpc:
        endpoint: 0.0.0.0:4317
```

This also needs to be done for the HTTP endpoint, as well as other components mentioned in open-telemetry/opentelemetry-collector-contrib/issues/30702.

There are a lot of pages that need to be updated. For example, you can find many instances on https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry.io+otlp%3A&type=code (some false positives as well).

I think the other components that may need updates are 
- the health check extension: https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry.io%20health_check&type=code, 
- the jaeger receiver https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry.io+jaeger%3A&type=code
- the opencensus receiver https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry.io+opencensus%3A&type=code

I may be missing some.

**Additional context**:  We don't have a explicitly deadline for when the change will happen, we will likely leave some time for this

theletterf:
Fixes #3839 

To check:

- Do we want to add a comment to all YAML edits? What should it say?
- Do we need to change this in instrumentation settings examples?
- Do we need to change `localhost` to `0.0.0.0` elsewhere in Collector configs?

---

**Preview**, e.g., see: https://deploy-preview-3847--opentelemetry.netlify.app/docs/collector/configuration/#endpoint-0.0.0.0-warning


content/en/docs/collector/building/receiver.md
```diff
@@ -117,7 +118,7 @@ exporters:
   debug:
     verbosity: detailed
   otlp/jaeger:
-    endpoint: localhost:14317
+    endpoint: 0.0.0.0:14317
```
svrnm, line 121: I think this only needs to be changed for receivers

content/en/docs/collector/building/receiver.md
```diff
@@ -1047,7 +1049,7 @@ exporters:
...
```
svrnm, line 1052: same here, I think this only needs to be changed for receivers

svrnm:
If I understand the issue correctly this is _only_ for receivers

mx-psi:
This is only on servers exposed by the OpenTelemetry Collector (the vast majority of which are receivers, we also have a couple extensions that may be featured in the docs). The intent is to mitigate https://cwe.mitre.org/data/definitions/1327.html

theletterf:
@svrnm @mx-psi I reverted all changes applied to things that are not receivers. Other than OTLP addresses I couldn't find other instances of receiver listening on localhost in the Collector docs.

I'm also unsure about adding a long note at all. I think this change can be transparent until the new standard comes into effect.

content/en/docs/collector/configuration.md
```diff
@@ -241,7 +249,7 @@ receivers:
         - job_name: otel-collector
           scrape_interval: 5s
           static_configs:
-            - targets: [localhost:8888]
+            - targets: [0.0.0.0:8888]
```
svrnm, line 252: I think this should be reverted as well since this is a scrapping info (pull, not push)

content/en/docs/collector/configuration.md
```diff
@@ -716,13 +724,14 @@ Collector acting as agent:
 ```yaml
 extensions:
   oidc:
-    issuer_url: http://localhost:8080/auth/realms/opentelemetry
+    issuer_url: http://0.0.0.0:8080/auth/realms/opentelemetry
```
svrnm, line 727: this should also remain localhost,

content/en/docs/collector/configuration.md
```diff
@@ -752,13 +761,13 @@ extensions:
   oauth2client:
     client_id: agent
     client_secret: some-secret
-    token_url: http://localhost:8080/auth/realms/opentelemetry/protocol/openid-connect/token
+    token_url: http://0.0.0.0:8080/auth/realms/opentelemetry/protocol/openid-connect/token
```
svrnm, line 764: this one should remain localhost

content/en/docs/demo/architecture.md
```diff
@@ -140,8 +140,8 @@ subgraph tdf[Telemetry Data Flow]
 
         subgraph oc[OTel Collector]
             style oc fill:#97aef3,color:black;
-            oc-grpc[/"OTLP Receiver<br/>listening on<br/>grpc://localhost:4317/"/]
-            oc-http[/"OTLP Receiver<br/>listening on <br/>http://localhost:4318/<br/>https://localhost:4318/"/]
+            oc-grpc[/"OTLP Receiver<br/>listening on<br/>grpc://0.0.0.0:4317/"/]
+            oc-http[/"OTLP Receiver<br/>listening on <br/>http://0.0.0.0:4318/<br/>https://0.0.0.0:4318/"/]
```
svrnm, line 144: I am not sure if the demo is changing anything here? Keep it as is

svrnm:
> I'm also unsure about adding a long note at all. I think this change can be transparent until the new standard comes into effect.

My suggestion is that we have something like this:

```
  otlp:
    protocols:
      grpc:
        # This will tell the collector to listen on all interfaces & addresses, 
        # make sure that you set this configuration to the value that is appropriate for your environment
        endpoint: 0.0.0.0:4317
```

theletterf:
@svrnm I've reverted the changes you pointed out as wrong and added a compact note to the rest. PTAL

content/en/docs/faas/lambda-collector.md
```diff
@@ -53,6 +53,9 @@ receivers:
   otlp:
     protocols:
       grpc:
+        # Tells the Collector to listen on all interfaces
+        # and addresses. Make sure to set the most
+        # appropriate value for your environment.
         endpoint: 'localhost:4317'
       http:
         endpoint: 'localhost:4318'
```
svrnm, line 61: is this still localhost on purpose?

content/en/docs/faas/lambda-collector.md
```diff
@@ -105,6 +108,9 @@ receivers:
...
```
svrnm, line 116: same question here

content/en/docs/kubernetes/helm/collector.md
```diff
@@ -53,6 +53,9 @@ receivers:
   jaeger:
     protocols:
       grpc:
+        # Tells the Collector to listen on all interfaces
+        # and addresses. Make sure to set the most
+        # appropriate value for your environment.
```
svrnm, line 58:
```suggestion
```

I don't think this is needed here

content/en/docs/kubernetes/helm/collector.md
```diff
@@ -61,6 +64,9 @@ receivers:
   otlp:
     protocols:
       grpc:
+        # Tells the Collector to listen on all interfaces
+        # and addresses. Make sure to set the most
+        # appropriate value for your environment.
```
svrnm, line 69:
```suggestion
```

I don't think this is needed here

svrnm (commented):
a few more things to fix, but overall LGTM

content/en/docs/kubernetes/operator/_index.md
```diff
@@ -49,6 +49,10 @@ spec:
       otlp:
         protocols:
           grpc:
+            # Tells the Collector to listen on all interfaces
+            # and addresses. Make sure to set the most
+            # appropriate value for your environment.
+            endpoint: 0.0.0.0:4317
           http:
```
mx-psi, line 56: I guess we should also add this endpoint

content/en/docs/faas/lambda-collector.md
theletterf: Fixed!

content/en/docs/faas/lambda-collector.md
theletterf: Fixed!

theletterf:
Comments addressed! PTAL

chalin (changes requested):
Oh, mi shivers! Can we find a way to avoid the repetition and keep things DRYer? E.g., add an anchored notes somewhere in the docs and link to it from the YAML excerpts. WDYT?

theletterf:
@chalin You're right, it's ugly and un-DRY. I suggested that we don't add the note at all, since it adds cruft that might be hard to maintain later. If we end up adding a Security doc to the Collector, that could be a home for this kind of recommendation. Troubleshooting could also be a good home.

As of now, I'd rather remove the note altogether, as it doesn't really hint at the security issue brought up by @mx-psi nor provides more information on what's appropriate. @svrnm WDYT?

mx-psi:
I am okay with removing the note for now, I think the important part is that we set the endpoints explicitly so that fewer people face a breaking change when we do this

theletterf:
@chalin @svrnm @mx-psi I've eliminated the note. LMK if this one is good to go -- thanks!

content/en/docs/collector/scaling.md
```diff
@@ -157,6 +157,9 @@ spec:
       otlp:
         protocols:
           grpc:
+            # Tells the Collector to listen on all interfaces 
+            # and addresses. Make sure to set the most 
+            endpoint: 0.0.0.0:4317
```
chalin, line 162:
```suggestion
            endpoint: 0.0.0.0:4317
```

chalin:
@mx-psi @theletterf @svrnm - my suggestion is to at least **add at least one note** in the **Collector config page**, something like this:

> #### :warning: **Important**
> While it is generally preferable to bind endpoints to `localhost` when all
> clients are local, our example configurations use the "unspecified" address
> `0.0.0.0` as a convenience. The Collector currently defaults to `0.0.0.0`, but
> the default will be changed to `localhost` in the near future. For details
> concerning either of these choices as endpoint configuration value, see
> [Safeguards against denial of service attacks].
> 
> [Safeguards against denial of service attacks]:
>   https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/security-best-practices.md#safeguards-against-denial-of-service-attacks

It might appear in the Collector config page just before the first endpoint config using `0.0.0.0`, like this:

> image

WDYT?

If you agree (possibly modulo some prose tweaks), let me know and I'll push a commit to this PR with this addition.

mx-psi:
Makes sense to me :+1: As I said above, fine for me to leave this to a different PR, but also fine to keep it here if okay with others :)

theletterf:
@chalin I echo @mx-psi. If you'd add the note with that same wording, that'd work for me.

chalin:
> @chalin I echo @mx-psi. If you'd add the note with that same wording, that'd work for me.

Added, see https://deploy-preview-3847--opentelemetry.netlify.app/docs/collector/configuration/#endpoint-0.0.0.0-warning.

chalin (approved):
If y'all agree with the added note, then the rest LGTM ✨

theletterf:
@chalin LGTM! Let's roll!

swiatekm-sumo (approved):
:+1: on the operator examples.

svrnm (approved):

svrnm:
awesome, this looks really good now. @mx-psi if you are happy as well we can merge :-)

mx-psi (approved):
Looks good to me! :)

PR #3873, opened 2024-01-29
Add Red Hat to vendors

pavolloffay:


data/ecosystem/vendors.yaml
```diff
@@ -251,6 +251,13 @@
   contact: ''
   oss: false
   commercial: true
+- name: Red Hat
+  distribution: true
+  nativeOTLP: true
+  url: 'https://docs.openshift.com/container-platform/4.14/otel/otel-release-notes.html'
+  contact: ''
```
svrnm, line 258: please add a point of contact, can be your github handle or an email address, we may use this eventually if we need to reach out

pavolloffay: Is it for reaching out to maintainers or sales?

svrnm: Maintainers.

PR #3831, opened 2024-01-19
Update getting-started.md to specify java version

arifcse019:
Without java 17, gradle builds fails

I got the following error when using Java 11 on Ubuntu 22

`ubuntu@ip-172-31-53-138:~/java-simple$ gradle assemble
Starting a Gradle Daemon (subsequent builds will be faster)

FAILURE: Build failed with an exception.

* What went wrong:
A problem occurred configuring root project 'java-simple'.
> Could not resolve all files for configuration ':classpath'.
   > Could not resolve org.springframework.boot:spring-boot-gradle-plugin:3.0.6.
     Required by:
         project : > org.springframework.boot:org.springframework.boot.gradle.plugin:3.0.6
      > No matching variant of org.springframework.boot:spring-boot-gradle-plugin:3.0.6 was found. The consumer was configured to find a library for use during runtime, compatible with Java 11, packaged as a jar, and its dependencies declared externally, as well as attribute 'org.gradle.plugin.api-version' with value '8.5' but:
          - Variant 'apiElements' capability org.springframework.boot:spring-boot-gradle-plugin:3.0.6 declares a library, packaged as a jar, and its dependencies declared externally:
              - Incompatible because this component declares a component for use during compile-time, compatible with Java 17 and the consumer needed a component for use during runtime, compatible with Java 11
              - Other compatible attribute:
                  - Doesn't say anything about org.gradle.plugin.api-version (required '8.5')
          - Variant 'javadocElements' capability org.springframework.boot:spring-boot-gradle-plugin:3.0.6 declares a component for use during runtime, and its dependencies declared externally:
              - Incompatible because this component declares documentation and the consumer needed a library
              - Other compatible attributes:
                  - Doesn't say anything about its target Java version (required compatibility with Java 11)
                  - Doesn't say anything about its elements (required them packaged as a jar)
                  - Doesn't say anything about org.gradle.plugin.api-version (required '8.5')
          - Variant 'mavenOptionalApiElements' capability org.springframework.boot:spring-boot-gradle-plugin-maven-optional:3.0.6 declares a library, packaged as a jar, and its dependencies declared externally:
              - Incompatible because this component declares a component for use during compile-time, compatible with Java 17 and the consumer needed a component for use during runtime, compatible with Java 11
              - Other compatible attribute:
                  - Doesn't say anything about org.gradle.plugin.api-version (required '8.5')
[...]

PR #3877, opened 2024-01-29
[Java] getting-started: tweak JDK version constraint

chalin:
- Followup to #3831


content/en/docs/languages/java/getting-started.md
```diff
@@ -16,7 +16,7 @@ console.
 
 Ensure that you have the following installed locally:
 
-- Java 17 JDK
+- Java 17+ JDK
```
jack-berg, line 19: Unrelated to this PR, but someone reading this may infer that we only support Java 17+. This is not the case. We support java 8+ but this specific example requires java 17+ because of spring. You also need java 17+ if you're going to build `opentelemetry-java` from source, but it is not needed to consume the artifacts.

chalin: Good point. Is this documented somewhere else? I think that we want to keep the "story" simple in the Getting Started page, but maybe link to somewhere else for the "entire story" :).

An alternative would be to add a footnote here. WDYT?

jack-berg: We document it [here](https://github.com/open-telemetry/opentelemetry-java?tab=readme-ov-file#requirements):

> Unless otherwise noted, all published artifacts support Java 8 or higher.

The JDK version compatibility is a pretty important point in the java ecosystem, so while the entire story may be documented elsewhere, I think any spot which mentions version requirement conflicting with java 8+ is prone to being misinterpreted. A footnote is probably best here.

Perhaps we introduce a brief section on the [manual instrumentation](https://opentelemetry.io/docs/languages/java/instrumentation/) page describing the java 8+ requirement, and link to it from this getting started page?

trask: > The JDK version compatibility is a pretty important point in the java ecosystem, so while the entire story may be documented elsewhere, I think any spot which mentions version requirement conflicting with java 8+ is prone to being misinterpreted. A footnote is probably best here.

👍

chalin: @jack-berg @trask - I've updated the prose. PTAL and let me know what you think. For now I'm trying to confine this change to the [**Getting started** page](https://deploy-preview-3877--opentelemetry.netlify.app/docs/languages/java/getting-started/#prerequisites):

> image

chalin:
@svrnm - PTAL at the copyedits to the introductory paragraph of [Example Application](https://deploy-preview-3877--opentelemetry.netlify.app/docs/languages/java/getting-started/#example-application). I've streamlined the text. I figured that you might want to comment on this since we're generally trying to keep the text similar across languages.

jack-berg (approved):

content/en/docs/languages/java/getting-started.md
```diff
@@ -9,23 +9,21 @@ weight: 10
...
-in such a way that [traces][], [metrics][] and [logs][] are emitted to the
+in such a way that [traces][], [metrics][], and [logs][] are emitted to the
 console.
 
 ## Prerequisites
 
 Ensure that you have the following installed locally:
 
-- Java 17 JDK
+- Java JDK 17+, due to the use of Spring Boot; [Java 8+ otherwise][java-vers]
```
trask, line 19: (Spring Boot 3 sort of famously made the leap to Java 17)

```suggestion
- Java JDK 17+, due to the use of Spring Boot 3; [Java 8+ otherwise][java-vers]
```

trask (approved):

content/en/docs/languages/java/getting-started.md
chalin: Works for me.

cartermp (approved):

PR #3881, opened 2024-01-29
Fix config example

austindrenski:
Add missing `tls` node to second and third example tabs.


theletterf (approved):
LGTM. Could you add a description with a bit of context? Thanks!

cartermp (approved):
mmm, good catch, this seems to have been missed a ~year or so ago when some other similar changes came through.

PR #3883, opened 2024-01-29
[CI/textlint] Enforce "backend" as one word

### LINKED ISSUE #3882: [textlint] Enforce "backend" vs "back end"
The docs use "backend" in all places except one:

```console
$ find content -type f -exec grep -nH -iEe "\bbackends?\b" {} + | wc -l            
     306
$ find content -type f -exec grep -nH -iEe "\bback ends?\b" {} + | wc -l
       1
```

So I'd suggest that we add a textlint rule for this. @svrnm et al. WDYT?

chalin:
- Fixes #3882

### DIFF
--- a/.textlintrc.yml
+++ b/.textlintrc.yml
@@ -113,6 +113,7 @@ rules:
       # https://github.com/sapegin/textlint-rule-terminology/blob/ca36a645c56d21f27cb9d902b5fb9584030c59e3/index.js#L137-L142.
       #
       - ['3rd[- ]party', third-party]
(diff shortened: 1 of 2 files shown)


theletterf (approved):
Ah, this one will hit me hard. :) Google does recommend "backend" as a noun, going the descriptive way, though grammar and dictionaries say otherwise. 

I'm OK with this though. I've been using and suggesting "back end" as noun myself due to Splunk's style guide. Will pay attention to this one going forward.

chalin:
> I'm OK with this though.

Great!

PR #3716, opened 2024-01-01
Supply example on Document baggage propagation

pegasas:
try to resolve: https://github.com/open-telemetry/opentelemetry-python/issues/3561

PR #3887, opened 2024-01-30
Drop unnecessary `api/propagation` aliases and local dictionary

chalin:
- Followup to:
  - #3716
  - #1951
- In each of the PRs listed above, a new `propagation.md` page is introduced. Generally for new pages we don't need an alias. This PR removes the superfluous aliases, and cleans up the page-local dictionary while we're there.

/cc @cartermp


theletterf (approved):

pegasas:
lgtm.
thanks @chalin for calling out.

svrnm (approved):

Issue #3891, opened 2024-01-30
[CI/textlint] Enforce autoconfigure and autoconfiguration as a single word


chalin:
Inspired, in particular, by https://github.com/open-telemetry/opentelemetry.io/pull/3758/files.

PR #3903, opened 2024-01-31
[CI/textlint] "Auto Config*" rewrite-rule capitalization fix

chalin:
- Completes the fix for #3891
- Fixes the "Auto Config*" rewrite-rule so that it forces the "c" in "config*" to be in lowercase
- Scrubs some page-local dictionaries

### DIFF
--- a/.cspell.yml
+++ b/.cspell.yml
@@ -33,6 +33,7 @@ words:
   - appender
   - appenders
   - aspecto
+  - autoconfigure
   - autoconfiguration
   - autoload
   - autoloaded
@@ -44,6 +45,7 @@ words:
   - cartservice
   - cassandra
   - checkoutservice
+  - classpath
   - cncf
   - currencyservice
   - daemonset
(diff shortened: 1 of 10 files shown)


svrnm (approved):

svrnm:
works for me, since this mainly affects java so far, would be good to hear if this makes sense?

theletterf (approved):

chalin:
> works for me, since this mainly affects java so far, would be good to hear if this makes sense?

@svrnm - the rule is already present in the TextLint config, this PR is just a fix to the rule. So, ... is this good to merge?

chalin:
> Would a sentence starting with Configure would be accepted with the new rule, for example:
> 
> > Configure this library in the following way:

@jeanbisutti - such an occurrence would not be affected by the rule.

svrnm:
> > works for me, since this mainly affects java so far, would be good to hear if this makes sense?
> 
> @svrnm - the rule is already present in the TextLint config, this PR is just a fix to the rule. So, ... is this good to merge?

Yes :-)

PR #3908, opened 2024-01-31
Added Dubbo in integrations.

steverao:
Added Dubbo information in integrations.


data/ecosystem/integrations.yaml
```diff
@@ -181,3 +181,8 @@
   docsUrl: https://farfetch.github.io/kafkaflow/docs/guides/open-telemetry
   components: [.NET]
   oss: true
+- name: Dubbo
+  url: https://github.com/apache/dubbo
+  docsUrl: https://cn.dubbo.apache.org/en/blog/2024/01/31/tracing-dubbo-with-opentelemetry/
```
svrnm, line 186: Is there any official documentation for that? A blog post may outdate, so I rather prefer a pointer to the docs that is kept up-to-date accordingly.

svrnm (changes requested):
thanks, great to hear that Dubbo is adopting OpenTelemetry! See my inline comment for feedback

svrnm:
@open-telemetry/java-approvers FYI

data/ecosystem/integrations.yaml
chickenlj: Hi, I am a member of the Dubbo community. 

It's been a while since we provided support for OpenTelemetry in Dubbo SDK, both apache/dubbo and apache/dubbo-go. But right now, the documentation of this part is only available in Chinese. We are in the process of building an English version of documentation for the whole site, which is expected to be publicly available in the next 2 months.

Can we accept this blog for now? I will update the official documentation when it's available ASAP.

svrnm: Thanks for clarification, @chickenlj !

Can we do the following: Add a comment above that line pointing to this comment (`https://github.com/open-telemetry/opentelemetry.io/pull/3908#discussion_r1473133919`), so we can keep track of it

Beyond that I am then OK with the blog post

svrnm (approved):

steverao:
> hi @steverao!
> 
> how does this relate to https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/instrumentation/apache-dubbo-2.7/library-autoconfigure?
> 
> also, does Dubbo require the use Micrometer Tracing and the Micrometer Tracing <-> OpenTelemetry API bridge? or can Dubbo use the OpenTelemetry API natively?

Dubbo achive

> hi @steverao!
> 
> how does this relate to https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/instrumentation/apache-dubbo-2.7/library-autoconfigure?
> 
> also, does Dubbo require the use Micrometer Tracing and the Micrometer Tracing <-> OpenTelemetry API bridge? or can Dubbo use the OpenTelemetry API natively?

Because of some history reasons, it uses Micrometer Tracing <-> OpenTelemetry API bridge way to collect observable data now. Maybe we can evaluate whether to directly integrate by the OpenTelemetry API natively in the future? @chickenlj

trask:
hi @steverao, I didn't follow above how this relates to https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/instrumentation/apache-dubbo-2.7/library-autoconfigure? thx!

steverao:
> https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/instrumentation/apache-dubbo-2.7/library-autoconfigure

There is no direct relationship between them. Due to some historical reasons, the current way the dubbo project integrates OpenTelemetry does not directly use the library mentioned above.

steverao:
> hi @steverao, I didn't follow above how this relates to https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/instrumentation/apache-dubbo-2.7/library-autoconfigure? thx!

There is no direct relationship between them. Due to some historical reasons, the current way the dubbo project integrates OpenTelemetry does not directly use the library mentioned above.

trask:
thanks, should we consider deprecating https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/instrumentation/apache-dubbo-2.7/library-autoconfigure if the Dubbo project is recommending a different approach?

steverao:
> thanks, should we consider deprecating https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/instrumentation/apache-dubbo-2.7/library-autoconfigure if the Dubbo project is recommending a different approach?

I think we can maintain the current status for the time being, because we were not particularly familiar with the implementation of OTel before, and we did not know the usage habits of external users. Should we conduct a follow-up survey in the community on the user acceptance of the two usages and then think it will be better？How do you think about it? @chickenlj

steverao:
Hi @trask, regarding the discussion above, if we have a clear hope to delete the implementation on Otel in the future, then we create related CR for processing it. Do you think it is okay to merge this change first?

steverao:
> thanks for the explanations, we can definitely continue discussing options in https://github.com/open-telemetry/opentelemetry-java-instrumentation

Ok, thank you very much!

PR #3927, opened 2024-02-02
Update Go guidance for attributes

cartermp:
In service towards https://cloud-native.slack.com/archives/CJFCJHG4Q/p1706816394957449


content/en/docs/languages/go/instrumentation.md
```diff
@@ -169,18 +169,39 @@ are useful for aggregating, filtering, and grouping traces. Attributes can be
 added at span creation, or at any other time during the lifecycle of a span
 before it has completed.
 
+Generally speaking, you should try to set attributes during span creation.
+This will incur less overhead compared to setting attributes after
```
cijothomas, line 173: Isn't that also the official recommendation of the API spec as well? Maybe this should be repeated in all languages doc.

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#span-creation

_The API documentation MUST state that adding attributes at span creation is preferred to calling SetAttribute later, as samplers can only consider information already present during span creation._

content/en/docs/languages/go/instrumentation.md
```diff
@@ -690,12 +711,37 @@ func init() {
 
 ### Adding attributes
 
+Attributes are keys and values that are applied as metadata to your metrics and
+are useful for aggregating, filtering, and grouping metrics. Attributes can be
+added at instrument creation, or at any other time during the lifecycle of an
+instrument.
+
+Generally speaking, it is advisable to add attributes during instrumentation
```
cijothomas, line 719:
```suggestion
Generally speaking, it is advisable to add attributes during instrument creation
```

content/en/docs/languages/go/instrumentation.md
```diff
@@ -690,12 +711,37 @@ func init() {
...
 options.
 
+For example, here's how you could add descriptive attributes during creation:
+
+```go
+counter, err := meter.Int64UpDownCounter(
+	"app.my-counter",
+	metric.WithDescription("My fancy counter."),
+	metric.WithAttributes(
+		attribute.String("my-important-key", "my-important-value"),
```
cijothomas, line 736: mm.. not sure if this is a good practice.. if the key and value are known ahead, are they better fitted as instrumentation scope attributes, set on the Meter itself, so they won't take up aggregation space, and are more efficiently transported over otlp too?

(Please ignore if this comment does not makes sense. I have limited Go experience)

cartermp: Yeah, I was thinking maybe there's three scenarios here:

- Common attributes at the meter level (perhaps resource level?)
- Attributes you want per instrument
- Attributes you need to add when you record a measurement

The first category is the weirdest one to me. But also there's I think a different mindset eto attributes for metrics that I'm not considering here.

cijothomas: The first category is already well defined - https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#get-a-meter
_[since 1.13.0] attributes: Specifies the instrumentation scope attributes to associate with emitted telemetry._

The 2nd one - instrument level attributes - they don't exist in the spec from what I can tell.

content/en/docs/languages/go/instrumentation.md
```diff
@@ -169,18 +169,39 @@ are useful for aggregating, filtering, and grouping traces. Attributes can be
...
+
+The following example adds attributes at creation:
+
 ```go
 // setting attributes at creation...
-ctx, span = tracer.Start(ctx, "attributesAtCreation", trace.WithAttributes(attribute.String("hello", "world")))
-// ... and after creation
-span.SetAttributes(attribute.Bool("isTrue", true), attribute.String("stringAttr", "hi!"))
+ctx, span = tracer.Start(ctx, "attributesAtCreation",
+	trace.WithAttributes(attribute.String("hello", "world"))
```
pellared, line 181: Have you checked if it works? AFAIK it does not. See https://github.com/open-telemetry/opentelemetry-go/issues/3368

content/en/docs/languages/go/instrumentation.md
pellared: > The 2nd one - instrument level attributes - they don't exist in the spec from what I can tell.

Correct. Have you checked if such code compiles?

pellared (commented):
This is the way for the users would like to reduce the amount of heap allocations: https://cloud-native.slack.com/archives/CJFCJHG4Q/p1706821596412959?thread_ts=1706816394.957449&cid=CJFCJHG4Q

content/en/docs/languages/go/instrumentation.md
cartermp: This works for me on the getting started guide:

```
	ctx, span := tracer.Start(
		r.Context(),
		"roll",
		trace.WithAttributes(attribute.String("game", "craps")),
	)
```

When run:

```
        "Attributes": [
                {
                        "Key": "game",
                        "Value": {
                                "Type": "STRING",
                                "Value": "craps"
                        }
                },
```

content/en/docs/languages/go/instrumentation.md
svrnm: good point, do you mind turning your comment in a new issue such that we can track it (and after this PR is merged we may turn it to a shortcode across all languages to be used for that)

cijothomas: https://github.com/open-telemetry/opentelemetry.io/issues/3947

content/en/docs/languages/go/instrumentation.md
pellared: You are right. My bad. The issue was about getting a tracer/meter with scope attributes and this creates a span.

content/en/docs/languages/go/instrumentation.md
```diff
@@ -690,12 +711,37 @@ func init() {
 
 ### Adding attributes
 
+Attributes are keys and values that are applied as metadata to your metrics and
+are useful for aggregating, filtering, and grouping metrics. Attributes can be
+added at instrument creation, or at any other time during the lifecycle of an
+instrument.
```
chalin, line 717:
```suggestion
[Attributes] are keys-values pairs applied as metadata to your metrics that
are useful for aggregating, filtering, and grouping metrics. Attributes can be
added at instrument creation, or at any other time during the lifecycle of an
instrument.

[Attributes]: /docs/concepts/signals/traces/#attributes
```

theletterf: s/key-values/key-value

content/en/docs/languages/go/instrumentation.md
cartermp: Reverted that portion

content/en/docs/languages/go/instrumentation.md
cartermp: I decided only to take this for the traces section, since the link takes you to the traces signal page.

content/en/docs/languages/go/instrumentation.md
```diff
@@ -164,23 +164,43 @@ Once a span has completed, it is immutable and can no longer be modified.
...
 ### Span Attributes
 
-Attributes are keys and values that are applied as metadata to your spans and
+[Attributes] are keys and values that are applied as metadata to your spans and
 are useful for aggregating, filtering, and grouping traces. Attributes can be
 added at span creation, or at any other time during the lifecycle of a span
 before it has completed.
 
+Generally speaking, you should try to set attributes during span creation. This
+will incur less overhead compared to setting attributes after span creation, and
```
cijothomas, line 173: > This
> will incur less overhead compared to setting attributes after span creation

^ Is this actually true? I think the answer is - it depends.!

If the sampler is ParentBased and the parent's decision is to Drop, then providing all attributes at startup time is wasted as the span is going to be dropped anyway...
If user choses to add it after span creation, he/she can avoid the cost of representing attributes completes, by short-circuiting if span is not recording.

The only reason spec recommended providing attributes at creation time is for the trace-sampler to access it, not perf. If Golang implementation makes it more performant, then this section makes sense - but I wanted to double check if that is indeed true.

content/en/docs/languages/go/instrumentation.md
```diff
@@ -164,23 +164,43 @@ Once a span has completed, it is immutable and can no longer be modified.
 
 ### Span Attributes
 
-Attributes are keys and values that are applied as metadata to your spans and
+[Attributes] are keys and values that are applied as metadata to your spans and
```
theletterf, line 167:
```suggestion
[Attributes] are key-value pairs that are applied as metadata to your spans and
```

content/en/docs/languages/go/instrumentation.md
```diff
@@ -164,23 +164,43 @@ Once a span has completed, it is immutable and can no longer be modified.
...
 are useful for aggregating, filtering, and grouping traces. Attributes can be
 added at span creation, or at any other time during the lifecycle of a span
 before it has completed.
 
+Generally speaking, you should try to set attributes during span creation. This
+will incur less overhead compared to setting attributes after span creation, and
+also allow these attributes to factor into SDK sampling decisions.
+
+The following example adds attributes at creation:
```
theletterf, line 176:
```suggestion
As a best practice, set attributes during span creation. This causes
less overhead compared to setting attributes after span creation and
allow attributes to contribute to SDK sampling decisions.

The following example adds attributes at span creation:
```

content/en/docs/languages/go/instrumentation.md
```diff
@@ -690,12 +710,17 @@ func init() {
 
 ### Adding attributes
 
+Attributes are keys and values that are applied as metadata to your metrics and
+are useful for aggregating, filtering, and grouping metrics.
```
theletterf, line 714:
```suggestion
Attributes are key-value pairs that are applied as metadata to your metrics and
are useful for aggregating, filtering, and grouping metrics.
```

theletterf (commented):
Some suggestions.

content/en/docs/languages/go/instrumentation.md
cartermp: This is the justification: https://cloud-native.slack.com/archives/CJFCJHG4Q/p1706816394957449 (SetAttributes also takes a variadic argument). My question would be if the Go compiler optimizes the single value case though.

pellared:
While I have nothing against the changes I just want to point out that in my opinion the description is misleading

> In service towards https://cloud-native.slack.com/archives/CJFCJHG4Q/p1706816394957449

See: https://github.com/open-telemetry/opentelemetry.io/pull/3927#pullrequestreview-1861131249

svrnm:
@cartermp @open-telemetry/go-approvers is this PR something we want to continue?

pellared:
The content and title LGTM.

However, I do think that the description is true.
More: https://github.com/open-telemetry/opentelemetry.io/pull/3927#issuecomment-1952311697.

CC @MrAlias

svrnm:
> However, I think that the description is not true. More: [#3927 (comment)](https://github.com/open-telemetry/opentelemetry.io/pull/3927#issuecomment-1952311697).

@cartermp does this help you to finalize this PR?

cartermp:
gonna close this one out, it's been rather long

PR #3940, opened 2024-02-05
Add default Collector configuration path to Configuration docs

### LINKED ISSUE #3929: Configuration
This page goes into a lot of detail about the content of the configuration file, but never states what the configuration file is called or where it is located.

theletterf:
Fixes #3929. Preview: https://deploy-preview-3940--opentelemetry.netlify.app/docs/collector/configuration/

I've added two new sections:

- Location of the default config file and how to set a different one (and validate it).
- How to override settings using the `--set` option.

Most of the information was taken from https://github.com/open-telemetry/opentelemetry-collector/blob/fa1b24184cb20ed0584457ea68c49c908da5f988/service/README.md?plain=1#L5, assuming that it applies to the `otelcol` binary as well, though confirmation is required. 

Some questions:

- Should we add the default location for Windows? I couldn't find what that is.
- Do the options described in the `otelcorecol` README file apply?

Note: It'd be really, really great to revive https://github.com/open-telemetry/opentelemetry-collector/issues/5654


content/en/docs/collector/configuration.md
```diff
@@ -8,14 +8,40 @@ cSpell:ignore: cfssl cfssljson fluentforward gencert genkey hostmetrics initca l
...
-- [Data collection concepts][dcc] in order to understand the repositories
-  applicable to the OpenTelemetry Collector.
+- [Data collection concepts][dcc], to understand the repositories applicable to
+  the OpenTelemetry Collector.
 - [Security guidance](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/security-best-practices.md)
 
+## Location {#location}
+
+By default, the Collector configuration is located in
+`/etc/otelcol/config.yaml`.
```
svrnm, line 22: I think this is only true for `core`, if I remember correctly for contrib it was `/etc/otelcol-contrib/config.yaml`?

cartermp: We'll definitely want to list the contrib location if it's different, since most people use contrib

svrnm: > We'll definitely want to list the contrib location if it's different, since most people use contrib

... although they shouldn't ...

cartermp: I disagree, they absolutely should. There's just too many useful things in there that you'll need, especially when there's several sources of data and a need to normalize or transform a bit.

content/en/docs/collector/configuration.md
```diff
@@ -843,3 +869,34 @@ This creates two certificates:
...
+# Translates to
+# key:
+#   - a
+#   - b
+#   - c
+otelcol --set "key=[a, b, c]"
+# Translates to
+# key:
+#   a: c
+otelcol --set "key={a: c}
```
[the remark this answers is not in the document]
theletterf: +CC @mx-psi This was taken from https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/README.md#how-to-override-config-properties

@svrnm You mean this last example? Or the whole snippet?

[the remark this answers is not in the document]
svrnm: This last examples, I would like to see some real application of that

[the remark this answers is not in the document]
mx-psi: I guess if you want to set a whole section of the configuration you could do something like:
```
otelcol --set "receivers::otlp::protocols::grpc={endpoint:localhost:4317, compression: gzip}"
```

maybe @bogdandrutu (author of the original docs) can share some other example?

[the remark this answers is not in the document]
svrnm: Ok, so it sounds like something I want to use to temporarily overwrite some parts of my config? Maybe setting the verbosity of the debug exporter could be a great example for endusers? Or disabling certain elements in their pipelines?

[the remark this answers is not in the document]
mx-psi: Yeah, I think those are both good examples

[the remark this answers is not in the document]
svrnm: @theletterf would you be able to add those examples?

content/en/docs/collector/configuration.md
theletterf: Would it help to write something like...

```
By default, the Collector configuration is located in `/etc/<otel-directory>/config.yaml`, where `<otel-directory>` can be `otelcol`, `otelcol-contrib`, or another value, depending on the Collector version or distribution you're using.
```

content/en/docs/collector/configuration.md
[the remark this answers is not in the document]
theletterf: Will add those tomorrow -- thanks!

content/en/docs/collector/configuration.md
svrnm: > I disagree, they absolutely should. There's just too many useful things in there that you'll need, especially when there's several sources of data and a need to normalize or transform a bit.

The recommendation we should eventually have for that is that people use `ocb` to have the collector tailored to exactly the components they need. This is mainly a security concern. So it's more about "Use otelcol-contrib to figure out what you need, then use `ocb` to create the collector that only contains that", but that's something we should put into a dedicated issue eventually

theletterf:
PTAL. I've made the override examples more realistic and edited the location paragraph to make it a bit more... flexible.

content/en/docs/collector/configuration.md
```diff
@@ -843,3 +871,23 @@ This creates two certificates:
...
+The following examples show how to override settings inside nested sections:
+
+```shell
+# The following example sets the verbosity
+# level of the debug exporter to 'detailed'
+otelcol --set "exporters::debug::verbosity=detailed"
+# The following example overrides gRPC
+# settings for the OTLP receiver
+otelcol --set "receivers::otlp::protocols::grpc={endpoint:localhost:4317, compression: gzip}"
+```
```
mx-psi, line 890: @bogdandrutu do you think having these examples make sense?

svrnm (approved):

content/en/docs/collector/configuration.md
theletterf: If they're syntactically correct, could we merge and review later?

mx-psi: :+1: let's do this

theletterf: Sweet! Could you approve? Thanks!

mx-psi (approved):

svrnm (approved):

PR #3971, opened 2024-02-09
Add Skyscanner to adopters lists

danielgblanco:
This change:

- Adds Skyscanner to the adopters list
- Sorts the list of adopters alphabetically. It seems it was sorted alphabetically and then at some point stopped being sorted alphabetically. Happy to revert that commit if it's not the intention of this list to be sorted alphabetically.

### DIFF
--- a/.cspell.yml
+++ b/.cspell.yml
@@ -144,6 +144,7 @@ words:
   - Sharma
   - shippingservice
   - shortcode
+  - Skyscanner
   - snmp
   - Socha
   - sqlite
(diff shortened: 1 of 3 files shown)


.cspell.yml
```diff
@@ -144,6 +144,7 @@ words:
   - Sharma
   - shippingservice
   - shortcode
+  - Skyscanner
```
svrnm, line 147: put Skycanner in line1 of the adopters.yaml, as long as this is not used across multiple pages we keep this local

svrnm:
Nice! Especially that this comes with a reference :-)

.cspell.yml
danielgblanco: Ah! I see! I completely missed the cspell comment. Will do now.

svrnm (approved):

PR #3991, opened 2024-02-13
go-automatic-instrumentation

nikolaydubina:
I have been using this for couple of years. I see many other people and companies are using this. I believe this was one of the first automatic instrumentations projects in Go too. Would like to mention it in OTEL documentation.

### DIFF
--- a/content/en/docs/languages/go/automatic.md
+++ b/content/en/docs/languages/go/automatic.md
@@ -4,7 +4,8 @@ linkTitle: Automatic
 weight: 30
 ---
 
-Automatic instrumentation in Go is possible by automatically modifying source code.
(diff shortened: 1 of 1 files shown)


nikolaydubina:
Got it. Will leave this to you guys which project include in the documentation.

Will see if I can join that work group 👍🏻

svrnm:
@nikolaydubina thanks for creating this PR. Unfortunately we can not provide documentation for pieces which are not official solutions by the opentelemetry community. As @dmathieu stated we should have a documentation for the official auto instrumentation.

Note, that you can add your solution to the registry: https://opentelemetry.io/ecosystem/registry/adding/

nikolaydubina:
Noted. Let me add it to registry.

PR #3992, opened 2024-02-13
Add rspec-otel to registry

dmathieu:
This adds the rspec-otel gem to the registry.
https://rubygems.org/gems/rspec-otel

### DIFF
--- a/data/registry/tools-ruby-rspec-matcher.yml
+++ b/data/registry/tools-ruby-rspec-matcher.yml
@@ -1,3 +1,4 @@
+# cSpell:ignore Damien Mathieu
 title: RSpec matchers
 registryType: utilities
 language: ruby
--- a/data/registry/tools-ruby-rspec-matcher.yml
+++ b/data/registry/tools-ruby-rspec-matcher.yml
@@ -10,7 +10,7 @@ description:
   The RSpec matchers tool provides a set of matchers to make it easier to write
   unit tests checking the presence of telemetry data.
 authors:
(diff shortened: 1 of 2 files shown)


data/registry/tools-ruby-rspec-matcher.yml
```diff
@@ -0,0 +1,20 @@
...
+language: ruby
+tags:
+  - ruby
+  - rspec
+license: MIT
+description:
+  The RSpec matchers tool provides a set of matchers to make it easier to write
+  unit tests checking the presence of telemetry data.
+authors:
+  - name: Damien MATHIEU # cspell:disable-line
```
svrnm, line 12:
```suggestion
  - name: Damien Mathieu
```
@dmathieu as suggested move the cspell ignore to the top of the file. Also if you don't mind I would prefer to not uppercase your last name (except this is how you commonly write it, then ignore that suggestion)

dmathieu: I usually write it in full uppercase, to show which is my name and which is my last name. But not doing so is fine too.

dmathieu:
I suppose the repo URL needs to be added to refcache.
But I don't see any documentation about doing so in the repository, and my attempt to add it manually don't resolve the failure.

svrnm:
> I suppose the repo URL needs to be added to refcache. But I don't see any documentation about doing so in the repository, and my attempt to add it manually don't resolve the failure.

yes, we need to document this better in our contributor guidelines

PR #3995, opened 2024-02-13
Add Caution Note re Experimental SemConvs

mattmccleary:
Following up on Slack Chat:
https://cloud-native.slack.com/archives/C041APFBYQP/p1707760677044609

---

**Preview**: https://deploy-preview-3995--opentelemetry.netlify.app/docs/languages/

### DIFF
--- a/content/en/docs/languages/_index.md
+++ b/content/en/docs/languages/_index.md
@@ -28,7 +28,7 @@ follows:
 {{% telemetry-support-table " " %}}
 
 {{% alert title="Note" color="warning" %}} Even though the API/SDK status is
-stable, changes to underlying [semantic conventions][SemConv] can interrupt your
+stable, changes to underlying [semantic conventions][sem-conv] can interrupt your
 data flow. If the status of a Semantic Convention that your instrumentation
 depends on is `Experimental`, it may be subject to future breaking changes.
 {{% /alert %}}
--- a/content/en/docs/languages/_index.md
+++ b/content/en/docs/languages/_index.md
@@ -25,15 +25,19 @@ Java, Node.js, Python, Go into your application.
 The current status of the major functional components for OpenTelemetry is as
 follows:
 
-{{% telemetry-support-table " " %}}
+{{% alert title="Warning" color="warning" %}}
 
-{{% alert title="Note" color="warning" %}} Even though the API/SDK status is
-stable, changes to underlying [semantic conventions][sem-conv] can interrupt your
-data flow. If the status of a Semantic Convention that your instrumentation
-depends on is `Experimental`, it may be subject to future breaking changes.
+Even though an API/SDK status is stable, changes to underlying [semantic
+conventions] can interrupt your data flow. If the status of a semantic
+convention that your instrumentation depends on is `Experimental`, it may be
+subject to future breaking changes.
+
+[semantic conventions]: /docs/concepts/semantic-conventions/
+
 {{% /alert %}}
+
+{{% telemetry-support-table " " %}}
 
 [auto]: /docs/kubernetes/operator/automatic/
 [instrumentation]: /docs/concepts/instrumentation/
 [otel-op]: /docs/kubernetes/operator/
-[sem-conv]: /docs/concepts/semantic-conventions/
--- a/content/en/docs/languages/_index.md
+++ b/content/en/docs/languages/_index.md
@@ -29,7 +29,7 @@ follows:
 
 {{% alert title="Note" color="warning" %}} Even though the API/SDK status
 is stable, changes to underlying [semantic conventions][SemConv] can interrupt
-your data flow. If the status of a Semantic Convention that your SDK/API
+your data flow. If the status of a Semantic Convention that your instrumentation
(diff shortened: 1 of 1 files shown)


chalin (commented):
@mattmccleary - FYI, I've added an `alert` closing tag so that this change will at least build.

svrnm (approved):

svrnm:
thanks @mattmccleary, this is an important call out. I think we will need to have it at a few other places as well for people to see and recognize it. But we can fix this as a follow up (we can turn your segment into a shortcode and put it at places where we think it is required)

content/en/docs/languages/_index.md
```diff
@@ -27,6 +27,13 @@ follows:
 
 {{% telemetry-support-table " " %}}
 
+{{% alert title="Note" color="warning" %}} Even though the API/SDK status is
+stable, changes to underlying [semantic conventions][SemConv] can interrupt your
```
svrnm, line 31:
```suggestion
stable, changes to underlying [semantic conventions][sem-conv] can interrupt your
```
Not sure why this is not rendering in hugo (see preview: https://deploy-preview-3995--opentelemetry.netlify.app/docs/languages/), maybe those tags have to be lower case, @chalin ?

content/en/docs/languages/_index.md
```diff
@@ -27,6 +27,13 @@ follows:
...
+data flow. If the status of a Semantic Convention that your instrumentation
+depends on is `Experimental`, it may be subject to future breaking changes.
+{{% /alert %}}
+
 [auto]: /docs/kubernetes/operator/automatic/
 [instrumentation]: /docs/concepts/instrumentation/
 [otel-op]: /docs/kubernetes/operator/
+[SemConv]: /docs/concepts/semantic-conventions/
```
svrnm, line 39:
```suggestion
[sem-conv]: /docs/concepts/semantic-conventions/
```

chalin:
Wrapping something up, will get back to this PR soon -- I have some extra comments beyond the markdown link reference issue.

jsuereth (approved):

svrnm:
> Wrapping something up, will get back to this PR soon -- I have some extra comments beyond the markdown link reference issue.

Sure, then let's not merge this before you follow up!

chalin (changes requested):
I've pushed a fix to the alert link-ref issue. The solution is to have the markdown link definition be placed inside the shortcode.

I've also moved the note to be before the table. It makes more sense to me that way.

All that being said, I'd like to suggest that we rework the note prose and/or the opening statement "The current status of the major functional components for OpenTelemetry is as follows", so that it is clear that we're talking about the same/related things.

That is, now I'm wondering what the status table is actually giving the status of. Can we make that clearer?

mattmccleary:
> I've pushed a fix to the alert link-ref issue. The solution is to have the markdown link definition be placed inside the shortcode.
> 
> I've also moved the note to be before the table. It makes more sense to me that way.
> 
> All that being said, I'd like to suggest that we rework the note prose and/or the opening statement "The current status of the major functional components for OpenTelemetry is as follows", so that it is clear that we're talking about the same/related things.
> 
> That is, now I'm wondering what the status table is actually giving the status of. Can we make that clearer?

I agree the table isn't clear, though the title of the page is "Language APIs & SDKs" so I inferred it's a composite of API and SDK. In the future, the table could be reworked to include semantic conventions but that would require some thought because it could quickly blow up the size. Since it's not clear to me how to improve the situation, I'm wondering if this could be a non-blocking follow-on item. I'm happy to think on it some more.

chalin:
I'd be generally in agreement with what you propose. I might want to tweak the text a bit though -- let me sleep on it :).

svrnm:
I agree that the table with the status per language needs to be improved, but this should not be a concern for this PR. I'll raise a follow up issue since I also have a few ideas here

content/en/docs/languages/_index.md
```diff
@@ -25,6 +25,17 @@ Java, Node.js, Python, Go into your application.
...
+{{% alert title="Warning" color="warning" %}}
+
+Even though an API/SDK status is stable, changes to underlying [semantic
+conventions] can interrupt your data flow. If the status of a semantic
+convention that your instrumentation depends on is `Experimental`, it may be
+subject to future breaking changes.
+
+[semantic conventions]: /docs/concepts/semantic-conventions/
+
+{{% /alert %}}
```
chalin, line 37:
```suggestion
{{% alert title="Important" color="warning" %}}

Regardless of an API/SDK's status, if your instrumentation relies on [semantic
conventions] that are marked as [Experimental] in the [semantic conventions
specification], your data flow might be subject to **breaking changes**.

[semantic conventions]: /docs/concepts/semantic-conventions/
[Experimental]: /docs/specs/otel/document-status/
[semantic conventions specification]: /docs/specs/semconv/

{{% /alert %}}
```

chalin (commented):
PTAL at the reworked note.

chalin (approved):
👍🏻

chalin:
Merging. I'll let @svrnm followup with the issue / proposed rework he mentioned.

PR #4011, opened 2024-02-16
[CI] Update `actions/{up,down}load-artifact` to v4

chalin:
- Supersedes and closes #3744
- Supersedes and closes #3746
- The problem is that both updates needed to occur at the same time


svrnm (approved):
thanks for investigating @chalin!

PR #4013, opened 2024-02-16
adding opentext to vendor list

kotagirisandeep:
We are introducing Application Observability as a new capability to OpenText Operations Bridge. Release announcement is at https://docs.microfocus.com/doc/OBS/SaaS/ReleaseNotes. Hence adding OpenText as one of the Vendors supporting OpenTelemetry.

### DIFF
--- a/data/ecosystem/vendors.yaml
+++ b/data/ecosystem/vendors.yaml
@@ -241,7 +241,7 @@
   distribution: true
   nativeOTLP: true
   url: 'https://docs.microfocus.com/doc/OBS/SaaS/ConfigureOTELCollector'
-  contact: ''
+  contact: 'skotagiri@opentext.com'
   oss: false
   commercial: true
 - name: Oracle
--- a/data/ecosystem/vendors.yaml
+++ b/data/ecosystem/vendors.yaml
@@ -237,6 +237,13 @@
   contact: 'hello@openobserve.ai'
   oss: true
   commercial: true
+- name: OpenText
+  distribution: true
+  nativeOTLP: true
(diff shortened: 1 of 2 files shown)


data/ecosystem/vendors.yaml
```diff
@@ -237,6 +237,13 @@
   contact: 'hello@openobserve.ai'
   oss: true
   commercial: true
+- name: OpenText
+  distribution: true
```
theletterf, line 241: I think this is not a distribution, as you're sending data through any OTel Collector, not your own distribution of it.

data/ecosystem/vendors.yaml
```diff
@@ -237,6 +237,13 @@
...
+  nativeOTLP: true
+  url: 'https://docs.microfocus.com/doc/OBS/SaaS/ConfigureOTELCollector'
+  contact: ''
```
theletterf, line 244: Could you add a contact email?

theletterf (changes requested):

data/ecosystem/vendors.yaml
kotagirisandeep: @theletterf at the moment we have our own distribution of OpenTelemetry Collector. Hence set distribution as true.

data/ecosystem/vendors.yaml
kotagirisandeep: I added my contact.

theletterf (approved):
Thanks — all questions clarified. I suggest that you open a separate PR to add the distribution.

PR #4024, opened 2024-02-18
Add github.com/thegeekyasian/go-chi-otel to registry

thegeekyasian:
Adding an adaption of OTEL for gochi/chi to the registry.

https://github.com/thegeekyasian/go-chi-otel


cartermp (approved):

cartermp:
Okay, I will pull this in as the change in `main` that is breaking isn't related to this change.

Isnor:
I don't think this should be included in the repository:

- The repository was an unattributed copy of [otel-chi](https://github.com/riandyrn/otelchi)
- I opened an issue in this repository, and the repository has since been deleted.

svrnm:
thanks for letting us know @Isnor

PR #4040, opened 2024-02-22
fix: adjust dead link

icco:
I noticed that https://community.cncf.io/end-user-community/ no longer existed, and couldn't find a related page, so pointed to the root of the domain.

~~I also removed all adopters with no blog post as defined in the `grace-period-2024-01-01` footnote.~~


icco:
> > I noticed that https://community.cncf.io/end-user-community/ no longer existed, and couldn't find a related page, so pointed to the root of the domain.
> 
> Thanks
> 
> > I also removed all adopters with no blog post as defined in the `grace-period-2024-01-01` footnote.
> 
> We might need to bump that grace period for a little bit (or remove it altogether), the hope was that some of the adopters get back to use with some links, but we didn't get there yet. Can you remove this from this PR for now? Thanks

Done!

svrnm (approved):

PR #4034, opened 2024-02-21
Try to really clarify that you need flask and wekzeug < 3

cartermp:
The evergreen problem that causes 50+ reply threads in slack persists. Herein I try to make it a little more clear what you need to do, until such a time that this resolves upstream.

PR #4047, opened 2024-02-24
Remove requirements for flask and werkzeug in Pythn docs

cartermp:
Flask 3 is now supported in python instrumentations, meaning we can remove the specific instructions for installation + various warnings about that.

Ref: https://github.com/open-telemetry/opentelemetry.io/pull/4034#issuecomment-1962066146

cc @aabmass


svrnm (approved):

swiatekm-sumo (changes requested):
The new instrumentation release isn't yet the default in the operator, so if we want to drop the notice in the operator docs, we should wait for its next release.

theletterf (approved):

swiatekm-sumo:
I'm the release manager for operator 0.95.0, so I'll just approve this PR once it's out, if that's ok with everyone.

cartermp:
@swiatekm-sumo when is the new operator release planned? If it's soon then we can just wait until that's done, but if not then I'll change the Operator docs

swiatekm-sumo:
@cartermp should be out by the end of this week. If you'd prefer to drop the operator change from this PR, I can submit it by myself once the operator release is out.

cartermp:
Sounds good @swiatekm-sumo. I've updated everything here, good to merge?

swiatekm-sumo (approved):

PR #4050, opened 2024-02-26
Improve documentation

Philly-B:
The documentation is mostly referencing the OpenTelemetry namespace classes, however, not for the current active span retrieval. Here is a small extension to the documentation to make this more clear or rather shows that it is also possible to retrieve the active span using the Tracer of the OpenTelemetry namespace.

### DIFF
--- a/content/en/docs/languages/net/instrumentation.md
+++ b/content/en/docs/languages/net/instrumentation.md
@@ -281,6 +281,13 @@ var activity = Activity.Current;
 Note that `using` is not used in the prior example. Doing so will end current
 `Activity`, which is not likely to be desired.
 
+Alternatively you can use the `Tracer` class to obtain the current `TelemetrySpan` to stay within the OpenTelemetry namespace.
+
+```csharp
+var span = Tracer.CurrentSpan;
+// may be null if there is none
+```
+
 ### Add tags to an Activity
 
 Tags (the equivalent of


theletterf (commented):
Thanks! 

+CC @open-telemetry/dotnet-maintainers Please have a look.

content/en/docs/languages/net/instrumentation.md
```diff
@@ -281,6 +281,13 @@ var activity = Activity.Current;
 Note that `using` is not used in the prior example. Doing so will end current
 `Activity`, which is not likely to be desired.
 
+Alternatively you can use the `Tracer` class to obtain the current `TelemetrySpan` to stay within the OpenTelemetry namespace.
```
cijothomas, line 284: if we are documenting this approach here, then it should be shown throughout the doc. Else it'll be inconsistent.

cijothomas (changes requested):
See https://github.com/open-telemetry/opentelemetry.io/pull/4050/files#r1503217209
Could we ensure consistency instead of ad-hoc fixes?

We need to either stick with Activity/friends in the terminology.
Or stick with Span terminology.
Or both.
Whatever we chose, it should be consistent across.

Docs in OTEL .NET repo exclusively shows the Activity terminology, with https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/examples/Console/TestOTelShimWithConsoleExporter.cs being the only example showing the Span terminology. Haven't heard much complaints about that, though there are people who prefer to use Span terminology (to align with rest of otel languages..)

content/en/docs/languages/net/instrumentation.md
reyang: I think having an alternative approach here in the doc makes things confusing to the user. @Philly-B would you elaborate what problem do you want to solve here? Thanks!

svrnm:
@cijothomas , @reyang this is something the .NET SIG needs to figure out. As of today we have 2 documents, the "normal" documentation uses the activity terminology and then there is [this "shim" documentation](https://opentelemetry.io/docs/languages/net/shim/), we can move that to a dedicated discussion if needed.

svrnm:
> > @cijothomas , @reyang this is something the .NET SIG needs to figure out. As of today we have 2 documents, the "normal" documentation uses the activity terminology and then there is [this "shim" documentation](https://opentelemetry.io/docs/languages/net/shim/), we can move that to a dedicated discussion if needed.
> 
> @open-telemetry/dotnet-approvers FYI.
> 
> Regardless of the discussion outcome, I think we should close this PR since there is no intention to have one single document mixing two paths. Would folks agree?

Can we express what is content of this PR with the "right path"?

content/en/docs/languages/net/instrumentation.md
cartermp: Since this is already documented in the shim doc I feel like it's fine not to document here: https://opentelemetry.io/docs/languages/net/shim/#get-the-current-span

Philly-B: I think you are right, that is indeed a bad spot to add that information.
I think my confusion was based on the first examples in the documentation being in the System.Diagnostics namespace and I only later found the shim documentation. 
Thank you for your reply, I think you are right and that documentation hint would not add anything but would rather make it more complicated.

PR #4052, opened 2024-02-26
Improve JS Instrumentation Docs

legalimpurity:
Since https://github.com/open-telemetry/opentelemetry-js/pull/4427 the method addMetricReader on MeterProvider is deprecated and will be removed in SDK 2.0. Would be better if we remove deprecated methods from documentation and change the sample code to reflect the new way to add metric reader.


theletterf (approved):
@open-telemetry/javascript-maintainers Please have a look!

cartermp:
It seems like this change is released, which would make it good, but I'd like @open-telemetry/javascript-approvers to confirm

pichlermarc (approved):
Yep, that's the new way of adding `MetricReader`s 
Thanks for taking care of this. :+1:

svrnm (approved):

svrnm:
thank you @legalimpurity !

PR #4053, opened 2024-02-26
Trace Exemplars to just "Exemplars"

cijothomas:
Rename the header to say "Exemplars". Was there any reason why it was called "Trace Exemplars" originally? After all, exemplars is a [Metrics feature](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#exemplar), and **one** use-case of it is to allow correlating to trace, so just "Exemplars" or "Metric Exemplars" is more suited here.


austinlparker:
I assume it's that because the only current implementation of exemplars is for metrics -> traces, but yeah, I'd agree that we should just say 'exemplars'

reyang (approved):

svrnm (approved):

PR #3996, opened 2024-02-14
Update instrumentation concept page

svrnm:
A try to improve the concept page about instrumentation. I tried to reduce the manual vs automatic language and also added a lot more benefits to why people should use OpenTelemetry.

PR #4066, opened 2024-02-28
Concepts pages: copyedits for instrumentation pages

chalin:
- Follow up to #3996
- Copyedits in situations where the phrases didn't make sense grammatically
- Includes other copyedits, such as reordering of solutions so zero-code is presented first.
- **Preview**: https://deploy-preview-4066--opentelemetry.netlify.app/docs/concepts/instrumentation/


content/en/docs/concepts/instrumentation/_index.md
```diff
@@ -11,25 +10,24 @@ from the system's components must emit [traces](/docs/concepts/signals/traces/),
...
 way to think of it is that they provide information about what's happening _at
 the edges_ of your application.
 
-It's generally recommended that you use both solutions when you can.
+**Code-based** solutions allow you to get deeper insight and rich telemetry from
+your application itself. They let you use the OpenTelemetry API to generate
+telemetry from your application, which acts as an essential complement to the
+telemetry generated by zero-code solutions.
+
+You can use both solutions simultaneously.
```
chalin, line 30: I'm unsure if this was on purpose, but IMHO, it makes more sense to me to mention zero-code first. WDYT @svrnm et al.?

svrnm: this was on purpose yes, we had a discussion on the original issue around that: https://github.com/open-telemetry/opentelemetry.io/pull/3996#discussion_r1493791120

Here is my reasoning:

> I rather prefer leading with code-based and then letting people know that there is zero-code as well, since not all languages have zero-code and for more and more use cases code-based should be what people are doing, that's why I moved that wording around that below and also tried to give a little bit more "when to use what" beyond "start with zero code then move forward to code-based"

svrnm (approved):

content/en/docs/concepts/instrumentation/_index.md
chalin: Thanks for the context. So, would you like for me to revert the order?
Or maybe we can add a note along the lines of what you shared: "not all languages have zero-code and for more and more use cases code-based should be what people are doing"?

cartermp: Let's go back to the original order I think. For now, at least.

chalin: Original order restored.

chalin:
Given that @svrnm approved, and the order was restored, I'm going to merge now.

PR #4067, opened 2024-02-28
[Hugo] Don't auto-linkify in prose & fix registry links

chalin:
- An upcoming semconv spec changes breaks the build because [Hugo is configured by default to linkify plain-text URLs](https://gohugo.io/getting-started/configuration-markup/#goldmark) -- there is a fictitious raw URL given as an example in prose that is treated as an external link and so link checking fails. That shouldn't happen.
  **Use the markdown sytnax `<URL>` in prose to introduce the raw URL as a link.** /cc @theletterf et al.
- Turns off the Goldmark [linkify](https://github.github.com/gfm/#autolinks-extension-) extension.
- Fixes some affected links in registry YAML files
- Other minor copyedits in a few registry YAML files
- **Preview**, e.g.:
  - https://deploy-preview-4067--opentelemetry.netlify.app/ecosystem/registry/?s=OpenTelemetry+Jaeger+context+propagation

### DIFF
--- a/data/registry/instrumentation-erlang-req.yml
+++ b/data/registry/instrumentation-erlang-req.yml
@@ -6,8 +6,7 @@ tags:
   - instrumentation
   - erlang
 license: Apache 2.0
-description:
-  See https://hex.pm/packages/opentelemetry_req for usage instructions.
+description: For details, see <https://hex.pm/packages/opentelemetry_req>.
 authors:
   - name: OpenTelemetry Authors
 urls:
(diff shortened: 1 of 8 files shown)


chalin:
FYI, here's a fix to the semconv page (well, registry entry) so that the error being addressed here won't actually appear. Still, the error was useful in helping us identify that we don't want the linkify extension enabled:

- https://github.com/open-telemetry/semantic-conventions/pull/783

theletterf (approved):
TIL! Thanks!

Issue #4064, opened 2024-02-28
Upgrade to Hugo 0.123+ & address breaking changes


chalin:
[Hugo 0.123.0](https://github.com/gohugoio/hugo/releases/tag/v0.123.0) comes with a **rewritten core** and some **breaking changes** -- and some bugs (there are [5 patch release of 0.123](https://github.com/gohugoio/hugo/releases/tag/v0.123.5) as of the time of writing). But it's not all bad news: the new core promises quicker build times for large sites, among other performance and feature improvements.

~But, I'm unable to build Docsy and it's User Guide properly yet using Hugo 0.123, so I'd suggest we wait until that is resolved before we upgrade for OTel.io~ --> Docsy has been upgraded via https://github.com/google/docsy/pull/1877.

**Bugs** impacting the website:

- [x] [Image block attribute disappears with 0.123.x](https://discourse.gohugo.io/t/image-block-attribute-disappears-with-0-123-x/48677)
  - https://github.com/gohugoio/hugo/issues/12203
- [x] [Subpage missing from section's page list when using 0.123.8](https://discourse.gohugo.io/t/subpage-missing-from-sections-page-list-when-using-0-123-8/48789)
  - https://github.com/gohugoio/hugo/issues/12263

**Differences** in Hugo behavior:

- [Anchor-only markdown links get prepended with index page path in 0.123.x](https://discourse.gohugo.io/t/anchor-only-markdown-links-get-prepended-with-index-page-path-in-0-123-x/48562)

**References**:

- [Hugo 0.123.0](https://github.com/gohugoio/hugo/releases/tag/v0.123.0), [Hugo 0.124.1](https://github.com/gohugoio/hugo/releases/tag/v0.124.1)
- https://github.com/google/docsy/issues/1873

---

Check for diffs using one of:

```console
$ (cd public && git diff -bw --ignore-blank-lines -I 'Hugo 0.12|navbar-dark|theme="dark"|code.jquery|sha512-|<img |var params =|class="text-') | grep ^diff | grep -vE '_print|\.xml$' | head
```

chalin:
With #4093, I think that we're good!

PR #4069, opened 2024-02-28
chore: update dependency

Arhell:
Try to update


Arhell:
> appreciated, @Arhell but we are currently not able to bump versions due to hugo concerns, see #4064

Got it, thanks, I'll close it for now

PR #4076, opened 2024-03-01
Add support for all fix commands to pr action

svrnm:
This way all `fix:*` commands in the package.json are recognized by the PR action, so we can also run /fix: dictname or /fix: markdown ...


svrnm:
ah... that doesn't work because the repo is checked out at a later stage

PR #4085, opened 2024-03-03
Update default value for controller and view common experimental telemetry

jeffmadrid:
Change the default value of the following properties for Controller and View Telemetry to match the default value in code.

- `otel.instrumentation.common.experimental.controller-telemetry.enabled` is set to `false` by default [here](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/fdeb032b33f1439f70ef4a239c0b08b8151757d9/javaagent-extension-api/src/main/java/io/opentelemetry/javaagent/bootstrap/internal/ExperimentalConfig.java#L37)
- `otel.instrumentation.common.experimental.view-telemetry.enabled` is set to `false` by default [here](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/fdeb032b33f1439f70ef4a239c0b08b8151757d9/javaagent-extension-api/src/main/java/io/opentelemetry/javaagent/bootstrap/internal/ExperimentalConfig.java#L42)


svrnm:
thanks @jeffmadrid !

@open-telemetry/java-approvers PTAL

laurit (approved):

PR #4086, opened 2024-03-03
updated connector readme

MinalAC:
Updated the connector.md to mention that servicegraph and spanmetrics processors are now deprecated and replaced by connectors as mentioned in the open-telemetry releases page.

### DIFF
--- a/content/en/docs/collector/building/connector.md
+++ b/content/en/docs/collector/building/connector.md
@@ -44,7 +44,8 @@ pipeline and send it to the metrics pipeline. Historically, some processors
 transmitted data by making use of a work-around that follows a bad practice
 where a processor directly exports data after processing. The connector
 component solves the need for this work-around and the processors that used the
-work around have been deprecated.
+work around have been deprecated.On the same line, above mentioned processors
+are also now deprecated in recent releases and are replaced by the connectors.
 
 Additional details about the connector's full capabilities can be found at the
 following links:


content/en/docs/collector/building/connector.md
```diff
@@ -44,7 +44,8 @@ pipeline and send it to the metrics pipeline. Historically, some processors
 transmitted data by making use of a work-around that follows a bad practice
 where a processor directly exports data after processing. The connector
 component solves the need for this work-around and the processors that used the
-work around have been deprecated.
+work around have been deprecated.On the same line, above mentioned processors
+are also now deprecated in recent releases and are replaced by the connectors.
```
svrnm, line 48:
```suggestion
work around have been deprecated. On the same line, above mentioned processors
are also now deprecated in recent releases and are replaced by the connectors.
```

svrnm:
thanks for your contribution @MinalAC ! 

I wonder if we need to make it clearer from the start of the paragraph that the processors are historically. Maybe we lead with 

> Before collector version X processors have been used to converte telemetry data of one type
into another one. A few examples are the now depreacted spanmetrics processor, as well as the
servicegraph processor.  [...]

MinalAC:
@svrnm 
Okay, but to mention this -- "Before collector version X processors have been used..." , is it the case that all the processors got replaced by collectors or only those two(servicegraph and spanmetrics ) got replaced by the collectors and there are no more processors used?

theletterf:
Could we merge this one?

mhausenblas (approved):
LGTM

Issue #4107, opened 2024-03-07
Updates to opentelemetry JS packages used in the website break netlify


svrnm:
This happened before: We update the opentelemetry-*-js packages used in tracing.js, and since tracing.js is disabled in the pre-prod built we do not catch errors. Currently builds are affected by that, see

https://app.netlify.com/sites/opentelemetry/deploys/65e9d0e4cf5af70008c332a1

> 3:37:24 PM: Error: error building site: JSBUILD: failed to transform "js/tracing.js" (text/javascript): "/opt/build/repo/assets/js/tracing.js:10:9": No matching export in "node_modules/@opentelemetry/semantic-conventions/build/esm/index.js" for import "SEMRESATTRS_SERVICE_NAME"

There are 2 things we need to fix:
* The current problem: updating the code with the changes to semconv
* The general problem: figure out how we can enable tracing.js safely for pre-prod, to verify that this does not happen, maybe we can influence the used exporter instead of disabling tracing by not including tracing.js ?

svrnm:
Closing this now, if this happens again we might want to revisit to address the general problem

PR #4109, opened 2024-03-07
Update how we do tracing

svrnm:
Addresses problem (2) in #4107 by always including tracing.js in the build. This PR includes an additional features:

Users can run `npm run serve` with OTEL_EXPORTER_OTLP_ENDPOINT or OTEL_TRACES_EXPORTER set to either define their own OTLP endpoint or to write spans to the console, helps debugging that code


cartermp (approved):

theletterf (approved):

layouts/partials/hooks/body-end.html
```diff
@@ -1,4 +1,25 @@
...
+<script>
+  const collectorType = 'otlp';
+  const otlpTracesEndpointUrl = '{{ getenv "OTEL_EXPORTER_OTLP_ENDPOINT" }}';
+  const tracingEnabled = true;
+</script>
+{{ else -}}
+<script>
+  const tracingEnabled = false;
+</script>
 {{ end -}}
```
chalin, line 23: How about assuming production, and making it a little more dry? Something like this could replace lines 1 to 23, inclusive:

```html
<script>
  var tracingEnabled = true;
  var collectorType = 'otlp';
  var otlpTracesEndpointUrl = 'https://otelwebtelemetry.com/v1/traces';
{{ if eq (getenv "OTEL_TRACES_EXPORTER") "console" -}}
  collectorType = 'console';
{{ else if getenv "OTEL_EXPORTER_OTLP_ENDPOINT" -}}
  otlpTracesEndpointUrl = '{{ getenv "OTEL_EXPORTER_OTLP_ENDPOINT" }}';
{{ else if not hugo.IsProduction -}}
  tracingEnabled = false;
{{ end -}}
</script>
```

chalin (approved):
Nice fix. See inline for a suggested code refactoring. Otherwise, LGTM if y'all are happy!

layouts/partials/hooks/body-end.html
svrnm: yeah, that's a good idea, let me do that.

svrnm: applied

.github/workflows/check-links.yml
```diff
@@ -20,7 +20,7 @@ jobs:
           jq '.devDependencies |= with_entries(
                   select(.key | test("prefix|hugo|css"))
                 )
-                | del(.dependencies, .optionalDependencies)' \
+                | del(.optionalDependencies)' \
```
svrnm, line 23: This breaks if tracing.js is always included and by that always build, so we need to decide what is more important for us (not sure yet exactly on the overall performance impact)

svrnm: This is  blocking the PR to be merged, @open-telemetry/docs-approvers please take a look

PR #4117, opened 2024-03-10
PoC: Add Vale linter with Google rules

theletterf:
Following https://github.com/open-telemetry/opentelemetry.io/discussions/3811, this PR adds the necessary files for checking docs locally against the Google style guide using the Vale linter. It also adds a GitHub workflow action that only tests new or modified files - the usefulness of this action in its current state is debatable though.

To test locally:

1. Install Vale: https://vale.sh/docs/vale-cli/installation/
2. Run `vale sync` from the repo.
3. Run `vale <file>` or use the [VS Code extension](https://marketplace.visualstudio.com/items?itemName=chrischinchilla.vale-vscode).

Things we could improve/add:

- If we want to tweak the styles or have our own rules (for example, for things like terminology), we could add a static styles folder to `.github/vale`, for example.
- The previous step is sort of required if we want to go forward with the GH workflow, as the error levels and how we want the pipeline to fail requires adjustments.
- I think Vale can coexist with cspell, given its emphasis on style and syntax. If we wanted to, we'd still consider replacing spell with a Vale [vocabulary ](https://vale.sh/docs/topics/vocab/) and a [spelling](https://vale.sh/docs/topics/styles/#spelling) rule. 
- Should we integrate this in the local test scripts?
- We'd have to document Vale usage in the Contributing guide (we can do it in this PR).

I think that just the first commit with the ini file and the .gitignore entry would allow us to test Vale while editing.


svrnm:
As discussed previously I am in favour of rolling out vale if it adds additional value above textlint & cspell. I will pull the PR later this week and take a look myself. A few comments

> think Vale can coexist with cspell, given its emphasis on style and syntax. If we wanted to, we'd still consider replacing spell with a Vale [vocabulary ](https://vale.sh/docs/topics/vocab/)and a [spelling](https://vale.sh/docs/topics/styles/#spelling) rule.

A minimum requirement for me would be replacing textlint and migrate all existing configuration to vale. The overlap of functionality is big, this also makes sure that we are not getting into a n+1 situation with all the additional tooling we introduce. If we can replace cspell as well, then I am also for doing it.

> Should we integrate this in the local test scripts?

Yes, a good state that we are in right now, is that we have some parity between local script, CI scripts, and /fix:* commands, let's try to keep that. 

> We'd have to document Vale usage in the Contributing guide (we can do it in this PR).

+1

theletterf:
FWIW, Vale is way faster than textlint:

https://github.com/errata-ai/vale?tab=readme-ov-file#benchmarks

How are we using textlint in this repo? Where could I see the rules?

Migrating from cSpell to Vale would require to move all our current cSpell directives to a single dictionary file and maintain that one, which sounds like too much overhead. I'm in favor of keeping cSpell and replace textlint for Vale.

svrnm:
> FWIW, Vale is way faster than textlint:
> 
> https://github.com/errata-ai/vale?tab=readme-ov-file#benchmarks
> 
> How are we using textlint in this repo? Where could I see the rules?

https://github.com/open-telemetry/opentelemetry.io/blob/main/.textlintrc.yml
https://github.com/open-telemetry/opentelemetry.io/blob/main/package.json#L15

> Migrating from cSpell to Vale would require to move all our current cSpell directives to a single dictionary file and maintain that one, which sounds like too much overhead. I'm in favor of keeping cSpell and replace textlint for Vale.

then let's keep cspell

svrnm:
@theletterf where are we with this PR, can you provide some updates on what is ready and what is missing? I think it would be good to talk about this at the next Comms Call to make sure everyone is onboard with this.

theletterf:
@svrnm Still haven't gotten around converting the textlint rules. I'd be happy to demo current progress in the next SIG though.

theletterf:
I need to temporarily halt this. Can't find the time. :(

svrnm:
> I need to temporarily halt this. Can't find the time. :(

No worries, we have to set our priorities right some times, thanks for giving it a try and also thanks for closing it when needed ❤️

PR #4143, opened 2024-03-12
Add Docker as a prerequisite to Collector quick start example

cthain:
This PR proposes to add Docker as a prerequisite for the Collector quick start example. Althought Docker is ubitquitous for running applications, if it is not installed on the user's machine, it requires setup prior to running this example.

### DIFF
--- a/content/en/docs/collector/quick-start.md
+++ b/content/en/docs/collector/quick-start.md
@@ -26,6 +26,7 @@ Make sure that your developer environment has the following. This page assumes
 that you're using `bash`. Adapt configuration and commands as necessary for your
 preferred shell.
 
+- [Docker](https://www.docker.com/)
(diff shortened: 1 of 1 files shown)


content/en/docs/collector/quick-start.md
```diff
@@ -26,6 +26,7 @@ Make sure that your developer environment has the following. This page assumes
 that you're using `bash`. Adapt configuration and commands as necessary for your
 preferred shell.
 
+- [Docker](https://www.docker.com/)
```
theletterf, line 29:
```suggestion
- [Docker](https://www.docker.com/) or any compatible containers' runtime.
```

theletterf (approved):
LGTM, but consider applying the suggestion. Thanks!

PR #4155, opened 2024-03-13
Remove distribution column from vendors

svrnm:
Since we have a dedicated page for distributions now, I think it is best to remove this column, it also often creates confusion when new entries are generated.


cartermp (approved):
Agree

Issue #2072, opened 2022-12-03
Document the new sub command "components"


Chinwendu20:
**What needs to be changed?**
A new sub-command has been added to the collector core. Lnk to merged [PR](https://github.com/open-telemetry/opentelemetry-collector/pull/6322)

**Additional context**
The command displays the components available in a given distribution. You can use the update in the readme as a guide for documentation. This is an example of how the command works:

```bash
   ./otelcorecol components
```
Sample output:

```yaml

buildinfo:
   command: otelcorecol
   description: Local OpenTelemetry Collector binary, testing only.
   version: 0.62.1-dev
receivers:
   - otlp
processors:
   - memory_limiter
   - batch
exporters:
   - otlp
   - otlphttp
   - logging
extensions:
   - zpages
   - memory_ballast


```

Chinwendu20:
Please the label bug is misleading. I do not know how to update that

svrnm:
@Chinwendu20 do you want to give it a try yourself to add this to the docs?

Chinwendu20:
That would be cool. I might want to do that but as of now, let us leave this open for grabs but I do have a question. Should the documentation be in line with the most recent release of the open telemetry API? I ask this because I know some projects have documentation for each API release.

cartermp:
@Chinwendu20 Latest API version please!

ishworg:
The `components` command is available from 0.67.0.

Chinwendu20:
Thanks @ishworg . Just got to see this message now @cartermp

nerudadhich:
Hi @svrnm @Chinwendu20 I will update the document but not able to find where should we add this.
There are two places that I feel we can add. 
1. In the Readme of component https://github.com/open-telemetry/opentelemetry-collector/tree/main/cmd/otelcorecol
2. Somewhere here in doc https://opentelemetry.io/docs/collector/

Please suggest and I will be happy to add the document.

svrnm:
Hm, I can make 2 suggestions:

https://opentelemetry.io/docs/collector/configuration/
or
https://opentelemetry.io/docs/collector/troubleshooting/

cc @open-telemetry/collector-approvers

mx-psi:
IMO this fits better in troubleshooting

nerudadhich:
[...]

PR #4156, opened 2024-03-13
Troubleshooting added to check available components in the collector

nerudadhich:
Fix for issue - https://github.com/open-telemetry/opentelemetry.io/issues/2072

### DIFF
--- a/content/en/docs/collector/troubleshooting.md
+++ b/content/en/docs/collector/troubleshooting.md
@@ -17,7 +17,8 @@ configured to output to local logs. For details, see
 
 ## Check available components in the collector
 
-Use below subcommand to Outputs available components in the collector distribution including their stability levels.
+Use below subcommand to Outputs available components in the collector
+distribution including their stability levels.
 
 ```sh
 otelcol components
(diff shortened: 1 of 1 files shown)


content/en/docs/collector/troubleshooting.md
```diff
@@ -15,6 +15,172 @@ network issues, it can be helpful to send a small amount of data to a collector
 configured to output to local logs. For details, see
 [Local exporters](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/troubleshooting.md#local-exporters).
 
+## Check available components in the collector
```
tiffany76, line 18:
```suggestion
## Check available components in the Collector
```

tiffany76 (commented):
I've made a few small suggestions to help with clarity and conform to the style guide. 

The sample output seems a little long to me. Maybe showing only a couple of each type of component would be clearer?  The Comms folks will know best.

nerudadhich:
> I've made a few small suggestions to help with clarity and conform to the style guide.
> 
> The sample output seems a little long to me. Maybe showing only a couple of each type of component would be clearer? The Comms folks will know best.

Thanks for comment, I agree with sample output. Let's wait for more inputs.

content/en/docs/collector/troubleshooting.md
svrnm: @theletterf do I remember correctly that we agreed that we write "Collector" if we are referring to the OpenTelemetry Collector?

mx-psi (commented):
Should we note that the output format may change across versions?

nerudadhich:
> Should we note that the output format may change across versions?

I have updated the sample output and added the note as well. Please review again and confirm if anything else need to add/update.

mx-psi (approved):

tiffany76 (commented):
LGTM!

theletterf (approved):

cartermp (approved):

PR #4163, opened 2024-03-14
Change wording to suit global audience

tiffany76:
This PR makes a tiny change to the reliability example on the [Observability Primer](https://opentelemetry.io/docs/concepts/observability-primer/) page. A pair of pants means something different in the UK. :)


cartermp (approved):
Thanks!

PR #4177, opened 2024-03-18
fix: erlang elixir samples

Graborg:
Before: 
`should_sample` only takes into account if the keys matched
config `%{"http.target": "/healthcheck"}` and a span `%{"http.target": "/something_else"}` would match and the span would be dropped.
After:
`should_sample` compares both keys and values
`%{"http.target": "/healthcheck"}` and a span `%{"http.target": "/something_else"}` => {:record_and_sample, [], []}
`%{"http.target": "/healthcheck"}` and a span `%{"http.target": "/healthcheck"}` => {:drop, [], []}

### DIFF
--- a/content/en/blog/2024/opentelemetry-announced-support-for-profiling.md
+++ b/content/en/blog/2024/opentelemetry-announced-support-for-profiling.md
@@ -75,9 +75,9 @@ companies. Across our most popular repositories, OpenTelemetry sees over 30
 million downloads a month<sup>2</sup>, and new open source projects are adopting
 our standards at a regular pace, including
 [Apache Kafka](https://cwiki.apache.org/confluence/display/KAFKA/KIP-714%3A+Client+metrics+and+observability),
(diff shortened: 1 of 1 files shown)


theletterf (approved):
Summoning @open-telemetry/erlang-approvers

tsloughter:
@theletterf thanks!

@Graborg can you explain what was wrong/fixed?

Graborg:
Hi! Sure! Before the function only took into account the that the keys matched: i.e if 
config `%{"http.target": "/healthcheck"}`  and a span `%{"http.target": "/something_else"}`; the span would be dropped. It now looks at the value as well.

tsloughter:
Oooh, hah! Thanks :)

Might be simpler code to just convert to a list and do a `sets:is_disjoint`.

Graborg:
Sure! Something like this maybe?

tsloughter (approved):
Perfect :+1:

Issue #4184, opened 2024-03-20
[Traces Documentation] Clarification on events vs span attributes


benhackett4:
I'll start with the motivating example. Consider a span that is responsible for publishing a bunch of messages. Should the number of failed publishes and successful publishes be attributes on the span, or should there be an event (named something like `FinishedAllPublishAttempts`) with those attributes attached?

More generally, can the documentation at https://opentelemetry.io/docs/concepts/signals/traces/#attributes clarify if it's best practice for span attributes to only include information known at the start of the span? Or is it fine if the span contains attributes with information known only after the span is almost complete?

cartermp:
Using your example, I'll say that:

- If the success or failure of message publishing has a timestamp that's meaningful -- i.e., you need to know the specific point in time that it attempted, succeeded, failed, etc. -- then a Span Event is appropriate to capture data about that operation
- If these sub-operations don't have any meaning in their timestamps, then they should just be attributes on the span

And yeah, attributes can absolutely contain data you don't know during span creation. The downside to that is you can't use those attribute vales when making SDK sampling decisions but that's pretty much it.

cartermp:
Okay, I created https://github.com/open-telemetry/opentelemetry.io/pull/4185

benhackett4:
@cartermp Awesome, thanks so much for the help!! Super appreciated. That all makes sense to me. And for what it's worth, the pull request looks good to me (but I assume I can't approve :P). (Though, do let me know if there's something I can do to help get the pull request approved). Thanks again!

PR #4185, opened 2024-03-20
Clarify span events vs. attributes guidance

cartermp:
fixes https://github.com/open-telemetry/opentelemetry.io/issues/4184

### DIFF
--- a/content/en/docs/concepts/signals/traces.md
+++ b/content/en/docs/concepts/signals/traces.md
@@ -238,6 +238,12 @@ For example, if a span tracks an operation that adds an item to a user's
 shopping cart in an eCommerce system, you can capture the user's ID, the ID of
 the item to add to the cart, and the cart ID.
 
+Attributes can be added to spans during span creation or after creation.
+If you can add an attribute during span creation, it's preferable to do
+that compared to after span creation. That's because it allows those attributes
+to be used in SDK sampling. However, if you only have a value to add after
+the span has been created, you can just add it to the span when it's available.
+
 Attributes have the following rules that each language SDK implements:
 
 - Keys must be non-null string values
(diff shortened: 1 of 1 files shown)


content/en/docs/concepts/signals/traces.md
```diff
@@ -238,6 +238,12 @@ For example, if a span tracks an operation that adds an item to a user's
 shopping cart in an eCommerce system, you can capture the user's ID, the ID of
 the item to add to the cart, and the cart ID.
 
+Attributes can be added to spans during span creation or after creation.
+If you can add an attribute during span creation, it's preferable to do
+that compared to after span creation. That's because it allows those attributes
+to be used in SDK sampling. However, if you only have a value to add after
+the span has been created, you can just add it to the span when it's available.
```
theletterf, line 245:
```suggestion
You can add attributes to spans during or after span creation. Always add
attributes at span creation to make the attributes available to SDK sampling.
If you have to add a value after span creation, update the span with the value.
```

content/en/docs/concepts/signals/traces.md
```diff
@@ -267,6 +273,19 @@ and an end.
...
+Span Events contain attributes as well, so you may wonder, when is it appropriate
+to use a Span Event or create a new Attribute on the current span?
+
+It all comes down to if a specific timestamp is meaningful.
+
+For example, if you are tracking an operation with a span and the operation completes,
+you may wish to add data from the operation to your telemetry. If the timestamp in
+which the operation completes is important to know, then a Span Event is an appropriate
+way to model things with telemetry. However, if the timestamp is not meaningful,
+then it's better to attach the data as a Span Attribute instead.
```
theletterf, line 287:
```suggestion
#### When to use span events versus span attributes

Since span events also contain attributes, the question of when to use events
instead of attributes might not always have an obvious answer. To inform your
decision, consider whether a specific timestamp is meaningful.

For example, when you're tracking an operation with a span and the operation
completes, you might want to add data from the operation to your telemetry.

- If the timestamp in which the operation completes is meaningful or relevant,
attach the data to a span event.
- If the timestamp isn't meaningful, attach the data as a span attribute.
```

theletterf (approved):
Some suggestions for clarity / structure.

theletterf:
@cartermp Whoops, you skipped my suggestions. :)

cartermp:
whoopsie, lemme get to that

PR #4190, opened 2024-03-21
Update span events/attributes guidance based on missed feedback

cartermp:
I missed feedback here: https://github.com/open-telemetry/opentelemetry.io/pull/4185


cartermp:
@theletterf I made tiiiny tweaks based on your suggestions but these should largely be the same.

theletterf (approved):
Thanks mate!

PR #4198, opened 2024-03-23
Add some common tags to aws sdk instrumentation metadata

cartermp:
This has come up several times at work (almost always when SQS is involved). This should make it more searchable.


cartermp:
This works on the preview

Issue #2548, opened 2023-03-28
[Meta] Sampling


svrnm:
Add basic documentation for all languages about Sampling:

* [x] JavaScript (https://opentelemetry.io/docs/instrumentation/js/sampling/) use as reference
* [x] Ruby (https://github.com/open-telemetry/opentelemetry.io/pull/2505)
* [ ] Python (WIP: https://github.com/open-telemetry/opentelemetry.io/pull/2485)
* [ ] C++
* [ ] .NET
* [x] Erlang/Elixir (https://github.com/open-telemetry/opentelemetry.io/pull/2556)
* [x] Go
* [ ] Java
* [ ] PHP
* [ ] Rust
* [ ] Swift

# Note

To implement this for a new language, please take a documentation from an existing language (https://opentelemetry.io/docs/instrumentation/go/sampling/ or https://opentelemetry.io/docs/instrumentation/js/sampling/) and adjust it for the language you picked

mahitha00:
is  issue still open

svrnm:
> is issue still open

yes, this still needs to be resolved for some of the languages.

SylvanasGr:
Hello @svrnm,

I hope you're doing well! I've just made my first contribution by opening PR #4199 . Whenever you find a moment, could you kindly review it?

Thank you so much for your time and assistance!

svrnm:
> I hope you're doing well! I've just made my first contribution by opening PR #4199 . Whenever you find a moment, could you kindly review it?

Thank you. Looks like @cartermp already took a look :-)

PR #4199, opened 2024-03-24
[feature]: Add sampling documentation for java

konstantinoskolios:
Hello :)

I've addressed issue #2548 by adding Java Sampling Documentation, referencing the 'Go Sampling' guide. Could you please review and provide feedback for improvements?

Thanks!


content/en/docs/languages/java/sampling.md
```diff
@@ -0,0 +1,36 @@
...
+
+`alwaysOn` and `alwaysOff` are self-explanatory values. `alwaysOn` means that every span is sampled, while alwaysOff means that no span is sampled. When you’re getting started, or in a development environment, use `alwaysOn`.
+
+Other samplers include:
+
+- `traceIdRatioBased`, which samples a fraction of spans, based on the fraction given to the sampler. If you set `0.5`, half of all the spans are sampled.
+- `parentBased`, which uses the parent span to make sampling decisions, if present.
+  By default, the tracer provider uses a parentBased sampler with the `alwaysOn` sampler.
+
+When in a production environment, consider using the `parentBased` sampler with the `traceIdRatioBased` sampler.
```
cartermp, line 36: So maybe this is the right place to figure out the guidance, but I think it's a little more complicated than this.

Consider a scenario where a service you don't control is upstream of your service. If that service always samples, then your service will always sample too, regardless of your value set for `traceIdRatioBased`. However, if you remove the `parentBased` sampler, then you can end up with messed up traces if a downstream service doesn't also sample in the exact same way.

There's...probably a whole article on production sampling best practices we should write here.

cartermp:
@SylvanasGr re my comment about production guidance -- we don't need to block on it, since the doc as you've written it is good w.r.t. our incomplete guidance on the topic. If we hash it out here quickly and want to add it to this doc, then we'll do that, but don't take the comment as blocking.

cartermp (approved):

cartermp:
@open-telemetry/java-approvers PTAL!

svrnm (approved):

trask (commented):
I think most Java users are using the Java agent, Spring Boot Starter, or SDK autoconfigure module (and these are generally our recommended solutions vs SDK manual configuration).

I wonder how we can make clear that this page doesn't apply to those users to avoid confusion?

svrnm:
> I think most Java users are using the Java agent, Spring Boot Starter, or SDK autoconfigure module (and these are generally our recommended solutions vs SDK manual configuration).
> 
> I wonder how we can make clear that this page doesn't apply to those users to avoid confusion?

good call out. On other pages we have a "Note" at the top of the page, e.g. https://opentelemetry.io/docs/languages/java/exporters/ and https://opentelemetry.io/docs/languages/java/resources/

cartermp:
For this:

> I wonder how we can make clear that this page doesn't apply to those users to avoid confusion?

I would propose having a top level section that focuses on the env vars / agent properties. Then we can still have a section below for controlling sampling in code as an alternative.

SylvanasGr:
I'm unsure about how to proceed with this pull request.

Any guidance would be greatly appreciated! 😄

zeitlinger:
Here's a page that has a working example for configuring samplers for the spring starter: https://opentelemetry.io/docs/languages/java/automatic/spring-boot/#exclude-actuator-endpoints-from-tracing

I'd suggest to make links both ways.

zeitlinger:
> I'm unsure about how to proceed with this pull request.
> 
> Any guidance would be greatly appreciated! 😄

I would go ahead with this PR.

It's true that we can do more, but as long as we have a [note](https://github.com/open-telemetry/opentelemetry.io/pull/4199#issuecomment-2022450169) I'd go forward with this PR.

svrnm:
@open-telemetry/java-approvers PTAL

breedx-splk (commented):
I think this is a fine start (tho it looks like it currently needs an update). If we want to merge it and improve from there, I think that's fine. We should also include Fabri's suggestions.

What I might call out as opportunities for future improvements:

* Emphasize that most the defaults suffice for most users, and that most users don't need to customize the sampler.
* Make it clear that if you don't use a parent-based sampler to wrap these other selections, you may get incomplete or broken traces.

Thanks @SylvanasGr !

breedx-splk (approved):

cartermp:
Merged since it's been a hot minute and this is much better as-is than the current state in production

PR #4207, opened 2024-03-26
Add IBMMQ instrumentation to registry

ibmmqmet:


svrnm:
Thanks. Are you able to sign the CLA? Is this a "first class" offering by IBM? If so we can also add it to the integrations page

ibmmqmet:
Still working out which of the many ibm CLAs is the right process here - will sign once I find the right person/internal approval mechanism. 

It's not currently a supported product offering so I wouldn't want to call it "first-class".

chalin (commented):
CI checks are green. I'll let the other maintainers give their final stamp of approval.

cartermp (approved):

PR #4218, opened 2024-03-27
Add "next steps" section for Java resources

### LINKED ISSUE #4217: Missing "Next Steps" section in Java resources page
**What needs to be changed?** 
The documentation around [OpenTelemetry Java Resources](https://opentelemetry.io/docs/languages/java/resources/) does not seem to have a `Next Steps` section like [the one in Javascript](https://opentelemetry.io/docs/languages/js/resources/#next-steps) that makes the users aware of the possibility of additional resource detectors along with a link to find them.

I think adding a similar short section to other languages would be a good idea. 

`Resources` located at `content/en/docs/languages/java/resources.md`

I'm willing to add this for `Java`, if it seems like a good idea to the community.

psx95:
Fixes #4217

### DIFF
--- a/content/en/docs/languages/java/resources.md
+++ b/content/en/docs/languages/java/resources.md
@@ -100,3 +100,8 @@ SdkLoggerProvider sdkLoggerProvider = SdkLoggerProvider.builder()
     ...
     .build();
 ```
+
+
+## Next steps
+
+There are more resource detectors that you can add to your configuration. You can find them in the [opentelemetry-java-contrib](https://github.com/open-telemetry/opentelemetry-java-contrib) repository.


content/en/docs/languages/java/resources.md
```diff
@@ -100,3 +100,10 @@ SdkLoggerProvider sdkLoggerProvider = SdkLoggerProvider.builder()
     ...
     .build();
 ```
+
+## Next steps
+
+There are more resource detectors that you can add to your configuration. You
+can find them in the
+[opentelemetry-java-contrib](https://github.com/open-telemetry/opentelemetry-java-contrib)
+repository.
```
breedx-splk, line 109: Could perhaps also add a note that most of those are in the [resource-providers](https://github.com/open-telemetry/opentelemetry-java-contrib/tree/main/resource-providers) module...I think that could be some additional steering.

breedx-splk (approved):
Thanks for the addition! I had one small note about mentioning the submodule. Cheers.

content/en/docs/languages/java/resources.md
psx95: Ah, good catch. The vendor specific resource detection modules like [aws-resources](https://github.com/open-telemetry/opentelemetry-java-contrib/tree/main/aws-resources) and [gcp-resources](https://github.com/open-telemetry/opentelemetry-java-contrib/tree/main/gcp-resources) are listed directly under the root repository, while the common ones are bundled under `resource-providers`. 

I will raise this question in the next SIG meeting to club them under a single folder for better discoverability.
 
However, given the current directory structure, do you think we should mention to specifically look for `resource-providers` module ? I think that might lead the users to think that it is the only one in the repo.

psx95: Had a discussion in the Java SIG meeting regarding this, it seems like it would be better to provide a list of locations where to find resource-detectors. I will update the PR to do so.

cartermp (approved):

PR #4285, opened 2024-04-11
Fix default collector deployment receiver endpoint

cwildman:
In the [default deployment example](https://opentelemetry.io/docs/collector/deployment/agent/#example) the export configuration is expecting to use OTLP over HTTP on port 4318.
```
export OTEL_EXPORTER_OTLP_ENDPOINT=http://collector.example.com:4318
```

 The collector examples though is then showing configuration for receivers on gRPC over 4317, instead of 4318.
 
 I assume this was just an oversight when HTTP became the default.


theletterf:
@open-telemetry/collector-approvers WDYT?

theletterf (approved):

codeboten (approved):

Issue #3895, opened 2024-01-30
Remove processor from glossary


cijothomas:
https://opentelemetry.io/docs/concepts/glossary/#processor talks about Instrumentation Library, but not sure what is the connection between Instrumentation Library and processor here....

cijothomas:
It may be better to remove it, as it is a collector specific term. (There are other processors like spanprocessor/logrecordprocessor, but not in the glossary, and don't think we should have it for the processor in the collector in the main glossary).

svrnm:
Thanks for calling this out. We had a few places in the past already that mixed instrumentation libraries with other components (here I would say it's language specific SDK)

Feel free to raise a PR

theletterf:
@cijothomas Will you open the PR? Thanks!

ricardoamaro:
https://github.com/open-telemetry/opentelemetry.io/pull/4307

svrnm:
@ricardoamaro please look above that @IrisDyr asked to pick this first and we should give people some time to provide their PR after volunteering. This is a trivial change, but I want to be respectful here. @IrisDyr please let us know if you are OK with us merging @ricardoamaro's PR.

IrisDyr:
It's okay for me to merge it. Thank you @svrnm

ricardoamaro:
Thank you, @IrisDyr 👍🏽! I would also be ok waiting more. Taking the opportunity here to learn the ropes :)

PR #4307, opened 2024-04-16
Remove processor from glossary

ricardoamaro:
https://opentelemetry.io/docs/concepts/glossary/#processor talks about Instrumentation Library, but we should remove it, as it is a collector specific term.


svrnm:
blocked, see https://github.com/open-telemetry/opentelemetry.io/issues/3895#issuecomment-2059389817

svrnm (approved):

PR #4313, opened 2024-04-17
Add pr labeler workflow

svrnm:
Copied from https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/.github/workflows/label-prs.yml (cc @pichlermarc) to enable auto-labeling of PRs additionally.

### DIFF
--- a/.github/component-label-map.yml
+++ b/.github/component-label-map.yml
@@ -74,4 +74,3 @@ sig:spec:
   - changed-files:
       - any-glob-to-any-file:
           - content/en/docs/specs/**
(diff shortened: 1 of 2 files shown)


theletterf (approved):
Nice! Could we also add a label for code related work (CSS, templates, etc.)?

svrnm:
> Nice! Could we also add a label for code related work (CSS, templates, etc.)?

probably yes, I have not looked into all the details but we can start with the labeling for SIGs and then work from there

PR #4317, opened 2024-04-18
Update java libraries page to use intro shortcode

svrnm:
This PR updates the libraries page for java to use the intro shortcode, which uses common wording across languages and adds some words about out of the box instrumented libraries. I also updated the integrations entry for elastic search api library since it's the one java library that has "native" otel (all other integrations require a 1st party extension)


cartermp (approved):

cartermp:
robot sad

cartermp:
well poo, I think the 'ole refache fixer doesn't work anymore

theletterf (approved):

PR #4246, opened 2024-04-04
Unify internal observability documentation - 1 of 3

tiffany76:
Partially fixes https://github.com/open-telemetry/opentelemetry-collector/issues/8886.

Relates to 4192. For ease of review, that larger PR is being split into three smaller ones. This is the first of three.

This PR addresses the first section of a new internal observability document on opentelemetry.io:
- How to enable observability

Follow up PRs (2 and 3 of 3) will be created once this one is merged.

cc: @mx-psi @codeboten @theletterf @svrnm

PR #4192, opened 2024-03-22
Unify internal observability documentation and move it to website

tiffany76:
Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/8886.

This PR brings together three different documents from the Collector repo: [Troubleshooting](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/troubleshooting.md), [OpenTelemetry Collector Observability](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/observability.md), and [Monitoring](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/monitoring.md). 

**NOTE: Work on this PR is broken up into three smaller pull requests. See bullet list below for corresponding PRs.**

Per @mx-psi, this new document should describe:

- How to enable observability (#4246)
- What is our observability offering and what each metric means (#4322)
- How to effectively use this telemetry to monitor the Collector (#4529)

The merged and reworked file will be located on opentelemetry.io, with its precise location in the nav hierarchy and title to be decided. 

cc @codeboten 

---

**Preview**: https://deploy-preview-4192--opentelemetry.netlify.app/docs/collector/internal-telemetry/

PR #4322, opened 2024-04-18
Unify internal observability documentation - 2 of 3

### LINKED ISSUE #4278: Add list of metrics by level to Collector's internal telemetry docs
**Desired feature or idea**: The documentation for configuring internal metrics for the Collector should include a list of the metrics available at each level (basic, normal, detailed, etc.). Work is being done on this list in https://github.com/open-telemetry/opentelemetry-collector/pull/9767. Once the list is finalized, it should be added to the website documentation.

**Additional context**: 
- See https://github.com/open-telemetry/opentelemetry.io/pull/4246#discussion_r1557114937 for additional context. 
- Page location: https://opentelemetry.io/docs/collector/internal-telemetry/#configure-internal-metrics

tiffany76:
This PR is the second of three. Together they will create a new internal observability document on opentelemetry.io, combining and migrating multiple markdown files from the Collector repo. This PR covers the section on **types of internal observability**.

Link to tracking issue: https://github.com/open-telemetry/opentelemetry-collector/issues/8886.

Also fixes #4278.

Also relates to:

- #4192. For ease of review, this larger PR is being split into three smaller ones. 
- #4246

Follow up PR (3 of 3) will be created once this one is merged.

---
**PREVIEW: https://deploy-preview-4322--opentelemetry.netlify.app/docs/collector/internal-telemetry/**


tiffany76:
Just getting started. I'll tag everyone when it's ready for review. Thanks!

tiffany76:
Hi @svrnm , @theletterf, @mx-psi, @codeboten,

This PR is ready for review. I'm also adding a few in-line comments and questions for the group. Thanks!

content/en/docs/collector/internal-telemetry.md
```diff
@@ -113,3 +118,148 @@ journalctl | grep otelcol | grep Error
...
+
+The OpenTelemetry Collector aims to be an exemplar of observable service by
+clearly exposing its own operational metrics. It also collects host resource
+metrics that can help you understand if problems are caused by a different
+process on the same host. In this section, you can explore the different types
+of observability emitted by the Collector itself.
+
+### Values observable with internal metrics
+
+The Collector emits internal metrics for the following **current values**:
```
tiffany76, line 132: Could the @open-telemetry/collector-approvers confirm that all of these metrics have been implemented? I did my best to check the repo and/or the metrics emitted by my own collector instance, but the lists definitely require validation. Thanks!

content/en/docs/collector/internal-telemetry.md
```diff
@@ -113,3 +118,148 @@ journalctl | grep otelcol | grep Error
...
+- Total incoming connection count, broken down by receiver.
+
+- Uptime since start.
+
+### List of internal metrics
+
+The following table identifies each internal metric by name and description.
+Each metric is also categorized by instrumentation type and level of verbosity.
+
+| Metric name                                       | Description                                                                               | Type      | Level      |
```
tiffany76, line 189: As you can see in the [preview](https://deploy-preview-4322--opentelemetry.netlify.app/), the table is a bit wide and might require scrolling in a narrow window. Does this concern us? If so, I have three thoughts on how to solve it.

1. I can remove the Type column from the table. That information is a nice-to-have, but it is available elsewhere.
2. I can remove the Level column and instead create tabbed panes in the [Configure internal metrics](https://opentelemetry.io/docs/collector/internal-telemetry/#configure-internal-metrics) section: a pane for Basic, Normal, Detailed that lists the corresponding metric names. 
3. I can do both 1 and 2.

WDYT?

content/en/docs/collector/internal-telemetry.md
```diff
@@ -113,3 +118,148 @@ journalctl | grep otelcol | grep Error
...
+| `otelcol_exporter_send_failed_metric_points`      | Number of metric points that exporter(s) failed to send to destination.                   | Counter   |            |
+| `otelcol_exporter_send_failed_spans`              | Number of spans that exporter(s) failed to send to destination.                           | Counter   |            |
+| `otelcol_exporter_sent_log_records`               | Number of logs successfully sent to destination.                                          | Counter   |            |
+| `otelcol_exporter_sent_metric_points`             | Number of metric points successfully sent to destination.                                 | Counter   |            |
+| `otelcol_exporter_sent_spans`                     | Number of spans successfully sent to destination.                                         | Counter   |            |
+| `otelcol_http_client_active_requests`             | Number of active HTTP client requests.                                                    | Counter   | `detailed` |
```
tiffany76, line 202: Tagging @dmitryax and @mx-psi for this question. I understand that the default level for internal metrics was changed from `basic` to `normal` in https://github.com/open-telemetry/opentelemetry-collector/pull/9767. Does that mean none of these metrics now have a default level of basic? And if so, for my own understanding, why were the `processor_batch` metrics separately updated to `normal`? I'd like to fill in the rest of the Level column, but I'm not sure whether to put `basic` or `normal`. Please advise. Thanks!

content/en/docs/collector/internal-telemetry.md
```diff
@@ -113,3 +118,148 @@ journalctl | grep otelcol | grep Error
...
+| `otelcol_rpc_server_request_size`                 | Measures the size of RPC request messages (uncompressed).                                 | Histogram | `detailed` |
+| `otelcol_rpc_server_requests_per_rpc`             | Measures the number of messages received per RPC. Should be 1 for all non-streaming RPCs. | Histogram | `detailed` |
+| `otelcol_rpc_server_response_size`                | Measures the size of RPC response messages (uncompressed).                                | Histogram | `detailed` |
+| `otelcol_rpc_server_responses_per_rpc`            | Measures the number of messages sent per RPC. Should be 1 for all non-streaming RPCs.     | Histogram | `detailed` |
+| `otelcol_scraper_errored_metric_points`           | Number of metric points the Collector failed to scrape.                                   | Counter   |            |
+| `otelcol_scraper_scraped_metric_points`           | Number of metric points scraped by the Collector.                                         | Counter   |            |
+
+### Events observable with internal logs
+
+The Collector logs the following internal events:
```
tiffany76, line 250: I tested on-the-fly reconfiguration with my own Collector instance, and it did not work, so I removed "The Collector reconfigured" from this list. @open-telemetry/collector-approvers, could you let me know if there are any other events in this list for which logging has not yet been implemented? Thanks!

content/en/docs/collector/internal-telemetry.md
svrnm: I like idea (2), can we try this first before dropping.

You might also consider to introduce line breaks for the long ones (via adding `<br>`), markdownlint will add a warning for that but not a breaking one.

svrnm:
This looks really good, I answered some of your questions. Thanks!

content/en/docs/collector/internal-telemetry.md
```diff
@@ -113,3 +118,148 @@ journalctl | grep otelcol | grep Error
 ```
 
 {{% /tab %}} {{< /tabpane >}}
+
+## Types of internal observability
+
+The OpenTelemetry Collector aims to be an exemplar of observable service by
```
theletterf, line 124:
```suggestion
The OpenTelemetry Collector aims to be a model of observable service by
```

theletterf (approved):
Lots of good stuff in here. Thanks so much, @tiffany76 !

content/en/docs/collector/internal-telemetry.md
tiffany76: Much better word choice! Thanks @theletterf.

content/en/docs/collector/internal-telemetry.md
tiffany76: I've added both the tabbed panes to the earlier section and line breaks to the table. Both data sets are still incomplete until I get confirmation of verbosity levels from the Collector folks, but in the meantime, let me know what you think! :)

svrnm: This looks good, thanks! I wonder if we have something in our pocket for column sorting, cc @chalin

content/en/docs/collector/internal-telemetry.md
```diff
@@ -34,15 +35,56 @@ service:
...
 - `none` indicates that no telemetry data should be collected.
 - `basic` is the recommended value and covers the basics of the service
   telemetry.
 - `normal` adds other indicators on top of basic.
 - `detailed` adds dimensions and views to the previous levels.
 
+By default, the Collector's internal metrics are distributed across verbosity
+levels, as shown here:
+
+{{< tabpane text=true >}} {{% tab "`basic`" %}}
```
tiffany76, line 51: Thinking about this metric breakdown from a Collector user's perspective, I think it would make more sense to include the complete list emitted at each level, rather than the delta. I'd like to use the [docsy shortcode for reusing documentation](https://github.com/google/docsy/blob/712aa05897c1abe239786522131e83d69e5e1b4e/userguide/content/en/docs/adding-content/shortcodes/index.md#reuse-documentation) so we don't have to repeat the hard-coded lists here in each tab. @svrnm or @chalin, would you mind if I tried this out? It would involve creating a few extra files (for the lists) in the docs/collector directory.

content/en/docs/collector/internal-telemetry.md
tiffany76: I spent some time working through this question on my own, and I've found the answer concerning the `processor_batch` metrics and the default level. (I misunderstood how the default level was being applied.) 

However, short of learning Go :smile:, I am really, truly stuck on how to find the level for specific metrics in the code. If @dmitryax or one of the other @open-telemetry/collector-approvers could let me know if any other* metrics are emitted at a level other than `basic`, I would be very grateful. Thank you!

*Other than:
`normal`:
`otelcol_processor_batch_batch_send_size`  
`otelcol_processor_batch_batch_size_trigger_send`  
`otelcol_processor_batch_metadata_cardinality`  
`otelcol_processor_batch_timeout_trigger_send`

`detailed`:
`otelcol_http_client_active_requests`  
`otelcol_http_client_connection_duration`  
`otelcol_http_client_open_connections`  
`otelcol_http_client_request_body_size`  
`otelcol_http_client_request_duration`  
`otelcol_http_client_response_body_size`  
`otelcol_http_server_active_requests`  
`otelcol_http_server_request_body_size`  
`otelcol_http_server_request_duration`  
`otelcol_http_server_response_body_size`  
`otelcol_rpc_client_duration`  
`otelcol_rpc_client_request_size`  
`otelcol_rpc_client_requests_per_rpc`  
`otelcol_rpc_client_response_size`  
`otelcol_rpc_client_responses_per_rpc`  
`otelcol_rpc_server_duration`  
`otelcol_rpc_server_request_size`  
`otelcol_rpc_server_requests_per_rpc`  
`otelcol_rpc_server_response_size`  
`otelcol_rpc_server_responses_per_rpc`

dmitryax: @tiffany76 this is the correct understanding. All the metrics other than those listed above are emitted in the `basic` level. `normal` is the level applied by default. It means that metrics with `basic` (most of the metrics) and `normal` (`otelcol_processor_batch_*` metrics) are emitted by default, while `otelcol_http_*` and `otelcol_rpc_*`are not. BTW the `otelcol_` prefix was recently dropped for them. Now they are emitted as `http_*` and `rpc_*`.

content/en/docs/collector/internal-telemetry.md
svrnm: There are 2 tabs right now and we talk about duplicating the ones in "normal" to "detailed", is this right?

content/en/docs/collector/internal-telemetry.md
chalin: @tiffany76 - will you be adding `basic` to the **Level** column value for the relevant entries in the table? That would be my preference (and I think it's yours too based on an earlier comment).

content/en/docs/collector/internal-telemetry.md
```diff
@@ -113,3 +155,149 @@ journalctl | grep otelcol | grep Error
...
+- Total dropped data due to invalid data received, broken down by data type.
+
+- Total incoming connection count, broken down by receiver.
+
+- Uptime since start.
+
+### List of internal metrics
+
+The following table identifies each internal metric by name and description.
+Each metric is also categorized by instrumentation type and level of verbosity.
```
chalin, line 225: This is a large table, but I'm ok leaving it as is for now -- even if readers need to scroll left-right on narrow displays.

A few comments and questions:

(1) Not all entries have a specified verbosity level / threshold. From a comment thread elsewhere, I think that those entries are `basic`. Let's make that explicit.

(2) Where is this list coming from? How quickly will it age? Will it be a maintenance burden? Can we alleviate any of this issues hinted at by the previous questions.

content/en/docs/collector/internal-telemetry.md
```diff
@@ -113,3 +155,149 @@ journalctl | grep otelcol | grep Error
...
+
+- Data dropping due to throttling stops.
+
+- Data dropping begins due to invalid data. A sample of the invalid data is
+  included.
+
+- Data dropping due to invalid data stops.
+
+- A crash is detected, differentiated from a clean stop. Crash data is included
+  if available.
```
chalin, line 303:
```suggestion
- A Collector instance starts or stops.
- Data dropping begins due to throttling for a specified reason, such as local
  saturation, downstream saturation, downstream unavailable, etc.
- Data dropping due to throttling stops.
- Data dropping begins due to invalid data. A sample of the invalid data is
  included.
- Data dropping due to invalid data stops.
- A crash is detected, differentiated from a clean stop. Crash data is included
  if available.
```

chalin (commented):
Nice work. See inline for questions, remarks, and suggestions. Thanks.

content/en/docs/collector/internal-telemetry.md
tiffany76: Hi @svrnm.  There are three tabs: basic (yet to be filled in; working on that now), normal, and detailed. The basic list is shortest. The normal list includes the basic list, plus a few more. And the detailed list includes the normal list plus several more metrics. If we decide to keep the tabbed panes (see below), I can hard-code those lists, but there will be duplication. Or I could try the docsy shortcode, but I'd need to create separate files for the lists. How about I hard-code the lists for now, and we can decide later to change it if the Markdown is too cluttered?

Thanks, @chalin. The original ask was for a list of metrics by verbosity level (see Issue #4278). I checked with a user (just one, but they had a better viewpoint than me) how they would want to see this information presented, and they responded with two use cases: wanting to see all the metrics emitted at a given level and wanting to know at which level a specific metric is emitted. If we could column-sort the table, I think it could serve for both use cases, but I couldn't find a way to sort without introducing CSS and JavaScript (there's an [open docsy issue](https://github.com/google/docsy/issues/604) for this). In the absence of sorting, I think the tabbed panes are the best way to capture the first use case. But if you think the information is redundant, I'm quite happy to remove the tabs and refer to the table.

content/en/docs/collector/internal-telemetry.md
tiffany76: > (1) Not all entries have a specified verbosity level / threshold. From a comment thread elsewhere, I think that those entries are basic. Let's make that explicit.

I'm working on that now. I was waiting for confirmation from the Collector folks that all the remaining metrics were, in fact, emitted at the `basic` level. 

> (2) Where is this list coming from? How quickly will it age? Will it be a maintenance burden? Can we alleviate any of this issues hinted at by the previous questions.

I used a grep command to find the metrics in the Collector core repository. I have no idea how often the list of metrics changes. The @open-telemetry/collector-approvers would know better.

content/en/docs/collector/internal-telemetry.md
```diff
@@ -34,16 +35,57 @@ service:
...
-- `basic` is the recommended value and covers the basics of the service
-  telemetry.
-- `normal` adds other indicators on top of basic.
+- `basic` covers the basics of the service telemetry.
+- `normal` is the default level and adds other indicators on top of basic.
 - `detailed` adds dimensions and views to the previous levels.
 
-For example:
+The Collector's internal metrics are distributed across verbosity levels, as
+shown here:
```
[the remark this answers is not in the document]
tiffany76: Fixed in 29d94a7.

content/en/docs/collector/internal-telemetry.md
tiffany76: Table is complete now. Resolving this conversation.

content/en/docs/collector/internal-telemetry.md
chalin: Cool. Maybe, at a minimum add a (doc-only) comment to that effect and even consider including the grep command(s) for reference?

content/en/docs/collector/internal-telemetry.md
chalin: Atm, I'd vote to drop these tabs, and direct the reader to the table towards the end of the page. I understand that this isn't ideal, but it's a good incremental step in the right direction to addressing #4278 and improving this page IMHO.

Then, I'd suggest commenting on https://github.com/google/docsy/issues/604 and upvoting the issue so that it is more likely to get worked on.

WDYT @svrnm et all?

If y'all vote to keep the tabs, I'm willing to go with the majority.

tiffany76: Another possibility to consider:
We drop the tabs but re-sort the table first by Level and then alphabetically by metric name.

content/en/docs/collector/internal-telemetry.md
tiffany76: Sure. Added a comment in 12399b4.

content/en/docs/collector/internal-telemetry.md
chalin: > Another possibility to consider: We drop the tabs but re-sort the table first by Level and then alphabetically by metric name.

Yes! That was what I was going to suggest too. Actually, we could even split the big table into three separate tables, one for each level, and then drop the level column (as Severin suggested at some point I believe). WDYT?

content/en/docs/collector/internal-telemetry.md
```diff
@@ -80,8 +223,11 @@ service:
...
 
-Self-monitoring is a risky practice. If an issue arises, the source of the
-problem is unclear and the telemetry is unreliable.
+When self-monitoring, the Collector collects its own telemetry and sends it to
+the desired backend for analysis. This can be a risky practice. If the Collector
+is unhealthy or performing poorly, its self-monitoring capability could be
+impacted. As a result, the self-monitored telemetry might not reach the backend
+in time for critical analysis.
 
 {{% /alert %}}
```
chalin, line 232: Ah, much clearer, thank you @tiffany76 for reworking the Caution prose.

I'd like to question whether we want such a callout or not, because, IMHO:

- Such a statements are true of any component emitting telemetry
- The phrase "if the Collector is unhealthy" unnecessarily sows doubt as to the reliability of the Collector

@svrnm @tiffany76 et all: WDYT?

tiffany76: > I'd like to question whether we want such a callout or not, because, IMHO:
Such a statements are true of any component emitting telemetry

The caution alert was requested by the Collector folks. There was [talk](https://github.com/open-telemetry/opentelemetry.io/pull/4246#discussion_r1555711946) of adding best practices for self-monitoring, but that work was beyond the scope of the first PR and I think it's beyond the scope of this one as well. I think we can keep the alert as is for now (subject to the below). 

> The phrase "if the Collector is unhealthy" unnecessarily sows doubt as to the reliability of the Collector

I agree with this assessment of "unhealthy", and I'll remove the word from the sentence.

mx-psi: > Such a statements are true of any component emitting telemetry

If you pass Collector self-telemetry through the Collector's pipeline, surely you increase the change of something going wrong vs exporting to some external system, right? I think that's what we are getting at with this note.

mhausenblas (approved):
LGTM, thank you!

content/en/docs/collector/internal-telemetry.md
```diff
@@ -34,16 +36,21 @@ service:
...
 - `detailed` adds dimensions and views to the previous levels.
 
-For example:
+Each verbosity level represents a threshold at which certain metrics are
+emitted. See the [lists of internal metrics](#lists-of-internal-metrics) for a
+breakdown of metrics by level.
+
+The default level for metrics output is `normal`. To see more or fewer metrics,
+you can change the default by modifying the level in the config
+`service::telemetry::metrics`. For example:
```
chalin, line 53:
```suggestion
The default level for metrics output is `normal`. To use another level, set `service.telemetry.metrics.level`:
```

Also, is it necessary to also set `address`? If not, remove it from the example, or maybe drop the example entirely -- assuming that the suggested prose given above is considered enough.

content/en/docs/collector/internal-telemetry.md
```diff
@@ -113,3 +123,143 @@ journalctl | grep otelcol | grep Error
...
 {{% /tab %}} {{< /tabpane >}}
+
+## Types of internal observability
+
+The OpenTelemetry Collector aims to be a model of observable service by clearly
+exposing its own operational metrics. Additionally, it collects host resource
+metrics that can help you understand if problems are caused by a different
+process on the same host. Specific components of the Collector can also emit
+their own custom telemetry. In this section, you can explore the different types
+of observability emitted by the Collector itself.
```
chalin, line 134:
```suggestion
their own custom telemetry. In this section, you will learn about the different types
of observability emitted by the Collector itself.
```

content/en/docs/collector/internal-telemetry.md
```diff
@@ -113,3 +123,143 @@ journalctl | grep otelcol | grep Error
...
+
+### Lists of internal metrics
+
+The following tables group each internal metric by level of verbosity: `basic`,
+`normal`, and `detailed`. Each metric is identified by name and description and
+categorized by instrumentation type.
+
+<!---To compile this list, configure a Collector instance to emit its own metrics to the localhost:8888/metrics endpoint. Select a metric and grep for it in the Collector core repository. For example, the `otelcol_process_memory_rss` can be found using:`grep -Hrn "memory_rss" .` Make sure to eliminate from your search string any words that might be prefixes. Look through the results until you find the .go file that contains the list of metrics. In the case of `otelcol_process_memory_rss`, it and other process metrics can be found in https://github.com/open-telemetry/opentelemetry-collector/blob/31528ce81d44e9265e1a3bbbd27dc86d09ba1354/service/internal/proctelemetry/process_telemetry.go#L92. Note that the Collector's internal metrics are defined in several different files in the repository.--->
+
+#### Metrics emitted at the `basic` level
```
chalin, line 182:
```suggestion
#### `basic`-level metrics
```

content/en/docs/collector/internal-telemetry.md
```diff
@@ -113,3 +123,143 @@ journalctl | grep otelcol | grep Error
...
+| `otelcol_receiver_accepted_`<br>`metric_points`        | Number of metric points successfully ingested and pushed into the pipeline.             | Counter   |
+| `otelcol_receiver_accepted_spans`                      | Number of spans successfully ingested and pushed into the pipeline.                     | Counter   |
+| `otelcol_receiver_refused_`<br>`log_records`           | Number of logs that could not be pushed into the pipeline.                              | Counter   |
+| `otelcol_receiver_refused_`<br>`metric_points`         | Number of metric points that could not be pushed into the pipeline.                     | Counter   |
+| `otelcol_receiver_refused_spans`                       | Number of spans that could not be pushed into the pipeline.                             | Counter   |
+| `otelcol_scraper_errored_`<br>`metric_points`          | Number of metric points the Collector failed to scrape.                                 | Counter   |
+| `otelcol_scraper_scraped_`<br>`metric_points`          | Number of metric points scraped by the Collector.                                       | Counter   |
+
+#### Additional metrics emitted at the `normal` level
```
chalin, line 219:
```suggestion
#### Additional `normal`-level metrics
```

content/en/docs/collector/internal-telemetry.md
```diff
@@ -113,3 +123,143 @@ journalctl | grep otelcol | grep Error
...
+
+| Metric name                                             | Description                                                     | Type      |
+| ------------------------------------------------------- | --------------------------------------------------------------- | --------- |
+| `otelcol_processor_batch_batch_`<br>`send_size`         | Number of units in the batch.                                   | Histogram |
+| `otelcol_processor_batch_batch_`<br>`size_trigger_send` | Number of times the batch was sent due to a size trigger.       | Counter   |
+| `otelcol_processor_batch_metadata_`<br>`cardinality`    | Number of distinct metadata value combinations being processed. | Counter   |
+| `otelcol_processor_batch_timeout_`<br>`trigger_send`    | Number of times the batch was sent due to a timeout trigger.    | Counter   |
+
+#### Additional metrics emitted at the `detailed` level
```
chalin, line 228:
```suggestion
#### Additional `detailed`-level metrics
```

chalin (commented):
Looking good! A few more comments, see inline. Thx

content/en/docs/collector/internal-telemetry.md
```diff
@@ -34,23 +36,26 @@ service:
...
+- `normal`: the default level, adds standard indicators on top of basic.
+- `detailed`: the most verbose level, includes dimensions and views.
 
-For example:
+Each verbosity level represents a threshold at which certain metrics are
+emitted. For the complete list of metrics, with a breakdown by level, see
+[Lists of internal metrics](#lists-of-internal-metrics).
+
+The default level for metrics output is `normal`. To use another level, set
+`service::telemetry::metrics::level`:
```
chalin, line 52: Why not use YAML notation given that this is YAML?

```suggestion
`service.telemetry.metrics.level`:
```

tiffany76: Sorry. The notation came from the developers, so I was trusting their judgment. I'll change all of the config paths to match this suggestion.

svrnm:
@open-telemetry/collector-approvers PTAL!

chalin (approved):
LGTM, thank you @tiffany76 ✨

content/en/docs/collector/internal-telemetry.md
```diff
@@ -25,7 +27,7 @@ endpoint to one specific or all network interfaces when needed. For
 containerized environments, you might want to expose this port on a public
 interface.
 
-Set the address in the config `service::telemetry::metrics`:
+Set the address in the config `service.telemetry.metrics`:
```
mx-psi, line 30: Why this change? `service::telemetry::metrics` actually works, while `service.telemetry.metrics` does not, that is to say `key::subkey: value` is equivalent to

```yaml
key:
  subkey: value
```

I think we should prefer the `::` variant because of that

```suggestion
Set the address in the config `service::telemetry::metrics`:
```

content/en/docs/collector/internal-telemetry.md
```diff
@@ -34,23 +36,26 @@ service:
...
+- `normal`: the default level, adds standard indicators on top of basic.
+- `detailed`: the most verbose level, includes dimensions and views.
 
-For example:
+Each verbosity level represents a threshold at which certain metrics are
+emitted. For the complete list of metrics, with a breakdown by level, see
+[Lists of internal metrics](#lists-of-internal-metrics).
+
+The default level for metrics output is `normal`. To use another level, set
+`service.telemetry.metrics.level`:
```
mx-psi, line 52:
```suggestion
`service::telemetry::metrics::level`:
```

content/en/docs/collector/internal-telemetry.md
```diff
@@ -80,15 +85,18 @@ service:
...
+result, the self-monitored telemetry might not reach the backend in time for
+critical analysis.
 
 {{% /alert %}}
 
 ### Configure internal logs
 
 You can find log output in `stderr`. The verbosity level for logs defaults to
-`INFO`, but you can adjust it in the config `service::telemetry::logs`:
+`INFO`, but you can adjust it in the config `service.telemetry.logs`:
```
mx-psi, line 99:
```suggestion
`INFO`, but you can adjust it in the config `service::telemetry::logs`:
```

content/en/docs/collector/internal-telemetry.md
chalin: > Why this change?

Because `service::telemetry::metrics` isn't valid formal YAML syntax or even a recognized informal YAML syntax.

> `service::telemetry::metrics` actually works, 

What do you mean "it works"? Where, in which context or language?

> ... that is to say `key::subkey: value` is equivalent to

Are you claiming that that is official YAML syntax? If so, I'm not seeing it in the [YAML spec](https://yaml.org/spec/1.2.2/).

chalin: Ooooh, I see that the double-colon notation is accepted by `otelcol --set`. Hmm, given that the prose is referring to YAML config, I'd suggest using the dot notation in the prose. WDYT?

content/en/docs/collector/internal-telemetry.md
```diff
@@ -34,23 +36,26 @@ service:
       address: '0.0.0.0:8888'
```
chalin, line 36:
```suggestion
      address: 0.0.0.0:8888
```

content/en/docs/collector/internal-telemetry.md
chalin, line 99: Should info be in lowercase?

```suggestion
`info`, but you can adjust it in the config `service.telemetry.logs`:
```

content/en/docs/collector/internal-telemetry.md
chalin: If you feel strongly about showing the double-colon notation, we should make it clear that it's the `otelcol --set` syntax.

But given that the set command has limitations (from the config page "Note tha the `--set` option doesn't support setting a key that contains a dot or an equal sign"), it might be better to stick to illustrating the YAML config syntax only?

content/en/docs/collector/internal-telemetry.md
```diff
@@ -34,23 +36,26 @@ service:
...
 
-- `none` indicates that no telemetry data should be collected.
-- `basic` is the recommended value and covers the basics of the service
-  telemetry.
-- `normal` adds other indicators on top of basic.
-- `detailed` adds dimensions and views to the previous levels.
+- `none`: no telemetry is collected.
+- `basic`: essential service telemetry.
+- `normal`: the default level, adds standard indicators on top of basic.
+- `detailed`: the most verbose level, includes dimensions and views.
```
chalin, line 45: Hmm, I just noticed that the `level` field is also describe, in the same detail in http://localhost:1313/docs/collector/configuration/#telemetry:

> - `level`: the level of telemetry metrics, default `basic`. The possible values
>   are:
>   - "none" indicates that no telemetry data should be collected.
>   - "basic" is the recommended and covers the basics of the service telemetry.
>   - "normal" adds some other indicators on top of basic.
>   - "detailed" adds dimensions and views to the previous levels.

We should avoid the duplication of the full details.

content/en/docs/collector/internal-telemetry.md
mx-psi: > Ooooh, I see that the double-colon notation is accepted by otelcol --set. Hmm, given that the prose is referring to YAML config, I'd suggest using the dot notation in the prose. WDYT?

@chalin Not only `--set`, this also works in the Collector configuration files. The following is a valid snippet of Collector configuration:

```yaml
service:
   telemetry::metrics::level: detailed
```

which is equivalent to:

```yaml
service:
  telemetry:
    metrics:
      level: detailed
```

content/en/docs/collector/internal-telemetry.md
tiffany76: I will fix this in a follow-up PR. There is more on that page to incorporate here than just the `level` definitions. Enough to warrant a separate PR. Thanks!

tiffany76: Put up an issue: #4476.

content/en/docs/collector/internal-telemetry.md
chalin: @mx-psi - ok, from [Collector Configuration], I see that `--set` and `--config="yaml:..."` accept the double-colon syntax.

As for the use of keys containing double colons (like the example you gave above) that are reinterpreted as nested structures, that is news to me and it isn't documented in [Collector Configuration] page, nor have I seen any example in the docs or elsewhere. But I'm willing to believe you.

Can you show me config files that use this syntax? 

The [Collector Configuration] page needs to be updated to reflect this non-standard interpretation of YAML keys then.

[Collector Configuration]: https://opentelemetry.io/docs/collector/configuration/

tiffany76: I read through the [Collector Configuration](https://opentelemetry.io/docs/collector/configuration/) page, and it does demonstrate the use of double colons when passing config settings at the command line:

> You can also provide configurations using environment variables, YAML paths, or HTTP URIs. For example:
> 
> `otelcol --config=env:MY_CONFIG_IN_AN_ENVVAR --config=https://server/config.yaml`
> `otelcol --config="yaml:exporters::debug::verbosity: normal"`

I propose to revert the paths in this PR to the double colon so we can move the PR forward. I will also create a new issue for adding more details about this syntax to the Configuration page.

WDYT?

chalin: Sounds good to me. Pls cross-link the new issue here. Thanks!

tiffany76: See issue #4484.

mx-psi (approved):

svrnm:
Yes!

chalin:
Great work @tiffany76 et all! 🙌🏻 🎉

PR #4345, opened 2024-04-23
Make use of PAT in pr-actions.yml

### LINKED ISSUE #4339: [CI] Running `/fix:cmd` over a PR rebuilds but skips GH action checks
For context, see:

- #4335

Unless we get a separate bot to handle `/fix:*` commands, we're stuck with this behavior: GH doesn't rerun actions if a workflow over the same repo committed a change to a PR. I usually work around this by forcing a change (via rebase, or by making a trivial edit) over the PR, but that isn't always practical.

Other possible solutions:
 
- Create a separate bot to handle `/fix:*` commands so that GH runs the full action checks. (This would be a lot of work but might be worth it in the long run.)
- Run a cron job that rebuilds the site after changes are made to the `main` branch. (This is easier to setup, and will at least notify us if build checks are broken).

Thoughts @open-telemetry/docs-maintainers?

svrnm:
Might fix #4339

### DIFF
--- a/.github/workflows/pr-actions.yml
+++ b/.github/workflows/pr-actions.yml
@@ -50,11 +50,11 @@ jobs:
         run: |
           gh pr comment $PR_NUM -b "You triggered fix:${PR_ACTION} action run at $GITHUB_SERVER_URL/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID"
         env:
-          GH_TOKEN: ${{ github.token }}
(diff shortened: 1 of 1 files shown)


chalin:
You meant "PAT" in the title, right? :)

chalin (approved):
Worth a try. Thanks for this @svrnm!

chalin:
Do you think that if we run a fix cmd here, it will use the PR's workflow file?

chalin:
That fix:all still runs. We'll need to test it on another PR to see if it actually fixes #4339.

PR #4351, opened 2024-04-24
Update CODEOWNERS with new SIG End User group

svrnm:
follow up to https://github.com/open-telemetry/community/issues/2004

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -22,7 +22,7 @@ content-modules/opentelemetry-proto      @open-telemetry/docs-maintainers @open-
 content-modules/opentelemetry-specification  @open-telemetry/docs-maintainers @open-telemetry/specs-approvers
 content-modules/semantic-conventions     @open-telemetry/docs-maintainers @open-telemetry/specs-semconv-approvers
 content/en/blog/                         @open-telemetry/docs-maintainers
-content/en/community/end-user/           @open-telemetry/docs-approvers @open-telemetry/end-user-wg
+content/en/community/end-user/           @open-telemetry/docs-approvers @open-telemetry/sig-end-user-approvers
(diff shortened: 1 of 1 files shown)


theletterf (approved):

.github/CODEOWNERS
```diff
@@ -22,7 +22,7 @@ content-modules/opentelemetry-proto      @open-telemetry/docs-maintainers @open-
 content-modules/opentelemetry-specification  @open-telemetry/docs-maintainers @open-telemetry/specs-approvers
 content-modules/semantic-conventions     @open-telemetry/docs-maintainers @open-telemetry/specs-semconv-approvers
 content/en/blog/                         @open-telemetry/docs-maintainers
-content/en/community/end-user/           @open-telemetry/docs-approvers @open-telemetry/end-user-wg
+content/en/community/end-user/           @open-telemetry/docs-approvers @open-telemetry/sig-end-user-approvers
```
danielgblanco, line 25: Should this be `sig-end-user-maintainers`? The end-user SIG only has maintainers at the moment.

svrnm: As far as I understand it maintainers are a subteam of the approvers, so if the approvers are tagged the maintainers will be tagged as well. We do this for consistency, but we can have the maintainers here if that's the preferred group

danielgblanco: Good point. I keep forgetting those are sub-teams!

danielgblanco (approved):
LGTM

.github/CODEOWNERS
cartermp: GitHub doesn't seen to think that this team exists.

svrnm: ah, we need to fix repo permissions as well

PR #4415, opened 2024-05-03
Fix vendors table (remove distros from table heading)

svrnm:
We removed the distribution column, but we didn't remove it in the heading which breaks the table, this PR fixes that:

https://opentelemetry.io/ecosystem/vendors/


chalin (approved):

cijothomas:
Thanks!

PR #4421, opened 2024-05-03
Fixes #4368

### LINKED ISSUE #4368: [docs/collector/scaling, docs/collector/deployment] Describe how to deal with single-writer principle
**What needs to be changed?** 

The OpenTelemetry metrics data model [requires each metric data stream to have a single writer](https://opentelemetry.io/docs/specs/otel/metrics/data-model/#single-writer). This spec section is exhaustive but it is written in spec-speak and it can be a bit hard to understand (and, possibly, to find).

Specifically in the Collector, it can be hard to understand that this is a requirement to avoid overwriting the same time series. Since we already have a couple pages dedicated to how to deploy the Collector, we can explicitly mention taking care of this in the documentation.

**What is the name + path of the page that needs changed?**  Either of `collector/scaling` or `collector/deployment` or one of its subpages would be good for this.

**Additional context**: Example related issue open-telemetry/opentelemetry-collector-contrib/issues/32043

michael2893:
## Summary

This change addresses the request for documentation on the Single-Writer principle. https://github.com/open-telemetry/opentelemetry.io/issues/4368

## Description

- add section on multiple collector deployments in `deployment/gateway`
- define single writer principle
- provide examples and context

## Open questions
- Can I provide examples from open issues to help better capture this problem?


content/en/docs/collector/deployment/single-writer-principle.md
```diff
@@ -0,0 +1,57 @@
+---
+title: The Single Writer Principle
+description: How to handle the single writer principle in Collector deployments
+weight: 1
```
theletterf, line 4: Alter the weight of the page so that it shows below "Gateway".

```suggestion
weight: 4
```

content/en/docs/collector/deployment/single-writer-principle.md
```diff
@@ -0,0 +1,57 @@
+---
+title: The Single Writer Principle
```
theletterf, line 2:
```suggestion
title: The Single Writer Principle
```

How can we better contextualize what this doc is about when it comes to Collector deployments? Think of a nav title that would fit the progression in the current section. Why should someone know about this principle?

content/en/docs/collector/deployment/single-writer-principle.md
```diff
@@ -0,0 +1,57 @@
...
+description: How to handle the single writer principle in Collector deployments
+weight: 1
+---
+
```
theletterf, line 6: As the doc is likely not self-contained, consider adding links from other topics.

content/en/docs/collector/deployment/single-writer-principle.md
theletterf: Also, could this be a section in an existing doc rather than its own doc?

theletterf (changes requested):
Thanks! Please have a look at the questions. Also, you'll need to sign the CLA.

@open-telemetry/collector-maintainers Please have a look, thanks!

content/en/docs/collector/deployment/single-writer-principle.md
michael2893: Ah, right. It might make sense in `collectors/scaling` as a subsection. What do you think?

PR #4433, opened 2024-05-07
Collector docs on single-writer principle

michael2893:
## Summary

This change addresses the request for documentation on the Single-Writer principle. https://github.com/open-telemetry/opentelemetry.io/issues/4368

## Description

- add section on multiple collector deployments in `deployment/gateway`
- define single writer principle
- provide examples and context

## Open questions
- Can I provide examples from open issues to help better capture this problem?


svrnm:
@open-telemetry/collector-approvers ptal

content/en/docs/collector/deployment/gateway.md
```diff
@@ -212,6 +212,47 @@ Cons:
...
+which is a violation of the Single-Writer principle. A potential consequence of this may be that
+these data streams overwrite each other, leading to inconsistent time series data.
+
+#### Example Scenario
+
+Consider the following configuration:
+
+There is a gateway deployment configured to handle all traffic for three other collectors in the same system.
+If the collectors are not uniquely identified and the SDK fails to distinguish between them, they may
+send identical data to the gateway collector from different points in time. In this scenario,
```
jpkrohling, line 238: Can you give a more concrete example here? Having multiple instances of a collector behind a load-balancer is certainly common practice, and there's no inherent problem in having a SDK sending data via this load-balancer, causing different data points for the same workload to land at different collector instances. 

There are a few situations that need to be accounted for when scaling, like using target-allocator for pull-based scraping (nothing to do with OTLP though), or tail-sampling (due to the statefulness characteristic of this component).

content/en/docs/collector/deployment/gateway.md
```diff
@@ -212,6 +212,47 @@ Cons:
...
+Concurrent access from multiple applications that modify or report on the same data can lead to data
+loss or, at least, degraded data quality. In gateway collector deployments, applying this principle guards against sending
+inconsistent data to the backend.
+
+#### Context in OTLP
+
+All metric data streams within OTLP must have a [single writer](https://opentelemetry.io/docs/specs/otel/metrics/data-model/#single-writer)
+This is because gateway collector deployments can involve multiple collectors in the same system.
+It is possible in this case for instances to receive and process data from the same resources,
+which is a violation of the Single-Writer principle. A potential consequence of this may be that
```
jpkrohling, line 229: I'm not sure I agree with this. Here's what the link above says:

> All metric data streams within OTLP MUST have one logical writer. This means, conceptually, that any Timeseries created from the Protocol MUST have one originating source of truth. In practical terms, this implies the following:
>
> - All metric data streams produced by OTel SDKs SHOULD have globally unique identity at any given point in time. [Metric identity is defined above.](https://opentelemetry.io/docs/specs/otel/metrics/data-model/#opentelemetry-protocol-data-model-producer-recommendations)
> - Aggregations of metric streams MUST only be written from a single logical source at any given point time. Note: This implies aggregated metric streams must reach one destination.

Multiple instances of the collector can certainly receive and process different OTLP requests for the same resource without problems. 

In the context of the collector, the single-writer principle is relevant for receivers that create metrics on their own, such as scraping receivers or things like the host metrics receiver. On that case, there should really be only one "host metrics receiver" instance per host.

content/en/docs/collector/deployment/gateway.md
```diff
@@ -212,6 +212,47 @@ Cons:
...
+send identical data to the gateway collector from different points in time. In this scenario,
+it would be possible to observe inconsistent values for the same series.
+
+
+#### Detection
+
+There are patterns in the data that may provide some insight into whether this is happening or not.
+For example, upon visual inspection, a series with unexplained gaps or jumps in the same series may be a clue that
+multiple collectors are sending the same samples. Unexplained behavior in a time series could potentially
+point to the backend scraping data from multiple sources.
```
jpkrohling, line 247: Another common way to find this out is when the backend complains about "out of order samples" -- if a data point for the state of a counter at T2 was received, and later a data point for the state of the same counter at T1 was received, a backend might say that the late data point is discarded.

content/en/docs/collector/deployment/gateway.md
michael2893: Yeah - what I have here isn't really specific enough. I can certainly provide an example here

content/en/docs/collector/deployment/gateway.md
michael2893: Ok, makes sense. This is a much more particular problem than my framing would suggest. 

In hindsight, 

> It is possible in this case for instances to receive and process data from the same resources,
which is a violation of the Single-Writer principle. A potential consequence of this may be that

Is not really correct. That's far too general, especially considering the case you've highlighted.

Thanks for the feedback, I will get to revising this.

content/en/docs/collector/deployment/multiple-collectors.md
```diff
@@ -0,0 +1,87 @@
...
+#### Host Metrics Receiver
+
+When creating metrics related to the host system via the [host metrics receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/hostmetricsreceiver),
+it is important to have only one host metrics receiver instance per host.
+A violation of the single-writer principle in this scenario would mean deploying more than one host metrics receiver
+on the same host. If both try to collect system data at the same time, this may result in
+inconsistent data or data loss. Collisions resulting from inconsistent timestamps may lead to an unstable or inconsistent
+representation of metrics, such as CPU usage.
+
+### Scaling considerations
```
michael2893, line 39: This selection doesn't seem like it would belong in `deployment` I don't think, actually.

I'm wondering if it's even necessary to mention. 

This is information basically already covered in `Scaling the Collector` and I'm wondering if it's redundant, or there should be some additional mention in that document about the single-writer principle there.

content/en/docs/collector/deployment/multiple-collectors.md
```diff
@@ -0,0 +1,85 @@
...
+loss or, at least, degraded data quality. In gateway collector deployments, applying this principle guards
+against sending inconsistent data to the backend.
+
+### Further Context
+
+All metric data streams within OTLP must have a [single writer](https://opentelemetry.io/docs/specs/otel/metrics/data-model/#single-writer).
+In a system with multiple collectors, the single-writer principle is most relevant for receivers that generate their 
+own metrics, such as scraping receivers or the host metrics receiver. 
+This is most important for receivers that create their own metrics, such as pull based scraping 
+that targets a specific metric source.
```
jpkrohling, line 26: I believe the two last statements are saying the same things. Perhaps complement the first with extra info from the second?

```suggestion
In a system with multiple collectors, the single-writer principle is most relevant for receivers that generate their
own metrics or target a specific metric source, such as scraping receivers, host metrics receiver, kubelet stats receiver, and similar.
```

content/en/docs/collector/deployment/multiple-collectors.md
jpkrohling: Scaling collectors is another way of saying that we need multiple collectors, so, this page is bound to repeat some/much of what's in the scaling page. How about we just add a "single-writer principle" to that one instead?

content/en/docs/collector/scaling.md
```diff
@@ -382,3 +382,99 @@ service:
...
+
+The Single-Writer Principle refers to employing a single logical writer for a
+particular resource. When scaling collectors horizontally in a system it's
+important to properly distinguish between targets using unique identities.
+
+##### Tail-based Sampling
+
+Partial or incomplete traces may be consequential for an implementation of
+tail-sampling, as the goal is to capture all or most of the spans within the
+trace in order to inform sampling decisions. When using the target allocator to
```
jpkrohling, line 396: the target allocator is not related to the tail-based sampling, this seems out of place here

content/en/docs/collector/scaling.md
```diff
@@ -382,3 +382,99 @@ service:
...
+
+```yaml
+- job_name: 'otel-collector'
+  static_configs:
+    - targets: ['0.0.0.0:8888']
+      labels:
+        service: 'my-service'
+```
+
+You can also use
```
jpkrohling, line 407: this is also out of place, it doesn't have anything to do with the tail-based sampling

content/en/docs/collector/scaling.md
```diff
@@ -382,3 +382,99 @@ service:
...
+
+In pull-based metric reporting it is important to maintain the concept of unique
+metric identities. When scaling scrapers, it's important to ensure that the
+targets have globally unique references. If the complexity in the environment is
+lower, you may be able to rely on sharding based on namespace or workload alone.
+As the system increases in complexity, consider adding custom labels related to
+the application or service to better delineate the targets.
+
+```yaml
+scrape_configs:
```
jpkrohling, line 433: This is a Prometheus scrape configuration, not an OTel Collector. Either provide the full configuration for the collector, possibly embedding this config you have here.

content/en/docs/collector/scaling.md
michael2893: Does it make sense to mention it at all here? It's relevant but I am a bit unsure if it actually needs mentioned in a document on collector scaling

svrnm:
@michael2893 do not worry too much about the markdown, link issues, etc. lets make sure that the content is right and then we can fix the PR accordingly.

Have you addressed the feedback by @jpkrohling, if so we need another round of reviews :-)

michael2893:
@svrnm Hi - yes I did address the content issues from the comments above

jpkrohling (commented):
I would like to get the opinion of another Collector approver, as well as an Operator approver. The reason is that I'm not seeing the benefit for most of this change, especially on the scaling.md: the point about scaling pull-based scrapers is already being covered in "Scaling the Scrapers". In general, I feel like the single-writer principle can be smaller mentions in existing docs, instead of new docs or new sections.

@mx-psi , is this PR following what you had in mind for #4368?

michael2893:
Ah, ok, right. I think with this change, I could enhance context in the existing docs with additional notes without restating as much as I have, depending on what the desired outcome for 4368 is!

mx-psi:
I have not had time to look into this PR yet, I will try to do it next week. What I had in mind in general terms is a page I can refer users to when they face trouble because they have an incorrect setup where the same metric stream is being produced by multiple writers. For an example, see https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/32043#issuecomment-2062329429. I'll try to give a more thorough review soon

svrnm:
any updates?

mx-psi:
Sorry, this slipped through the cracks, let me add it to my queue :+1:

content/en/docs/collector/deployment/multiple-collectors.md
```diff
@@ -0,0 +1,71 @@
...
+With a Prometheus backend, an example error is:
+`Error on ingesting out-of-order samples`.
+
+This could indicate that identical targets exist in two jobs, and the order of
+the timestamps is incorrect.
+
+Ex:
+
+- Metric T2 received at time 13:56:04
+- Metric T1 received at time 13:56:07 for the same state as T2
```
mx-psi, line 60: I am unsure if this is a good example. I am not very familiar with Prometheus, but this seems like a different issue: you can have out of order points (I believe the OTLP data model does not explicitly forbid this) while still not running into single-writer issues.

svrnm:
@michael2893 do you know what to do next?

michael2893:
@svrnm yep! working on condensing this and moving it into 'deployment'

content/en/docs/collector/deployment/multiple-collectors.md
```diff
@@ -0,0 +1,71 @@
...
+
```
iblancasa, line 61: I think a better example would be:

```suggestion
- Metric `M1` received at time 13:56:04 with value `100`
- Metric `M1` received at time 13:56:24 with value `120`
- Metric `M1` received at time 13:56:04 with value `110`
```

michael2893:
@mx-psi - just to clarify, can this can go under `gateway` in collector/deployment? I thought that could be an appropriate place for the note, but just wanted to check before pushing anymore changes

mx-psi:
> @mx-psi - just to clarify, can this can go under `gateway` in collector/deployment? I thought that could be an appropriate place for the note, but just wanted to check before pushing anymore changes

Yeah, I think it can fit there :)

cartermp:
@michael2893 unfortunately one of your commits came from a different email not covered by the CLA. I'd recommend squashing that commit.

michael2893:
there was an issue with squashing the commits from the invalid email here, so I just moved the change to here:
https://github.com/open-telemetry/opentelemetry.io/pull/5166

PR #4454, opened 2024-05-09
Update K8s operator _index.md

gsmini:
In my actual operation, I found that the namespace attribute needs to be declared, so I modified the document. Please review the modifications.


swiatekm-sumo:
Didn't it create the resource in the namespace from your active context? What was the error message?

gsmini:
> Didn't it create the resource in the namespace from your active context? What was the error message?

Sorry, I read my operation log carefully. I was trying to solve another problem at the time. I thought it was caused by this namespace. Now I have tried again and it is normal that there is no namesapce attribute. Moreover, this resource object namesapce only restricts tenants. I made a mistake, sorry.

PR #4481, opened 2024-05-14
[zh] Add new pages: what-is-opentelemetry and its _index.md

### LINKED ISSUE #4478: [zh] Add a first docs page
@windsonsea - please submit a PR with the translated docs page, which was originally a part of #4341 but that I (@chalin) asked you to remove so that I can setup the `en` page fallback.

windsonsea:
- Fix #4478 
- Adds a new zh page after the zh homepage.

**Preview** https://deploy-preview-4481--opentelemetry.netlify.app/zh/docs/


content/zh/_index.md
```diff
@@ -22,7 +22,7 @@ developer_note:
 
 <div class="l-primary-buttons mt-5">
 
-- [了解更多](/docs/what-is-opentelemetry/)
+- [了解更多](/zh/docs/what-is-opentelemetry/)
```
svrnm, line 25: @chalin is there a way to express "go to the language specific docs and if you do not find them go to the english ones"? We will need this more often in the future and eventually it will be hard to track which links are pointing where

svrnm (approved):

content/zh/what-is-opentelemetry.md
svrnm, line 1: I think this file needs to be named `content/zh/docs/what-is-opentelemetry.md`, for that you might also need a stub `content/zh/docs/_index.md`

svrnm (changes requested):

content/zh/_index.md
windsonsea: Yes, updated.

chalin: Thanks for the first few doc pages @windsonsea.

@svrnm - I still need to experiment with Hugo so see what the best way forward is. We might have a few alternatives with tradeoffs to consider. I'll report back here as soon as I can.

content/zh/docs/_index.md
```diff
@@ -0,0 +1,15 @@
+---
+title: 文档
+linkTitle: Docs
+menu: { main: { weight: 10 } }
+aliases: [/zh/docs/workshop/*]
```
chalin, line 5: These aliases were for historical reasons, but those pages never existed in `zh` forms, so we don't need aliases here:

```suggestion
```

content/zh/docs/what-is-opentelemetry.md
```diff
@@ -0,0 +1,92 @@
+---
+title: 什么是 OpenTelemetry？
+description: 简短说明 OpenTelemetry 是什么，不是什么。
+aliases: [/zh/about, /zh/docs/concepts/what-is-opentelemetry, /zh/otel]
```
chalin, line 4: For the same reasons given earlier, I'd drop all of these aliases, though I wonder if there'd be value in keeping `/zh/about`. WDYT @svrnm?

```suggestion
```

chalin:
All, this is what the Docs section will look like with an English-page fallback:

> image
> ...

If you like this, then I propose that we move forward with this approach. I still need to fix the broken links. I'll do that soon, but wanted to get the previous out to you first.

I also need to fix canonical link URLs, but I don't want to hold this PR back because of it.

content/zh/_index.md
chalin: @svrnm - with the config I've setup, the proper thing to do in the `zh` pages is to lead all URLs with `/zh/` _unless_ there is a wish to intentionally target an `en` page. Does that make sense to you all?

The fallback is implemented via Hugo mounts.

content/zh/docs/_index.md
```diff
@@ -0,0 +1,15 @@
+---
+title: 文档
+linkTitle: Docs
```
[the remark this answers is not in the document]
chalin: I've dropped `linkTitle`.

content/zh/docs/_index.md
chalin: FYI, I've gone ahead and committed this suggested change.

chalin:
All checks are passing, and I've implemented all changes I wanted to put in place. This is good to go, and ready for a final approval @svrnm et al.

svrnm (approved):

PR #4511, opened 2024-05-20
Updated receiver tutorial

zimny:
Updated tutorial for building receiver with:
- [x] Safe check for a nil value in case the port is taken, to avoid unexpected panic. Otherwise, the start method will not be called, thus nil will be referenced.
- [x] Updated instruction to reflect the correct function name


svrnm (approved):
lgtm, @open-telemetry/collector-approvers ptal

mx-psi (approved):

PR #4518, opened 2024-05-21
Update getting-started.md with Bandit

pdgonzalez872:


content/en/docs/languages/erlang/getting-started.md
```diff
@@ -59,7 +65,10 @@ def deps do
     {:opentelemetry_api, "~> {{% param versions.otelApi %}}"},
     {:opentelemetry_exporter, "~> {{% param versions.otelExporter %}}"},
     {:opentelemetry_phoenix, "~> {{% param versions.otelPhoenix %}}"},
-    {:opentelemetry_cowboy, "~> {{% param versions.otelCowboy %}}"},
+    # for Cowboy
+    {:opentelemetry_cowboy, "~> {{% param versions.otelCowboy %}}"}
+    # for Bandit
+    {:opentelemetry_bandit, "~> 0.1.4"},
```
pdgonzalez872, line 71: hmm, this doesn't seem correct. Will wait for a review here.

svrnm:
@pdgonzalez872 thanks for raising this PR! Can you provide some details on the background for this PR? From my point of view the getting started should only make use of one webserver and then exercise the whole example with it. In other languages we have a section at the top that calls out that instrumentation is available for other languages, we can add this for sure.

tsloughter:
@svrnm The framework Phoenix only in the last release changed its default to the web server Bandit. So I think this was meant to cover the case for users on the latest Phoenix as well as still on previous versions who are looking to add OpenTelemetry.

Maybe just leaning in on the latest makes sense though.

svrnm:
Thanks @tsloughter that helps! Then this change makes sense.

pdgonzalez872:
@tsloughter is spot on.

I'd still try to have both the old and the new web servers as options if possible. Even though it's the new default, it is still somewhat rare. The docs, especially, the getting started, should help folks hit the ground running, that's why I went the old/new approach.

WDYT?

svrnm (approved):

tsloughter (approved):

pdgonzalez872:
hey, that's a cool check! 
image


fixed

bryannaegele:
@pdgonzalez872 can we please get this PR reverted? These updates were premature and should not have been approved. Apologies for the late request.

pdgonzalez872:
@bryannaegele absolutely. I tried creating a revert automatically, but it didn't like it:
image

Has this been done already?

PR #4522, opened 2024-05-22
Add docs-zh-approvers to CODEOWNERS

svrnm:
Adds @open-telemetry/docs-zh-approvers as codeowners for docs/zh, since @open-telemetry/docs-zh-maintainers is a subteam they inherit that ownership and do not need to be mentioned


.github/CODEOWNERS
```diff
@@ -42,3 +42,4 @@ content/en/docs/specs/                   @open-telemetry/docs-approvers @open-te
 content/en/docs/security/                @open-telemetry/docs-approvers @open-telemetry/sig-security-maintainers
 content/en/ecosystem/demo/               @open-telemetry/demo-approvers @open-telemetry/demo-approvers
 content/en/docs/contributing/       @open-telemetry/docs-approvers @open-telemetry/docs-maintainers
+content/zh/                         @open-telemetry/docs-zh-approvers
```
chalin, line 45: I'd like to suggest that `content/zh/` also have @open-telemetry/docs-approvers, if only initially, as we iron out the wrinkles in the process. @open-telemetry/docs-approvers can always RSLGTM for PRs that are pure content, but as we ramp up there are likely to be "structural" / Hugo / Docsy / tooling related issues.

PR #4554, opened 2024-05-28
Reorganize Spring Boot content

jeanbisutti:
This PR reorganizes the Spring Boot content following the discussions in #4324.

_Further improvements could certainly be made to the Spring Boot content. This PR focuses on the reorganization discussed in #4324._

Preview: https://deploy-preview-4554--opentelemetry.netlify.app/docs/zero-code/java/spring-boot-starter/

Fix #4324

PR #4562, opened 2024-05-30
fix java redirects

zeitlinger:


content/en/docs/zero-code/java/agent/_index.md
```diff
@@ -1,10 +1,11 @@
 ---
 title: Java Agent
 linkTitle: Agent
-aliases:
```
zeitlinger, line 4: the same alias exists one level up where it makes more sense

content/en/docs/zero-code/java/agent/_index.md
```diff
@@ -1,10 +1,11 @@
 ---
 title: Java Agent
 linkTitle: Agent
```
breedx-splk, line 3: Why remove the aliases as part of this change?

```suggestion
linkTitle: Agent
aliases:
  - /docs/java/automatic_instrumentation
  - /docs/languages/java/automatic_instrumentation
```

breedx-splk (approved):

content/en/docs/zero-code/java/agent/_index.md
zeitlinger: the same alias exists one level up where it makes more sense

content/en/docs/zero-code/java/spring-boot-starter/out-of-the-box-instrumentation.md
```diff
@@ -1,7 +1,7 @@
 ---
 title: Out of the box instrumentation
 description: Out of the box instrumentation for the starter
-aliases: [../annotations]
+aliases: [../instrumentation]
```
trask, line 4: do we want the alias to stay under `/spring-boot-starter/*`, e.g.
```suggestion
aliases: [instrumentation]
```

trask (approved):

cartermp (approved):
LGTM but I will defer to @chalin

chalin:
@open-telemetry/docs-approvers et al: I think that we need to fine tune our PR review and merging process. I wasn't done reviewing #4554, and that PR wasn't ready to be merged IMHO. In particular (as some of you have noted) because more aliases were needed. I didn't comment on the need for aliases on my first review pass because in my first review pass I suggested some _structural changes_ -- so it didn't make sense to suggest specific aliases until the section structure stabilized.

Any suggestions on how to improve our process? Maybe if on @open-telemetry/docs-approvers makes some suggested changes, wait for them (a reasonable amount of time) to give a final approval?

No big deal, and I'm really just hoping to make our contribution approval process smoother. Lemme know what y'all think.

cartermp:
I would propose some kind of disclaimer if reviews are being done in stages.

FWIW my general philosophy for this kind of stuff is to bias towards merging when content looks good and keep the flow of PRs going, but I missed that this re-org of pages would need redirects put into place as well.

chalin (changes requested):
Thanks for this PR @zeitlinger. I've submitted an alternative (simpler) fix, see #4563 -- by adding a single extra alias we can address the broken-link issue you faced, without having to introduce a wildcard redirect rule. Also #4563 drops some unnecessary aliases and includes some preliminary copyedits.

All: PTAL at #4563.

content/en/docs/zero-code/java/spring-boot-starter/out-of-the-box-instrumentation.md
chalin: That alias shouldn't be there in the first place. It's been dropped in #4563.

zeitlinger:
> Thanks for this PR @zeitlinger. I've submitted an alternative (simpler) fix, see #4563 -- by adding a single extra alias we can address the broken-link issue you faced, without having to introduce a wildcard redirect rule. Also #4563 drops some unnecessary aliases and includes some preliminary copyedits.
> 
> All: PTAL at #4563.

thanks @chalin - much simpler than my fix!

Issue #4560, opened 2024-05-30
MassTransit Messaging Library


KaydenMiller:
The MassTransit library supports a large portion of the Open Telemetry tooling in .NET https://masstransit.io/documentation/configuration/observability#open-telemetry

Noticed on the https://opentelemetry.io/docs/languages/net/libraries/ page that you were looking for libraries that support OTEL.

svrnm:
Thanks @KaydenMiller. [This PR](https://github.com/open-telemetry/opentelemetry.io/pull/4337) by @mgurrie should have added it to the list but I think there is a minor change missing to make this happen, let me check

PR #4586, opened 2024-06-03
fix missing native key for MassTransit integrations #4560

KaydenMiller:
Fixing the missing `native` key to the `integrations.yaml` file to mark native support for OTEL in MassTransit. Should solve #4560


svrnm:
Awesome, thanks!

PR #4595, opened 2024-06-04
[CI] Add script to update NPM pkgs to latest, and update them

chalin:
- Adds NPM script `update:pkgs` that can be used to update all NPM packages to `@latest`
- In a separate commit, this PR updates (almost) all packages
  - Gulp is excluded because @chalin hasn't fully debugged the problem with v5
- **Note the `@opentelemetry/*` version updates** <-- @austinlparker @cartermp is this ok?
- There are no changes to the generated site files (other than time stamps the Hugo version)


cartermp (approved):
The OTel dep updates are fine, we will probably want to always be on latest packages.

PR #4648, opened 2024-06-10
[CI & i18n] i18n script and doc improvements

### LINKED ISSUE #4641: [CI] i18n GH action check doesn't always work
For example, see: https://github.com/open-telemetry/opentelemetry.io/actions/runs/9448595703/job/26022945739?pr=4616

I have an improved script that is almost ready that I'll submit soon.

chalin:
- Fixes #4641 
- Improves i18n script
- Adds NPM scripts for `check:i18n` and `fix:i18n*`
- Renames scripts to `check-i18n*` rather than `i18n-check*`, for better consistency with other check script names


chalin:
Ok, I think that this is working as expected. The i18n check failure is due to / block waiting on:

- #4645

svrnm (approved):

PR #4653, opened 2024-06-11
Update component-label-map.yml

svrnm:
This adds a `docs` label to all english content and a `translation` label to all non-english content which is translated from it, this way approvers/maintainers can filter by new additions to the documentation or by translations

### DIFF
--- a/.github/component-label-map.yml
+++ b/.github/component-label-map.yml
@@ -2,10 +2,18 @@ blog:
   - changed-files:
       - any-glob-to-any-file:
           - content/en/blog/**
-registry:
+docs:
   - changed-files:
       - any-glob-to-any-file:
-          - data/registry/**
+          - content/en/**
+docs:
+  - changed-files:
+      - any-glob-to-any-file:
+          - content/en/**
+translation:
+  - changed-files:
+      - any-glob-to-any-file:
+          - '!content/en/**'
 lang:zh:
   - changed-files:
       - any-glob-to-any-file:
@@ -14,6 +22,10 @@ lang:ja:
   - changed-files:
       - any-glob-to-any-file:
           - content/ja/**
+registry:
+  - changed-files:
+      - any-glob-to-any-file:
+          - data/registry/**
 sig:cpp:
   - changed-files:
       - any-glob-to-any-file:


.github/component-label-map.yml
```diff
@@ -2,10 +2,18 @@ blog:
...
-registry:
+docs:
   - changed-files:
       - any-glob-to-any-file:
-          - data/registry/**
+          - content/en/**
+docs:
+  - changed-files:
+      - any-glob-to-any-file:
+          - content/en/**
```
chalin, line 12: Is there something I'm missing or do we have duplicate entries here?

.github/component-label-map.yml
```diff
@@ -2,10 +2,18 @@ blog:
...
+translation:
```
chalin, line 13: How about:
```suggestion
i18n:
```

.github/component-label-map.yml
chalin: For consistency, should `docs` be triggered on all `content/**`
And then for consistency add `lang:XX` for any `content/XX/**`, even English?

svrnm: No duplicate entries, the diff is just looking strange since I also moved registry down.

The goal is to distinguish between translations and non-translations, such that I can have dedicated views with one label vs doing a "-lang:1 -lang:2 -lang:3..."

.github/component-label-map.yml
svrnm: then it would be l10n from my understanding. I am also not happy with "translation", I also was thinking about "lang"

.github/component-label-map.yml
chalin: If I look at the file source, I see two entries for `docs` that have the same fields, so I think that there is duplication. That is, in the following, lines 5 to 8 are the same as 9 to 12.

> image

Anyhow, maybe we can drop the `docs` tagging rules and use `lang:en` for English pages? Would that work for the use cases you have in mind (I think that it does). That is, use `lang:en` and `-lang:en`. WDYT?

.github/component-label-map.yml
chalin: Right, `l10n`. Btw, I think that the rule should be

```yaml
  - changed-files:
      - any-glob-to-any-file:
          - content/**
          - '!content/en/**'
```

.github/component-label-map.yml
svrnm: oh... then I dropped the registry: somewhere, let me fix that, `lang:en` is a good idea, becaue there should (rarely) be a PR that touches multiple languages.

svrnm:
I need to redo this PR

PR #4658, opened 2024-06-11
Update opentelemetry-java version to v1.39.0, take 2

chalin:
Followup to:

- #4621 

I can't recall for sure why the change in this PR wasn't included by 4621, other than because the auto-update script had not yet been adjusted yet.

Trigger the auto-version update script has no effect, because it won't submit "duplicate" PRs when one exists already. Hence this manually generated PR.

The only place that I see this variable used is in the Sprint Boot starter > Getting started > Dependency management, in the Maven config. PTAL to confirm that the version value is correct for:

```xml
        <dependency>
            <groupId>io.opentelemetry</groupId>
            <artifactId>opentelemetry-bom</artifactId>
            <version>1.39.0</version>
            <type>pom</type>
            <scope>import</scope>
        </dependency>
```


zeitlinger:
yes, that's correct

PR #4676, opened 2024-06-13
Add Kubewarden to integrations.yaml

flavio:
Kubewarden is using OpenTelemetry inside of its Rust and Go code both for tracing and metrics.


data/ecosystem/integrations.yaml
```diff
@@ -238,3 +238,8 @@
   docsUrl: https://istio.io/latest/docs/tasks/observability/distributed-tracing/opentelemetry/
   components: [Go]
   oss: true
+- name: Kubewarden
```
theletterf, line 241: @flavio Could you add `Kubewarden` to the list of cSpell exceptions on line 1? Thanks!

flavio: done, change force pushed

flavio:
There's still a check failing. I've run `npm run test-and-fix` as suggested [here](https://github.com/open-telemetry/opentelemetry.io/actions/runs/9515656884/job/26230480884?pr=4676), but no file was changed locally.

Can someone help with that?

cartermp (approved):

Issue #4467, opened 2024-05-10
More i18n support


chalin:
Carry-over items from #4443:

- [ ] https://github.com/gohugoio/hugoDocs/pull/2564
- [ ] Ensure that search works as expected (a concern raised for other Docsy sites but that might apply here)

chalin:
From https://github.com/open-telemetry/opentelemetry.io/issues/2402#issuecomment-2109153206:

> @chalin> Should English banners show up on the zh homepage? That's how it is currently setup. It might seem reasonable to include the English banners when the Chinese versions aren't available?

@windsonsea> Could we provide a placeholder-alike? If no localized zh text, use the en banner; if localized, use the zh banner.

Andygol:
I would like you to consider and implement CrowdIn translation management system (or any other similar system) as part of this  initiative. It will greatly simplify and speed up the work of translation and maintenance teams. 🙏 

RE https://crowdin.com/page/open-source-project-setup-request

PR #4689, opened 2024-06-14
[i18n] Make absolute paths locale-specific by default

### LINKED ISSUE #4688: [i18n][design] Interpret absolute paths as locale specific
### Design l10n proposal

In summary, I’d like to propose that “absolute” link paths like `/some-path` be interpreted relative to the current localization site root, whether it is the default localization or not. So, it will be `otel.io/some-path` for `en` pages and `otel.io/xx/some-path` for non-en pages with language code `xx`.

### Here’s why…

With fallback pages in place, each localization is like its own (sub) site — and in fact, Hugo treats each localization as a full Site object.

For the default localization (which is `en`), link paths like `/some-path` implicitly refer to pages under `content/en`. If an English page wants to refer to say, a `ja` page, then the link path would be written as `/ja/some-path`.

Currently, localization authors for language `xx` who translate an English page, need to remember to change **all paths** of the form `/some-path` to `/xx/some-path`, otherwise they’ll be linking out to `en` pages. That is a burden on writers and maintainers.
I’d like to put a link render hook in place so that `/some-path` refers to the **localization-specific version of the page**, whether it is `en` or not. If a non-`en` localization wants to explicitly refer to an `en` page, they’d use `/en/some-path` as a path.

How does that sound?

This was originally posted on Slack at https://cloud-native.slack.com/archives/C06EDFPQ5EH/p1718358356363719.

@theletterf and @svrnm gave a thumbs up

chalin:
- Contributes to #4467
- Fixes #4688
- Removes unnecessary `/zh/` path prefix from link paths in `zh` pages

**Preview**: https://deploy-preview-4689--opentelemetry.netlify.app/docs/contributing/localization/#translation-tips

### DIFF
--- a/content/en/docs/contributing/localization.md
+++ b/content/en/docs/contributing/localization.md
@@ -17,6 +17,28 @@ English is the default language, with US English as the default (implicit) local
 A growing number of other localizations are supported, as can be seen from the languages
 dropdown menu in the top nav.
 
+## Translation tips
+
+When you translate page content, follow this guidance:
+
+- To ensure that heading anchor targets are uniform across localizations, when
+  translating headings:
+  - If the heading has an explicit ID (which is of the form `{ #some-id }` and
+    comes after the heading text), then preserve the given ID
+  - Otherwise, include a heading ID corresponding to the original English
+    heading text.
+- For link references that are local paths, preserve the path _as is_.
+
+{{% alert title="Note" %}}
+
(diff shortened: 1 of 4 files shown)


svrnm (approved):

chalin:
All checks are green. Given that the only change to the `zh` file was due to path normalization, I'm going to merge now.

Issue #3674, opened 2023-12-11
Baggage doc should be modified to make it less tied to tracing singnal


cijothomas:
**What needs to be changed?**

Baggage doc is tied too much to tracing. We need to modify it making it clear that Baggage is applicable across all signals, and potentially show examples on Baggage can be used across signals.

**What is the name + path of the page that needs changed?** 
https://github.com/open-telemetry/opentelemetry.io/blob/main/content/en/docs/concepts/signals/baggage.md

**Additional context**: Noticed in https://github.com/open-telemetry/opentelemetry.io/pull/3657#discussion_r1420851652.

svrnm:
Thanks for raising this, I agree, this page needs some attention

theletterf:
Thanks for raising this. We had a similar conversation on this @flands and I.

svrnm:
Yeah, the baggage doc needs some attention...

Issue #1878, opened 2022-10-19
Baggage content is too trace related, .NET specific example is confusing


cijothomas:
Couple suggestions:
1. The contents are too trace related. Baggage can be used for all signals, not just as span attributes. Metrics can have attributes populated from baggage. Same with logs
2. I'd recommend to use a diff. example as `Activity.Current?.SetTag("AccountId", accountId)` is .NET specific, which can confuse users as .NET calls its spans "Activity". Users will find it confusing. If we keep the same example, mention in code comment that Activity is .NET's representation of Span.

cartermp:
Agreed overall, although I'd say that Baggage is currently the _most_ useful in tracing since it lets you propagate metadata along the rest of a trace, whereas this kind of capability is less important outside the context of tracing. So while it's true that it's independent from tracing, it's also much more niche to consider it outside the context of tracing, since going through the pains of propagating baggage without getting the benefit of tracing is kinda weird.

theletterf:
@cijothomas Are you up for it? :)

cijothomas:
> @cijothomas Are you up for it? :)

I'll help review any PRs addressing this. I just don't have the bandwidth to work on an actual PR itself.

cartermp:
Closing as per doc update

PR #4692, opened 2024-06-15
Rework baggage page to be less trace-specific

cartermp:
Should fix https://github.com/open-telemetry/opentelemetry.io/issues/3674 and https://github.com/open-telemetry/opentelemetry.io/issues/1878#issuecomment-2158675491

Tried to capture how Baggage isn't trace-specific while still honoring the reality that it's usually used the most in conjunction with tracing.

cc @cijothomas


austinlparker:
Would it be worth calling out somehow that baggage is only intended for Observability data, and that it's not recommended to use it to side channel values for other purposes (control flow, etc)?

cartermp:
@austinlparker added

content/en/docs/concepts/signals/baggage.md
```diff
@@ -4,46 +4,49 @@ weight: 4
 description: Contextual information that is passed between signals.
 ---
 
-In OpenTelemetry, Baggage is contextual information that’s passed between spans.
-It's a key-value store that resides alongside span context in a trace, making
-values available to any span created within that trace.
+In OpenTelemetry, Baggage is contextual information that resides next to
+context. Baggage is a key:value store, which means it lets you propagate any
```
svrnm, line 8:
```suggestion
context. Baggage is a key-value store, which means it lets you [propagate](/docs/concepts/context-propagation/#propagation) any
```

content/en/docs/concepts/signals/baggage.md
```diff
@@ -4,46 +4,49 @@ weight: 4
...
+data you like alongside context.
```
svrnm, line 9:
```suggestion
data you like alongside [context](/docs/concepts/context-propagation/#context).
```

content/en/docs/concepts/signals/baggage.md
```diff
@@ -4,46 +4,49 @@ weight: 4
...
 
-For example, imagine you want to have a `ClientId` attribute on every span in
-your trace, which involves multiple services; however, `ClientId` is only
-available in one specific service. To accomplish your goal, you can use
-OpenTelemetry Baggage to propagate this value across your system.
+Baggage means you can pass data across services and processes, making it
+available to add to trace, metrics, or logs in those services.
```
svrnm, line 12:
```suggestion
available to add to [trace](/docs/concepts/signals/traces/), [metrics](/docs/concepts/signals/metrics/), or [logs](/docs/concepts/signals/logs/) in those services.
```

content/en/docs/concepts/signals/baggage.md
```diff
@@ -52,23 +55,35 @@ then this risk may not apply, but keep in mind that downstream services could
...
 ```
 
+Because one of the most common use cases for Baggage is to add data to
+[Span Attributes](/docs/concepts/signals/traces/#attributes) across a whole
+trace, several languages have Baggage Span Processors that automatically read
+Baggage values and add them as attributes for you.
+
+## Baggage is only for observability data
+
+Although Baggage does support propagating arbitrary key:value pairs across
```
svrnm, line 80:
```suggestion
Although Baggage does support propagating arbitrary key-value pairs across
```

content/en/docs/concepts/signals/baggage.md
```diff
@@ -52,23 +55,35 @@ then this risk may not apply, but keep in mind that downstream services could
...
+services, it is not a replacement for programmatic data flow. Because Baggage is
+stored in HTTP headers, not only does this mean the data is available for anyone
+who can read those headers to consume, but it would be a poor choice to
+propagate general program data given the practical limits HTTP servers place on
+header lengths.
```
svrnm, line 85: I like this section. To make the disadvantage maybe even more apparent you could give an example where headers get dropped by a proxy inbetween services, which then **only** affects observability but not the functionality of your app

svrnm (changes requested):
Thank you! Some remarks

content/en/docs/concepts/signals/baggage.md
```diff
@@ -4,46 +4,49 @@ weight: 4
...
-Baggage provides a uniform way to store and propagate information across a trace
-and other signals. For example, you may want to attach information from your
-application to a span and retrieve that information much later and use it later
-on with another span. However, spans in OpenTelemetry are immutable once
-created, and can be exported before you need information on them later on.
-Baggage allows you to work around this problem by providing a place to store and
-retrieve information.
+By using
+[Context Propagation](/docs/concepts/signals/traces/#context-propagation) to
+pass Baggage across these services, the `clientId` is available to add to any
```
theletterf, line 26:
```suggestion
pass baggage across these services, the `clientId` is available to add to any
```

content/en/docs/concepts/signals/baggage.md
```diff
@@ -4,46 +4,49 @@ weight: 4
...
-Common use cases include information that’s only accessible further up a stack.
-This can include things like Account Identification, User IDs, Product IDs, and
-origin IPs, for example. Passing these down your stack allows you to then add
-them to your Spans in downstream services to make it easier to filter when
-you’re searching in your Observability backend.
+Baggage is best used to include information typically available only at the
+start of a request further downstream. This can include things like Account
+Identification, User IDs, Product IDs, and origin IPs, for example.
+
+Propagating this information via Baggage allows for deeper analysis of telemetry
```
theletterf, line 38:
```suggestion
Propagating this information using baggage allows for deeper analysis of telemetry
```

content/en/docs/concepts/signals/baggage.md
```diff
@@ -52,23 +55,35 @@ then this risk may not apply, but keep in mind that downstream services could
...
 
-## Baggage is not the same as Span attributes
+## Baggage is not the same as attributes
 
-One important thing to note about Baggage is that it is not a subset of the
-[Span Attributes](/docs/concepts/signals/traces/#attributes). When you add
-something as Baggage, it does not automatically end up on the Attributes of the
-child system’s spans. You must explicitly take something out of Baggage and
-append it as Attributes.
+An important thing to note about Baggage is that it is a separate key:value
```
theletterf, line 62:
```suggestion
An important thing to note about baggage is that it is a separate key-value
```

content/en/docs/concepts/signals/baggage.md
```diff
@@ -52,23 +55,35 @@ then this risk may not apply, but keep in mind that downstream services could
...
 
+Because one of the most common use cases for Baggage is to add data to
+[Span Attributes](/docs/concepts/signals/traces/#attributes) across a whole
+trace, several languages have Baggage Span Processors that automatically read
+Baggage values and add them as attributes for you.
+
+## Baggage is only for observability data
+
+Although Baggage does support propagating arbitrary key:value pairs across
+services, it is not a replacement for programmatic data flow. Because Baggage is
```
theletterf, line 81:
```suggestion
services, it is not a replacement for programmatic data flow. Because baggage is
```

theletterf (changes requested):

content/en/docs/concepts/signals/baggage.md
cartermp: Done

content/en/docs/concepts/signals/baggage.md
```diff
@@ -52,23 +55,37 @@ then this risk may not apply, but keep in mind that downstream services could
...
-[Span Attributes](/docs/concepts/signals/traces/#attributes). When you add
-something as Baggage, it does not automatically end up on the Attributes of the
-child system’s spans. You must explicitly take something out of Baggage and
-append it as Attributes.
+An important thing to note about baggage is that it is a separate key-value
+store and is unassociated with attributes on spans, metrics, or logs.
 
-For example, in .NET you might do this:
+To add Baggage entries to attributes, you need to explicitly read that data and
+add it. For example, in .NET you might do this:
```
cijothomas, line 66: could we switch to use any language here other than .NET? ("Activity", "Tag" are unknown terms outside .NET, so best to keep this example neutral)

content/en/docs/concepts/signals/baggage.md
```diff
@@ -52,23 +55,37 @@ then this risk may not apply, but keep in mind that downstream services could
...
 
 ```csharp
 var accountId = Baggage.GetBaggage("AccountId");
 Activity.Current?.SetTag("AccountId", accountId);
 ```
 
+Because one of the most common use cases for Baggage is to add data to
```
cijothomas, line 73: I am not sure if the part _"Because one of the most common use cases"_ is relevant here.

content/en/docs/concepts/signals/baggage.md
```diff
@@ -52,23 +55,37 @@ then this risk may not apply, but keep in mind that downstream services could
...
+[Span Attributes](/docs/concepts/signals/traces/#attributes) across a whole
+trace, several languages have Baggage Span Processors that automatically read
```
cijothomas, line 75: could we add links to them? Or mention something like "consult language specific doc to learn how to use them".

content/en/docs/concepts/signals/baggage.md
```diff
@@ -52,23 +55,37 @@ then this risk may not apply, but keep in mind that downstream services could
...
+Baggage values and add them as attributes for you.
+
+## Baggage is only for observability data
```
cijothomas, line 78: any reason why we need to mention that? 

For example, if a user is using baggage to pass something like ("real-user","false") to all downstreams, and downstream services could be leveraging this to make decisions like which traffic to prioritize (eg: prioritize ("real-user","true")), without ever recoding the baggage in any telemetry data (i.e Baggage is not ending up on Observability data), is that wrong?

OTel Baggage spec (or W3CBaggage) does not have any such wordings, so I don't think we should add it for the docs here.

content/en/docs/concepts/signals/baggage.md
```diff
@@ -4,46 +4,49 @@ weight: 4
...
+In OpenTelemetry, Baggage is contextual information that resides next to
+context. Baggage is a key-value store, which means it lets you [propagate](/docs/concepts/context-propagation/#propagation) any
+data you like alongside [context](/docs/concepts/context-propagation/#context).
 
-For example, imagine you want to have a `ClientId` attribute on every span in
-your trace, which involves multiple services; however, `ClientId` is only
-available in one specific service. To accomplish your goal, you can use
-OpenTelemetry Baggage to propagate this value across your system.
+Baggage means you can pass data across services and processes, making it
+available to add to [trace](/docs/concepts/signals/traces/), [metrics](/docs/concepts/signals/metrics/), or [logs](/docs/concepts/signals/logs/) in those services.
```
cijothomas, line 12:
```suggestion
available to add to [traces](/docs/concepts/signals/traces/), [metrics](/docs/concepts/signals/metrics/), or [logs](/docs/concepts/signals/logs/) in those services.
```
```suggestion
available to add to [trace](/docs/concepts/signals/traces/), [metrics](/docs/concepts/signals/metrics/), or [logs](/docs/concepts/signals/logs/) in those services.
```

content/en/docs/concepts/signals/baggage.md
cartermp: Yeah, it's not mentioned in the spec but I really think we don't want people using this as a replacement for a data flow library or framework. Storing potentially large amounts of data in HTTP headers via a mechanism designed for observability feels wrong.

austinlparker: we should mention it because we don't want people to do it. i've seen people abuse baggage then get confused by it. if someone wants to use w3c baggage for _other_ reasons, fine, but they should do it outside of otel namespaces

content/en/docs/concepts/signals/baggage.md
cartermp: I removed it -- decided it's not worth putting code samples in here right now.

content/en/docs/concepts/signals/baggage.md
cartermp: Reworded a bit. It is common, maybe not the most common

content/en/docs/concepts/signals/baggage.md
cartermp: I decided against this right now since we don't have docs in the languages on using baggage APIs yet. If/when we do, we should absolutely link them though.

content/en/docs/concepts/signals/baggage.md
cijothomas: if doc owners agree with this, I have no objection to add it here. But I'd suggest to first add such wording in the spec itself, instead of just adding in the docs.

One interesting wording I noticed [in the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#get-a-tracer):
_"OpenTelemetry API may also ignore this name and return a default instance for all calls, if it does not support "named" functionality **(e.g. an implementation which is not even observability-related**)"_

content/en/docs/concepts/signals/baggage.md
svrnm: Do we have any languages that mention the baggage API already? If so we might consider having some listing similar to the sampling page?

cartermp: Not that I'm aware of, no.

content/en/docs/concepts/signals/baggage.md
cijothomas: @kalyanaj To share any comments from W3CBaggage perspective, which could be different from OTel's perspective.

svrnm (approved):

content/en/docs/concepts/signals/baggage.md
kalyanaj: From the W3C baggage spec perspective, we think of two categories of use cases for distributed context propagation / baggage: 

1) use cases that improve observability and 
2) use cases that improve controllability. 

We have documented some example use cases here: https://github.com/w3c/baggage/blob/main/baggage/README.md, and this is inspired from @yurishkuro 's post on the same topic.  Also, ccing: @dyladan and @SergeyKanzhelev. 

So, if someone wants to accomplish the use cases in category 2 above, my understanding until now had been that OTel Baggage APIs could be used for it: if the customer rolls out their own libraries in multiple languages instead of using the standard OTel libraries, I feel that will lead to even more problems. Are the possibilities of misuse specific to the non-observability use cases, or can they happen in either category?

So, want to learn/understand if there are other mechanisms to prevent abuse - e.g., by having suitable limits (which I believe OTel already has), semantic conventions for well-known use cases etc.

yurishkuro: I know many use cases from real large scale systems that use baggage for things outside of observability.  Just recently saw a huge overload protection win in a team that used baggage to propagate request priority.

cartermp: Okay. Sounds good to me. I think we can just nix this section.

cijothomas (approved):

cartermp:
@theletterf any final thoughts?

Issue #4230, opened 2024-04-02
Incorrect display of "<" and ">" characters


progxaker:
**What happened?**
The characters "<" and ">" are displayed as "&lt;" and "&gt;".
https://github.com/open-telemetry/opentelemetry.io/blob/db091baf9937a69d4c165aeaf40cb17fd9e0125b/content/en/docs/languages/java/automatic/configuration.md?plain=1#L136
![image](https://github.com/open-telemetry/opentelemetry.io/assets/26653354/97fad1f6-e6f8-42cb-aa2d-0516a2d88f60)

**What did you expect would happen?**
The same result as in the Markdown code.

**What is the name + path of the page where you encountered the issue?**
Path: Language APIs & SDKs -> Java -> Automatic -> Configuration
Name: Agent Configuration
Link: https://opentelemetry.io/docs/languages/java/automatic/configuration/#peer-service-name

**What browser, OS, and platform were you using when you encountered the
problem?**
Waterfox G6.0.11 (64-bit), Xubuntu 22.04.4 LTS x86_64, Desktop.

chalin:
Thanks for reporting this. I'll take a look as soon as I can.

progxaker:
Hello @chalin. Have you had time to look at this? Thanks in advance.

svrnm:
@progxaker thanks for bringing this back to our attention. I quickly looked into it and was able to identify the root cause. @chalin can look at this when he is available again.

progxaker:
@svrnm thank you for the investigation and prepared (draft) changes 🙂

svrnm:
Referenced page reorganized/removed

PR #4709, opened 2024-06-18
fix bug in config_option shortcode

svrnm:
This addresses #4230. Note that this PR is not yet finalized, but to begin with it demonstrates the underlying issue.

There seems to be an issue with hugo "re-markdownifying" (see https://github.com/gohugoio/hugo/issues/5774#issuecomment-486425783), so this applies the same suggestions:

1. Use {{< request >}} instead of {{% request %}}, so Hugo stops trying to re-markdownify
2. Manually markdownify the content

cc @chalin

### DIFF
--- a/content/en/docs/zero-code/java/agent/configuration.md
+++ b/content/en/docs/zero-code/java/agent/configuration.md
@@ -123,7 +123,7 @@ is the name of a remote service to which a connection is made. It corresponds to
 `service.name` in the [resource](/docs/specs/semconv/resource/#service) for the
 local service.
 
-{{% config_option name="otel.instrumentation.common.peer-service-mapping" %}}
+{{< config_option name="otel.instrumentation.common.peer-service-mapping" >}}
 
 Used to specify a mapping from host names or IP addresses to peer services, as a
 comma-separated list of `<host_or_ip>=<user_assigned_name>` pairs. The peer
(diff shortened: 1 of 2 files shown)


chalin:
Thanks, I appreciate the time you put into investigating this. I can't recall the reason why I made the shortcode an HTML file (other than it mainly contains only HTML :)). What you've implemented here feels like a workaround. I'd like to try to fix the source of the problem. The first thing that I'd try is to rename the shortcode to `config_option.md` and try to get it to work from there. I don't know when I'll have the time to do that.

Have you double checked that this "fix" doesn't break other invocations of `config_option`?

svrnm:
> Thanks, I appreciate the time you put into investigating this. I can't recall the reason why I made the shortcode an HTML file (other than it mainly contains only HTML :)). What you've implemented here feels like a workaround. I'd like to try to fix the source of the problem. The first thing that I'd try is to rename the shortcode to `config_option.md` and try to get it to work from there. I don't know when I'll have the time to do that.

That would be great!

> Have you double checked that this "fix" doesn't break other invocations of `config_option`?

No, I haven't, that's why this is a draft PR, I didn't want to add this without further investigation from your site.

svrnm:
@chalin I will close this PR now, since it only was a proof of concept, it's linked to the original issue

PR #4741, opened 2024-06-25
[ja] translate Semantic convention and Resource

ymotongpoo:
translate the following files into Japanese

- content/ja/docs/concepts/resources/index.md
- content/ja/docs/concepts/semantic-conventions.md


katzchang (approved):

content/ja/docs/concepts/resources/screenshot-jaeger-resources.png
svrnm, line 1: Just tagging this (non blocking for the merge): it might become worth to investigate how we want to handle images in localized documentation. It's probably not trivial to solve, because sometimes the image needs to be localized as well, but somtimes we also might just want to use the existing English one (as we do here)

cc @chalin

svrnm (approved):

content/ja/docs/concepts/resources/screenshot-jaeger-resources.png
chalin: Hugo is now setup to reuse the default language image -- the exact same file, no duplication (as was the case for static resources a few Hugo versions ago).

**Images should _not_ be duplicated.** I'd suggest using the default image except where there's too much English text in the image, of course.

chalin, line 1: Has this image been changed (it doesn't seem so), in which case it should not be included in this PR.

 I might have to do some tests to ensure that this works as expected. Putting this PR in draft mode until then.

chalin (changes requested):

content/ja/docs/concepts/resources/screenshot-jaeger-resources.png
chalin: Thanks for this first PR that surfaces the problem with shared images @ymotongpoo. 
I'll need some time to investigate how to address this issue.

ymotongpoo: ok, then I remove this file from this PR.

ymotongpoo: I removed duplicated files from this PR.

chalin (approved):

PR #4749, opened 2024-06-25
Update blog command with correct year

tiffany76:
The [commands for creating a blog post template file](https://opentelemetry.io/docs/contributing/blog/#fork-and-write-locally) still use the year 2023. This PR updates the year to 2024.


svrnm:
good catch! We might consider replacing this with some hugo magic to print the current year, but let's fix it that way first

svrnm (approved):

PR #4752, opened 2024-06-26
Add a note about the Spring Boot BOM

jeanbisutti:


zeitlinger (approved):
I wonder if we should add a similar note about maven. Most users use the parent project and that seems to work fine

jeanbisutti:
> I wonder if we should add a similar note about maven. Most users use the parent project and that seems to work fine

The issue was observed with Maven and would probably also happen with Gradle. For clarity, I have just moved the position of the note.

svrnm (approved):

PR #4756, opened 2024-06-27
Elastic APM  is oss

shalk:
Elastic APM server and agent are oss https://github.com/elastic/apm-server/

storage and ui elasticsearch and kibana are oss.


svrnm:
@AlexanderWert I think this is correct but can you verify that this is the case for sure? Thanks

svrnm (approved):

svrnm:
apologies for the delay @shalk, merging this now based on your pointers that it is OSS for sure. if this is not correct, someone can provide a PR to revert this change

PR #4793, opened 2024-07-04
Update getting-started.md

Hill11235:
Slight typo, suggesting a small improvement

### DIFF
--- a/content/en/docs/languages/go/getting-started.md
+++ b/content/en/docs/languages/go/getting-started.md
@@ -262,7 +262,7 @@ func newLoggerProvider() (*log.LoggerProvider, error) {
 ```
 <!-- prettier-ignore-end -->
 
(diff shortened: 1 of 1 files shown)


content/en/docs/languages/go/getting-started.md
```diff
@@ -262,7 +262,7 @@ func newLoggerProvider() (*log.LoggerProvider, error) {
 ```
 <!-- prettier-ignore-end -->
 
-If you're only using tracing or metrics, you can omit the code the corresponding
+If you're only using tracing or metrics, you can omit from the code the corresponding
```
dmathieu, line 265: I wouldn't even specify the `from the code` bit.

```suggestion
If you're only using tracing or metrics, you can omit the corresponding
```

dmathieu (approved):

cartermp (approved):

PR #4804, opened 2024-07-08
add missing java http settings

zeitlinger:


svrnm (approved):

jeanbisutti (approved):

theletterf (approved):

chalin:
As is mentioned in #4815 (and @svrnm pointed out again recently in some other thread that I can't recall), care must be taken when running the slash fix commands: since the GH actions are _not_ rerun, the responsibility falls back on the invoker of the command to ensure that the previously failed GH actions are _all_ actually _really_ solved.

For a following PR to this PR see:

- #4815

zeitlinger:
> are _all_ actually _really_ solved.

is there a way to make this less fragile?

PR #4805, opened 2024-07-08
Introduce schema for registry entries

svrnm:
This PR introduces a JSON schema for registry entries. We can use this schema eventually to validate that registry entries provide all the data needed. This PR does not yet include that step as a CI action, but there is already a github friendly output


theletterf (approved):

package.json
```diff
@@ -103,9 +105,14 @@
     "update:submodule": "set -x && git submodule update --remote ${DEPTH:- --depth 1}"
   },
   "devDependencies": {
+    "ajv": "^8.16.0",
+    "ajv-errors": "^3.0.0",
+    "ajv-formats": "^3.0.1",
     "autoprefixer": "^10.4.19",
     "cspell": "^8.12.1",
     "gulp": "^5.0.0",
+    "hugo-extended": "0.128.0",
```
chalin, line 114: @svrnm - this added a _second_ entry for `hugo-extended`; the other is at line 116 below. I'll file an issue for this

PR #4829, opened 2024-07-12
Update vendors.yaml with APImetrics

SophieMcK:


tiffany76:
Hi @SophieMcK, can you sign the CLA? We need your signature before we can merge this PR. Thanks!

SophieMcK:
Hi @tiffany76. CLA has been signed now

svrnm:
@SophieMcK can you provide us with some more details what your platform is doing? I read the following in your docs:

> Send data from APImetrics to any platform that intakes OpenTelemetry with our OpenTelemetry Webhook

This sounds more like your platform sends data OUT in OpenTelemetry, and not that you take OpenTelemetry data IN?

SophieMcK:
Yes. That is the case

svrnm:
> Yes. That is the case

Then I would argue that this might be an integration instead. I know that the lines between the different categories are blurry, but "vendors" are mostly backends (or observability pipelines) that **consume** opentelemetry data.

SophieMcK:
Okay. I'll close this and submit to the integrations page

Issue #4795, opened 2024-07-05
Highlight first party ownership in the ecosystem registry


codefromthecrypt:
**Desired feature or idea**:

Similar to how there is highlighting of code maintained by us, I'd like to see highlighting of first party ownership of instrumentation, exporters, etc in the ecosystem registry.

For example, this project is maintained by us
![image](https://github.com/open-telemetry/opentelemetry.io/assets/64215/f927b9a2-9207-43fc-931e-c1d28e52260d)

These two are maintained by the upstream project (first party), something you can tell by name comparison, but it isn't annotated otherwise
![image](https://github.com/open-telemetry/opentelemetry.io/assets/64215/8e7df296-11b7-4d54-832f-95731a2e5ca7)

I'd love to see a stylesheet or otherwise change based on a flag in the registry data of "isUpstream" negate of "isThirdParty" or similar, so that end users can know something in the registry is authoritatively owned.

**Additional context**:

While it is easy to tell which instrumentation are done by us, it is not easy to tell which are done by the upstream owners. We make a call that is in my opinion correct, [that the very first priority is that the instrumentation is upstream](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/instrumentation#new-instrumentation).  By highlighting we encourage the same behaviors in our registry as our policy, and make it easier for end users to know if something is neither maintained by us, nor by the project owner. For example, they can prefer something that isn't a vendor-specific codebase which is neither subject to the conduct of the target technology, nor ours.

svrnm:
We have functionality it's just not enabled for all those upstream instrumentations, we can distinguish in the registry between "first party integrations" and "native":

first party example: https://opentelemetry.io/ecosystem/registry/?s=kafkaflow
native example: https://opentelemetry.io/ecosystem/registry/?s=next.js

We also have the integrations page to feature those:

https://opentelemetry.io/ecosystem/integrations/

A piece missing is that this list of integrations shares the data from the registry, right now those are 2 independent data sets. We also leverage that second list to promote native (and only native) integrations on their language pages around libraries, e.g. https://opentelemetry.io/docs/languages/js/libraries/#use-natively-instrumented-libraries

All of this also has a dependency with this terminology discussion (and everytime I tried to improve the situation I struggeld with finding the right words): https://github.com/open-telemetry/opentelemetry-specification/issues/4089

So what can you do to help with that:
* Tag registry entries which are missing the native/first party flag
[...]

PR #4836, opened 2024-07-14
Add native/firstParty flag to collector exporters

haidong:
To address #4795, this is a series of PRs that add native or first party flags to registries that lack them. This batch touches collector-builder and all collector-exporter-* registries.

### DIFF
--- a/layouts/partials/ecosystem/registry/entry.html
+++ b/layouts/partials/ecosystem/registry/entry.html
@@ -48,7 +48,7 @@
                 "icon" "fab fa-rust")
     -}}
 
-    {{ $isNativ := and (eq .registryType "instrumentation") (.isNative) }}
+    {{ $isNative := and (eq .registryType "instrumentation") (.isNative) }}
     {{ $isFirstParty := and (eq .registryType "instrumentation") (.isFirstParty) }}
     {{ $currentTime := (time now) -}}
     {{ $delta := $currentTime.Sub (time.AsTime .createdAt) -}}
@@ -64,7 +64,7 @@
     {{ if $isNew -}}
         {{ $highlightStyle = "border-info" -}}
     {{ end -}}
-    {{ if $isNativ -}}
+    {{ if $isNative -}}
         {{ $highlightStyle = "border-success" -}}
     {{ end -}}
     {{ if $isFirstParty -}}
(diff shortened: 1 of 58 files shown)


cartermp (approved):

data/registry/collector-builder.yml
```diff
@@ -14,6 +14,7 @@ urls:
   repo: https://github.com/open-telemetry/opentelemetry-collector/tree/main/cmd/builder
   docs: /docs/collector/custom-collector/
 createdAt: 2023-12-18
+isFirstParty: true
```
svrnm, line 17: Apologies, I should have been more specific in the original issue. This flag is not relevant for collector components it refers to instrumentations that are "first party", i.e. provided by the same org that created a certain app or library. 

We could consider to have them for receivers/exporters hosted outside the community repos by orgs that create such modules fur their solution but I don't know if this is what we want them to do va adopting otlp

svrnm (changes requested):
Please see my inline comment. IsFirstParty should not be used for collector components.

data/registry/collector-builder.yml
haidong: I was unsure about this change. Thanks for the explanation.

I've made the adjustment to your change request, rebased and adjusted the commit message a bit, then did a force push.

Thanks again!

haidong: Just to clarify, Do you mean that _all_ collector-exporters don't need this change? In other words, this PR should be abandoned?

svrnm: Depends on how we want to use that flag. We had a thirdParty flag in the past that we removed because it created so much confusion. Right now the first party flag is used in the registry to say what I outlined above. I don't know if we water down the value of that if all otel first party components have that label as well

layouts/partials/ecosystem/registry/entry.html
svrnm, line 1: Oh, thanks, typo by me! Can you move this in a separate PR since the other topic will need some time to be discussed?

svrnm:
Moving the discussion out from my outdated comment:

> Depends on how we want to use that flag. We had a thirdParty flag in the past that we removed because it created so much confusion. Right now the first party flag is used in the registry to say what I outlined above. I don't know if we water down the value of that if all otel first party components have that label as well

I don't think that we should add the isFirstParty flag to non-instrumentation entries of the registry, especially not for things build by the opentelemetry community. Otherwise 50%+ of our entries would have this flag (it is currently not showing up because `layouts/partials/ecosystem/registry/entry.html` is implementing that flag exactly that way that it only works for instrumentation)

We already highlight components by the opentelemetry community by distinguishing the "by The OpenTelemetry Authors" from other others by highlighting them.

Again, @haidong, I appreciate the effort you put into creating this PR, and I apologies for the confusion that my comment in #4795 might have created, but I would like to not implement this.

cartermp:
Agreed, let's keep this scoped to just instrumentations then.

haidong:
I've been busy with other stuff and didn't get a chance to attend to this until today.

Thanks all for your input. I've created a new PR that touch files under data/registry. Please let me know your suggestions/comments there.

PR #4845, opened 2024-07-18
[Go] Add zap log bridge

pellared:
Fixes https://github.com/open-telemetry/opentelemetry-go-contrib/issues/5191

I think it would be better to reuse the OTel Registry, but it is currently impossible:
- https://github.com/open-telemetry/opentelemetry.io/issues/4846


pellared:
CC @khushijain21

theletterf (approved):

PR #4865, opened 2024-07-23
Update integrations.yaml with APImetrics

SophieMcK:


data/ecosystem/integrations.yaml
```diff
@@ -310,3 +310,9 @@
   components: [C++, Java, Go, Python]
   oss: true
   cncf: incubating
+- name: APImetrics
+  url: https://apicontext.com/
+  docsUrl: https://docs.apimetrics.io/docs/export-with-opentelemetry
+  components: [Exporter]
```
svrnm, line 316: would it be possible to know how those exporters are implemented? so are you using opentelemetry collectors, or is this implemented with one of the language SDKs?

neil-gall: They're implemented using the Python SDK.

svrnm:
```suggestion
  components: [Python]
```

We need to do better in the future on how we distinguish these integrations, but let's have it like that for now.

svrnm (approved):

SophieMcK:
Is there any chance that this could be listed as APIContext and not APImetrics

svrnm:
> Is there any chance that this could be listed as APIContext and not APImetrics

You can provide a PR to update the file. Note that we have changed the way how we store integrations, you will find them in data/registry now

PR #4940, opened 2024-08-02
[CI] Only report registry format errors by default

chalin:
- Followup to #4805
- When running `check:registery` locally, only display warnings and notices when the `--debug` flag is provided. To do so run the script like this:
  `npm run check:registry -- --debug`

/cc @svrnm


tiffany76 (approved):

chalin:
Thanks for the review @tiffany76 ✨ 
@svrnm - going to merge now. We can always readjust if you'd prefer another solution.

PR #4970, opened 2024-08-06
Add community members page

svrnm:
This PR adds a page under /community that shows all members in their roles across the OpenTelemetry project. It also includes a script to update this data, if we want to add this to the page, I can turn that into a scheduled workflow. 

Preview: https://deploy-preview-4970--opentelemetry.netlify.app/community/members/

cc: @open-telemetry/sig-contributor-experience-approvers


theletterf:
Wow! I love it!

Any chance we'd split members by project using headings?

svrnm:
> Wow! I love it!
> 
> Any chance we'd split members by project using headings?

I purposfully didn't do that, because I wanted to have every person only represented once (except GC/TC), and also not make things to complicated. The per-subproject info is available on their repositories. What I was thinking about as a follow up feature is that there is a filter button and you can change that few to only show you maintainers by SIG

mx-psi (commented):
I think this is a great idea! The current CSS styling for the tables doesn't work well on small screens though, this is how it looks on a Pixel 7 screen 
image

(screenshot from devtools but it looks the same as in my personal device)

svrnm:
@mx-psi thanks! I just applied some grid options from bootstrap to address that:

```
<div class="col-6 col-md-4 col-lg-3 col-xxl-2">
```

mx-psi:
You can see the images now :) 
image

There are a couple of profile pictures that seem to be lower resolution and look odd, maybe we can enforce the same size for all profile pictures as well?

codefromthecrypt:
I like the progress here, might make the icons a bit smaller but no need to action that.

Is there any way to have text of the github ID here without making it messy? Right now I found myself hovering over each image ;)

.cspell.yml
```diff
@@ -8,6 +8,7 @@ import:
 caseSensitive: true
 ignorePaths:
   - '*.svg'
+  - 'data/community/members.yaml'
```
chalin, line 11:
```suggestion
  - data/community/members.yaml
```

gulp-src/lint-md.js
```diff
@@ -14,6 +14,7 @@ const markdownFiles = [
   '!layouts/**',
   '!node_modules/**',
   '!scripts/registry-scanner/node_modules/**',
+  '!scripts/generate-community-data/node_modules/**',
```
chalin, line 17: This is fine as is. I'm curious if we could get away with `!**/node_modules/**`.

chalin (approved):
Wow, nice work! ✨  RSLGTM

danielgblanco (approved):
Love it!

gulp-src/lint-md.js
svrnm: We can, fixed it :-)

mx-psi (approved):

svrnm:
> There are a couple of profile pictures that seem to be lower resolution and look odd, maybe we can enforce the same size for all profile pictures as well?

The flexible layout provided by bootstrap makes this not a trivial task, but I also do not want to write something on my own. It's not feature perfect, so I would suggest we iterate on it later.

> Is there any way to have text of the github ID here without making it messy? Right now I found myself hovering over each image ;)

done

codefromthecrypt (approved):
:shipit: thanks a lot!

svrnm:
FYI, I want to present this to the maintainers first (via slack/maintainers call) before merging to get a broad feedback for that.

data/community/members.yaml
```diff
@@ -0,0 +1,2125 @@
+technical-committee:
+  - name: jsuereth
```
cijothomas, line 2: could we show name along with github alias?

svrnm: I would guess so, the name should be accessible via the github API

svrnm:
Note, that I will need to remove the images from this PR, the images are sourced from github and we do not have full control what does images show and represent, so we have a risk of showing something on the website we don't want to have there.

codefromthecrypt:
> we have a risk of showing something on the website we don't want to have there.

honestly I did think about this, too, and wondered how we might deal with it

svrnm:
> > we have a risk of showing something on the website we don't want to have there.
> 
> honestly I did think about this, too, and wondered how we might deal with it

Not showing them is the only viable option. Another one would pulling and storing the images locally, and then reviewing them every time we update the list but that sounds like too much effort for me.

Note, that with the latest commit I fixed that and turned the lists in to html unordered lists and added tags per user with the SIGs

data/community/members.yaml
svrnm: from what I understand unfortunately that would need another API call per user, I will leave that out for now and use the github handles

data/community/members.yaml
```diff
@@ -0,0 +1,2211 @@
...
+  - name: Oberon00
+    html_url: https://github.com/Oberon00
+    avatar_url: https://avatars.githubusercontent.com/u/849039?v=4
+  - name: bogdandrutu
+    html_url: https://github.com/bogdandrutu
+    avatar_url: https://avatars.githubusercontent.com/u/1373887?v=4
+  - name: dyladan
+    html_url: https://github.com/dyladan
+    avatar_url: https://avatars.githubusercontent.com/u/1612643?v=4
+  - name: lmolkova
```
cijothomas, line 51: Looks like TC list is duplicated into spec-sponsors. Is that the intend? I think its best to remove TC from spec-sponsor list.

svrnm: good catch! Let me fix that

Issue #4981, opened 2024-08-06
[i18n] Spell-check non-en pages other than ja and zh


chalin:
- [x] `pt-BR` - #4982
- [x] `es` -- ref. #4887
- [ ] Add a note to the localization docs stating that spell checking should be set up if a dictionary is available for the newly proposed language

PR #4982, opened 2024-08-06
[CI][i18n] Spell check content/pt, and enable other dicts

chalin:
- Contributes to #4981
- Enables checking of `content/pt` pages
- Excludes `ja` and `zh` pages
- Adds other support dictionaries like `companies`, etc.
- Drops words from our local dictionary that are now covered by the newly added custom dictionaries


theletterf (approved):
Would I need to open a similar PR for `es`?

svrnm (approved):

chalin:
> Would I need to open a similar PR for `es`?

All we'd need to do is add an `es` dictionary (as an NPM package, and then to the cSpell config). Generally speaking, I'd suggest that we do that as part of the "on boarding" of the new language — that is, with their first PR that contains the translation of the homepage. WDYT?

Issue #4987, opened 2024-08-07
Add a filter to the registry for "is native" and "is first party"


svrnm:
Based on feedback from the last Maintainers call: it would be good if the registry would allow to filter entries by the "is native" and "is first party" flag. 

cc @arminru

svrnm:
If we merge #5328 this issue can be closed.

PR #4992, opened 2024-08-07
make native and first party searchable in registry

svrnm:
Partially addresses #4987: this PR allows users to search for "native" and "first party" and find the registry entries.

This PR also includes an update of the used search engine (minisearch) + pinning the hash


chalin (approved):
RSLGTM

PR #5008, opened 2024-08-09
[CI] Install cspell/dict before running cspell action

chalin:
- Contributes to #5007

PR #5010, opened 2024-08-09
[CI] Simplify cSpell dict install and other script tweaks

chalin:
Follow up to:

- #5008


theletterf (approved):

chalin:
Note that the job reports a `EBADENGINE` error, but I think that we can ignore that for now, since the checking works. Let me see if I can fix that.

chalin:
Fixed

PR #5014, opened 2024-08-09
[CI] Push back htmltest refcache expiry date, and fix a link

chalin:
I was refreshing cache entries, but I think that my IP might have been flagged by Cloudflare. So for now, I'll just push back the refcache expiry date. I managed to detect and fix one link.


chalin:
Letting GH fix the refcache since my machine seems to have been blocked :-/

chalin:
Got some updates over the refcache afer all \o/, so I force pushed those. Not sure why the refcache is being reported as stale?

svrnm (approved):

PR #5032, opened 2024-08-13
Update and rename application-integration-python-apimetrics.yml to ap…

SophieMcK:
…plication-integration-python-apicontext.yml

Change APImetrics to APIContext

### DIFF
--- a/data/registry/application-integration-python-apicontext.yml
+++ b/data/registry/application-integration-python-apicontext.yml
@@ -1,4 +1,4 @@
-# cspell:ignore apicontext
+# cspell:ignore apicontext, apimetrics
 title: APIContext
 registryType: application integration
 language: python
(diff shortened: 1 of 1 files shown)


data/registry/application-integration-python-apicontext.yml
```diff
@@ -1,16 +1,16 @@
-# cspell:ignore apimetrics
-title: APImetrics
+# cspell:ignore apicontext, apimetrics
+title: APIContext
```
svrnm, line 2: https://docs.apimetrics.io/docs/introducing-apimetrics

is the product called "APImetrics by APIContext" or will it be renamed to "APIContext" and also the docs link will change?

SophieMcK: The product is migrating over to the name APIContext. Docs will change eventually but not in the near future

svrnm: then I am fine with renaming it to APIContext already, just wanted to verify.

svrnm (approved):

PR #5092, opened 2024-08-22
Collector telemetry lifecycle

codeboten:
This is an initial attempt at documenting the lifecycle of telemetry emitted by the Collector.

Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/10003

Pinging @open-telemetry/collector-approvers for reviews on this 

---

**Preview**: https://deploy-preview-5092--opentelemetry.netlify.app/docs/collector/internal-telemetry/#telemetry-lifecycle


content/en/docs/collector/internal-telemetry.md
```diff
@@ -260,6 +260,53 @@ categorized by instrumentation type.
 | `rpc_server_response_size`        | Measures the size of RPC response messages (uncompressed).                                | Histogram |
 | `rpc_server_responses_per_rpc`    | Measures the number of messages sent per RPC. Should be 1 for all non-streaming RPCs.     | Histogram |
 
+### Telemetry lifecycle
+
+#### Tracing
+
+Tracing instrumentation is still under active development, and may change in a variety of ways. This includes span names, attached attributes, instrumented endpoints, etc. Until this feature graduates to stable, there are no guarantees of backwards compatibility for tracing instrumentation.
```
tiffany76, line 267:
```suggestion
Tracing instrumentation is still under active development, and changes might be made to span names, attached attributes, instrumented endpoints, or other aspects of the telemetry. Until this feature graduates to stable, there are no guarantees of backwards compatibility for tracing instrumentation.
```

content/en/docs/collector/internal-telemetry.md
```diff
@@ -260,6 +260,53 @@ categorized by instrumentation type.
...
+
+#### Metrics
+
+The lifecycle of metrics in the Collector follows the following cycle, each stage is documented below:
```
tiffany76, line 271:
```suggestion
The Collector's metrics follow a four-stage lifecycle:
```

content/en/docs/collector/internal-telemetry.md
```diff
@@ -260,6 +260,53 @@ categorized by instrumentation type.
...
+##### Alpha
+
+Alpha metrics have no stability guarantees. These metrics can be modified or deleted at any time.
+
+##### Stable
+
+Stable metrics are guaranteed to not change. This means:
+
+* A stable metric without a deprecated signature will not be deleted or renamed
+* A stable metric's type and attributes will not be modified
```
tiffany76, line 284:
```suggestion
- A stable metric without a deprecated signature will not be deleted or renamed.
- A stable metric's type and attributes will not be modified.
```

content/en/docs/collector/internal-telemetry.md
```diff
@@ -260,6 +260,53 @@ categorized by instrumentation type.
...
+
+    # HELP some_counter this counts things
+    # TYPE some_counter counter
+    otelcol_exporter_queue_size 0
+
+    After deprecation
+
+    # HELP some_counter (Deprecated since 1.15.0) this counts things
+    # TYPE some_counter counter
+    some_counter 0
```
tiffany76, line 300: This should help with the linter issue.

```suggestion
```sh
Before deprecation

# HELP some_counter this counts things
# TYPE some_counter counter
otelcol_exporter_queue_size 0

After deprecation

# HELP some_counter (Deprecated since 1.15.0) this counts things
# TYPE some_counter counter
some_counter 0
```
```

File: content/en/docs/collector/internal-telemetry.md [hunk 5]
```diff
@@ -260,6 +260,53 @@ categorized by instrumentation type.
...
+
+##### Deleted
+
+Deleted metrics are no longer published and cannot be used.
+
+#### Logs
+
+Individual log entries and their formatting may change from one release to the next, there are no stability guarantees at this time.
```
tiffany76, line 308:
```suggestion
Individual log entries and their formatting might change from one release to the next. There are no stability guarantees at this time.
```

content/en/docs/collector/internal-telemetry.md
```diff
@@ -260,6 +260,53 @@ categorized by instrumentation type.
...
+##### Stable
+
+Stable metrics are guaranteed to not change. This means:
+
+* A stable metric without a deprecated signature will not be deleted or renamed
+* A stable metric's type and attributes will not be modified
+
+##### Deprecated
+
+Deprecated metrics are slated for deletion, but are still available for use. These metrics include an annotation about the version in which they became deprecated. For example:
```
tiffany76, line 288:
```suggestion
Deprecated metrics are slated for deletion but are still available for use. These metrics include an annotation about the version in which they became deprecated. For example:
```

tiffany76 (commented):
Thanks for putting this together! I've made a few inline suggestions.

I also think it makes sense to shift this section up a hierarchical level, so the first heading would be "## Telemetry lifecycle". And then we can position it below [Events observable with internal logs](https://opentelemetry.io/docs/collector/internal-telemetry/#events-observable-with-internal-logs), so we don't break up the Types of internal telemetry section. Happy to discuss.

content/en/docs/collector/internal-telemetry.md
```diff
@@ -260,6 +260,53 @@ categorized by instrumentation type.
 | `rpc_server_response_size`        | Measures the size of RPC response messages (uncompressed).                                | Histogram |
 | `rpc_server_responses_per_rpc`    | Measures the number of messages sent per RPC. Should be 1 for all non-streaming RPCs.     | Histogram |
 
+### Telemetry lifecycle
```
chalin, line 263: Hmm, what about clarifying the kind of lifecycle? Also, IMHO it isn't a cycle per se, it's a maturity path. Maybe just drop "lifecycle"?

The official terminology from https://github.com/open-telemetry/oteps/blob/main/text/0232-maturity-of-otel.md#maturity-levels is "maturity levels":

So, how about this:

```suggestion
### Telemetry maturity levels
```

Btw, should this instead be the following instead?

```suggestion
### Signal maturity levels
```

content/en/docs/collector/internal-telemetry.md
chalin: If you do you fenced code notation, the language should be something other than `sh` IMHO. Maybe just `text`?

Also, is it meant to be all in a code block? Could it instead be what is shown below:

---

Before deprecation:

```text
# HELP some_counter this counts things
# TYPE some_counter counter
otelcol_exporter_queue_size 0
```

After deprecation:

```text	
# HELP some_counter (Deprecated since 1.15.0) this counts things
# TYPE some_counter counter
some_counter 0
```

content/en/docs/collector/internal-telemetry.md
codeboten: I wouldn't want to use Signal here as it may be confused with the signals that the collector receives/processes/exports

content/en/docs/collector/internal-telemetry.md
```diff
@@ -260,6 +260,57 @@ categorized by instrumentation type.
 | `rpc_server_response_size`        | Measures the size of RPC response messages (uncompressed).                                | Histogram |
 | `rpc_server_responses_per_rpc`    | Measures the number of messages sent per RPC. Should be 1 for all non-streaming RPCs.     | Histogram |
 
+### Telemetry maturity levels
+
+#### Tracing
```
tiffany76, line 265:
```suggestion
#### Traces
```

tiffany76:
Reviving my comment about the information architecture, I think this section needs to be shifted one hierarchical level higher and positioned **below** the [Events observable with internal logs](https://deploy-preview-5092--opentelemetry.netlify.app/docs/collector/internal-telemetry/#events-observable-with-internal-logs) section. 

Right now, it is inserted into the Types of internal telemetry section. I think it works better as its own ##-level section, right before Use internal telemetry to monitor the Collector.

Screenshot from 2024-09-09 17-06-39

content/en/docs/collector/internal-telemetry.md
```diff
@@ -260,6 +260,57 @@ categorized by instrumentation type.
...
+
+Tracing instrumentation is still under active development, and changes might be made to span names, attached attributes, instrumented endpoints, or other aspects of the telemetry. Until this feature graduates to stable, there are no guarantees of backwards compatibility for tracing instrumentation.
+
+#### Metrics
+
+The Collector's metrics follow a four-stage lifecycle:
+
+> Alpha metric → Stable metric → Deprecated metric → Deleted metric
```
mx-psi, line 273: How does a user identify the stage in which a metric is at?

codeboten: This could be included in the documentation generated for each metric via the metadata file, wdyt?

mx-psi: Yeah that makes sense. I would also like to have the stage in the metrics description

content/en/docs/collector/internal-telemetry.md
```diff
@@ -260,6 +260,57 @@ categorized by instrumentation type.
...
+- A stable metric's type and attributes will not be modified.
+
+##### Deprecated
+
+Deprecated metrics are slated for deletion but are still available for use. These metrics include an annotation about the version in which they became deprecated. For example:
+
+Before deprecation:
+
+```sh
+# HELP otelcol_exporter_queue_size this counts things
```
bixu, line 293: Do we only consider Prometheus metrics or is there also a convention we want if deprecating an Otel metric?

codeboten: I'd expect the description of the metric to include a note about deprecation. The description will be rendered differently based on the exporter configured

content/en/docs/collector/internal-telemetry.md
```diff
@@ -260,6 +260,57 @@ categorized by instrumentation type.
...
+##### Stable
+
+Stable metrics are guaranteed to not change. This means:
+
+- A stable metric without a deprecated signature will not be deleted or renamed.
+- A stable metric's type and attributes will not be modified.
+
+##### Deprecated
+
+Deprecated metrics are slated for deletion but are still available for use. These metrics include an annotation about the version in which they became deprecated. For example:
```
codeboten, line 288:
```suggestion
Deprecated metrics are slated for deletion but are still available for use. The description of these metrics include an annotation about the version in which they became deprecated. For example:
```

content/en/docs/collector/internal-telemetry.md
codeboten: Would you expect to see all the stages in the description or would it be enough to include it only in the deprecation stage? I wouldn't want to end up sending unnecessary prefixes for every metric at every stage.... maybe any stage that isn't stable?

mx-psi: Only on those that are not stable IMO

codeboten: Opened a draft PR to start work on this here: https://github.com/open-telemetry/opentelemetry-collector/pull/11160

mx-psi (approved):
I think this looks good to me for a first version. At some point I would like to define how the transition between states works, but I don't think we need that right now

codeboten:
> Right now, it is inserted into the Types of internal telemetry section. I think it works better as its own ##-level section, right before Use internal telemetry to monitor the Collector.

@tiffany76 I don't have a strong preference here, happy to move this if it will move this PR along

tiffany76:
> I don't have a strong preference here, happy to move this if it will move this PR along

@codeboten You can move it in this PR, or I can follow-up with an IA PR once this one is merged. It's up to you.

tiffany76 (approved):
Approving and setting this as "ready to be merged." Once a docs maintainer gets to it, I'll follow up with a PR to move the section. No need to hold this one up.

PR #5100, opened 2024-08-25
Fix name of jaeger_remote polling interval property

drewhammond:
This PR simply corrects the name of the polling interval property used by the `jaeger_remote` and `parentbased_jaeger_remote` samplers.

Ref: https://github.com/open-telemetry/opentelemetry-java/blob/c90175f1ae7d15fef6bac8eb6c95492c981daa61/sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/JaegerRemoteSamplerProvider.java#L22

### DIFF
--- a/content/en/docs/languages/sdk-configuration/general.md
+++ b/content/en/docs/languages/sdk-configuration/general.md
@@ -92,11 +92,11 @@ be set as follows:
 - For `jaeger_remote` and `parentbased_jaeger_remote`: The value is a comma
   separated list:
   - Example:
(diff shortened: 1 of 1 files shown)


theletterf:
Thanks, @drewhammond !

theletterf (approved):

drewhammond:
After discussion in https://github.com/open-telemetry/opentelemetry-specification/pull/4195, I think it makes sense to revert this doc change and fix the java implementation instead.

PR #5110, opened 2024-08-27
Update integrations.md

tedder:
fix tense. Yeah, single-character commit.


theletterf (approved):
One char commits are entirely fine. :) Thanks!

PR #5116, opened 2024-08-28
Mention sync gauge in Otel Go metrics docs

cijothomas:
Sync gauge was not listed as supported, but the later section of the doc show that it is indeed supported.


tiffany76 (approved):

MrAlias (approved):

chalin:
All checks are green. Merging after I test the new update to the `fix:*` workflow.

chalin (approved):

PR #5122, opened 2024-08-29
[CI] After a `fix:*` command, tell user to rerun full checks

chalin:
- A step towards aleveating the pain of #4339
- Reports successful runs of `fix:*` commands, and instructs the user to run `fix:all` to ensure that all checks are passing. This isn't a perfect solution, but it will hopefully help avoid issues like https://github.com/open-telemetry/opentelemetry.io/pull/5119#issuecomment-2318012101.

/cc @cartermp @svrnm

### DIFF
--- a/.github/workflows/pr-actions.yml
+++ b/.github/workflows/pr-actions.yml
@@ -101,9 +101,16 @@ jobs:
         env:
           GH_TOKEN: ${{ secrets.OPENTELEMETRYBOT_GITHUB_TOKEN }}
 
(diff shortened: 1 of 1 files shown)


chalin:
I don't know if running a `fix:*` command over this PR will use the script in the PR (I'm thinking that it won't), but it's worth a try.

chalin:
It doesn't seem like it. So we might just have to merge this PR and test it afterwards.

tiffany76 (approved):
RSLGTM

PR #5130, opened 2024-08-30
Replace "dynamically injects bytecode" in Python zero-code instrumentation

CFly17:
Update/clarify the automatic instrumentation method for Python zero-code instrumentation

- replace "dynamically injects bytecode" in Introduction section with a brief description of monkey patching
- add details to Setup section about how the `opentelemetry-bootstrap -a install` command uses monkey patching


cartermp (approved):

cartermp:
@open-telemetry/python-approvers PTAL!

content/en/docs/zero-code/python/_index.md
```diff
@@ -37,7 +38,10 @@ packages installed in your active `site-packages` folder, and installs the
 corresponding instrumentation libraries for these packages, if applicable. For
 example, if you already installed the `flask` package, running
 `opentelemetry-bootstrap -a install` will install
-`opentelemetry-instrumentation-flask` for you.
+`opentelemetry-instrumentation-flask` for you. The OpenTelemetry Python agent
+will use
+[monkey patching](https://docs.pytest.org/en/stable/how-to/monkeypatch.html#monkeypatching-functions)
```
xrmx, line 43: I would like to wikipedia article about monkey patching instead. We are not using `monkeypatch`.

xrmx (changes requested):

content/en/docs/zero-code/python/_index.md
emdneto: Also, I would add the reference to the first time where it was mentioned

static/refcache.json
```diff
@@ -1711,6 +1711,10 @@
     "StatusCode": 200,
     "LastSeen": "2024-01-18T19:55:51.433612-05:00"
   },
+  "https://docs.pytest.org/en/stable/how-to/monkeypatch.html#monkeypatching-functions": {
```
emdneto, line 1714: Should update this too?

xrmx (approved):
With the refcache fixed LGTM

static/refcache.json
xrmx, line 1714: I think this chunk is not needed, looks like fixrefcache only adds entries

CFly17: Thanks for the review @xrmx. It means a lot; this is my first contribution to open source. 

I was unfamiliar with comment-triggered actions like `/fix:refcache`. Should only maintainers do this?

I am also reading the [contributing process](https://github.com/open-telemetry/community/blob/main/guides/contributor/processes.md) and wonder if there a 'checklist' to clean up my local before committing, e.g. `npx prettier --write`?

svrnm: Hey @CFly17! Thanks for your contribution, happy to read that you choose otel for being your first!

The `/fix:*` commands are just automations we use run the `npm fix:*` commands without pulling your PR, running it locally and pusing it back. You may want to run `npm fix:all` before pushing your PR, but since we sometimes do inline changes, etc. we have no issues with a PR that has failing CI checks. The thing that counts is that those checks are green shortly before merging the PR, and this can either be done by you or by a maintainer. 

Note, that this is how we (OTel Docs Maintainer) are doing it, other OSS projects might choose a different approach

PR #5141, opened 2024-09-03
correct metric type for http req/res sizes

jamesmoessis:
It took me a while to find out these docs were wrong, reading the OTLP receiver code, then the obs reporter helper, then the opentelemetry-go-contrib which holds the instrumentation for the HTTP servers. Thought I would save someone else the time.

These are counters, not histogram. Here is the code that records this: https://github.com/open-telemetry/opentelemetry-go-contrib/blob/18eaff6c5acc512a4b53748c78bda978844c4fe4/instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go#L154 

Interestingly, the gRPC code uses a histogram: https://github.com/open-telemetry/opentelemetry-go-contrib/blob/18eaff6c5acc512a4b53748c78bda978844c4fe4/instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go#L154

### DIFF
--- a/content/en/docs/collector/internal-telemetry.md
+++ b/content/en/docs/collector/internal-telemetry.md
@@ -242,13 +242,13 @@ categorized by instrumentation type.
 | `http_client_active_requests`     | Number of active HTTP client requests.                                                    | Counter   |
 | `http_client_connection_duration` | Measures the duration of the successfully established outbound HTTP connections.          | Histogram |
 | `http_client_open_connections`    | Number of outbound HTTP connections that are active or idle on the client.                | Counter   |
-| `http_client_request_size`        | Measures the size of HTTP client request bodies.                                          | Histogram |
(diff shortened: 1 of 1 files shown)


theletterf (approved):
@open-telemetry/collector-approvers Please have a look.

jpkrohling (approved):
LGTM. IIRC, there's an issue open with the underlying Go SDK to make this a histogram, so this might get outdated soon again (to the state before this PR), but this PR reflects the current situation of the Collector.

jamesmoessis:
Thankyou @jpkrohling!

jamesmoessis:
For what it's worth, a counter here is more useful for me because I want to know total bytes rather than distribution over requests. I can see use cases for both. This is probably not the right place to discuss it, though.

PR #5142, opened 2024-09-03
chore(docs): Update getting started to not use Bandit as it is not re…

pdgonzalez872:
…ady. Reverts #4518, per https://github.com/open-telemetry/opentelemetry.io/pull/4518#issuecomment-2326950585


bryannaegele (approved):

content/en/docs/languages/erlang/getting-started.md
```diff
@@ -51,15 +51,9 @@ We'll need a few other dependencies that Phoenix doesn't come with.
...
-  to Bandit while pre 1.7.11 default to Cowboy. Both choices are valid. Use one
-  of the below options based on the web server your Phoenix application uses:
-  - `opentelemetry_cowboy`: creates OpenTelemetry spans from the Elixir
-    `:telemetry` events created by the Cowboy web server
-  - `opentelemetry_bandit`: creates OpenTelemetry spans from the Elixir
-    `:telemetry` events created by the Bandit web server
-
+- `opentelemetry_cowboy`: creates OpenTelemetry spans from the Elixir
+  `:telemetry` events created by the Cowboy web server (which is used by
+  Phoenix).
```
tiffany76, line 56:
```suggestion
  `:telemetry` events created by the Cowboy web server, which is used by
  Phoenix.
```

tiffany76 (commented):
One small change.

Also, could you confirm the reference to "the last three" in Line 81 is still accurate?

pdgonzalez872:
@tiffany76 thanks for helping! The linter seems to be super helpful, neat!

When I added this a couple months ago, seems like it was 2 only (line 75 below). 
image


But it seems like it got introduced [here](https://github.com/open-telemetry/opentelemetry.io/commit/65c06ed8223f25b906a84be0e7d3669bd9846288#diff-312f8408ae4ee5cf752780f6f858e8ada359b577edd86a90981cf4d6ab6cdf1dR81), so I'll likely defer to folks that know more about this?

Thanks again for helping.

pdgonzalez872:
@tiffany76, it's only two things being set up, I'm pretty sure two is the correct. Pushed an update for that.

I'm not sure if you all squash commits automatically or not, if not, just please let me know and I'll do it.

Thanks!

tiffany76 (approved):
LGTM! Thanks @pdgonzalez872! The commits will be squashed when the maintainers merge into main.

bryannaegele:
Thank you @pdgonzalez872 !

PR #5150, opened 2024-09-04
Add MLflow entry to the OpenTelemetry integrations registry

daniellok-db:
[MLflow](https://github.com/mlflow/mlflow/) has a tracing feature built using OpenTelemetry, and we'd like to add an entry to the registry.

### DIFF
--- a/data/registry/application-integration-python-mlflow.yml
+++ b/data/registry/application-integration-python-mlflow.yml
@@ -1,0 +1,20 @@
+title: MLflow
+registryType: application integration
+language: python
+tags:
+  - python
+  - MLflow
+  - LLMs
+license: Apache-2.0
+description:
+  The MLflow tracing feature is built using OpenTelemetry, and allows
+  users to trace the execution of their generative AI applications.
+authors:
+  - name: MLflow Authors
+    url: https://github.com/mlflow
+urls:
+  website: https://mlflow.org/
+  repo: https://github.com/mlflow/mlflow/
+  docs: https://mlflow.org/docs/latest/llms/tracing/index.html
+createdAt: '2024-09-04'
+isNative: true


theletterf:
@daniellok-db Please sign the CLA, thanks!

daniellok-db:
Signed!

svrnm (changes requested):
The documentation does not mention OpenTelemetry at all (I only see it in an error) and I also do not see a way to export data into a 3rd party solution, it seems all to be included within MLFlow, is this correct?

cc @open-telemetry/semconv-llm-approvers FYI

svrnm:
> Let me edit the docs and request another review after that! 

That would be great!

> At the moment the trace data is saved in a JSON file that the user can use however they want, but we haven't described this in the documentation yet.

That's fairly unusual, is there a particular reason why you do not make use of exporters that allows people to extract telemetry into the tool of their choice?

nirga:
@daniellok-db looks like you guys instrument otel, but [you're not using semantic conventions](https://github.com/mlflow/mlflow/blob/9af1b29effbcb48cf8dfd233fd7a14d72b344b54/mlflow/llama_index/tracer.py#L332). Do you have plans to switch over to semantic conventions?

daniellok-db:
@svrnm we're looking into adding an exporter to allow users to export it to wherever they wish, but at the moment it hasn't been implemented

@nirga i think we weren't aware of the semantic conventions at the time we implemented, but migrating to use them is definitely something we're thinking about!

Thanks all for the comments! This is a relatively new feature for us and I think there are still a number of gaps to be filled before we can really be considered an integration 😓. Let me close this for now and reopen when we've addressed these!

svrnm:
> Thanks all for the comments! This is a relatively new feature for us and I think there are still a number of gaps to be filled before we can really be considered an integration 😓. Let me close this for now and reopen when we've addressed these!

Don't worry! It's great to see that MLFlow makes the effort to integrate OpenTelemetry into their project. Take a look at the semantic conventions that @nirga called out, you will massively profit from that and also take a look what other projects provide in the GenAI/LLM space around OpenTelemetry. 

Note, that we do not have super strict requirements what an integration needs to provide, but for user expectations it would be good if your documentation calls out that it is build around OpenTelemetry and that data can be exported via OTLP if needed. So if you add that to your docs + project, happy to take another PR :-)

PR #5173, opened 2024-09-10
Fix wording in index.md

monteiro-renato:
English is not my mother tongue so there might be a better way to word this still.
Feel free to suggest a better way.
Just wanted to raise awareness.


tiffany76 (approved):
Your wording sounds good to me. Thanks!

Issue #5168, opened 2024-09-09
Demo Architecture Diagram shows "unsupported markdown"


candlerb:
Suggested improvements for page: https://opentelemetry.io/docs/demo/architecture/

There is a rendering problem in the images, where there are currently lots of URLs shown as `"Unsupported markdown: link"` - both inside nodes and on edges. This makes the images hard to interpret.

Examples:

![image](https://github.com/user-attachments/assets/c5dfe83e-2678-483e-8369-217434f800c4)

The text inside the graph source is much clearer, e.g.

```
            pr-http[/"Prometheus HTTP<br/>listening on<br/>http://localhost:9090"/]
```

svrnm:
@open-telemetry/demo-approvers PTAL>

@candlerb thanks for reporting

julianocosta89:
@svrnm has anything changed in the docs?
I'm pretty sure this was working before 🤔

theletterf:
[...]

PR #5175, opened 2024-09-10
Update architecture.md to remove incompatible link notation from diagram

theletterf:
Fixes https://github.com/open-telemetry/opentelemetry.io/issues/5168

### DIFF
--- a/content/en/docs/demo/architecture.md
+++ b/content/en/docs/demo/architecture.md
@@ -128,8 +128,8 @@ subgraph tdf[Telemetry Data Flow]
 
         subgraph oc[OTel Collector]
             style oc fill:#97aef3,color:black;
-            oc-grpc[/"OTLP Receiver<br/>listening on<br/>grpc://localhost:4317/"/]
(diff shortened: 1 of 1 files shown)


mviitane (approved):
LGTM

mviitane:
Not directly related to this PR, but this Telemetry Data Flow image looks too large compared to the Service Diagram (on the same page). Is there any setting for that?

theletterf:
@mviitane We'd only control that through CSS classes, I'm afraid. Could you post a screenshot and your screen res?

mviitane:
> @mviitane We'd only control that through CSS classes, I'm afraid. Could you post a screenshot and your screen res?

I have 2560 x 1440 screens, but in this screenshot, I'm using a somewhat narrow browser window. The font in the lower picture is much bigger than the upper pic (similar font size would work better). Also, the lower picture is so high that I cannot fit it on the screen at once.
Screenshot 2024-09-10 at 12 49 55

monteiro-renato:
FYI: Looks like it's not completely fixed yet:
bilde

tiffany76 (approved):

theletterf:
Let's merge till the bug gets fixed in Mermaid.

PR #5176, opened 2024-09-10
Fix the endering of http urls in architecture.md's mermaid diagram

monteiro-renato:
Hey,

The diagram here is currently broken > https://opentelemetry.io/docs/demo/architecture/

What I did (and I think this is probably a terrible idea) was adding a Zero-width space before the colon in the urls.

I also tested adding a <span> and it also allows the rending to take place.

I'm not super familiar with Mermaid so there might exist a better way to do this (I hope so at least).

Raising this for awareness, if you are aware of an easy fix let me know and I can fix it quickly. Otherwise I will try to look into it whenever I have time available.


theletterf:
@monteiro-renato Please check https://github.com/open-telemetry/opentelemetry.io/issues/5168

PR #5266, opened 2024-09-28
Contributing: improved guidance for quick-fixers, mentees, and more

chalin:
- "Contributing" landing page - call out / cater to:
  1. "quick-fixers" (for typos)
  2. first-timers (read "mentees"), and 
  3. the rest of the contributors
- Recovers the first-timers' instructions that didn't make the restructuring of the `CONTRIBUTING.md1 page (from here https://github.com/open-telemetry/opentelemetry.io/pull/4654/files#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055L50), which emphasized that:
  1. Look for "good first issue" issues
  2. We don't cater to "please assign to me" comments
- Some page and title rename & rework

PR #5271, opened 2024-09-30
Fix typo in Contributing / issues.md

chalin:
- Followup to #5266

### DIFF
--- a/content/en/docs/contributing/issues.md
+++ b/content/en/docs/contributing/issues.md
@@ -44,7 +44,7 @@ One of the best ways to help make OTel docs better is to fix an existing issue.
    need for clarification by posting comments over the issue.
 5. Share your intention to work on the issue by adding add a comment to this
    effect.
-6. Work on fixing the issue. Let maintainers now if you run into any problems.
+6. Work on fixing the issue. Let maintainers know if you run into any problems.
 7. When ready, [submit your work through a pull request](../pull-requests) (PR).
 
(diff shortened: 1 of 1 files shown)


content/en/docs/contributing/issues.md
```diff
@@ -44,7 +44,7 @@ One of the best ways to help make OTel docs better is to fix an existing issue.
    need for clarification by posting comments over the issue.
 5. Share your intention to work on the issue by adding add a comment to this
    effect.
-6. Work on fixing the issue. Let maintainers now if you run into any problems.
+6. Work on fixing the issue. Let maintainers know if you run into any problems.
```
chalin, line 47: @open-telemetry/docs-approvers - regarding steps 5 & 6: should we explicitly say that contributors don't have to wait for our thumbs up, or should they wait for at least a thumbs up from a docs approver?

theletterf: I'd encourage them not to wait.

chalin: That works for me. In that case, I'd leave the step prose as is for now (by not stating that contributors have to wait, they have to liberty of moving on from steps 5 to step 6 directly -- we can always make it explicit later if the need arises).

chalin: So, shall we merge this typo fix as is?

theletterf (approved):

PR #5272, opened 2024-09-30
API reference: use /api as path and config as redirect

### LINKED ISSUE #5261: Implement `/api-docs` as a redirect and possibly change path
Given that `/api-docs` is just a covenience, as @svrnm mentioned, and give that the same info appears in https://opentelemetry.io/docs/languages/#api-references, I'd like to propose that we keep things DRY and create a simple redirect rule instead of a separate page.

Also, I think that path of the redirect should be either `/docs/api`, or just `/api`. WDYT?

### LINKED ISSUE #5274: API reference list: missing path
The Tip prose at https://opentelemetry.io/docs/languages/#api-references current reads: "Tip
You can find a list of available API references at ."

> image

That is, the path is missing.

chalin:
- Fixes #5261
- Fixes #5274

### Preview and redirect tests

- https://deploy-preview-5272--opentelemetry.netlify.app/api
- https://deploy-preview-5272--opentelemetry.netlify.app/api-docs


chalin:
@svrnm et al. - before merging, pls comment on https://github.com/open-telemetry/opentelemetry.io/issues/5261#issuecomment-2383067092, though we could handle that separately too.

svrnm (approved):

PR #5282, opened 2024-10-01
Update gateway.md

muralikanagala:
NGINX config example is using only collector1 and collector2


muralikanagala:
There is a typo which I fixed in another PR

PR #5339, opened 2024-10-07
Fix telemetry config example for public interfaces

rrjjvv:
The original version results in error:
```
error decoding 'service.telemetry': decoding failed due to the following error(s):

'metrics.readers': source data must be an array or slice, got map
```

Converting that inner element to an array, similar to other examples in the page, results in the telemetry working again.

### DIFF
--- a/content/en/docs/collector/internal-telemetry.md
+++ b/content/en/docs/collector/internal-telemetry.md
@@ -36,11 +36,11 @@ service:
   telemetry:
     metrics:
       readers:
-        pull:
-          exporter:
-            prometheus:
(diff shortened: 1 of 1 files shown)


tiffany76:
@open-telemetry/collector-approvers, PTAL. Thanks.

tiffany76:
Hi @rrjjvv, with apologies, it appears this fix has already been made in [another PR](https://github.com/open-telemetry/opentelemetry.io/pull/5397). We sincerely thank you for your efforts and hope you will contribute again.

Issue #5317, opened 2024-10-04
[outreachy] Registry: Put items into sections (plugins, libraries, instrumentation, ...)


svrnm:
Outreachy task for @beccamak

Based on your user experience report, please provide the following suggested improvement via a PR:

> we can put the page into sections(plugins, libraries,instrumentation ...) whilst we leave the search bar and filter, and then add like a view all link or button to the different sections.

**Note**: This is an assigned issue for a participant that has completed the user experience report, if you are an outreachy participant as well, find instructions here: https://github.com/open-telemetry/opentelemetry.io/issues/5290

Beccamak:
@svrnm Thank you. I'm on it.

PR #5348, opened 2024-10-08
Put items into sections

Beccamak:
#5317 
@svrnm I have created sections for the major categories on the registry page. 
For view all links, I had to use the search functionality that was already implemented. Initially, I thought that on the backend, there was an identifier or tag for entries that easily identifies them as plugins, libraries, or instrumentations. The plan was to have all these categories filtered out and then on clicking "view all" they are populated. 

However, after some interactions with the codebase, I didn't see that there. I noticed that if I had to get a category out (for example, the libraries), it had to be through a search (the fuzzy search the system is using), and I couldn't find a way of passing the filtered results back to the template,  which I think explains why the whole entries were initially inserted on the registry page. The keys were compared to filter out the keys (entries) that didn't match the keys (entries) in the returned search results on the DOM. 

Kindly let me know your thoughts on this.


content/en/ecosystem/registry/_index.md
svrnm, line 1: this change is unrelated, please undo it

layouts/partials/ecosystem/registry/entry.html
```diff
@@ -2,7 +2,7 @@
 
 {{ with .value -}}
     {{ 
-        $remoteRegistries := dict
+        $remoteRegistries := dict
```
svrnm, line 5: revert this unrelated change

layouts/shortcodes/ecosystem/registry/search-form.html
```diff
@@ -108,3 +109,10 @@
 
 <script src="https://cdn.jsdelivr.net/npm/minisearch@7.1.0/dist/umd/index.min.js" integrity="sha384-58p1D2xuw/76VRFUgDMRcrtnJAW5LWLE02ihn8aITu65c4vgo+BLy7raRKNC7KaL" crossorigin="anonymous"></script>
 {{ partial "script.html" (dict "src" "js/registrySearch.js") -}}
+
+
+
+
+
+
+
```
svrnm, line 118: remove

layouts/shortcodes/ecosystem/registry/search-form.html
```diff
@@ -92,13 +92,14 @@
   </form>
 </div>
 
+
```
svrnm, line 95: remove

svrnm (changes requested):
This is a really good start, thank you.

A few inline comments. Most of them are related to changes you have applied to files that shouldn't be there. You might be able to revert all of them by running `npm run fix:format` locally.

A few more comments:
* I understand why you did it that way (we say "libraries, plugins, integrations"), but by applying a search on those items you get a selection of items that not always fits the description. This is OK for now, but we need to think about addressing that differently. I will think about that
* The links at the bottom of an item are off, can you take a look that they are in line with the rest of the element?
* If I make the screen size smaller the elements become very thin, I think it should switch to 1 item per row with an earlier width-selector

content/en/ecosystem/registry/_index.md
Beccamak: @svrnm Thank you for taking the time to review and for your feedback. 

- Just to be sure I understand you well, you said, "...but by applying a search on those items you get a selection of items that not always fits the description. This is OK for now, but we need to think about addressing that differently...", were you referring to the already existing search functionality?
- For the links, I'll take a look at them and make fixes as soon as possible.
- Sincere apologies for the responsiveness, I only checked on Desktop and mobile. I'll work on that too.
- Also, for [content/en/ecosystem/registry/_index.md](https://github.com/open-telemetry/opentelemetry.io/pull/5348/files/997d3c550e9c79dadcfda2f9a6a2c74c821d9522#diff-a6511056b88cb6dd90f17707ea99f2de5886b5088aff728c3cc0c6c23849f52f) you said, "this change is unrelated, please undo it", do you mean I should delete this "{{< ecosystem/registry/categories>}}?

svrnm: >     * Just to be sure I understand you well, you said, "...but by applying a search on those items you get a selection of items that not always fits the description. This is OK for now, but we need to think about addressing that differently...", were you referring to the already existing search functionality?

What I tried to express is the following: you are currently applying searches for "plugin", "library", etc to generate the catagories. Which is great because it matches the description above. But the underlying data is not matching that, so we need to address that, otherwise the lists per category are not matching what end users might expect. I'll think about that and come back with a suggestion

>     * Sincere apologies for the responsiveness, I only checked on Desktop and mobile. I'll work on that too.

No need to apologize, this is a normal review, where another person is calling out places to improve/requirements that should be met. This is great progress!

>     * Also, for [content/en/ecosystem/registry/_index.md](https://github.com/open-telemetry/opentelemetry.io/pull/5348/files/997d3c550e9c79dadcfda2f9a6a2c74c821d9522#diff-a6511056b88cb6dd90f17707ea99f2de5886b5088aff728c3cc0c6c23849f52f) you said, "this change is unrelated, please undo it", do you mean I should delete this "{{< ecosystem/registry/categories>}}?

Oh no, I marked too much, remove the additional empty lines (should be fixed via the fix:format)

Beccamak: @svrnm Thank you, this is well understood. I'll be looking forward to your feedback.

Beccamak:
@svrnm I trust you're doing fine. Happy weekend.

I have made the changes you requested. I noticed some of those cards weren't displaying the quick install code so I worked on it. Also, would you advise that all the cards being displayed have a "quick install" to ensure all components are aligned together? 

This is also a gentle reminder for the suggestion regarding the "search that returns items that do not always fit the description".

svrnm:
Thank you @beccamak! This looks really good, I consider this as done! 

The changes required for making this work with the current solution is out of scope for your work. When the application phase for outreachy is over, we can revisit this and talk about making this work, if you want to, but since we do not merge PRs until the end of the application phase, let's keep it as is.

Beccamak:
@svrnm  Thank you so much. This is well noted and understood. I've learned a lot by just trying to implement some of those things. I'll move on to selecting an optional task now. 
Thank you so much for your time.

svrnm:
While I still like the idea of this PR, I am yet unsure how to incorporate it. The categories "Libraries, Plugins, Integrations" do not work, since they do not match any of the categories we have as of today, but we could do "Instrumentation Libraries, "Collector Components", "Application Integrations" and a few other ones more fitting. This would also allow us to add some words and explaining them better.

cc @open-telemetry/docs-approvers WDYT?

Beccamak:
@svrnm Yes, I'm still interested in helping out.

svrnm:
A few thoughts on that:
1. we should feature Instrumentation Library, Collector Components and maybe Utilities. Each one of them should also have a small top paragraph letting the end user know what will get from them, what they are etc (I can help with that)
2. the items for section should be minimal, such that it is visible to a user that there are multiple sections. maybe only the name, the description and a "more" link that can expand the details?

Beccamak:
@svrnm Looking at the first suggestion, we want to have a small top paragraph for Instrumentation Library, Collector Components and Utilities letting the end user know what they will get from them, what they are etc... right?

So that would come before the search bar?

svrnm:
@Beccamak any thoughts on my previous comment?

PR #5363, opened 2024-10-09
Update custom-collector.md

jackgopack4:
update providers in builder-config.yaml to match stable versions.

Per https://github.com/open-telemetry/opentelemetry-collector/releases/tag/v0.111.0, more `providers` are marked as stable. Updating custom collector documentation to use the latest stable components by default

### DIFF
--- a/content/en/docs/collector/custom-collector.md
+++ b/content/en/docs/collector/custom-collector.md
@@ -166,11 +166,11 @@ receivers:
       go.opentelemetry.io/collector/receiver/otlpreceiver {{% version-from-registry collector-receiver-otlp %}}
 
 providers:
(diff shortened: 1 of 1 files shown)


cartermp (approved):

svrnm:
I just wanted to suggest that we use the versions from the registry for those entries as well, but then I saw that the providers are not in the registry. I wonder if we should add them?

svrnm:
> I think confmap providers should be added to the registry: they are conceptually similar to other kinds of components in the Collector in that they provide extra functionality to the Collector, can be added on a Collector builder manifest, can be 'official' or come from a third-party repository...

then let's do that, in the meantime, would you mind approving this PR, so we have it up-to-date and then we add them to the registry?

mx-psi (approved):

PR #5377, opened 2024-10-10
fix: add qoutation marks around bash variables

### LINKED ISSUE #5359: Build fails when within a folder with spaces
**What needs to be changed?**
Add a note that `npm run build` will fail if the cloned repo is within a folder that has spaces

content/en/docs/contributing/development.md

**Additional context**:
This might only be an issue if you run the build command within a terminal that you've changed to (cd into). That is, if you are within the parent folder and you cd'd into the repo via the terminal then run `npm run build` from there.

FatumaA:
Added quotation marks around variables. 
Builds but leaves lots of warnings:

```
WARN  docs/specs/otel/common/_index.md: use a local path, not an external URL, for the following reference to a local specification page: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/attribute-naming.md
WARN  docs/specs/otel/common/_index.md: use a local path, not an external URL, for the following reference to a local specification page: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/attribute-requirement-level.md
WARN  docs/specs/otel/common/attribute-naming.md: use a local path, not an external URL, for the following reference to a site local page: https://opentelemetry.io/docs/specs/semconv/general/attribute-naming/
WARN  docs/specs/otel/common/attribute-requirement-level.md: use a local path, not an external URL, for the following reference to a site local page: https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/
```

Fixes #5359


chalin (changes requested):
Hi @FatumaA. Thanks for this attempt at a fix, but this breaks the build rather than fixing it. As you've noted already, the warnings (which are a symptom of the build not working), were not there before. https://github.com/open-telemetry/opentelemetry.io/actions/runs/11372802351/job/31638107804?pr=5377

Let me take a look at a fix and report back here.

PR #5406, opened 2024-10-15
make language an enum in registry schema

### LINKED ISSUE #1355: Registry: validate language and component-type fields
The registry allows folks to add languages and component types on the fly. I think that's fine, but we need a way to ensure that this addition doesn't slip by unnoticed by reviewers.

Adding a cross-check against a fixed list/text file might be all we need. For example we could track the output of the following:

```console
$ cd opentelemetry.io 
$ find content/en/registry/*.* -exec grep -e 'registryType:' {} \; | sort | uniq
registryType: core
registryType: exporter
registryType: extension
registryType: instrumentation
registryType: processor
registryType: receiver
registryType: utilities
$ find content/en/registry/*.* -exec grep -e 'language:' {} \; | sort | uniq    
language: kotlin
language: collector
language: cpp
language: dotnet
language: elixir
language: erlang
language: go
language: java
language: js
language: lua
language: php
language: python
language: ruby
language: rust
language: swift
```

And advantage of doing this outside of the build cycle is that it doesn't slow down the build. A disadvantage is that it might not be easy to run this check for users under Windows.

An alternative is to create data files with the currently supported values, and check each registry entry as it's page is generated.

Related:

- #900

svrnm:
fixes #1355, cc @chalin

### DIFF
--- a/.cspell/en-words.txt
+++ b/.cspell/en-words.txt
@@ -69,7 +69,6 @@ lifecycles
 lightstep
 Loffay
 Mancuso
-matlab
 microservices
 mongodb
 ndjson
--- a/.cspell/en-words.txt
+++ b/.cspell/en-words.txt
@@ -69,6 +69,7 @@ lifecycles
 lightstep
 Loffay
 Mancuso
+matlab
 microservices
 mongodb
 ndjson
--- a/data/registry-schema.json
+++ b/data/registry-schema.json
@@ -1,4 +1,5 @@
 {
+  "cspell:ignore": "matlab",
   "$schema": "http://json-schema.org/draft-07/schema#",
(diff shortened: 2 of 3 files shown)


chalin (approved):
LGTM, once the failed check is addressed. Thx

.cspell/en-words.txt
```diff
@@ -69,6 +69,7 @@ lifecycles
 lightstep
 Loffay
 Mancuso
+matlab
```
svrnm, line 72: @chalin since I can not add a cspell ignore section to the JSON file I added it globally. I hope this is a valid exception.

chalin: If you can add arbitrary fields to the JSON file, which I think we can, then add it via:

```json
  "spelling": "cSpell:ignore words-to-ignore -"
```

PR #5412, opened 2024-10-15
Add image minifcation task 

### LINKED ISSUE #1647: Tooling: add image compression
_Originally posted by @mic-max in https://github.com/open-telemetry/opentelemetry.io/issues/869#issuecomment-965749094_

Making this part of the PR checks would be cool.
One way to implement it would be something like:
1. The PR either modifies or creates a PNG
2. Run a compression program on all the modified and created PNGs
3. If all the compressed images have the same file sizes in the PR, succeed. Otherwise fail.

Then you'd likely want to add some guidance that before uploading they run the **same** compression program locally to a README

svrnm:
fixes #1647 by adding a gulp task that minifies all images. It also introduces a cache such that we can validate if images have been minified and fix it/check it via CI.

**Note**: This PR touches **all** images that's why there are 280 changed files. The relevant code is in [gulp-src/image-min.js](https://github.com/open-telemetry/opentelemetry.io/pull/5412/files#diff-108185f779806befbc92f2c430697b4d38be32ea8cab790daa6f008f16affb4d)


svrnm:
> Thanks for this. My first broad comment is that I consider SVG files as source files, they should not be minimized in their source form, but rather only when we generate the website. WDYT?

good point. Should we run this only on production build then?

Issue #5361, opened 2024-10-09
Review visualizations for dark/light mode compatibility


svrnm:
While we have dark mode for a while now, some of the visualizations on our website have a white background and/or colors that are made for light mode, e.g.

* https://opentelemetry.io/img/otel-diagram.svg on https://opentelemetry.io/docs/
* https://opentelemetry.io/img/waterfall-trace.svg on https://opentelemetry.io/docs/concepts/observability-primer/
* https://opentelemetry.io/docs/collector/img/otel-collector.svg on https://opentelemetry.io/docs/collector/
* ...

It would be great if at a minimum the background for those images could be set to be transparent, so that they either are black or white, and then to check if some colors need to be adjusted to still work with light&dark mode at the same time. The overall color schema & theme should remain untouched.

**Special Note for Outreachy Contributors**: if you want to work on this issue, please make sure that you have completed the small steps and at least one assigned issue as outlined here: https://github.com/open-telemetry/opentelemetry.io/issues/5290

Tharanishwaran:
@svrnm can i work on this issue ?

svrnm:
@SuvineetSingh @Tharanishwaran we are currently running the participation phase for [outreachy](https://www.outreachy.org/), and we need to reserve some good first issues for the participants. If you participate in outreachy please take a look in #5290, otherwise I can currently not assign this issue to you. When the participation phase ends, and this issue is still available, I am happy to assign it.

SusanGithaigaN:
@svrnm, I am an outreachy participant, can I work on this issue?

svrnm:
> @svrnm, I am an outreachy participant, can I work on this issue?

@SusanGithaigaN let's finish your PR and then you can work on this. Since this is similar to your current work, I can imagine well that this is something you can help with. Thank you.

Naggayi-Daphne-Pearl:
hey @svrnm, 
can i work on this issue

svrnm:
fixed

PR #5434, opened 2024-10-18
Adjust visualizations for dark/light mode compatibility

SusanGithaigaN:
This pr resolves issue #5361 and adjusts the background color of the visualizations, ensuring they adapt to both light and dark modes. The overall color scheme and theme remain unchanged.

## Demo:

[Review visualizations for dark_light mode compatibility #5361.webm](https://github.com/user-attachments/assets/bd8a6a49-0468-4621-883a-98de7cc4a708)


SusanGithaigaN:
Hi @svrnm, could you please review my pr for this issue.
Thank you

chalin (changes requested):
Very interesting work, thanks @SusanGithaigaN.

I've noted that the new SVG files actually contain many embedded PNG elements when there were none before. It is as if parts (probably the icons) were converted to PNGs. So now, it's like we have a collection of PNGs wrapped in an SVG.

I'd rather preserve the original SVG. Make minimal updates to the SVG text (if any is required at all), so that the new color dependent CSS classes can kick in. How does that sound @SusanGithaigaN?

SusanGithaigaN:
> kick

Thanks for the feedback. Lemme take a look and make the necessary changes

chalin:
@SusanGithaigaN @svrnm - Here's an example of the kind of minimal changes I had in mind. I've done the update for only one of the images, you can work on the other images if you'd like:

- #5436

PR #5450, opened 2024-10-21
[outreachy] updated registry with inline help tooltips

### LINKED ISSUE #5420: [outreachy] Registry: inline help
Outreachy task for @bhekidaweti

Thank you for filling out the user experience report, please provide the following improvement via a PR:

> Like I mentioned, providing a little explanation further would help first time users understand clearly what the app is about. 

Come up with 1 or 2 ideas how inline help could be added to the page. This is not about adding a link that leads to a page explaining everything, but about small interactive elements that show how one can use the registry, and what they can expect from the registry. We can then discuss what you can implement in a PR.

Note: This is an assigned issue for a participant that has completed the user experience report, if you are an outreachy participant as well, find instructions here: https://github.com/open-telemetry/opentelemetry.io/issues/5290

bhekidaweti:
This PR resolves #5420. I implemented hover hints on key action buttons to make clear to users what they can expect when they click these buttons. The buttons show users exactly how to interact with different parts of the OpenTelemetry registry.

I also added a hover hint on the alert text that describes what OpenTelemetry Registry is, further describing what the Telemetry is, without leaving the page itself. 

### Example of the hover hint in one of the action buttons.
programming-language-hint

### Hover hint in the alert description
oTel-hint

### DIFF
--- a/layouts/shortcodes/ecosystem/registry/search-form.html
+++ b/layouts/shortcodes/ecosystem/registry/search-form.html
@@ -34,10 +34,54 @@
 {{ end -}}
 {{ $types = $types | uniq | sort -}}
 
+<!--Custom css for the inline help-->
+<style>
+  .btn-hint:hover::after {
+    position: absolute;
+    background-color: #0b5394;
+    border-radius: 3px;
+    color: white;
+    font-size: 0.85rem;
+    bottom: 120%;
+    left: 0;
+    white-space: break-spaces;
+}
+
+.btn-outline-success:hover::after {
+    content: "Submit your search query.";
+}
+
+.btn-outline-danger:hover::after {
+    content: "Clear search and reset filters.";
+}
+
+#languageDropdown:hover::after {
+    content: "Select the programming language.";
+}
+
+#componentDropdown:hover::after {
+    content: "Choose the component type.";
+}
+.hover-hint {
+    position: relative; 
+    cursor: pointer;
+    text-decoration: underline dotted #0b5394;
+}
+
+.hover-hint:hover::after {
+    content: attr(title); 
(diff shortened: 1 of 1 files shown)


bhekidaweti:
Hello, @svrnm  I have made another PR with the changes and the correct files to work on as per our last discussion.

layouts/shortcodes/ecosystem/registry/search-form.html
```diff
@@ -34,10 +34,54 @@
 {{ end -}}
 {{ $types = $types | uniq | sort -}}
 
+<!--Custom css for the inline help-->
```
svrnm, line 37: Please move custom CSS into this file:

https://github.com/open-telemetry/opentelemetry.io/blob/main/assets/scss/_registry.scss

Doing so, make sure that the CSS is only applied on that page, e.g. right now if we use "btn-outline-success" class on any other page the hover::after would also be applied

svrnm (changes requested):
minor comment, otherwise lgtm

layouts/shortcodes/ecosystem/registry/search-form.html
bhekidaweti: Thanks mentor I will update it.

svrnm: Please make sure that you provide the update today, such that I still can review it. I will not be available for more feedback afterwards until the end of the application phase

svrnm:
@bhekidaweti thanks for working on this during the outreachy application phase. I will close this PR for now, since we have not yet decided on how we want to proceed with tooltips on the registry.

PR #5470, opened 2024-10-24
Add MLflow entry to the OpenTelemetry integrations registry

daniellok-db:
Hi all! This is a follow-up to #5150. We've built an OTel exporter and have [instructions in the doc page](https://mlflow.org/docs/latest/llms/tracing/index.html#using-opentelemetry-collector-for-exporting-traces) on how to use it 😄. As for semantic conventions, due to historical reasons (some downstream applications are relying on the existing data format) it's a little difficult to migrate immediately, but we're keeping an eye on the spec and where it makes sense to integrate.


daniellok-db:
cc @svrnm

svrnm (approved):

PR #5337, opened 2024-10-07
Add MetricsHub to vendor list #5314

iguitton:
Added MetricsHub to vendor list.

PR #5488, opened 2024-10-28
add Dash0 to vendor list

basti1302:


data/ecosystem/vendors.yaml
```diff
@@ -90,6 +90,12 @@
   contact: hello@cribl.io
   oss: false
   commercial: true
+- name: Dash0
+  nativeOTLP: true
+  url: https://www.dash0.com/
```
basti1302, line 95: FYI, I've seen in [earlier PR for this list](https://github.com/open-telemetry/opentelemetry.io/pull/5337#discussion_r1790337177) that the provided link should "highlight that your product supports OTLP natively" -- being OTel-native is very much at the core of Dash0's messaging, so this is highlighted all over our landing page; hence this should check out I think.

theletterf (approved):
Thanks!

PR #5489, opened 2024-10-28
Add missing step to howto guide for adding a vendor

basti1302:
The required check "Links / REFCACHE updates? (pull_request)" will fail if you follow the current instructions to add a new vendor to the list of OTel vendors; because a required step is missing - updating the file static/refcache.json. This change adds instructions on how to update the file after updating `data/ecosystem/vendors.yaml`.

### DIFF
--- a/.cspell/en-words.txt
+++ b/.cspell/en-words.txt
@@ -108,6 +108,7 @@ quarkus
 quoteservice
 recommendationservice
 redis
+refcache
 relref
 Rexed
 runbook
--- a/content/en/ecosystem/vendors.md
+++ b/content/en/ecosystem/vendors.md
@@ -32,6 +32,9 @@ The entry should include the following:
   source distribution does not qualify your offering to be marked "open source".
 - GitHub handle or email address as a point of contact so that we can reach out
   in case we have questions
+- make sure you have [set up](/docs/contributing/development/) your local clone
+  for development (`nvm install`/`nvm use` & `npm install`) and run
+  `npm run log:check:links` to update the file `static/refcache.json`
 
 Note that this list is for organizations that consume OpenTelemetry and offer
 Observability to [end users](https://community.cncf.io/end-user-community/).


tiffany76:
Hi @basti1302, thanks for making this suggestion. Several GitHub actions run on docs PRs, not just refcache, so I don't think it makes sense to add instructions for just one. Maybe we could add a link to the [contributing documentation](https://opentelemetry.io/docs/contributing/pull-requests/#fix-content-issues-automatically). 

@open-telemetry/docs-approvers, WDYT?

basti1302:
> Hi @basti1302, thanks for making this suggestion. Several GitHub actions run on docs PRs, not just refcache, so I don't think it makes sense to add instructions for just one. Maybe we could add a link to the [contributing documentation](https://opentelemetry.io/docs/contributing/pull-requests/#fix-content-issues-automatically).

My intent is not to document which GH actions run as this is of no immediate concern to the person submitting the PR.

The intent is to document the manual steps that are required to submit a PR for adding a vendor to the list and not have any checks fail. For the record, just before this PR I created https://github.com/open-telemetry/opentelemetry.io/pull/5488 and the refcache GH action was the only one that failed. The typical vendor list PR will only add an entry to `data/ecosystem/vendors.yaml`, which is explained specifically here in the instructions. But it is not enough. The mentioned `npm run` script also needs to be run, but that step is missing. No further steps are required for a vendor list submission.

To be honest I searched for quite a while until I found how to update the refcache file, I think it would improve the experience if this necessary step is just added to the guide on how to add a vendor entry.

cartermp:
I think this is a reasonable addition to the instructions. That said, I think it's worth mentioning that we have some easy fixes we can run at PR time here that use actions to solve this. The refcache in particular is one of the more surprising things that people wouldn't necessarily expect they'd need to update. I think it's a good idea to be both proactive here (run the check locally if you can) and reactive here (don't worry if it fails, we have a convenient way to catch you).

chalin (changes requested):
I disagree with this addition. Such guidance isn't specific to adding a vendor. I'll make some specific suggestions in a followup comment.

chalin (commented):
Now that I've had time to read through all the comments. I agree with @tiffany76: we should instead just add a link to [Submitting content](/docs/contributing/pull-requests/), and contributors can decide on their own if they're content doing a simple submission from GH directly, or whether they choose to work from a local dev environment (and follow the detailed steps in that page on how to more extensively conduct tests).

I'd suggest changing:

> To have your organization listed, submit a PR with an entry added to the

to

> To have your organization listed, [submit a PR](/docs/contributing/pull-requests/) with an entry added to the

and drop the added bullet point.

chalin:
By the way, thanks for raising this issue @basti1302. You are right that our docs need to be improved. Let me dig a bit deeper into this and report back here. I'll likely submit a separate PR. If you'd like to make the change I suggested earlier, we can get this PR merged (even though it doesn't fully yet address your original concern). If you could open an issue to track the required change, that would be appreciated; otherwise I'll do so. Thanks again!

basti1302:
> [Submitting content](https://github.com/docs/contributing/pull-requests/), and contributors can decide on their own if they're content doing a simple submission from GH directly, or whether they choose to work from a local dev environment (and follow the detailed steps in that page on how to more extensively conduct tests).

Seems that the consensus is to not add this. Feel free to close the PR then.

Just to reiterate: I was mainly coming from my own experience here, because seeing the failing refcheck GH action on my other PR sent me down a rabbit hole of finding out how to fix. Having all necessary instructions in one place would have saved me a lot of time. I personally still think it is good to have all instructions for a specific use case in one place, or to at least point out that people submitting this kind of PR do not need to take care of failing checks. Only linking to the general PR submission guidelines will still leave people wondering what's up with the refcache check. (The info is there but it is kind of hard to find with all the other stuff that is explained there.)

@chalin feel free to follow up as you see fit, I don't plan to change this PR or create an issue for it, thanks.

chalin:
Thanks again for bringing this to our attention. I've done a minimal followup via:

- #5571

@svrnm is working on something more substantive via #5544.

PR #5499, opened 2024-10-29
Remove duplicate text

Expecho:
The removed text is already included by linking the span-status-preamble


cartermp (approved):
Thanks, good catch!

PR #5504, opened 2024-10-30
Update Node LTS + Hugo patch and other pkg vers

### LINKED ISSUE #5503: [CI] Update Node LTS version
The current Node LTS version, since 2024-10-29, is v22. Update the constraint in our package.json file.

Reference:

- https://nodejs.org/en/about/previous-releases

chalin:
- Fixes #5503

### DIFF
--- a/package.json
+++ b/package.json
@@ -118,16 +118,16 @@
     "ajv-errors": "^3.0.0",
     "ajv-formats": "^3.0.1",
     "autoprefixer": "^10.4.20",
-    "cspell": "^8.15.4",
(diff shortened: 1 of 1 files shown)


chalin:
Hmm, it seems that the environment for GH actions doesn't think that v22 is the latest `lts/*`, so let's wait a bit.

chalin:
Oh, our build scripts don't check whether the Node version is `lts/*` -- which is fine. Let's merge this since GH should update their `lts/*` version soon.

tiffany76 (approved):
RSLGTM.

PR #5522, opened 2024-11-04
Group java performance, javadoc, examples, registry under Learn More

### LINKED ISSUE #5211: Finish refactoring Java information architecture
Followup on #4966 to finish the ideas in #4853. If / when complete, we can decide if its useful for other languages to follow suit.

- [] Refactor [Instrumentation page](https://opentelemetry.io/docs/languages/java/instrumentation/) into pages discussing how to use the API, guide for library developers, context propagation, using semantic conventions
- [] Nest [Registry](https://opentelemetry.io/docs/languages/java/registry/), [Performance](https://opentelemetry.io/docs/languages/java/performance/), [API (Javadoc)](https://opentelemetry.io/docs/languages/java/api/) under new "Learn more" section
- [] Add new Learn the Basics page, introducing users to the otel java ecosystem and routing them to various pages for more information
- [] Add page describing instrumentation landscape in java and guide to installing instrumentation in applications

jack-berg:
Resolves #5211.

Reduces the prominence of the Java Performance, Javadoc, Examples, and Registry by collecting them under a "Learn More" section. This reduces the number of entries under the Java section by 3, focusing user attention on the pages which are more likely to be useful and hopefully making the navigation less overwhelming. 

Screenshot 2024-11-04 at 11 25 46 AM

### DIFF
--- a/layouts/shortcodes/apidocs.md
+++ b/layouts/shortcodes/apidocs.md
@@ -7,6 +7,10 @@
     {{ with $.Site.GetPage "/docs/languages/net/metrics-api" -}}
         {{ $pages = $pages | append (dict "lang" $value "page" .) }}
     {{ end }}
+    {{ else if eq $key "java" -}}
+    {{ with $.Site.GetPage "/docs/languages/java/learn-more/api" -}}
+        {{ $pages = $pages | append (dict "lang" $value "page" .) }}
+    {{ end }}
     {{ else -}}
     {{ with $.Site.GetPage (printf "/docs/languages/%s/api" $key) -}}
         {{ $pages = $pages | append (dict "lang" $value "page" .) }}
--- a/.htmltest.yml
+++ b/.htmltest.yml
@@ -20,6 +20,8 @@ IgnoreInternalURLs: # list of paths
 IgnoreURLs: # list of regexs of paths or URLs to be ignored
   - ^/api$
   - ^((/..)?/docs/languages/\w+|\.\.)/(api|examples|registry)/$
+  # Java nests ./api/, ./examples/, ./registry/ under ./learn-more/
+  - ^((/..)?/docs/languages/\w+|\.\.)/learn-more/(api|examples|registry)/$
   - ^(/..)?/docs/collector/registry/$
   - ^(/..)?/docs/languages/net/(metrics-api|traces-api)/
   - ^((/..)?/docs/migration/)?opencensus/$
--- a/content/en/docs/languages/java/intro.md
+++ b/content/en/docs/languages/java/intro.md
@@ -61,8 +61,9 @@ The OpenTelemetry Java documentation is organized as follows:
   Most users will use this page like an encyclopedia, consulting the index of
   sections as needed, rather than reading front to back.
 - **Learn More**: Supplementary resources, including end-to-end
-  [examples](../examples/), [Javadoc](../api/), component
-  [registry](../registry/), and a [performance reference](../performance/).
+  [examples](../learn-more/examples/), [Javadoc](../learn-more/api/), component
+  [registry](../learn-more/registry/), and a
+  [performance reference](../learn-more/performance/).
 
 ## Repositories
 
--- a/content/en/docs/languages/java/learn-more/_index.md
+++ b/content/en/docs/languages/java/learn-more/_index.md
@@ -1,0 +1,4 @@
+---
+title: Learn More
+description: Supplementary resources for the OpenTelemetry Java ecosystem
+---
--- a/content/en/docs/languages/java/learn-more/api.md
+++ b/content/en/docs/languages/java/learn-more/api.md
@@ -2,6 +2,8 @@
 title: Javadoc API reference
 linkTitle: Javadoc
 redirect: https://javadoc.io/doc/io.opentelemetry
+aliases:
+  - /docs/java/api
 manualLinkTarget: _blank
 _build: { render: link }
 weight: 210
--- a/content/en/docs/languages/java/learn-more/examples.md
+++ b/content/en/docs/languages/java/learn-more/examples.md
@@ -5,6 +5,7 @@ manualLinkTarget: _blank
 aliases:
   - /docs/java/instrumentation_examples
   - /docs/languages/java/instrumentation_examples
+  - /docs/java/examples
 _build: { render: link }
 weight: 220
 ---
--- a/content/en/docs/languages/java/learn-more/performance.md
+++ b/content/en/docs/languages/java/learn-more/performance.md
@@ -3,6 +3,8 @@ title: Performance
 description: Performance reference for the OpenTelemetry Java agent
 weight: 75
 cSpell:ignore: Dotel
+aliases:
+  - /docs/java/performance
 ---
 
 The OpenTelemetry Java agent instruments your application by running inside the
@@ -40,7 +42,7 @@ the Java agent.
 
 The volume of spans processed by the instrumentation might impact agent
 overhead. You can configure trace sampling to adjust the span volume and reduce
-resource usage. See [Sampling](../sdk/#sampler).
+resource usage. See [Sampling](../../sdk/#sampler).
 
 ### Turn off specific instrumentations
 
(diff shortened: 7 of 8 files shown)


trask (approved):

cartermp (approved):
Approved content and moving around, pending approval from others on the infra changes

jaydeluca (approved):

layouts/shortcodes/apidocs.md
```diff
@@ -7,6 +7,10 @@
     {{ with $.Site.GetPage "/docs/languages/net/metrics-api" -}}
         {{ $pages = $pages | append (dict "lang" $value "page" .) }}
     {{ end }}
+    {{ else if eq $key "java" -}}
+    {{ with $.Site.GetPage "/docs/languages/java/learn-more/api" -}}
+        {{ $pages = $pages | append (dict "lang" $value "page" .) }}
+    {{ end }}
```
jack-berg, line 13: Realized that without this, there's no java entry in the "Language APIs & SDKs -> API References" section. Moving the javadoc to be nested under "Learn more" means that we need to add a special case to this shortcode like dotnet.

On a separate note, while I've linked to the javadoc page for consistency with the other languages that have an external link, it would be more useful to link to the [Record Telemetry with API](http://localhost:1313/docs/languages/java/api-components/) page, since that includes explanations, examples, and javadoc links for all the key API surface area.

What we link to is a question of whether we value consistency or usefulness more.

Screenshot 2024-11-05 at 8 55 55 AM

svrnm: the purpose of this list is to make the API references (read: JavaDoc) accessible more easily. So while I get your point that people should read the api components page first, this is more for people that have read that page and need quick access to JavaDocs (or the equivalent in another language)

I think this is the only thing that bothers me with this change, that we break that consistent like `/docs/languages/<language>/api` to point to the available documentation

I wonder if we could have a canonical URL **outside** the structure, like `/api/<language>`

cc @chalin

svrnm: > > I think this is the only thing that bothers me with this change, that we break that consistent like /docs/languages//api to point to the available documentation
> 
> To be fair, dot net already breaks that uniformity. I understand your point though.

Yes, but I would call that "for historical" reasons. We had a separation for JS in the past as well since the published their JsDocs for API & SDK on different places, but I think that's not the case anymore so we have only one link.

jack-berg: What do you think about:

- Remove the external javadoc redirect page at `/docs/languages/java/learn-more/api`
- Rename the "Record Telemetry with API" from `/docs/languages/java/api-components` to `/docs/languages/java/api`. This page becomes Java's canonical API page.

If we do this, then the only inconsistency is that java's API link is internal rather than external.

chalin (changes requested):
I'd vote against this PR. The main motivation seems to be:

> **Reduces the prominence** of the Java Performance, Javadoc, Examples, and Registry

The prominence of those entries is already low, by the fact that they are at the end of the section ToC. 

Questions:

- Why does the prominence of those entries need to be reduced further? It's not like the ToC is overpopulated atm.
- At what cost will it be done -- in both UX and infrastructure/coding "special casing"?
- What will the resulting "delta" in reduced prominence be, and is it worth it?

Here are the ToCs side-by-side to help you all get a feel:

| Before | After |
|--------|--------|
| image | image<br><br><br><br> |

Pro of before: I can see all ToC entries at a glimpse.

There may be other ways to reduce the prominence of those entries, by styling them differently (e.g., faded font, italics). I'm unsure that it's worth the special casing atm, but I'm willing to entertain the discussion.

jack-berg:
Fair points @chalin. I'm not married to this proposal I've made, but wanted to at least open up the discussion to wrap up #5211.

When I take a step back and consider my motivation here, I realize a few things:

* I don't really like 3/4 of the pages under "Learn more", but not enough to delete them.
* The "Performance" page is sort of java agent centric, but some of the ideas are applicable to the API / SDK as well. I'd really like the Java SIG (including myself) to come up with a more complete guide on how to think about performance of the SDK, including references to performance benchmarks with explanations. We're not there yet.
* The "Javadoc" link is useless now that "Manage Telemetry with SDK" and "Record Telemetry with API" exist and point to much more specific sections of the Javadoc. I wonder if we can just get rid of this page today?
* The "Examples" link is useful. No real issue with it.
* The "Registry" is a great concept, but falls short right now. I've manually added tables which I think could one day be served by the registry all over "Manage Telemetry with SDK" ([for example](https://opentelemetry.io/docs/languages/java/sdk/#spanexporter)). 

If we delete "Javadoc" (and use the "Record Telemetry with API" page as the canonical java API page), and have future plans to improve performance and registry, then it becomes more palatable to not nest pages under the additional "Learn ore" section.

chalin:
In summary: 

- I'm ok with what you propose: to use "Record Telemetry with API" as the `api` page.
- But, I'd like to strongly suggest that you a banner or note to the top of the `api` page giving readers who are looking for the Javadocs a direct link to the Javadocs.

More context: my original vision (coming from other projects where this is implemented), is to have `api`, `examples`, etc be (relatively) uniform across implementation languages. That being said, I'm ok with what you propose.

svrnm:
If Performance is agent-centric we should move it to the agent for now, if this changes in the future or there is a dedicated page on the SDK we can have it as well.

Leaves us with the other pages:
* Examples: No discussion here, I agree, we keep that 
* Registry: as of today this link serves the registry more than it serves the java documentation. We added those links to the language docs across all languages to promote the availability of the registry. If we find better ways in the future to integrate it, I am all in, but right now I would like to keep it.

API: 

> I'm ok with what you propose: to use "Record Telemetry with API" as the api page.
> But, I'd like to strongly suggest that you a banner or note to the top of the api page giving readers who are looking for the Javadocs a direct link to the Javadocs.

+1 for this. At some point developers will look for the Javadoc (or the equivalent in their language) as an additional resource to learn more about the available APIs. Especially if some of them are not part of the documentation, so we should make them discoverable for them. I am fine with the proposal to remove the "Javadoc" link from the menu and have it highlighted in the "Record Telemetry with API" page, but I want to keep the link from https://opentelemetry.io/api to the Javadoc directly, 
which could be https://opentelemetry.io/api/java if we want to move it out of the existing hierarchy. 

It looks like we do not need the "Learn More" sub-menu for now. We can keep that in the back of our heads if the number of pages continues to grow, but I think for now let's make the adjustment discussed here and keep the remaining pages (Examples, Registry) where they are.

trask:
> At some point developers will look for the Javadoc

will they? I only ever read Javadoc inside of my IDE 🤷‍♂️

svrnm:
> > At some point developers will look for the Javadoc
> 
> will they? I only ever read Javadoc inside of my IDE 🤷‍♂️

Fair point ;-)

chalin:
> > At some point developers will look for the Javadoc
> 
> will they? I only ever read Javadoc inside of my IDE 🤷‍♂️

By that reasoning, we should not even be hosting API reference documentation :) ... but we do, so someone must feel that it's useful ;).

While I agree with what you say @trask (and do the same), having API reference docs hosted comes in handy now and again IMHO.

Of course, if we had access to the hosted API docs analytics, we could have a better idea about usage. Does anyone have access?

theletterf (approved):

trask:
> if we had access to the hosted API docs analytics, we could have a better idea about usage. Does anyone have access?

I couldn't find anything about analytics. FWIW the https://javadoc.io site isn't anything official, it's just run as someone's personal project, scraping javadocs from maven central.

svrnm:
> > if we had access to the hosted API docs analytics, we could have a better idea about usage. Does anyone have access?
> 
> I couldn't find anything about analytics. FWIW the [javadoc.io](https://javadoc.io) site isn't anything official, it's just run as someone's personal project, scraping javadocs from maven central.

oh, for whatever reason I was under the assumption that this was something "official" we participate in. Is this a concern or is this good enough for us? As an alternative we can follow the examples of other languages and publish it to github pages/read the docs and link to that. WDYT?

I agree with @chalin that there is some value to have them, while not big advantages, still something worth to consider:

* there are people not doing java all the time, so their IDE might not be setup 100% to consume javadoc, so they can read it there
* people that write some similar code and want to take from that (e.g. they write an instrumentation library but want to see how others are structuring it)
* most importantly: search engine optimization, and that we might want to link out to the API Docs from our docs eventually, so people can look up signatures etc while reading the docs (and mybe not yet writing code)

At the end it's SIG Java's decision. If you think that this does not deliver value we can remove it

jack-berg:
I opened #5590 as an alternative to this, and have marked this PR as a draft because I'm leaning in the direction of this new PR now.

jack-berg:
Closing since we all seem to be in favor of #5590.

PR #5540, opened 2024-11-06
Update community members

danielgblanco:
Updates members using `scripts/generate-community-data/generate.js`

My only required update was to update GC members after election, however running the script above seems to have reshuffled things and added more changes that have occurred since it was originally run. Let me know if I'm missing anything.


svrnm (approved):
lgtm overall, I need to review the script to make sure that the ordering in the YAML file is consistent but that's not a concern of this PR.

thanks

PR #5548, opened 2024-11-07
Add new page feedback issue template

svrnm:
We discussed this a while back that with github's feature of YML-based form-based issue templates we could streamline the page feedback, PTAL

Preview of the form: https://github.com/svrnm/opentelemetry.io/blob/add-page-feedback-template-issue/.github/ISSUE_TEMPLATE/PAGE_FEEDBACK.yml


.github/ISSUE_TEMPLATE/PAGE_FEEDBACK.yml
```diff
@@ -0,0 +1,27 @@
...
+  if a page was helpful.
+title: '[Page Feedback]: Your feedback'
+body:
+  - type: input
+    id: url
+    attributes:
+      label: URL
+      description: >-
+        The URL of the website where you are facing an issue. If not
+        auto-populated please copy the URL from the browser address bar.
```
tiffany76, line 13:
```suggestion
        The URL of the webpage where you are facing an issue. If not
        auto-populated, please copy the URL from the browser address bar.
```

.github/ISSUE_TEMPLATE/PAGE_FEEDBACK.yml
```diff
@@ -0,0 +1,27 @@
...
+      required: true
+  - type: textarea
+    id: description
+    attributes:
+      label: Description
+      description:
+        'How has the page not been helpful? What information is missing? '
+      placeholder: >-
+        Provide details on the problem. The more information you share with us
+        the easier it is to help and to provide a solution
```
tiffany76, line 25:
```suggestion
        Provide details on the problem. The more information you share with us
        the faster we can resolve the issue.
```

tiffany76 (approved):
Couple small suggestions. I think we should emphasize that we want the page URL. Otherwise, LGTM.

.github/ISSUE_TEMPLATE/PAGE_FEEDBACK.yml
```diff
@@ -0,0 +1,27 @@
+name: Page Feedback
+description: >-
+  Used for page feedback from users visiting opentelemetry.io and clicking "no"
+  if a page was helpful.
+title: '[Page Feedback]: Your feedback'
```
chalin, line 5: Is this the template title or the title that will be used when the template is instantiated? If the latter, can we replace "Your feedback" with "INSERT PAGE TITLE HERE" ... or just use "Page feedback". Or, add an optional field entitled "Page title".

chalin (approved):
Very nice. LGTM

Should you be linking to this template from the "no" button text in this PR too?

svrnm:
> Very nice. LGTM
> 
> 
> 
> Should you be linking to this template from the "no" button text in this PR too?

I would do this in a separate PR. I had it as part of this PR initially but couldn't verify it working before it is merged.

.github/ISSUE_TEMPLATE/PAGE_FEEDBACK.yml
chalin: Sure.

theletterf (approved):

.github/ISSUE_TEMPLATE/PAGE_FEEDBACK.yml
svrnm:
```suggestion
title: '[Page Feedback]: summary of your feedback'
```

PR #5554, opened 2024-11-07
[i18n] Render-link fix: prepend lang code when page exists

### LINKED ISSUE #5553: Render-link hook isn't generating locale specific paths when possible
While PR #4689 was a good step in the right direction of solving

- #4688

It doesn't work as expected.

The simplest solution would be to always prefix a language code to the absolute paths of non-en section pages, that causes problems for our link checker, which requires the use of canonical links (in other words, it knows nothing and can't deal with redirects). I'll post more later.

chalin:
- Fixes #5553
- Render-link hook: adds the language code prefix to absolute paths of links that appear in non-`en` pages
  - As a consequence of this fix, link-check failures are now reported and will need to be fixed.
- Adds `IgnoreDir` regexs so that the offending non-`en` are temporarily excluded from link-checking

### DIFF
--- a/.htmltest.yml
+++ b/.htmltest.yml
@@ -10,10 +10,44 @@ IgnoreDirs:
   # DO NOT EDIT! IgnoreDirs list is auto-generated from markdown file front matter.
   # Ignore blog index pages for all locales and in all blog sections (top-level and years)
   - ^(../)?blog/(\d+/)?page/\d+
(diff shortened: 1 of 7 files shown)


tiffany76 (approved):
RSLGTM

chalin:
In summary, this PR will help:

- Identify missing heading IDs in new non-`en` pages as soon as the new page is submitted.
- Provide a better UX: non-`en` locale page navigation will stay in the locale even if fallback pages are visited.

PR #5561, opened 2024-11-08
[CI] Add GH workflow to run dev build using submodules at head

chalin:
- Adds GH workflow that we can run manually to build the website from the HEAD of selected submodules. 

  This is proposed as a solution to **mitigating OTel spec build failures** when a new version of the spec is released. The workflow could be **run by OTel spec maintainers** _before_ they publish a release.

- Cleans up and improves various other the NPM scripts, including `test-and-fix`

### DIFF
--- a/.github/workflows/build-dev.yml
+++ b/.github/workflows/build-dev.yml
@@ -1,0 +1,88 @@
+name: Dev build
+
+on:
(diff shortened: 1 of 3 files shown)


svrnm (approved):
lgtm

PR #5567, opened 2024-11-08
Updated summary and table of internal telemetry

### LINKED ISSUE #5468: Document internal metrics being renamed when exporting with Prometheus
**What needs to be changed?**

The [Collector "Internal telemetry" page](https://opentelemetry.io/docs/collector/internal-telemetry) at `/docs/collector/internal-telemetry` mentions metrics emitted by the RPC and HTTP clients/servers, such as `rpc.server.duration`, but gives their names with the dots replaced by underscores (`rpc_server_duration`). This is presumably because the Prometheus exporter performs such a renaming, and it is the default method for exposing internal telemetry. I believe it would be useful to explicitly document this behavior when using Prometheus on the page, as well as the original names of these metrics, for the benefit of users exporting internal telemetry in some other way.

### LINKED ISSUE #5483: Remove mention of internal telemetry that does not exist
**What needs to be changed?**

The ["Values observable with internal metrics" section in the Collector "Internal telemetry" page](https://opentelemetry.io/docs/collector/internal-telemetry/#values-observable-with-internal-metrics) (`docs/collector/internal-telemetry`) mentions a number of metrics which (as far as I know) are not actually implemented, or listed in the detailed table of metrics below. To list some: data reception / export rates in bytes, data drop rates caused by throttling / invalid data, incoming connection rate, queue size in bytes, end-to-end and per-component latency, etc.

It looks like this list was taken from an old "Observability Vision" document [added to the core repo years ago](https://github.com/open-telemetry/opentelemetry-collector/pull/188), and the wording somehow changed from "these values need to be observable" to "these values *are* observable" along the way. (Thank you to @mx-psi for finding this PR.)

The simplest solution would be to delete the section, and let people rely on the table below it. A more involved solution would be to rewrite the section to only include things that actually ended up being implemented. We may want to keep the current contents around somewhere to serve as a roadmap for future new internal metrics.

jade-guiton-dd:
I rewrote the "Values observable with internal metrics" section of the "Collector > Internal telemetry" page to accurately summarize existing telemetry instead of being partly wishful thinking. This should resolve #5483.

While doing that, I noticed that the table below was also inaccurate to the telemetry currently implemented in the Collector, so I updated it. (This involved replacing the `processor_accepted/dropped_*` metrics by `processor_incoming/outgoing_items`, and putting the `processor_batch` metrics under the correct level.)

I also added a note below the table which may be helpful in understanding it. It may be enough to resolve #5468. I will remove it if it is deemed unnecessary.

### DIFF
--- a/content/en/docs/collector/internal-telemetry.md
+++ b/content/en/docs/collector/internal-telemetry.md
@@ -2,7 +2,7 @@
 title: Internal telemetry
 weight: 25
 # prettier-ignore
-cSpell:ignore: alloc journalctl kube otecol pprof tracez underperforming zpages
+cSpell:ignore: alloc journalctl kube otecol pprof tracez underperforming zpages batchprocessor
 ---
 
 You can inspect the health of any OpenTelemetry Collector instance by checking
@@ -214,7 +214,7 @@ The Collector emits internal metrics for at least the following values:
 - For exporters: Queue size and capacity.
 - Count, duration, and size of HTTP/gRPC requests and responses.
 
-A more detailed list is available below.
+A more detailed list is available in the following sections.
 
 ### Lists of internal metrics
 
--- a/content/en/docs/collector/internal-telemetry.md
+++ b/content/en/docs/collector/internal-telemetry.md
@@ -2,7 +2,7 @@
 title: Internal telemetry
 weight: 25
 # prettier-ignore
-cSpell:ignore: alloc journalctl kube otecol pprof tracez underperforming zpages batchprocessor
+cSpell:ignore: alloc batchprocessor journalctl kube otecol pprof tracez underperforming zpages
 ---
 
 You can inspect the health of any OpenTelemetry Collector instance by checking
--- a/content/en/docs/collector/internal-telemetry.md
+++ b/content/en/docs/collector/internal-telemetry.md
@@ -201,43 +201,20 @@ process on the same host. Specific components of the Collector can also emit
 their own custom telemetry. In this section, you will learn about the different
 types of observability emitted by the Collector itself.
 
-### Values observable with internal metrics
+### Summary of values observable with internal metrics
 
-The Collector emits internal metrics for the following **current values**:
+The Collector emits internal metrics for at least the following values:
 
-- Resource consumption, including CPU, memory, and I/O.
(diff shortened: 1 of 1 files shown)


content/en/docs/collector/internal-telemetry.md
```diff
@@ -201,43 +201,20 @@ process on the same host. Specific components of the Collector can also emit
...
+- For exporters: Queue size and capacity.
+- Count, duration, and size of HTTP/gRPC requests and responses.
 
-- Total received data, broken down by receivers.
-- Total exported data, broken down by exporters.
-- Total dropped data due to throttling, broken down by data type.
-- Total dropped data due to invalid data received, broken down by data type.
-- Total incoming connection count, broken down by receiver.
-- Uptime since start.
+A more detailed list is available below.
```
theletterf, line 217:
```suggestion
A more detailed list is available in the following sections.
```

content/en/docs/collector/internal-telemetry.md
```diff
@@ -274,74 +251,80 @@ files in the repository.
...
+| `rpc_server_requests_per_rpc`                         | Measures the number of messages received per RPC. Should be 1 for all non-streaming RPCs. | Histogram |
+| `rpc_server_response_size`                            | Measures the size of RPC response messages (uncompressed).                                | Histogram |
+| `rpc_server_responses_per_rpc`                        | Measures the number of messages sent per RPC. Should be 1 for all non-streaming RPCs.     | Histogram |
+
+{{% alert title="Note" color="info" %}} The `http_` and `rpc_` metrics come from
+instrumentation libraries. Their original names use dots (`.`), but when
+exposing internal metrics with Prometheus, they are translated to use
+underscores (`_`) to match Prometheus' naming constraints.
+
+The `otelcol_processor_batch_` metrics are unique to the `batchprocessor`.
```
theletterf, line 323: Add `batchprocessor` to the cSpell exceptions in the frontmatter.

theletterf (changes requested):

evan-bradley (approved):

svrnm:
@theletterf please re-review

theletterf (approved):

PR #5568, opened 2024-11-08
[CI] Simplify test-and-fix script for build-dev workflow

chalin:
- Another followup to #5561
- Simplifies `test-and-fix` by simply listing all scripts to run with `seq`.


tiffany76 (approved):
RSLGTM.

PR #5573, opened 2024-11-08
[CI] Update pkgs, including Hugo to 0.138.0; use new `comment` shortcode

chalin:
- Updates to Hugo [v0.138.0](https://github.com/gohugoio/hugo/releases/tag/v0.138.0)
- Illustrates use of the new Hugo [comment](https://gohugo.io/content-management/shortcodes/#comment) shortcode
- There are no changes to generated site files other than:
  - The updated Hugo version as part of the generator ID header element
  - The HTML comment removed from the page where this PR uses the `comment` shortcode instead.


tiffany76 (approved):
Note this might create a small merge conflict for #5567.

chalin:
Actually, that PR doesn't touch the same region of the file so we should be fine.

Issue #3493, opened 2023-11-04
Don't update old blogs, and possibly warn readers


chalin:
I think that we need to define an **update policy for blogs**. It is understood that blog entries, in contrast to doc pages, are fixed in time -- they rot. For example, Kubernetes doesn't allow updates for blogs older than 12 months. In fact, [I implemented the addition of the following banner-warning support (PR#31594)](https://github.com/kubernetes/website/pull/31594) (screenshot from K8s):
> <img width="1082" alt="image" src="https://github.com/open-telemetry/opentelemetry.io/assets/4140793/5f6a29af-75bb-444b-a02a-10bd79933afd">

We could implement something similar here -- WDYT? I'd keep it simple and only permit the update of blog post in the current year.

Of course, when pages move or are deleted from the website, we can fix any blog posts to ensure that links are still valid (or add a link-checker ignore rule).

Context, this was prompted by the merge of the following PR, which updates a blog entry over a year old:

- #3484

### Tasks

- [ ] Add warning banner
  - [ ] #5883
- [ ] Don't check links?
  - [x] #5874

svrnm:
https://github.com/kubernetes/website/blob/d0cf73f310eedb198203a6109fd8accfac60e954/layouts/blog/baseof.html

This is how Kubernetes is doing it. If I understand it correctly we would need to take the baseof for blog posts out of docsy and modify it, I would rather prefer not doing that.

I wonder if this is a feature we can upstream to docsy?

chalin:
Thanks for creating the issue over on the Docsy repo. I'll see if we can get this setup more quickly here first though.

PR #5575, opened 2024-11-09
Blog post for OpenTelemetry Generative AI updates

drewby:
Title: OpenTelemetry for Generative AI

This blog post introduces enhancements to OpenTelemetry specifically tailored for generative AI technologies, focusing on the development of Semantic Conventions and the Python Instrumentation Library. 

Samples are in Python

SIG: [GenAI Observability](https://github.com/open-telemetry/community/blob/main/projects/gen-ai.md)

Sponsors: @tedsuo @lmolkova 

Closes: #5581 

---

**Preview**: https://deploy-preview-5575--opentelemetry.netlify.app/blog/2024/otel-generative-ai/


content/en/blog/2024/otel-generative-ai/index.md
```diff
@@ -0,0 +1,89 @@
+---
+title: OpenTelemetry for Generative AI
+linkTitle: OpenTelemetry for Generative AI
+date: 2024-11-09
+author: >-
+  [Drew Robbins](https://github.com/drewby) (Microsoft),  [Luidmila
```
lmolkova, line 6:
```suggestion
  [Drew Robbins](https://github.com/drewby) (Microsoft),  [Liudmila
```

:)

content/en/blog/2024/otel-generative-ai/index.md
```diff
@@ -0,0 +1,89 @@
...
+
+The **Instrumentation Library** is being developed within the OpenTelemetry Python Contrib project to automate telemetry collection for generative AI applications. The first release is a Python library, given Python’s widespread use in AI development. Designed to integrate seamlessly with OpenAI’s API, this library captures spans, metrics, and events, gathering essential data like model inputs, response metadata, and token usage in a structured format.
+
+## Key Signals for Generative AI
+
+The [Semantic Conventions for Generative AI](https://github.com/open-telemetry/semantic-conventions/tree/main/docs/gen-ai) focus on capturing insights into AI model behavior through three primary signals: Spans, Metrics, and Events.
```
lmolkova, line 20: nit: 

it's probably best to pin specific version (in case things move around)

```suggestion
The [Semantic Conventions for Generative AI](https://github.com/open-telemetry/semantic-conventions/tree/v1.28.0/docs/gen-ai) focus on capturing insights into AI model behavior through three primary signals: Spans, Metrics, and Events.
```

content/en/blog/2024/otel-generative-ai/index.md
```diff
@@ -0,0 +1,89 @@
...
+
+Together, these signals provide a comprehensive monitoring framework, enabling better cost management, performance tuning, and request tracing.
+
+**Extending Observability with Vendor-Specific Attributes**
+
+The Semantic Conventions also define vendor-specific attributes for platforms like OpenAI and Azure Inference API, ensuring telemetry captures both general and provider-specific details. This added flexibility supports multi-platform monitoring and in-depth insights. 
+
+## Building the Python Instrumentation Library for OpenAI
+
+This Python-based library for OpenTelemetry captures key telemetry signals for OpenAI models, providing developers with an out-of-the-box observability solution tailored to AI workloads. The library, [hosted within the OpenTelemetry Contrib project](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/instrumentation-genai/opentelemetry-instrumentation-openai-v2), automatically collects telemetry from OpenAI model interactions, including request and response metadata, token usage, and operational metrics.
```
lmolkova, line 39: another nit on pinning versions

```suggestion
This Python-based library for OpenTelemetry captures key telemetry signals for OpenAI models, providing developers with an out-of-the-box observability solution tailored to AI workloads. The library, [hosted within the OpenTelemetry Python Contrib repo](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/opentelemetry-instrumentation-openai-v2%3D%3D2.0b0/instrumentation-genai/opentelemetry-instrumentation-openai-v2), automatically collects telemetry from OpenAI model interactions, including request and response metadata and token usage.
```

no metrics yet 🤐

content/en/blog/2024/otel-generative-ai/index.md
```diff
@@ -0,0 +1,89 @@
...
+```
+
+Then include the following code in your python application.
+
+```python
+from openai import OpenAI
+from opentelemetry.instrumentation.openai_v2 import OpenAIInstrumentor
+
+OpenAIInstrumentor().instrument()
+
```
lmolkova, line 59: we probably need to add the code that configure otel SDK (or add a comment pointing to docs on how to set it up).

The code snippet:

```python
# NOTE: OpenTelemetry Python Logs and Events APIs are in beta 
from opentelemetry import trace, _logs, _events
from opentelemetry.sdk.trace import TracerProvider
from opentelemetry.sdk._logs import LoggerProvider
from opentelemetry.sdk._events import EventLoggerProvider

from opentelemetry.sdk.trace.export import BatchSpanProcessor
from opentelemetry.sdk._logs.export import BatchLogRecordProcessor
from opentelemetry.exporter.otlp.proto.grpc.trace_exporter import OTLPSpanExporter
from opentelemetry.exporter.otlp.proto.grpc._log_exporter import OTLPLogExporter

trace.set_tracer_provider(TracerProvider())
trace.get_tracer_provider().add_span_processor(
    BatchSpanProcessor(OTLPSpanExporter())
)

_logs.set_logger_provider(LoggerProvider())
_logs.get_logger_provider().add_log_record_processor(BatchLogRecordProcessor(OTLPLogExporter()))
_events.set_event_logger_provider(EventLoggerProvider())
```

an existing example that does something similar - https://github.com/open-telemetry/opentelemetry-python/blob/stable/docs/examples/logs/example.py

/cc @lzchen @xrmx in case they have some suggestions

content/en/blog/2024/otel-generative-ai/index.md
```diff
@@ -0,0 +1,89 @@
...
+)
+
+# The library captures telemetry, including request and response metadata, token usage, and more.
+```
+
+With this simple instrumentation, one can begin capture traces from their generative AI application. Here is an example from the Aspire Dashboard for local debugging.
+
+![Chat trace in Aspire Dashboard](aspire_dashboard_trace.png)
+
+Here is a similar trace captured in Jaeger:
```
lmolkova, line 73:
```suggestion
Here is a similar trace captured in [Jaeger](https://www.jaegertracing.io/docs/next-release-v2/getting-started/#running):
```

content/en/blog/2024/otel-generative-ai/aspire_dashboard_trace.png
lmolkova, line 1: nit: maybe crop the image to remove non-essential details?

image

content/en/blog/2024/otel-generative-ai/index.md
drewby: That's a lot of extra code. Should I link to that code example or is there a better doc?

content/en/blog/2024/otel-generative-ai/index.md
```diff
@@ -0,0 +1,142 @@
...
+linkTitle: OpenTelemetry for Generative AI
+date: 2024-11-09
+author: >-
+  [Drew Robbins](https://github.com/drewby) (Microsoft),  [Liudmila
+  Molkova](https://github.com/lmolkova) (Microsoft)
+---
+
+As organizations increasingly adopt Large Language Models (LLMs) and other
+generative AI technologies, ensuring reliable performance, efficiency, and
+safety is essential. Enhanced observability—tracking AI operations, behaviors,
```
theletterf, line 12: > ensuring reliable performance, efficiency, and safety is essential.

Why?

content/en/blog/2024/otel-generative-ai/index.md
```diff
@@ -0,0 +1,142 @@
...
+## Key Signals for Generative AI
+
+The
+[Semantic Conventions for Generative AI](https://github.com/open-telemetry/semantic-conventions/tree/v1.28.0/docs/gen-ai)
+focus on capturing insights into AI model behavior through three primary
+signals: Spans, Metrics, and Events.
+
+**Spans: Tracing Model Interactions**  
+Spans track each model interaction’s lifecycle, covering input parameters (e.g.,
+temperature, top_p) and response details like token count or errors. They
```
theletterf, line 43:
```suggestion
Spans track each model interaction’s lifecycle, covering input parameters (for
example, temperature, top_p) and response details like token count or errors. They
```

content/en/blog/2024/otel-generative-ai/index.md
```diff
@@ -0,0 +1,142 @@
...
+**Events: Capturing Detailed Interactions**  
+Events log detailed moments during model execution, such as user prompts and
+model responses, providing a granular view of model interactions. These insights
+are invaluable for debugging and optimizing AI applications where unexpected
+behaviors may arise.
+
+Together, these signals provide a comprehensive monitoring framework, enabling
+better cost management, performance tuning, and request tracing.
+
+**Extending Observability with Vendor-Specific Attributes** The Semantic
```
theletterf, line 62: Is the part in bold meant to be a title? Consider using ### markers for each of the subsections.

theletterf (commented):
Cool one! Added some comments.

content/en/blog/2024/otel-generative-ai/index.md
lmolkova: let me add it to https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2988 and, once merged, let's just link it.

content/en/blog/2024/otel-generative-ai/index.md
drewby: I updated the intro to answer Why.

codefromthecrypt:
The below is about if we can make it zero code or not, and what's remaining to do so:

---

I want to run `dotenv run -- opentelemetry-instrument python main.py` and not have any telemetry code in the foreground or even noticed (zero code approach)

I have a similar example, I've tried locally, and I don't see a way to implicitly configure the logging provider, yet. I'm not sure if we want to make a hybrid to reduce the amount of code or just leave the explicit tracing and logging stuff in until logging can be env configured.

cc @anuraaga and @xrmx in case I got below wrong also

requirements
```
openai~=1.54.3

opentelemetry-sdk==1.28.1
opentelemetry-exporter-otlp-proto-http==1.28.1
opentelemetry-distro==0.49b1
opentelemetry-instrumentation-openai-v2==2.0b0
```

env
```
# Override default ENV variables for Ollama
OPENAI_BASE_URL=http://localhost:11434/v1
OPENAI_API_KEY=unused
CHAT_MODEL=qwen2.5:0.5b

# Set to false or remove to disable log events
OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT=true

OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4318
OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf
OTEL_SERVICE_NAME=python-opentelemetry-openai
OTEL_METRIC_EXPORT_INTERVAL=100
OTEL_BSP_SCHEDULE_DELAY=100
```

Best I could manage was to add hooks only for the log/event stuff

```python
import os

from openai import OpenAI

# NOTE: OpenTelemetry Python Logs and Events APIs are in beta
from opentelemetry import _logs, _events
from opentelemetry.sdk._logs import LoggerProvider
from opentelemetry.sdk._events import EventLoggerProvider

from opentelemetry.sdk._logs.export import BatchLogRecordProcessor
from opentelemetry.exporter.otlp.proto.http._log_exporter import OTLPLogExporter

_logs.set_logger_provider(LoggerProvider())
_logs.get_logger_provider().add_log_record_processor(BatchLogRecordProcessor(OTLPLogExporter()))
_events.set_event_logger_provider(EventLoggerProvider())

def main():
    client = OpenAI()
    messages = [
        {
            "role": "user",
            "content": "Answer in up to 3 words: Which ocean contains the falkland islands?",
        },
    ]
    model = os.getenv("CHAT_MODEL", "gpt-4o-mini")
    chat_completion = client.chat.completions.create(model=model, messages=messages)
    print(chat_completion.choices[0].message.content)


if __name__ == "__main__":
    main()

```

then, I get a warning about overriding the event provider, but at least the events do show up
```bash
$ dotenv run -- opentelemetry-instrument python main.py
Overriding of current EventLoggerProvider is not allowed
Indian Ocean
```

codefromthecrypt (commented):
I guess what's out of scope is metrics, as they aren't implemented yet. One concern is that folks follow this, then later when metrics are implemented, they need more instructions, or people need to remember to go back and change the docs etc.

Since there are recent developments like asyncapi etc going in fairly quickly, if metrics were added quickly also, would it make sense to hold the blog until they are released? or would it make more sense to do a second blog and revisit the setup instructions once that's supported?

drewby:
> I guess what's out of scope is metrics, as they aren't implemented yet. One concern is that folks follow this, then later when metrics are implemented, they need more instructions, or people need to remember to go back and change the docs etc.
> 
> Since there are recent developments like asyncapi etc going in fairly quickly, if metrics were added quickly also, would it make sense to hold the blog until they are released? or would it make more sense to do a second blog and revisit the setup instructions once that's supported?

Metrics are discussed as part of the semantic conventions, but correct, they are not yet implemented in the library yet. I think it's worth getting an article out there earlier than later. We might even attract some contributors for the Metrics implementation.

content/en/blog/2024/otel-generative-ai/aspire-dashboard-content-capture.png
svrnm, line 1: Is it possible to use Jaeger for this screenshot? See: ["_prefer CNCF projects over non-CNCF projects (e.g. use Jaeger for trace visualization, and Prometheus for metric visualization)_"](https://opentelemetry.io/docs/contributing/blog/#before-submitting-a-blog-post)

content/en/blog/2024/otel-generative-ai/aspire-dashboard-trace.png
svrnm, line 1: Is it possible to use Jaeger for this screenshot? See: ["_prefer CNCF projects over non-CNCF projects (e.g. use Jaeger for trace visualization, and Prometheus for metric visualization)_"](https://opentelemetry.io/docs/contributing/blog/#before-submitting-a-blog-post)

content/en/blog/2024/otel-generative-ai/index.md
```diff
@@ -0,0 +1,147 @@
...
+generative AI technologies, ensuring reliable performance, efficiency, and
+safety is essential to meet user expectations, optimize resource costs, and
+safeguard against unintended outputs. Effective observability for AI operations,
+behaviors, and outcomes can help meet these goals. OpenTelemetry is being
+enhanced to support these needs specifically for generative AI.
+
+Two primary assets are in development to make this possible: **Semantic
+Conventions** and an **Instrumentation Library**.
+
+**Semantic Conventions** establish standardized guidelines for how telemetry
```
svrnm, line 20:
```suggestion
[**Semantic Conventions** ](https://opentelemetry.io/docs/concepts/semantic-conventions/) establish standardized guidelines for how telemetry
```

I'm fine with the inline explanation (and maybe we can even take some if and improve the existing concept page), but link to docs pages where possible, so people can dig deeper if they want to learn more

codefromthecrypt:
> We might even attract some contributors for the Metrics implementation.

good point!

content/en/blog/2024/otel-generative-ai/index.md
```diff
@@ -0,0 +1,147 @@
...
+data is structured and collected across platforms, defining inputs, outputs, and
+operational details. For generative AI, these conventions streamline monitoring,
+troubleshooting, and optimizing AI models by standardizing attributes such as
+model parameters, response metadata, and token usage. This consistency supports
+better observability across tools, environments, and APIs, helping organizations
+track performance, cost, and safety with ease.
+
+The **Instrumentation Library** is being developed within the OpenTelemetry
+Python Contrib project to automate telemetry collection for generative AI
+applications. The first release is a Python library, given Python’s widespread
```
svrnm, line 30:
```suggestion
The [**Instrumentation Library**](https://opentelemetry.io/docs/specs/otel/overview/#instrumentation-libraries) is being developed within the [OpenTelemetry Python Contrib](https://github.com/open-telemetry/opentelemetry-python-contrib) project to automate telemetry collection for generative AI
applications. The first release is a Python library, given Python’s widespread
```

content/en/blog/2024/otel-generative-ai/index.md
```diff
@@ -0,0 +1,147 @@
...
+
+As organizations increasingly adopt Large Language Models (LLMs) and other
+generative AI technologies, ensuring reliable performance, efficiency, and
+safety is essential to meet user expectations, optimize resource costs, and
+safeguard against unintended outputs. Effective observability for AI operations,
+behaviors, and outcomes can help meet these goals. OpenTelemetry is being
+enhanced to support these needs specifically for generative AI.
+
+Two primary assets are in development to make this possible: **Semantic
+Conventions** and an **Instrumentation Library**.
```
svrnm, line 18:
```suggestion
Conventions** and an **Instrumentation Libraries**.
```

Is there only one library that is developed or are there multiple libraries?

content/en/blog/2024/otel-generative-ai/index.md
```diff
@@ -0,0 +1,147 @@
...
+operational details. For generative AI, these conventions streamline monitoring,
+troubleshooting, and optimizing AI models by standardizing attributes such as
+model parameters, response metadata, and token usage. This consistency supports
+better observability across tools, environments, and APIs, helping organizations
+track performance, cost, and safety with ease.
+
+The **Instrumentation Library** is being developed within the OpenTelemetry
+Python Contrib project to automate telemetry collection for generative AI
+applications. The first release is a Python library, given Python’s widespread
+use in AI development. Designed to integrate seamlessly with OpenAI’s API, this
```
svrnm, line 31: Same question as above, is there only one library (for OpenAI) or multiple? If there is only one make it much clearer that "the instrumentation library" is "the instrumentation library for OpenAI"

content/en/blog/2024/otel-generative-ai/index.md
```diff
@@ -0,0 +1,147 @@
...
+library captures spans and events, gathering essential data like model
+inputs, response metadata, and token usage in a structured format.
+
+## Key Signals for Generative AI
+
+The
+[Semantic Conventions for Generative AI](https://github.com/open-telemetry/semantic-conventions/tree/v1.28.0/docs/gen-ai)
```
svrnm, line 38:
```suggestion
[Semantic Conventions for Generative AI](/docs/specs/semconv/gen-ai/)
```

content/en/blog/2024/otel-generative-ai/index.md
```diff
@@ -0,0 +1,147 @@
...
+focus on capturing insights into AI model behavior through three primary
+signals: Spans, Metrics, and Events.
```
svrnm, line 40:
```suggestion
signals: [Traces](/docs/concepts/signals/traces/), [Metrics](/docs/concepts/signals/metrics/), and [Logs](/docs/concepts/signals/logs/).
```

Re events: I am fine with that as well, but people will stumble upon that and ask themselves "why events and not logs", so you would need to add some explanation and links about it. I know that this is a highly debated subject within the community but this blog post will be consumed by a lot of people that do not know all the things happening within OTel so we need to be cautious to put words out there without adding some more context.

content/en/blog/2024/otel-generative-ai/index.md
```diff
@@ -0,0 +1,147 @@
...
+
+### Spans: Tracing Model Interactions
```
svrnm, line 42:
```suggestion
### Traces: Tracing Model Interactions
```

content/en/blog/2024/otel-generative-ai/index.md
```diff
@@ -0,0 +1,147 @@
...
+
+Spans track each model interaction’s lifecycle, covering input parameters (for
+example, temperature, top_p) and response details like token count or errors.
```
svrnm, line 45:
```suggestion
Traces track each model interaction’s lifecycle, covering input parameters (for
example, temperature, top_p) and response details like token count or errors.
```

content/en/blog/2024/otel-generative-ai/index.md
```diff
@@ -0,0 +1,147 @@
...
+They provide visibility into each request, aiding in identifying bottlenecks and
+analyzing the impact of settings on model output.
+cd
```
svrnm, line 48:
```suggestion

```

content/en/blog/2024/otel-generative-ai/index.md
```diff
@@ -0,0 +1,147 @@
...
+token counts, essential for managing costs and performance. This data is
+particularly critical for API-dependent AI applications with rate limits and
+cost considerations.
+
+### Events: Capturing Detailed Interactions
+
+Events log detailed moments during model execution, such as user prompts and
+model responses, providing a granular view of model interactions. These insights
+are invaluable for debugging and optimizing AI applications where unexpected
+behaviors may arise.
```
svrnm, line 61: See my comment above around logs vs events

content/en/blog/2024/otel-generative-ai/index.md
```diff
@@ -0,0 +1,147 @@
...
+
+Together, these signals provide a comprehensive monitoring framework, enabling
+better cost management, performance tuning, and request tracing.
```
svrnm, line 64: Since you used sub headings this feels like a part of "Events: Capturing Detailed Interactions", you can move it up or think about explaining those three in bullet points instead.

content/en/blog/2024/otel-generative-ai/index.md
```diff
@@ -0,0 +1,147 @@
...
+
+Here’s an example of using the OpenTelemetry Python library to monitor a
+generative AI application with the OpenAI client. Make sure you first install
+the library:
+
+```bash
+pip install opentelemetry-instrumentation-openai-v2
+```
+
+Then include the following code in your Python application.
```
svrnm, line 98:
```suggestion
Then include the following code in your Python application:
```

content/en/blog/2024/otel-generative-ai/index.md
```diff
@@ -0,0 +1,147 @@
...
+generative AI application. Here is an example from the
+[Aspire Dashboard](https://learn.microsoft.com/dotnet/aspire/fundamentals/dashboard/standalone?tabs=bash)
+for local debugging.
+
+![Chat trace in Aspire Dashboard](aspire-dashboard-trace.png)
+
+Here is a similar trace captured in
+[Jaeger](https://www.jaegertracing.io/docs/next-release-v2/getting-started/#running):
+
+![Chat trace in Jaeger](jaeger-trace.png)
```
svrnm, line 125: Limit this to the Jaeger example (see my comment above around prefering CNCF projects over other projects).

content/en/blog/2024/otel-generative-ai/index.md
```diff
@@ -0,0 +1,147 @@
+---
+title: OpenTelemetry for Generative AI
+linkTitle: OpenTelemetry for Generative AI
+date: 2024-11-09
```
svrnm, line 4: Putting this comment here to keep an eye on setting the date right when we finally publish. Do not resolve.

```suggestion
date: 2024-11-09
```

svrnm (changes requested):
Great read, thanks for writing this!

content/en/blog/2024/otel-generative-ai/index.md
```diff
@@ -0,0 +1,147 @@
...
+author: >-
+  [Drew Robbins](https://github.com/drewby) (Microsoft),  [Liudmila
+  Molkova](https://github.com/lmolkova) (Microsoft)
+---
```
svrnm, line 8:
```suggestion
issue: 
sig: SIG GenAI Observability
---
```

@drewby can you create an issue with the same comment as you have here in the PR description, so we can link it to this blog post. Since this PR has everything needed (SIG + sponsors) we can keep it, but we want to establish this practice so let's set a good example

content/en/blog/2024/otel-generative-ai/aspire-dashboard-content-capture.png
drewby: The Events (which are Logs, not Spans) that show the captured content do not show in Jaeger which only supports spans (AFAIK).

content/en/blog/2024/otel-generative-ai/aspire-dashboard-trace.png
drewby: We wanted to show two different tools. We have both the Aspire trace view and the Jaeger trace view in the document.

content/en/blog/2024/otel-generative-ai/index.md
drewby: There is only one at this moment. We focused on Python first.

content/en/blog/2024/otel-generative-ai/index.md
drewby: will do

content/en/blog/2024/otel-generative-ai/index.md
drewby: Jeager doesn't show the Events/Logs we are using for Content Capture. Unless I'm missing how to view these events in Jaeger. @lmolkova, do you know a way?

content/en/blog/2024/otel-generative-ai/index.md
drewby: We were thinking to pin the version here to prevent potential broken links in the future. cc @lmolkova

content/en/blog/2024/otel-generative-ai/index.md
```diff
@@ -0,0 +1,163 @@
...
+considerations.
+
+### Events: Capturing Detailed Interactions
+
+Events log detailed moments during model execution, such as user prompts and
+model responses, providing a granular view of model interactions. These insights
+are invaluable for debugging and optimizing AI applications where unexpected
+behaviors may arise.
+
+> Note that we decided to use the newer Events API
```
drewby, line 75: @svrnm I've added this blockquote as a note on our use of the Events API. Does this resolve the issue you are raising around logs vs events?

content/en/blog/2024/otel-generative-ai/aspire-dashboard-content-capture.png
svrnm: thanks, if this is not possible with Jaeger, then there is no concern from my site.

content/en/blog/2024/otel-generative-ai/aspire-dashboard-trace.png
svrnm: See my comment above, if jaeger can not show the data you want to represent in your image, then it's fine to use a different tool. 

To add some context to this: we get a lot of blog posts suggested for blog(dot)opentelemetry(dot)io and we need to make sure that those blog posts are in line with our [social media guidelines](https://github.com/open-telemetry/community/blob/main/social-media-guide.md) and especially our project's goal to be vendor neutral. So if a blog makes use of a non-CNCF-project and a project that is affiliated with the vendor of the blog post authors, we need to verify that there is a reason for that beyond linking (and by that advertising) this project. Otherwise it becomes much harder for us (Docs Maintainers) to hold up our principles in future discussions in future blog posts. That jaeger (or another CNCF) project can not represent the data you'd like to show as of today, is a valid reason for me.

content/en/blog/2024/otel-generative-ai/index.md
svrnm: then maybe write this here already:

```suggestion
Conventions** and an **Instrumentation Library** for OpenAI's APIs:
```

content/en/blog/2024/otel-generative-ai/index.md
svrnm: Yes, that's perfect 🙌

content/en/blog/2024/otel-generative-ai/index.md
```diff
@@ -0,0 +1,163 @@
...
+> (https://opentelemetry.io/docs/specs/otel/logs/event-api/) specification in
+> the Semantic Conventions for Generative AI. The events API allows for us to
+> define specific
+> [semantic conventions](https://opentelemetry.io/docs/specs/semconv/general/events/)
+> for the user prompts and model responses that we capture.
```
svrnm, line 80:
```suggestion
{{% alert title="Note" color="info" %}}

Note that we decided to use the newer Events API
(https://opentelemetry.io/docs/specs/otel/logs/event-api/) specification in
the Semantic Conventions for Generative AI. The events API allows for us to
define specific
[semantic conventions](https://opentelemetry.io/docs/specs/semconv/general/events/)
for the user prompts and model responses that we capture.

{{% /alert %}}
```

While the `>` accomplish something similar, let's use the [alert shortcode](https://www.docsy.dev/docs/adding-content/shortcodes/#alert)

content/en/blog/2024/otel-generative-ai/index.md
```diff
@@ -0,0 +1,163 @@
...
+For generative AI, these conventions streamline monitoring, troubleshooting, and
+optimizing AI models by standardizing attributes such as model parameters,
+response metadata, and token usage. This consistency supports better
+observability across tools, environments, and APIs, helping organizations track
+performance, cost, and safety with ease.
+
+The
+[**Instrumentation Library**](https://opentelemetry.io/docs/specs/otel/overview/#instrumentation-libraries)
+is being developed within the
+[OpenTelemetry Python Contrib](https://github.com/open-telemetry/opentelemetry-python-contrib)
```
gyliu513, line 34: how about link to https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/instrumentation-genai?

content/en/blog/2024/otel-generative-ai/index.md
codefromthecrypt: If we made this change, it would conflate the GenAI working group assets with OpenAI, which I think doesn't follow later. Also, it feels less attractive to folks who are not OpenAI based, yet should be welcome into the team. Also, the specs are used for non-openai right now. It is just that the first library instrumentation in the shared codebase is openai.

If we want to hint what's first, I think we should make that a second sentence. wdyt?

```suggestion
Conventions** and an **Instrumentation Library**. The first instrumentation library targets OpenAI in Python.
```

content/en/blog/2024/otel-generative-ai/index.md
codefromthecrypt: agree on pinning as this has drifted in the past, unless there's a process in the website to find bad links and automatically raise PRs to retarget them.

content/en/blog/2024/otel-generative-ai/index.md
codefromthecrypt: I feel strongly we should say "Log events API" as it is a subclass of logging, and if you don't configure logging properly nothing will work. This is particularly important because status quo before was span events.

codefromthecrypt:
update on https://github.com/open-telemetry/opentelemetry.io/pull/5575#issuecomment-2467118675 based on feedback from @lzchen. We can pare down the code impact to a few lines, assuming you have env like below:

```
OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4318
OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf
OTEL_SERVICE_NAME=python-opentelemetry-openai

# Set to false or remove to disable log events
OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT=true
OTEL_PYTHON_LOGGING_AUTO_INSTRUMENTATION_ENABLED=true
OTEL_LOGS_EXPORTER=otlp_proto_http
```

This part will have to stick around until we have an ENV variable to control the event logger provider (or it is enabled by default).
```python
# NOTE: OpenTelemetry Python Log Events APIs is in beta
from opentelemetry import _events
from opentelemetry.sdk._events import EventLoggerProvider
_events.set_event_logger_provider(EventLoggerProvider())
```

Correct me, if I'm wrong, but in any case this seems very close in terms of making zero code work.

codefromthecrypt:
raised https://github.com/open-telemetry/opentelemetry-python/issues/4269 to hopefully sort out the manual code last mile

content/en/blog/2024/otel-generative-ai/aspire-dashboard-trace.png
drewby: Got it, I understand the concern around the guidelines. Thanks for the explanation and if we find a way to show in jaeger before we publish, I'll update the reference.

content/en/blog/2024/otel-generative-ai/index.md
drewby: We have the link later in the doc when we talk about the specifics of the library.

content/en/blog/2024/otel-generative-ai/index.md
drewby: @codefromthecrypt I hear you, however I don't see it referred to that way elsewhere? Maybe it is, I just didn't see it.

svrnm: The [Events API](https://opentelemetry.io/docs/specs/otel/logs/event-api/) states that _"this existing Events API and once it is defined and implemented, the Events API usage will be migrated, deprecated, renamed and eventually removed."_ 

Reading this note (we use the new events api) and then reading that it is going to be deprecated, is confusing for an end user. 

@lmolkova @tedsuo any guidance here?

content/en/blog/2024/otel-generative-ai/index.md
svrnm: We have a process to find bad links automatically, also this is a blog post not a documentation page, so it's expected for them to decay, see https://github.com/open-telemetry/opentelemetry.io/issues/3493, cc @chalin

content/en/blog/2024/otel-generative-ai/index.md
codefromthecrypt: @drewby it is in the log directory and many places it has the word log in the doc. We should fix the heading in the doc in my opinion if that's the only way to sort out this blog to highlight that this is log events not span events

codefromthecrypt: signal wise, we have log, metrics, spans, profile. If we want instead to literally say "events api" we should introduce it as a log signal. I personally jumping straight to "events api" doesn't prepare folks that this uses log signals, and a single word helps untangle this. EOF!

svrnm:
> raised [open-telemetry/opentelemetry-python#4269](https://github.com/open-telemetry/opentelemetry-python/issues/4269) to hopefully sort out the manual code last mile

just to verify, is this issue blocking this PR?

content/en/blog/2024/otel-generative-ai/index.md
gyliu513: Yes, Line 94 is a link for a specified link for openai instrumentation, but here should be the folder link of `instrumentation-genai`?

```suggestion
[OpenTelemetry Python Contrib](https://github.com/open-telemetry/opentelemetry-python-contrib) under `[instrumentation-genai](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/instrumentation-genai)`
```

drewby:
> > raised [open-telemetry/opentelemetry-python#4269](https://github.com/open-telemetry/opentelemetry-python/issues/4269) to hopefully sort out the manual code last mile
> 
> just to verify, is this issue blocking this PR?

No, I think it may get simpler with a no code instrumentation, but the code in the blog will continue to work.

codefromthecrypt:
> just to verify, is this issue blocking this PR?

not my call, but I think putting best foot forward is getting the "zero code" approach in the blog. Having otel infra code in the foreground of the project is not ideal, nor is having something look like it might work without code, then to try the demo and find out it doesn't work. I would go for best==zero code, and push for a patch. It is a small change to get us over the line and worth doing.,

content/en/blog/2024/otel-generative-ai/index.md
```diff
@@ -0,0 +1,163 @@
...
+
+Here’s an example of using the OpenTelemetry Python library to monitor a
+generative AI application with the OpenAI client. Make sure you first install
+the library:
+
+```bash
+pip install opentelemetry-instrumentation-openai-v2
+```
+
+Then include the following code in your Python application:
```
codefromthecrypt, line 114: this example is currently misleading, as it looks like it is zero code. The large setup logic is not visible in the blog. This is why I strongly don't agree with the optimization of time vs clean.

There are new facets people need to plumb for this to work as expected. People in genai expect to see prompt/completion data. We should make it work (zero code approach) or make very visible the code required to have that happen.

.cspell/en-words.txt
svrnm, line 1: Please move those words into a cspell ignore of your index.md

content/en/blog/2024/otel-generative-ai/index.md
```diff
@@ -0,0 +1,163 @@
+---
+title: OpenTelemetry for Generative AI
+linkTitle: OpenTelemetry for Generative AI
+date: 2024-11-09
+author: >-
+  [Drew Robbins](https://github.com/drewby) (Microsoft),  [Liudmila
+  Molkova](https://github.com/lmolkova) (Microsoft)
+issue: [#5581](https://github.com/open-telemetry/opentelemetry.io/issues/5581)
+sig: SIG GenAI Observability
+---
```
svrnm, line 10:
```suggestion
sig: SIG GenAI Observability
cSpell:ignore: genai liudmila molkova
---
```

content/en/blog/2024/otel-generative-ai/index.md
```diff
@@ -0,0 +1,163 @@
...
+automatically collects telemetry from OpenAI model interactions, including
+request and response metadata and token usage.
+
+As generative AI applications grow, additional instrumentation libraries for
+other languages will follow, extending OpenTelemetry support across more tools
+and environments. The current library’s focus on OpenAI highlights its
+popularity and demand within AI development, making it a valuable initial
+implementation.
+
+### Example Usage
```
dasiths, line 104: Would be good to show the metrics being captures as well. The name of the metric and the attributes captured along with it.

content/en/blog/2024/otel-generative-ai/index.md
```diff
@@ -0,0 +1,163 @@
...
+![Content Capture Aspire Dashboard](aspire-dashboard-content-capture.png)
+
+## Join Us in Shaping the Future of Generative AI Observability
+
+Community collaboration is key to OpenTelemetry’s success. We invite developers,
+AI practitioners, and organizations to contribute, share feedback, or
+participate in discussions. Explore the OpenTelemetry Python Contrib project,
+contribute code, or help shape observability for AI as it continues to evolve.
+More information can be found at the
+[Generative AI Observability project page](https://github.com/open-telemetry/community/blob/main/projects/gen-ai.md).
```
gyliu513, line 163: @drewby How about highlighting the current contributors at the end of the blog? This way, readers can quickly see who’s actively involved in this SIG.

```suggestion
[Generative AI Observability project page](https://github.com/open-telemetry/community/blob/main/projects/gen-ai.md), we now have contributors from OpenLIT, Langtrace, Elastic, MicroSoft, Traceloop, IBM etc., welcome to join the community!
```

Rutledge: Happy to include Scorecard as a supporter and adopter of GenAI OTEL!

drewby: > @drewby How about highlighting the current contributors at the end of the blog? This way, readers can quickly see who’s actively involved in this SIG.

It's a good suggestion and I'd be delighted to include how we've had great collaboration and contribution across all these companies. 

@svrnm I'm not sure how this fits with the community/blog guidelines?

content/en/blog/2024/otel-generative-ai/index.md
drewby: We are not capturing metrics in the instrumentation library yet. Good place to contribute! ;)

content/en/blog/2024/otel-generative-ai/index.md
```diff
@@ -0,0 +1,183 @@
...
+
+# The library captures telemetry, including request and response metadata, token usage, and more.
+```
+
+With this simple instrumentation, one can begin capture traces from their
+generative AI application. Here is an example from the
+[Aspire Dashboard](https://learn.microsoft.com/dotnet/aspire/fundamentals/dashboard/standalone?tabs=bash)
+for local debugging.
+
+![Chat trace in Aspire Dashboard](aspire-dashboard-trace.png)
```
codefromthecrypt, line 156: I noticed both of these traces are not from the sample code, as they have spans that it wouldn't produce. Is this ok?

drewby:
> update on [#5575 (comment)](https://github.com/open-telemetry/opentelemetry.io/pull/5575#issuecomment-2467118675) based on feedback from @lzchen. We can pare down the code impact to a few lines, assuming you have env like below:

@codefromthecrypt I have updated the code example. Can you take a look?

content/en/blog/2024/otel-generative-ai/index.md
```diff
@@ -0,0 +1,186 @@
...
+_events.set_event_logger_provider(EventLoggerProvider())
+
+from openai import OpenAI
+from opentelemetry.instrumentation.openai_v2 import OpenAIInstrumentor
+
+OpenAIInstrumentor().instrument()
+
+client = OpenAI()
+response = client.chat.completions.create(
+    model="gpt-4-mini",
```
codefromthecrypt, line 147: At least for me, I had to switch this for my account to work

```suggestion
    model="gpt-4o-mini",
```

codefromthecrypt (commented):
I think this needs more work and another person to double-check. I checked personally noticing the images don't match the main code. It failed in jaeger, but I don't think it is about jaeger as the failure also happens in [otel-tui](https://github.com/ymtdzzz/otel-tui). I think the manual setup has a glitch in it and we shouldn't publish this blog until the code here matches the images and also it doesn't produce exceptions exporting logs. If it is my mistake, please let me know what I did wrong.

I'm wondering if someone else can double-check this example as-is because it failed for me against jaeger v2 like so:

```bash
$ OPENAI_API_KEY=not-tellin python main.py
Overriding of current EventLoggerProvider is not allowed
Exception while exporting logs.
Traceback (most recent call last):
  File "/Users/adriancole/oss/otel-ollama/examples/python/opentelemetry/openai/.venv/lib/python3.12/site-packages/opentelemetry/sdk/_logs/_internal/export/__init__.py", line 307, in _export_batch
    self._exporter.export(self._log_records[:idx])  # type: ignore
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/adriancole/oss/otel-ollama/examples/python/opentelemetry/openai/.venv/lib/python3.12/site-packages/opentelemetry/exporter/otlp/proto/grpc/_log_exporter/__init__.py", line 111, in export
    return self._export(batch)
           ^^^^^^^^^^^^^^^^^^^
  File "/Users/adriancole/oss/otel-ollama/examples/python/opentelemetry/openai/.venv/lib/python3.12/site-packages/opentelemetry/exporter/otlp/proto/grpc/exporter.py", line 300, in _export
    request=self._translate_data(data),
            ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/adriancole/oss/otel-ollama/examples/python/opentelemetry/openai/.venv/lib/python3.12/site-packages/opentelemetry/exporter/otlp/proto/grpc/_log_exporter/__init__.py", line 108, in _translate_data
    return encode_logs(data)
           ^^^^^^^^^^^^^^^^^
  File "/Users/adriancole/oss/otel-ollama/examples/python/opentelemetry/openai/.venv/lib/python3.12/site-packages/opentelemetry/exporter/otlp/proto/common/_internal/_log_encoder/__init__.py", line 37, in encode_logs
    return ExportLogsServiceRequest(resource_logs=_encode_resource_logs(batch))
                                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/adriancole/oss/otel-ollama/examples/python/opentelemetry/openai/.venv/lib/python3.12/site-packages/opentelemetry/exporter/otlp/proto/common/_internal/_log_encoder/__init__.py", line 71, in _encode_resource_logs
    pb2_log = _encode_log(sdk_log)
              ^^^^^^^^^^^^^^^^^^^^
  File "/Users/adriancole/oss/otel-ollama/examples/python/opentelemetry/openai/.venv/lib/python3.12/site-packages/opentelemetry/exporter/otlp/proto/common/_internal/_log_encoder/__init__.py", line 57, in _encode_log
    body=_encode_value(log_data.log_record.body),
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/adriancole/oss/otel-ollama/examples/python/opentelemetry/openai/.venv/lib/python3.12/site-packages/opentelemetry/exporter/otlp/proto/common/_internal/__init__.py", line 90, in _encode_value
    raise Exception(f"Invalid type {type(value)} of value {value}")
Exception: Invalid type <class 'NoneType'> of value None
```

I used the following requirements.txt
```
openai~=1.54.3

# 1.28.1 is required for Log Events API/SDK
opentelemetry-sdk~=1.28.1
opentelemetry-exporter-otlp-proto-grpc~=1.28.1
opentelemetry-distro~=0.49b1
opentelemetry-instrumentation-openai-v2~=2.0b0
```

and I used exactly the code as listed here:
```python
from opentelemetry import trace, _logs, _events
from opentelemetry.sdk.trace import TracerProvider
from opentelemetry.sdk._logs import LoggerProvider
from opentelemetry.sdk._events import EventLoggerProvider

from opentelemetry.sdk.trace.export import BatchSpanProcessor
from opentelemetry.sdk._logs.export import BatchLogRecordProcessor
from opentelemetry.exporter.otlp.proto.grpc.trace_exporter import OTLPSpanExporter
from opentelemetry.exporter.otlp.proto.grpc._log_exporter import OTLPLogExporter

trace.set_tracer_provider(TracerProvider())
trace.get_tracer_provider().add_span_processor(
    BatchSpanProcessor(OTLPSpanExporter())
)

_logs.set_logger_provider(LoggerProvider())
_logs.get_logger_provider().add_log_record_processor(BatchLogRecordProcessor(OTLPLogExporter()))
_events.set_event_logger_provider(EventLoggerProvider())

from openai import OpenAI
from opentelemetry.instrumentation.openai_v2 import OpenAIInstrumentor

OpenAIInstrumentor().instrument()

client = OpenAI()
response = client.chat.completions.create(
    model="gpt-4o-mini",
    messages=[{"role": "user", "content": "Write a short poem on OpenTelemetry."}],
)
```

For my jaeger setup I needed to use the following config because you can no longer listen on all points via ENV variable. I actually just wanted to override the listen ports, but don't know how to do that so I merged the receiver config with a file in the repo. cc @yurishkuro 

```yaml
service:
  extensions: [jaeger_storage, jaeger_query, remote_sampling, healthcheckv2, expvar]
  pipelines:
    traces:
      receivers: [otlp, jaeger, zipkin]
      processors: [batch]
      exporters: [jaeger_storage_exporter]
  telemetry:
    resource:
      service.name: jaeger
    metrics:
      level: detailed
      address: 0.0.0.0:8888
    logs:
      level: info
    # TODO Initialize telemetery tracer once OTEL released new feature.
    # https://github.com/open-telemetry/opentelemetry-collector/issues/10663

extensions:
  jaeger_query:
    storage:
      traces: some_storage

  jaeger_storage:
    backends:
      some_storage:
        memory:
          max_traces: 100000

  remote_sampling:
    # We can either use file or adaptive sampling strategy in remote_sampling
    file:
      path: ./cmd/jaeger/sampling-strategies.json
    # adaptive:
    #   sampling_store: some_store
    #   initial_sampling_probability: 0.1
    http:
    grpc:

  healthcheckv2:
    use_v2: true
    http:
      endpoint: "0.0.0.0:13133"
    grpc:

  expvar:
    port: 27777

receivers:
  otlp:
    protocols:
      grpc:
        endpoint: 0.0.0.0:4317
      http:
        endpoint: 0.0.0.0:4318

  jaeger:
    protocols:
      grpc:
      thrift_binary:
      thrift_compact:
      thrift_http:

  zipkin:

processors:
  batch:

exporters:
  jaeger_storage_exporter:
    trace_storage: some_storage
```

content/en/blog/2024/otel-generative-ai/index.md
```diff
@@ -0,0 +1,186 @@
...
+
+With this simple instrumentation, one can begin capture traces from their
+generative AI application. Here is an example from the
+[Aspire Dashboard](https://learn.microsoft.com/dotnet/aspire/fundamentals/dashboard/standalone?tabs=bash)
+for local debugging.
+
+![Chat trace in Aspire Dashboard](aspire-dashboard-trace.png)
+
+Here is a similar trace captured in
+[Jaeger](https://www.jaegertracing.io/docs/next-release-v2/getting-started/#running):
```
yurishkuro, line 162:
```suggestion
[Jaeger](https://www.jaegertracing.io/docs/2.0/getting-started/#running):
```

yurishkuro:
There should be no need for this config, jaeger has it by default

drewby:
> I think this needs more work and another person to double-check. I checked personally noticing the images don't match the main code. It failed in jaeger, but I don't think it is about jaeger as the failure also happens in [otel-tui](https://github.com/ymtdzzz/otel-tui). I think the manual setup has a glitch in it and we shouldn't publish this blog until the code here matches the images and also it doesn't produce exceptions exporting logs. If it is my mistake, please let me know what I did wrong.
> 
> I'm wondering if someone else can double-check this example as-is because it failed for me against jaeger v2 like so:

I've updated the code sample to something similar to yours. Would love to reduce the amount of code, but also want to make sure it works. 

Do you think we need to include the requirements and/or pip install instructions for openai, opentelemetry, etc?

codefromthecrypt:
So, the last code I pasted works fine, but only if you also set `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT=true`

Through trial and error, what I figured out is that if you have the boilerplate python to configure logging, but you don't also set `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT=true`, then the stack trace below comes out. 
This seems like a bug as it should fallback safely right?

```
Overriding of current EventLoggerProvider is not allowed
Exception while exporting logs.
Traceback (most recent call last):
  File "/Users/adriancole/oss/otel-ollama/examples/python/opentelemetry/openai/.venv/lib/python3.12/site-packages/opentelemetry/sdk/_logs/_internal/export/__init__.py", line 307, in _export_batch
    self._exporter.export(self._log_records[:idx])  # type: ignore
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/adriancole/oss/otel-ollama/examples/python/opentelemetry/openai/.venv/lib/python3.12/site-packages/opentelemetry/exporter/otlp/proto/grpc/_log_exporter/__init__.py", line 111, in export
    return self._export(batch)
           ^^^^^^^^^^^^^^^^^^^
  File "/Users/adriancole/oss/otel-ollama/examples/python/opentelemetry/openai/.venv/lib/python3.12/site-packages/opentelemetry/exporter/otlp/proto/grpc/exporter.py", line 300, in _export
    request=self._translate_data(data),
            ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/adriancole/oss/otel-ollama/examples/python/opentelemetry/openai/.venv/lib/python3.12/site-packages/opentelemetry/exporter/otlp/proto/grpc/_log_exporter/__init__.py", line 108, in _translate_data
    return encode_logs(data)
           ^^^^^^^^^^^^^^^^^
  File "/Users/adriancole/oss/otel-ollama/examples/python/opentelemetry/openai/.venv/lib/python3.12/site-packages/opentelemetry/exporter/otlp/proto/common/_internal/_log_encoder/__init__.py", line 37, in encode_logs
    return ExportLogsServiceRequest(resource_logs=_encode_resource_logs(batch))
                                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/adriancole/oss/otel-ollama/examples/python/opentelemetry/openai/.venv/lib/python3.12/site-packages/opentelemetry/exporter/otlp/proto/common/_internal/_log_encoder/__init__.py", line 71, in _encode_resource_logs
    pb2_log = _encode_log(sdk_log)
              ^^^^^^^^^^^^^^^^^^^^
  File "/Users/adriancole/oss/otel-ollama/examples/python/opentelemetry/openai/.venv/lib/python3.12/site-packages/opentelemetry/exporter/otlp/proto/common/_internal/_log_encoder/__init__.py", line 57, in _encode_log
    body=_encode_value(log_data.log_record.body),
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/adriancole/oss/otel-ollama/examples/python/opentelemetry/openai/.venv/lib/python3.12/site-packages/opentelemetry/exporter/otlp/proto/common/_internal/__init__.py", line 90, in _encode_value
    raise Exception(f"Invalid type {type(value)} of value {value}")
Exception: Invalid type <class 'NoneType'> of value None

```

codefromthecrypt:
so current status is I can get the example working on otel-tui, but only if I export that variable:

```bash
$ OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT=true  OPENAI_API_KEY=sk-not_telling python main.py
Overriding of current EventLoggerProvider is not allowed
```

Screenshot 2024-11-14 at 1 21 36 PM


That said, I get a failure on jaeger v2, but @yurishkuro might want to double-check me as maybe I messed something up.
```bash
$ OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT=true  OPENAI_API_KEY=sk-not_telling python main.py
Overriding of current EventLoggerProvider is not allowed
Failed to export logs to localhost:4317, error code: StatusCode.UNIMPLEMENTED
```

content/en/blog/2024/otel-generative-ai/index.md
```diff
@@ -122,29 +122,47 @@ appropriate:
 OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4318
 OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf
 OTEL_SERVICE_NAME=python-opentelemetry-openai
+OPENAI_API_KEY=<replace_with_your_openai_api_key>
 
 # Set to false or remove to disable log events
 OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT=true
```
codefromthecrypt, line 128: main thing here is that if you set this to false, it will crash (at least for me)

content/en/blog/2024/otel-generative-ai/index.md
```diff
@@ -0,0 +1,204 @@
...
+# NOTE: OpenTelemetry Python Log Events APIs is in beta
+from opentelemetry.sdk.trace import TracerProvider
+from opentelemetry.sdk._logs import LoggerProvider
+from opentelemetry.sdk._events import EventLoggerProvider
+from opentelemetry.trace import set_tracer_provider
+from opentelemetry._logs import set_logger_provider
+from opentelemetry._events import set_event_logger_provider
+from opentelemetry.sdk.trace.export import SimpleSpanProcessor
+from opentelemetry.sdk._logs.export import SimpleLogRecordProcessor
+from opentelemetry.exporter.otlp.proto.grpc.trace_exporter import OTLPSpanExporter
```
codefromthecrypt, line 144: while grpc is ok I'm not sure if all vendors support that. Also it requires platform deps to build. Should we switch this to http instead?

codefromthecrypt (commented):
@drewby FYI I re-pasted the code from your latest commit and it doesn't change the error when content capture is set to false.

The main thing that's a discrepancy are the traces. The code only produces a single span, and both examples have multiple spans. Probably least confusing to reproduce the traces based on the code in the example, so people don't wonder why their trace has only one vs the three in the images. WDYT?

on pip install vs requirements, it is fine either way as long as the end results work. The main issue is that pip install is only doing the `opentelemetry-instrumentation-openai-v2` part, so it is missing the other dependencies I mentioned in my requirements.txt

Make sense?

As mentioned earlier, best case is we can use the zero code bootstrap thing which reduces the amount of commands and visible code. So the above are mostly around the manual trace infra code approach.

content/en/blog/2024/otel-generative-ai/index.md
```diff
@@ -0,0 +1,204 @@
...
+
+```bash
+pip install opentelemetry-instrumentation-openai-v2
+```
+
+Set the following environment variables, updating the endpoint and protocol as
+appropriate:
+
+```bash
+OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4318
```
codefromthecrypt, line 122: TL;DR; IMHO whatever we paste in a bash codeblock, people will expect to work by pasting on their own. This doesn't work, yet even if we fix the pip deps due to not exporting the variables.

maybe worth removing these two `OTEL_EXPORTER` variables and leave it to defaults based on the dependency like `opentelemetry-exporter-otlp-proto-grpc` or `opentelemetry-exporter-otlp-proto-http`. wdyt?

Then, we have only the service name, api key and capture content ones. Also, these need to be exported or set inline. If you don't the python command won't work later. Many people put these into a .env file and use dotenv to load it.

e.g. 

```bash
$ pip install "python-dotenv[cli]"
$ dotenv run -- python main.py
```

However, we can also just inline them like

```bash
$ OTEL_SERVICE_NAME=python-opentelemetry-openai OPENAI_API_KEY=sk-not_tellin OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT=true python main.py
```
```

File: content/en/blog/2024/otel-generative-ai/index.md (hunk 34 above)
Comment (codefromthecrypt) [reply content/en/blog/2024/otel-generative-ai/index.md:128]: @lzchen @xrmx can one of you double-check the stack trace which happens when this is set to false or unset? I'm not sure if this is already fixed in main or not.

https://github.com/open-telemetry/opentelemetry.io/pull/5575#issuecomment-2475434466

Comment (codefromthecrypt) [reply content/en/blog/2024/otel-generative-ai/index.md:128]: https://github.com/open-telemetry/opentelemetry-python/pull/4276 or similar will fix this, though not sure if all backends expect a log with no body or not.

**Comment by codefromthecrypt** [comment]
follow-up for jaeger https://github.com/jaegertracing/documentation/issues/778

**Comment by codefromthecrypt** [comment]
here's example code which would stay in draft until the upcoming release of [1.28.2](https://github.com/open-telemetry/opentelemetry-python/issues/4281) and any feedback https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3006

**Review by codefromthecrypt — COMMENTED** [review]
Here's some food for thought on jaeger which only accepts span data. Since the demo can send logs and traces, configuration care is needed.

Notably, `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT=false` doesn't disable log export, rather it just pares down the log events such that they have no body. So, this means you will still get errors as it attempts to send log messages to jaeger where they aren't implemented.

One way is to make an ENV that comments out everything here, so that it works with trace-only backends, but still shows how to use it in aspire. That said, this makes more steps for folks to see the prompt/completion data and that's a big part of this instrumentation. Another way is to leave the demo setup in a way that is optimized for collectors that accept logs, and mention this implies commenting this part out if yours doesn't.
```
# Comment out the following if your backend doesn't support OTLP logs
OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT=true
OTEL_PYTHON_LOGGING_AUTO_INSTRUMENTATION_ENABLED=true
OTEL_LOGS_EXPORTER=otlp_proto_http
```

Regardless, when the code settles someone should give second eyes to the Aspire approach which appears to accept all signals. I've not yet.

File: content/en/blog/2024/otel-generative-ai/index.md (hunk 35 above)
Comment (codefromthecrypt) [reply content/en/blog/2024/otel-generative-ai/index.md:144]: I see now, it is because aspire only supports grpc https://learn.microsoft.com/en-us/dotnet/aspire/fundamentals/dashboard/standalone?tabs=bash#configure-opentelemetry-sdk

Comment (codefromthecrypt) [reply content/en/blog/2024/otel-generative-ai/index.md:144]: @drewby that link isn't correct afterall, as aspire dashboard also supports http. I suggest changing to http in the final example as that lets you use smaller images like alpine without installing gcc which is needed for the grpc package.

e.g. this works if we use http

```
# Use an alpine image to make the runtime smaller
FROM docker.io/python:3.12.7-alpine3.20
RUN python -m pip install --upgrade pip

COPY /requirements.txt /tmp/requirements.txt
RUN pip install -r /tmp/requirements.txt

COPY main.py /

CMD [ "opentelemetry-instrument", "python", "main.py" ]
```

Comment (codefromthecrypt) [reply content/en/blog/2024/otel-generative-ai/index.md:144]: specifically, the following let's you use http on localhost:4318

```bash
docker run --rm -it  -p 18888:18888     -p 4318:18890  --name aspire-dashboard     mcr.microsoft.com/dotnet/aspire-dashboard:9.0
```

codefromthecrypt:
example code is ready to go and tested three ways. Note, until the PR is merged links won't work https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3006

Feel free to revise this PR based on feedback around the example and otherwise.

content/en/blog/2024/otel-generative-ai/index.md
```diff
@@ -0,0 +1,187 @@
...
+    model=os.getenv("CHAT_MODEL", "gpt-4o-mini"),
+    messages=[
+        {
+            "role": "user",
+            "content": "Write a short poem on OpenTelemetry.",
+        },
+    ],
+)
+print(chat_completion.choices[0].message.content)
+```
```
codefromthecrypt, line 150: maybe you need a "finally, run the example with `opentelemetry-instrument`" part, because the zero code requires prefixing python with that

content/en/blog/2024/otel-generative-ai/index.md
```diff
@@ -0,0 +1,187 @@
...
+[OpenTelemetry Python Contrib](https://github.com/open-telemetry/opentelemetry-python-contrib)
+under
+[instrumentation-genai](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/instrumentation-genai)
+project to automate telemetry collection for generative AI applications. The
+first release is a Python library for instrumenting OpenAI client calls, given
+Python's widespread use in AI development and the popularity of OpenAI. Designed
+to integrate seamlessly with OpenAI's API, this library captures spans and
+events, gathering essential data like model inputs, response metadata, and token
+usage in a structured format.
+
```
SergeyKanzhelev, line 45: can we say that this article is not covering the model server? And maybe call to action for model server authors to look into standardizing metrics they expose?

codefromthecrypt (commented):
notes

content/en/blog/2024/otel-generative-ai/index.md
```diff
@@ -0,0 +1,193 @@
...
+        {
+            "role": "user",
+            "content": "Write a short poem on OpenTelemetry.",
+        },
+    ],
+)
+print(chat_completion.choices[0].message.content)
+```
+
+And then run the example using `opentelemetry-instrument`:
```
lzchen, line 152: I don't believe this example will work that is without installing all of the correct packages. Users would need `opentelemetry-distro` package as well. To make this truly zero code and easier for users to digest, we could add instructions similar to the current [zero-code](https://opentelemetry.io/docs/zero-code/python/#setup) example for Python. Specifically:

```
pip install opentelemetry-distro
opentelemetry-bootstrap -a install
```
Following [this-pr](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2996), openai instrumentation is automatically added to `opentelemetry-bootstrap`, which is designed to automatically install instrumentation libraries if users have the underlying instrumented library installed (openai in this case). 

tldr: The current example will not actually work properly without `opentelemetry-distro` installed. Replace `pip install opentelemetry-instrumentation-openai-v2` with `pip install opentelemetry-distro` and call `opentelemetry-bootstrap -a install`.

drewby: Updated the instructions. Thanks!

drewby:
@svrnm, I think we have all the conversations resolved. Any more feedback? 

cc @lmolkova

codefromthecrypt:
FYI I mentioned to drew offline my 2p: ready to go when others say so

content/en/blog/2024/otel-generative-ai/index.md
svrnm: I think that's fine, but I would suggest to order them alphabetically

svrnm (changes requested):
lgtm overall, some final comments

content/en/blog/2024/otel-generative-ai/index.md
gyliu513: Awesome, thanks @drewby !

content/en/blog/2024/otel-generative-ai/index.md
```diff
@@ -0,0 +1,200 @@
+---
+title: OpenTelemetry for Generative AI
+linkTitle: OpenTelemetry for Generative AI
```
chalin, line 3: Drop the stuttered `linkTitle` or shorten it, e.g.:

```suggestion
title: OpenTelemetry for Generative AI
linkTitle: OTel for Generative AI
```

My preference would be to drop it.

content/en/blog/2024/otel-generative-ai/index.md
```diff
@@ -0,0 +1,200 @@
...
+date: 2024-11-09
+author: >-
+  [Drew Robbins](https://github.com/drewby) (Microsoft),  [Liudmila
+  Molkova](https://github.com/lmolkova) (Microsoft)
+issue: [#5581](https://github.com/open-telemetry/opentelemetry.io/issues/5581)
```
chalin, line 8: This is breaking the build I believe. Here's a fix:

```suggestion
issue: https://github.com/open-telemetry/opentelemetry.io/issues/5581
```

chalin (changes requested):
Initial drive-by comments, with a suggestion for fixing the build:

```console
> hugo --cleanDestinationDir -e dev --buildDrafts --baseURL https://deploy-preview-5575--opentelemetry.netlify.app/ --minify
...
Error: error building site: process: readAndProcessContent: "/opt/build/repo/content/en/blog/2024/otel-generative-ai/index.md:8:1": failed to unmarshal YAML: yaml: line 8: did not find expected ',' or ']'
Failed during stage 'building site': Build script returned non-zero exit code: 2 (https://ntl.fyi/exit-code-2)
...
```

content/en/blog/2024/otel-generative-ai/index.md
lmolkova: Event API is being merged into Logs API - https://github.com/open-telemetry/opentelemetry-specification/blob/v1.39.0/specification/logs/api.md#emit-an-event.

So a future-proof way would be to say 

> we decided to use the log-based events

or

> we decided to use events emitted with Logs API

content/en/blog/2024/otel-generative-ai/index.md
```diff
@@ -0,0 +1,200 @@
...
+        },
+    ],
+)
+print(chat_completion.choices[0].message.content)
+```
+
+And then run the example using `opentelemetry-instrument`:
+
+```shell
+opentelemetry-instrument python main.py
```
lzchen, line 154: Is it expected for this to be exporting to console or otlp? Otlp is used by default but if we want a simple get started for customers perhaps this would be better:

```
opentelemetry-instrument --traces_exporter console --metrics_exporter console python .\server_automatic.py
```

lzchen: btw, if user does NOT set up collector and leaves the example as it is to use otlp by default they will get error messages.

content/en/blog/2024/otel-generative-ai/index.md
drewby: thanks!

content/en/blog/2024/otel-generative-ai/index.md
drewby: I added it as an additional example. We don't have metrics yet, so I just included traces option. Thanks!

lzchen: Nice! If you don't want to include metrics, you have to explicitly set to None.

`opentelemetry-instrument --traces_exporter console --metrics_exporter None python main.py`

drewby: In that case, I'll just make it console

content/en/blog/2024/otel-generative-ai/index.md
drewby: I've updated the Alert text based on @lmolkova feedback.

chalin:
This PR will need to be manually rebased (after pulling the latest from GitHub):

> image

drewby:
> This PR will need to be manually rebased (after pulling the latest from GitHub):
> 

Up to date now.

content/en/blog/2024/otel-generative-ai/index.md
```diff
@@ -0,0 +1,209 @@
...
+opentelemetry-instrument --traces_exporter console --metrics_exporter console python main.py
+```
+
+There is a complete example
+[available here](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/instrumentation-genai/opentelemetry-instrumentation-openai-v2/example).
+
+With this simple instrumentation, one can begin capture traces from their
+generative AI application. Here is an example from the
+[Aspire Dashboard](https://learn.microsoft.com/dotnet/aspire/fundamentals/dashboard/standalone?tabs=bash)
+for local debugging.
```
lmolkova, line 172: since we're using `4318` port and `otlp_proto_http` in the example, we should provide the (non-default) way to run aspire dashboard with

```bash
docker run --rm -it -d \
    -p 18888:18888 \
    -p 4317:18889 \
    -p 4318:18890 \
    --name aspire-dashboard \
    mcr.microsoft.com/dotnet/aspire-dashboard:9.0
```

The `-p 4318:18890` is not mentioned in the Aspire docs (as most of otel defaults to grpc protocol on 4317 port).

The alternative is to update the sample here to match python example being updated - 
https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2988/

It should be fine either way, let's just make sure we provide consistent guidance.

lmolkova (approved):

dmitryax (approved):

content/en/blog/2024/otel-generative-ai/index.md
drewby: I added a docker command for aspire dashboard and jaeger to keep the examples consistent.

svrnm: @lmolkova can you re-review the changes for that?

svrnm (approved):

content/en/blog/2024/otel-generative-ai/index.md
svrnm: I would like to have @chalin re-review the PR, and we shouldn't publish this week anyways due to thanksgiving

lmolkova (approved):
Thanks for following up! It looks great!

drewby:
@chalin can you give this one a review? I think we are ready to go once you approve.

codefromthecrypt:
hey @drewby seems the screen shots are still out of date as they don't match the code as mentioned before. Do you need a hand getting a new copy of these? or you can grab one from this https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3006

chalin:
Oops, wrong strategy for resolving the refcache conflict:

> image

Let me try again.

chalin (approved):
I defer to @svrnm et al. for the blog content. "Infrastrucutre" wise, this LGTM once all GH actions pass.

PR #5586, opened 2024-11-11
Add maven to registry auto update workflow

svrnm:
Companion PR for #5330: this updates the script auto updating the versions for the registry to also look into maven central. It includes one updated file for verification that it works.

Note that this also adds a minor fix to the script where it runs `npm run fix:format` before creating the PR, since this is required in ~95% of the cases and we have to manually run it.


chalin (approved):
LGTM on quick review.

PR #5552, opened 2024-11-07
Update opentelemetry-specification version to v1.39.0

opentelemetrybot:
Update opentelemetry-specification version to `v1.39.0`.

See https://github.com/open-telemetry/opentelemetry-specification/releases/tag/v1.39.0.

PR #5587, opened 2024-11-11
[CI] adjust-pages: report obsolete patch-code as INFO msgs

chalin:
- Followup to #5552
- Report obsolete patch-code as INFO msgs so that it doesn't fail the build -- so that we can test new spec version builds using the `buid-dev` workflow. The patch code should be removed after the spec is updated so that we can confirm that the generated site files are the same.


tiffany76 (approved):
RSLGTM.

Issue #5555, opened 2024-11-07
[i18n] Link-check failures due to missing heading IDs


chalin:
Heading IDs are missing in the non-`en` pages identified below, as can be seen from the log of https://github.com/open-telemetry/opentelemetry.io/actions/runs/11732757332/job/32685831165?pr=5554.

These must be fixed, and the temporary ignore rules removed from the `IgnoreDirs` section of https://github.com/open-telemetry/opentelemetry.io/blob/main/.htmltest.yml by deleting the corresponding rules from the appropriate page **front matter** (search for the regexs to find them). The following locales are affected:

- [ ] `es`
- [ ] `fr`
- [ ] `ja`
- [ ] `pt`
- [ ] `zh`

Details:

```
ja/docs/zero-code/php/index.html
  hash does not exist --- ja/docs/zero-code/php/index.html --> /ja/docs/concepts/context-propagation/#propagation
ja/docs/languages/ruby/sampling/index.html
  hash does not exist --- ja/docs/languages/ruby/sampling/index.html --> /ja/docs/concepts/sampling#head-sampling
ja/docs/languages/erlang/sampling/index.html
  hash does not exist --- ja/docs/languages/erlang/sampling/index.html --> /ja/docs/concepts/sampling#head-sampling
  hash does not exist --- ja/docs/languages/erlang/sampling/index.html --> /ja/docs/concepts/sampling/#head-sampling
ja/docs/concepts/glossary/index.html
  hash does not exist --- ja/docs/concepts/glossary/index.html --> /ja/docs/what-is-opentelemetry/#history
  hash does not exist --- ja/docs/concepts/glossary/index.html --> /ja/docs/what-is-opentelemetry/#history
ja/docs/languages/js/sampling/index.html
  hash does not exist --- ja/docs/languages/js/sampling/index.html --> /ja/docs/concepts/sampling#head-sampling
ja/docs/concepts/signals/baggage/index.html
  hash does not exist --- ja/docs/concepts/signals/baggage/index.html --> /ja/docs/concepts/context-propagation/#context
  hash does not exist --- ja/docs/concepts/signals/baggage/index.html --> /ja/docs/concepts/context-propagation/#propagation
ja/docs/concepts/components/index.html
  hash does not exist --- ja/docs/concepts/components/index.html --> /ja/docs/concepts/sampling/#head-sampling
pt/docs/zero-code/php/index.html
[...]

PR #5596, opened 2024-11-13
[zh] Update dev.md

youngledo:
style：优化翻译（仪器化被翻译成插桩）。


my-git9:
I think '插桩' can better express its true meaning in Chinese compared to '仪器 ' here, and it may need to be evaluated by the @open-telemetry/docs-zh-approvers  together

chalin:
Reminder to @open-telemetry/docs-zh-approvers that we are awaiting a fix of #5555 from your localization team before we can accept further `zh` page updates.

youngledo:
> I think '插桩' can better express its true meaning in Chinese compared to '仪器 ' here, and it may need to be evaluated by the @open-telemetry/docs-zh-approvers together

Perhaps so, I am considering closing this issue.

PR #5599, opened 2024-11-13
Update getting-started.md file(set FLASK_APP env before flask run.)

woyaodeyangguang:
set FLASK_APP env before flask run.

### DIFF
--- a/content/en/docs/languages/python/getting-started.md
+++ b/content/en/docs/languages/python/getting-started.md
@@ -78,6 +78,7 @@ Run the application with the following command and open
 <http://localhost:8080/rolldice> in your web browser to ensure it is working.
 
 ```sh
+export FLASK_APP=app.py
 flask run -p 8080
 ```
 


woyaodeyangguang:
@svrnm

svrnm:
> A module named `app` is the default for `flask run` so no need to add it explicitly I think

but it still works if it is set explicitly? People following the tutorial might not be well informed about flask so maybe this helps if they want to do something related but didn't name their file `app.py`. WDYT?

xrmx:
> > A module named `app` is the default for `flask run` so no need to add it explicitly I think
> 
> but it still works if it is set explicitly? People following the tutorial might not be well informed about flask so maybe this helps if they want to do something related but didn't name their file `app.py`. WDYT?

If you are going to add it explicitly, which I don't suggest, it would be calling `flask run app` I guess, not exporting an env variable.

tammy-baylis-swi:
> A module named app is the default for flask run so no need to add it explicitly I think

Yes, the Flask CLI should [automatically pick up](https://flask.palletsprojects.com/en/stable/cli/#application-discovery) the `app.py` that the getting-started is [currently describing](https://github.com/open-telemetry/opentelemetry.io/pull/5599/files#diff-cb920a8ad1b1daf71773962b67489a1bd1d7d99e452ec1ba2162e71f9b4822b0R50).

For the case of a new Flask/OTel user deviating from that instruction and naming the file something else, a better cue might be to specify `--app app` at the [call to run](https://github.com/open-telemetry/opentelemetry.io/pull/5599/files#diff-cb920a8ad1b1daf71773962b67489a1bd1d7d99e452ec1ba2162e71f9b4822b0L119). But I think it's sufficient that we're already [pointing to the Flask docs](https://github.com/open-telemetry/opentelemetry.io/pull/5599/files#diff-cb920a8ad1b1daf71773962b67489a1bd1d7d99e452ec1ba2162e71f9b4822b0R23) for more info about that framework.

Issue #5423, opened 2024-10-17
[ja|pt] Drop temporary link-checking rules


chalin:
- #5276 required that we disable link checking for some `ja` and `pt` pages
- This issue is a reminder to drop the temporary rule in https://github.com/open-telemetry/opentelemetry.io/blob/main/.htmltest.yml after the `ja` and `pt` pages are updated

### Tasks

- [x] `ja` page patch - #5609
- [x] `pt` page patch - #5603

/cc 
  @open-telemetry/docs-pt-approvers
  @open-telemetry/docs-ja-approvers

chalin:
@open-telemetry/docs-ja-approvers can you submit a PR similar to what the `pt` team did?

katzchang:
@chalin Thank you for your heads-up, I'll take a look!

PR #5603, opened 2024-11-13
[pt] Temporarily patch page to avoid link-check failure

emdneto:
Address #5423

### DIFF
--- a/content/pt/docs/concepts/instrumentation/libraries.md
+++ b/content/pt/docs/concepts/instrumentation/libraries.md
@@ -4,9 +4,7 @@ description: Aprenda como adicionar instrumentação nativa à sua biblioteca.
 weight: 40
 default_lang_commit: a570a00c7a238ffe26528d7bfb20efdbaf939c39
 htmltest:
-  IgnoreDirs:
-    # TODO drop next line after https://github.com/open-telemetry/opentelemetry.io/issues/5423 is fixed for pt pages:
-    - ^pt/docs/concepts/instrumentation/libraries/
+  IgnoreDirs: {}
 ---
 
 O OpenTelemetry fornece [bibliotecas de instrumentação][] para várias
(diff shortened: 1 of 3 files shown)


edsoncelio (approved):

EzzioMoreira (approved):

content/pt/docs/concepts/instrumentation/libraries.md
```diff
@@ -4,9 +4,7 @@ description: Aprenda como adicionar instrumentação nativa à sua biblioteca.
 weight: 40
 default_lang_commit: a570a00c7a238ffe26528d7bfb20efdbaf939c39
 htmltest:
-  IgnoreDirs:
-    # TODO drop next line after https://github.com/open-telemetry/opentelemetry.io/issues/5423 is fixed for pt pages:
-    - ^pt/docs/concepts/instrumentation/libraries/
+  IgnoreDirs: {}
```
svrnm, line 7: I think you can drop this attribute if it is empty, cc @chalin

chalin: Yes. Drop the `htmltest` entry entirely.

chalin: There's a broader issue to be addressed though. I'll comment separately from this thread.

chalin (changes requested):
The intent of #5423 is remind the two named locales to catch up with the updates made to the Java pages, not by eliminating the link, but by catching up with the `en` versions of the pages.

By making the change to the page without updating the `a570a00c7a238ffe26528d7bfb20efdbaf939c39`, you are breaking the correspondence.

chalin:
> The intent of #5423 is remind the two named locales to catch up with the updates made to the Java pages, not by eliminating the link, but by catching up with the `en` versions of the pages.
> 
> By making the change to the page without updating the `a570a00c7a238ffe26528d7bfb20efdbaf939c39`, you are breaking the correspondence.

@open-telemetry/docs-maintainers - we should decide if the "quick fix" in this PR, which introduces a drift/deviation from `a570a00c7a238ffe26528d7bfb20efdbaf939c39` is an acceptable deviation. I would prefer not, but given that there are more significant changes in the page, it might be an acceptable "temporary" patch so that we get restore full link checking. Thoughts on setting this precedent?

chalin (approved):
After some discussion with @open-telemetry/docs-maintainers, we've agreed to accept this "patch" PR. I've added a note to that effect to the `default_lang_commit` front-matter line.

Thanks for the PR @emdneto, and for your patience as we iron out protocol relative to i18n PRs.

chalin:
The i18n check is failing. Let me see if I can fix the checker to support comments on the same line as the hash.

PR #5602, opened 2024-11-13
Registry: add missing URL for Traefik

chalin:
- Contributes to #5601

/cc @svrnm

PR #5605, opened 2024-11-13
Update registry schema to allow any https URL for authors

svrnm:
https://github.com/open-telemetry/opentelemetry.io/pull/5602#discussion_r1840247361 -- this way we can also allow URLs for author.url in registry entries

### DIFF
--- a/data/registry-schema.json
+++ b/data/registry-schema.json
@@ -109,9 +109,9 @@
           },
           "url": {
             "type": "string",
-            "description": "The URL of the author",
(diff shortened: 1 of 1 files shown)


chalin (approved):

chalin:
Or you can maintain the requirement that it be a URL from GH, either of a user, or an org.

svrnm:
> Or you can maintain the requirement that it be a URL from GH, either of a user, or an org.

I think that's not sustainable in the long run. If we bring in vendors into the registry as well at some point we will need this anyways and their might be projects not hosted on github, etc.

PR #5619, opened 2024-11-18
Sort and update community members page

svrnm:
This adds a small fix to the script that updates the community membership pages. The results are sorted within each group before they are written to a file, that way we create some stability and future updates are easier to validate, since names are not jumping within the groups based on the updates and based on how the API returns results

cc @danielgblanco


chalin (approved):
LGTM once checks are green.

PR #5635, opened 2024-11-19
Add `uv` package manager instructions to Python zero-code docs

jord1e:
When using https://docs.astral.sh/uv/ the default instructions fail because the shell syntax and method of running scripts is a bit different.

pip install also doesnt work like it would normally, and you wouldn't really want to use it anyways because uv is faster and uses its own venvs

This PR adds instructions to the troubleshooting section.

### DIFF
--- a/content/en/docs/zero-code/python/_index.md
+++ b/content/en/docs/zero-code/python/_index.md
@@ -135,16 +135,15 @@ uv pip install opentelemetry-distro opentelemetry-exporter-otlp
 
 Now, you can install the auto instrumentation:
 
-```
+```sh
 uv run opentelemetry-bootstrap -a requirements | uv pip install --requirement -
 ```
 
 Finally, use `uv run` to start your application (see [Configuring the agent](#configuring-the-agent)):
 
-```
+```sh
 uv run opentelemetry-instrument python myapp.py
 ```
 
 Please note that you have to reinstall the auto instrumentation every time you run `uv sync` or update existing packages.
 It is therefore recommended to make the installation part of your build pipeline.
-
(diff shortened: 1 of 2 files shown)


tiffany76:
Thanks for writing up the solution, @jord1e! 

@open-telemetry/python-approvers, PTAL.

xrmx (approved):
LGTM a couple of nits

jord1e:
Fixed the nits, thanks @xrmx

chalin (commented):
This PR needs to be rebased before the `/fix` commands will work, and I think that we (maintainers) can't rebase it because of permissions issues, so you're going to have to do it @jord1e:

> image

emdneto (approved):

svrnm:
@jord1e we advise against using a github organization for creating forks, since they are not editable by maintainers by default, so we can neither rebase your branch nor use automation to fix any styling issues

There are 3 options now:
* Check if it is possible to make the PR editable: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork
* Do the rebase locally and run the fix commands locally (`npm run fix:all`)
* Close this PR and reopen it not using an organization (i.e. it comes from your user directly)

jord1e:
> @jord1e we advise against using a github organization for creating forks, since they are not editable by maintainers by default, so we can neither rebase your branch nor use automation to fix any styling issues

You learn something new every day -_-

---

I rebased locally and force-pushed (couldn't easily make it work otherwise) after running `npm run fix:all`

I'll move the fork to my normal account for any next PR's, apologies.

svrnm (approved):

Issue #5387, opened 2024-10-11
I think confmap providers should be added to the registry: they are conceptually similar to other kinds of components in the Collector in that they provide extra functionality to the Collector, can be added on a Collector builder manifest, can be 'official' or come from a third-party repository...


svrnm:
I think confmap providers should be added to the registry: they are conceptually similar to other kinds of components in the Collector in that they provide extra functionality to the Collector, can be added on a Collector builder manifest, can be 'official' or come from a third-party repository...

_Originally posted by @mx-psi in https://github.com/open-telemetry/opentelemetry.io/issues/5363#issuecomment-2407473568_

jackgopack4:
we also have added the ability to specify custom confmap converters with the latest release of v0.114.0 OpenTelemetry Collector Builder tool; I think it would make sense to add this type as well

PR #5637, opened 2024-11-19
add confmap provider to registry.json

jackgopack4:
As mentioned in #5387, there are more components that are compatible with the OpenTelemetry Collector.
These include `confmap.Converter` and `confmap.Provider` types. 

confmap.Converter: https://github.com/open-telemetry/opentelemetry-collector/issues/11582

confmap.Provider: https://github.com/open-telemetry/opentelemetry-collector/issues/4567

There are currently [five confmap.Providers that are maintained in the OpenTelemetry Collector core repository](https://github.com/open-telemetry/opentelemetry-collector/tree/main/confmap/provider) and while there is only one Converter there and it is deprecated, there are vendors that make use of converters and it is possible they would choose to publish them in the registry once this is available.

As I am not super familiar with the registry and changes necessary to update the same, I may have missed file(s) that need to be changed and will happily make these changes if suggested.

Additionally, I don't feel strongly about calling them "confmap converter" vs. "converter" or provider, but it seemed slightly more descriptive. Open to feedback there as well.

### DIFF
--- a/data/registry-schema.json
+++ b/data/registry-schema.json
@@ -17,7 +17,6 @@
       "description": "The type of the entry",
       "enum": [
         "application integration",
-        "confmap converter",
         "confmap provider",
         "core",
         "exporter",
--- a/data/registry-schema.json
+++ b/data/registry-schema.json
@@ -17,6 +17,8 @@
       "description": "The type of the entry",
       "enum": [
         "application integration",
+        "confmap converter",
+        "confmap provider",
         "core",
         "exporter",
         "extension",
--- a/data/registry/collector-confmap-provider-envprovider.yml
+++ b/data/registry/collector-confmap-provider-envprovider.yml
@@ -1,0 +1,21 @@
+# cSpell:ignore confmap, envprovider
+title: Collector Environment Variable Provider
+registryType: confmap provider
+language: collector
+tags:
+  - go
+  - confmap
+  - provider
+  - collector
+  - env
+license: Apache 2.0
+description: Environment variable provider for OpenTelemetry Collector confmaps.
+authors:
+  - name: OpenTelemetry Authors
+urls:
+  repo: https://github.com/open-telemetry/opentelemetry-collector/tree/main/confmap/provider/envprovider
+createdAt: 2024-11-20
+package:
+  registry: go-collector
+  name: go.opentelemetry.io/collector/confmap/provider/envprovider
+  version: v1.20.0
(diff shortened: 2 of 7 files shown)


jackgopack4:
I will also plan to submit a separate PR to add the existing five confmap.providers to the registry once this change is approved.

jackgopack4:
> Can you put all of them in one PR? Makes it easier to validate things and how we are going to name those values.

Yes, done.

songy23 (approved):
LGTM

data/registry-schema.json
```diff
@@ -17,6 +17,8 @@
       "description": "The type of the entry",
       "enum": [
         "application integration",
+        "confmap converter",
+        "confmap provider",
```
svrnm, line 21:
```suggestion
        "confmap provider",
```

let's add them incrementally when we need them so skip the `converter` for now. Also are there other types of providers as well? In the [OCB Documentation](https://opentelemetry.io/docs/collector/custom-collector/) they are loaded via `providers`, so I think we should make it consistent.

data/registry/collector-confmap-provider-envprovider.yml
```diff
@@ -0,0 +1,21 @@
...
+registryType: confmap provider
+language: collector
+tags:
+  - go
+  - confmap
+  - provider
+  - collector
+  - env
+license: Apache 2.0
+description: Environment variable provider for OpenTelemetry Collector confmaps.
```
svrnm, line 12:
```suggestion
description: Environment variable provider for OpenTelemetry Collector configuration maps.
```
Is `confmap` something a reader without intimate knowledge about the collector understands, knows that it is? Can we write it out as "configuration maps" to make it clearer?

svrnm (changes requested):
Thanks, see some questions and suggestions inlined

data/registry/collector-confmap-provider-envprovider.yml
jackgopack4: I think that's a good idea

data/registry-schema.json
jackgopack4: the providers are exclusively used in the confmap struct; we reference them as providers in some places and confmapProviders in other places in the codebase.
I am not aware of any other type of providers

jackgopack4:
ah shoot, this shows "confmap providers" as the yaml tag; is there any way to have the type listed as "Confmap provider" in the dropdown but show "providers" in the sample manifest?

Screenshot 2024-11-22 at 11 07 07 AM

PR #5653, opened 2024-11-21
add yottabyte in vendor list

vector1983:
add yottabyte in vendor list

### DIFF
--- a/data/ecosystem/vendors.yaml
+++ b/data/ecosystem/vendors.yaml
@@ -451,9 +451,9 @@
   contact: 'hello@groundcover.com'
   oss: false
   commercial: true
-- name: YottaByte
+- name: rizhiyi lunaxee by YottaByte
   nativeOTLP: true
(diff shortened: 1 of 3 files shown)


data/ecosystem/vendors.yaml
```diff
@@ -451,3 +451,9 @@
   contact: 'hello@groundcover.com'
   oss: false
   commercial: true
+- name: YottaByte
+  nativeOTLP: true
+  url: https://rizhiyi.com/lunaxee/
```
svrnm, line 456: what's the difference between YottaByte, rizhiyi and lunaxee? For an end user it is surprising to click a link that says "YottaByte" and then landing on a place with a very different name. If YottaByte is the company and the other ones are the product you can update the name to something like "Rizhiyi Lunaxee by YottaByte" or similar

vector1983: ok, I will change yottabyte to "rizhiyi lunaxee by yottabyte", thanks

PR #5666, opened 2024-11-22
Update instrumentation.md

EkkoLee98:
import express, { Request, Express } from 'express';

I think 'Request' is not needed.  To mislead readers, we should delete it.


maryliag (approved):
Thank you for your contribution! Good catch!
LGTM!

svrnm (approved):

PR #5679, opened 2024-11-25
adding LogEase's product GuanChaYi in vendor list

vector1983:
The LogEase Observability Monitoring Platform(GuanChaYi) is an independently developed monitoring platform by LogEase, which can support otel natively. Adding it in the vendor list.

### DIFF
--- a/data/ecosystem/vendors.yaml
+++ b/data/ecosystem/vendors.yaml
@@ -457,3 +457,9 @@
   contact: contact@rizhiyi.cn
   oss: false
   commercial: true
+- name: GuanChaYi by LogEase
+  nativeOTLP: true
+  url: https://logease.cn/en-logease-observability-platform/
+  contact: contact@yottabyte.cn
+  oss: false
+  commercial: true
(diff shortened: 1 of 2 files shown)


data/ecosystem/vendors.yaml
```diff
@@ -457,3 +457,9 @@
   contact: contact@rizhiyi.cn
   oss: false
   commercial: true
+- name: GuanChaYi by LogEase
+  nativeOTLP: true
+  url: https://logease.cn/en-logease-observability-platform/
+  contact: contact@yottabyte.cn
+  oss: false
+  commercial: true
```
svrnm, line 465: how is this different from the entry for rizhiyi? the pages use the same language and if I switch languages I can go back and forth between LogEase and rizhiyi

svrnm (changes requested):

data/ecosystem/vendors.yaml
vector1983: Already updated, would you please help to review it, thanks @svrnm

svrnm (approved):

PR #5683, opened 2024-11-25
fix confmap provider name

jackgopack4:
This changes the name for Collector Confmap Providers in the registry from "confmap provider" to "provider".

The way the registry shows how to add a component, `confmap provider`, is not the correct way to include a provider with your collector build. While the name is more descriptive, without knowing a way to have a different display name in the drop-down, it seems better to name it `provider` instead to ensure a correct code snippet.

Screenshot 2024-11-22 at 11 07 07 AM

### DIFF
--- a/data/registry-schema.json
+++ b/data/registry-schema.json
@@ -17,13 +17,13 @@
       "description": "The type of the entry",
       "enum": [
         "application integration",
-        "confmap provider",
         "core",
         "exporter",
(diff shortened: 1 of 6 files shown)


cartermp (approved):
Ah, that's a good catch. I agree that it's much better to show the correct code sample than to have a slightly more descriptive name. Thanks.

PR #5697, opened 2024-11-26
fix: Update filelog receiver operators in kubernetes docs

TylerHelmuth:
This filelog receiver example was out-of-date now that the container operator exists.  This brings the example in line with the default value we use in the helm chart.


cartermp (approved):
sick

PR #5703, opened 2024-11-28
Fix ansible role name misspelling

Troels51:
The ansible role is named opentelemetry_collector without the extra r

I found this spelling error while searching for an example of setting up the collector with ansible.


tiffany76 (approved):
Good eye! Thanks for the fix, @Troels51.

svrnm (approved):

PR #5709, opened 2024-11-30
Update java config docs for OTLP endpoint

jaydeluca:
## Context

My intent is to clarify the configuration options for the `otel.exporter.otlp.endpoint` configuration option. I think the current description could be interpreted to mean that It is needed to supply the `v1/{signal}` postfix for the endpoint, when that is only required when using the `otel.exporter.otlp.{signal}.endpoint` option.

## What is being changed

Updated description for `otel.exporter.otlp.endpoint` removing references to the `v1/{signal}` postfix, and updated the reference to the default value of `http://localhost:4318` instead of `http://localhost:4318/v1/{signal}`

Updated the description for `otel.exporter.otlp.{signal}.endpoint` to include the note about needing to include the postfix to the provided endpoint


breedx-splk (approved):
I agree, this helps to clarify the settings. Thanks @jaydeluca !

tiffany76 (approved):

PR #5729, opened 2024-12-05
Add Docker and Kubernetes examples to DoS section

### LINKED ISSUE #5706: Include examples on how to set up endpoints in security best practices doc
**What needs to be changed?** Add open-telemetry/opentelemetry-collector/pull/10562 changes to opentelemetry.io

**What is the name + path of the page that needs changed?**  /docs/security/config-best-practices/#protect-against-denial-of-service-attacks

@tiffany76 Do you think this makes sense to copy as is?

tiffany76:
This PR adds to the security documentation several end-user examples for setting up secure Collector endpoints in Docker and Kubernetes. The content is pulled from [this README](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/security-best-practices.md#safeguards-against-denial-of-service-attacks).

Fixes #5706


content/en/docs/security/config-best-practices.md
```diff
@@ -115,10 +115,129 @@ addresses so the network functions properly in dual-stack environments and
...
+```
+
+In your `docker run` command, use the `--hostname` argument to bind the
+Collector to the `my-hostname` address. You can access the Collector from
+outside that Docker network (for example, on a regular program running on the
+host) by connecting to `127.0.0.1:4567`. Here is an example `docker run`
+command:
+
+```shell
+docker run --hostname my-hostname --name container-name -p 127.0.0.1:4567:4317 otel/opentelemetry-collector:0.104.0
```
tiffany76, line 142: @mx-psi Should we update this command with the most recent Collector version? Same question for lines 155 and 196.

svrnm (approved):

content/en/docs/security/config-best-practices.md
mx-psi: Yes, that makes sense to me, we can always point to the latest version

mx-psi (approved):

content/en/docs/security/config-best-practices.md
svrnm: you should be able to pick up the version with a shortcode, I don't have it in my head right now, but maybe it's used in some of the other collector pages already and you can copy it here

tiffany76: Thanks, I found the short code.

content/en/docs/security/_index.md
```diff
@@ -1,5 +1,7 @@
 ---
 title: Security
+cascade:
+  vers: 0.115.1
```
chalin, line 4: If this is the collector version then:

- The key should make that clear; e.g., use a name like `collector_vers`
- You'll need to modify the auto-version update script(s) to ensure that this parameter gets updated.

chalin (commented):
Drive-by comments, see inline.

content/en/docs/security/_index.md
tiffany76: Thank you! I was wondering how this shortcode worked. I think I updated the correct script, but if you could take a look, I'd appreciate it. 

For a straightforward copy-paste PR, I learned a lot! :smile:

svrnm (approved):
lgtm

PR #5733, opened 2024-12-05
Update accepted exporters in SDK configuration

macno:
Add missing `console` exporter to the accepted list of exporters in all signals.   
Also add a note in the description that the value may be a comma-separated list


cartermp (approved):
Thanks, good catch!

PR #5734, opened 2024-12-05
Add Delivery Hero to OpenTelemetry adopters

ekovalenk0:
Delivery Hero uses OpenTelemetry collectors and SDKs since at least a year (some parts of the company longer), hence it makes sense to add us to the list of OpenTelemetry adopters ;)


data/ecosystem/adopters.yaml
```diff
@@ -20,6 +20,11 @@
   components: [Collector, Python]
   reference: ''
   contact: ''
+- name: Delivery Hero
+  url: https://www.deliveryhero.com/
+  components: [Collector, Go, Java]
+  reference: ''
```
svrnm, line 26: Would you have a reference like a talk or blog post we can link out to? This helps potential adopters to learn from your experience

svrnm (changes requested):
Thanks.

data/ecosystem/adopters.yaml
```diff
@@ -20,6 +20,11 @@
...
+  contact: 'https://github.com/ekovalenk0'
```
chalin, line 27: Btw, here and elsewhere, we don't need quotes around URLs.
```suggestion
  contact: https://github.com/ekovalenk0
```

data/ecosystem/adopters.yaml
ekovalenk0: There is no blogpost. There was a talk at the OTel meetup in Berlin but AFAIK it was not recorded. Once there is any material available, I will link it here ;)

svrnm (approved):

Issue #5745, opened 2024-12-09
Rust Getting Started Docs Error - the trait bound `opentelemetry_stdout::SpanExporter: opentelemetry_sdk::export::trace::SpanExporter` is not satisfied


TownCube:
**What needs to be changed?**
Update Rust Getting Started example so it compiles.

**What is the name + path of the page that needs changed?** https://opentelemetry.io/docs/languages/rust/getting-started/

**Additional context**: The error I get after copy/pasting the getting started example:

```
error[E0277]: the trait bound `opentelemetry_stdout::SpanExporter: opentelemetry_sdk::export::trace::SpanExporter` is not satisfied
   --> src/dice_server.rs:40:31
    |
40  |         .with_simple_exporter(SpanExporter::default())
    |          -------------------- ^^^^^^^^^^^^^^^^^^^^^^^ the trait `opentelemetry_sdk::export::trace::SpanExporter` is not implemented for `opentelemetry_stdout::SpanExporter`
    |          |
    |          required by a bound introduced by this call
    |
help: trait impl with same name found
   --> /home/TownCube/.cargo/registry/src/index.crates.io-6f17d22bba15001f/opentelemetry-stdout-0.27.0/src/trace/exporter.rs:33:1
    |
33  | impl opentelemetry_sdk::export::trace::SpanExporter for SpanExporter {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    = note: perhaps two different versions of crate `opentelemetry_sdk` are being used?
note: required by a bound in `opentelemetry_sdk::trace::Builder::with_simple_exporter`
   --> /home/TownCube/.cargo/registry/src/index.crates.io-6f17d22bba15001f/opentelemetry_sdk-0.22.1/src/trace/provider.rs:161:36
    |
161 |     pub fn with_simple_exporter<T: SpanExporter + 'static>(self, exporter: T) -> Self {
    |                                    ^^^^^^^^^^^^ required by this bound in `Builder::with_simple_exporter`
```

Bumping the versions seems to resolve it but didn't want to blindly raise a PR to update without understanding the ramifications in case this break is a sign of something bigger not working correctly.

```suggestion
opentelemetry = "0.27.1"
opentelemetry_sdk = "0.27.1"
opentelemetry-stdout = { version = "0.27.0", features = ["trace"] }
```

[...]

PR #5760, opened 2024-12-11
add crates support to registry version auto updater

svrnm:
fixes https://github.com/open-telemetry/opentelemetry.io/issues/5745 via automation. This PR will update the workflow and the workflow will update the versions 

cc cijothomas


cijothomas (approved):

cijothomas:
@svrnm Just realized this is not going to work well until we declare 1.0 stable! As public APIs are still being changed, so if we bump version alone without code changes as required, we'll be broken...

cijothomas (changes requested):
Need to hold off until 1.0 stable!

PR #5776, opened 2024-12-12
Update metrics.md - fix typo on Gauge (a-synchronous)

### LINKED ISSUE #5774: Metrics
In https://opentelemetry.io/docs/concepts/signals/metrics/:

> **Gauge**: Measures a current value at the time it is read. An example would be the fuel gauge in a vehicle. Gauges are **asynchronous**.
> **Asynchronous Gauge**: Same as the Gauge, but is collected once for each export. Could be used if you don’t have access to the continuous changes, but only to the aggregated value.

Are Gauge asynchronous or synchronous ? It looks like it's a typo ?

c2main:
Description of Gauge was misleading due to a typo between asynchronous and synchronous.

Fix #5774


tiffany76 (approved):

svrnm:
thanks @c2main

Issue #5786, opened 2024-12-13
[CI] Rework and optimize workflows


chalin:
- Once the following lands, we should make a pass over the GH workflows to cleanup and optimize them to use `npm ci`
- This would also be an opportunity to use the trick employed in #5780 to other workflows, such as text and markdown lint jobs

## Tasks

- [ ] #5787
- [ ] ...

svrnm:
There are 2 other things I'd like to see with the workflows, and since the issue title is generic enough for that, here they are:

* move all scripts from .github/workflows/scripts to ./scripts
* use https://github.com/peter-evans/create-pull-request/ instead of per-script custom logic to create and maintain pull requests

wdyt?

chalin:
Yes, I've been wanting to move the workspace scripts into `./scripts` for a while now too.

I'm not familiar with that GH action, but I'm willing to give it a try! Thanks for suggesting it.

svrnm:
> I'm not familiar with that GH action, but I'm willing to give it a try! Thanks for suggesting it.

See the community update action where it is in use. What I like about it that it is also able to update existing PRs, which would especially be helpful for the daily registry updates, so instead of having 1 per day there would be one until merged and then a new one. 

The other advantage would be that by removing the per-script logging of committing and creating PRs the scripts are easier to use outside of the workflows. 

I would try to find some time next year to start with the registry workflow and then we can look into the other ones if it is worth the effort

vitorvasc:
@chalin @svrnm, is this still valid?

svrnm:
> [@chalin](https://github.com/chalin) [@svrnm](https://github.com/svrnm), is this still valid?

I suspect yes, but @chalin would know better than me

PR #5787, opened 2024-12-13
[CI] Commit package-lock.json

### LINKED ISSUE #5784: [CI] Commit package-lock.json file?
I've resisted the idea of committing the `package-lock.json` file out of concern that it would cause headaches during merge and rebases, especially for long-standing PRs (which we have enough of to be of concern.

But experience with a file with a similar profile (`refcache.json`) has proven to be not as bad as I expected. Given that, I'd like to suggest that we give it a try: i.e., commit `package-lock.json`. We should probably agree on the (maleable) convention that only site maintainers and approvers should update the lock file.

Of course, after a trial period, we can always fall back to our current situation, or even pushing things further and making `package.json` into a substitute lock file by pinning packages to exact version, which is what we do currently for Hugo.

**Pros** of committing the lock file: it can help us optimize our CI, and simplify some workflows. No extra lock file to maintain since it gets generated dynamically as needed.

**Cons**: as mentioned before, it can complicate merges and rebases, esp. for long-standing PRs. Another con is that we need to "maintain" the lock file.

Thoughts?

chalin:
- Fixes #5784
- Contributes to #5786

### DIFF
--- a/.github/workflows/check-format.yml
+++ b/.github/workflows/check-format.yml
@@ -18,16 +18,10 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Create NPM cache-hash input file
-        run: |
-          mkdir -p tmp
-          jq '{devDependencies, engines, gitHubActionCacheKey}' package.json > tmp/package-ci.json
-
       - uses: actions/setup-node@v4
         with:
           node-version-file: .nvmrc
           cache: npm
-          cache-dependency-path: tmp/package-ci.json
 
       - name: Install package(s)
         run: |
--- a/.cspell.yml
+++ b/.cspell.yml
@@ -55,3 +55,4 @@ dictionaries:
   - companies
 words: # Valid words across all locales
   - htmltest
+  - nvmrc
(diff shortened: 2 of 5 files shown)


.github/workflows/check-format.yml
```diff
@@ -18,16 +18,10 @@ jobs:
...
-      - name: Create NPM cache-hash input file
-        run: |
-          mkdir -p tmp
-          jq '{devDependencies, engines, gitHubActionCacheKey}' package.json > tmp/package-ci.json
-
       - uses: actions/setup-node@v4
         with:
           node-version-file: .nvmrc
           cache: npm
-          cache-dependency-path: tmp/package-ci.json
```
chalin, line 30: `package-lock.json` will be used by default (IIRC)

chalin:
Ah, more care will be needed in using this approach. Putting this in draft, I'll come back to it later.

vitorvasc:
Since this dates back to 2024 (and a lot has changed since then), I'll go ahead and close it for now. We can move the discussion back to the issue and open a new PR if it's still needed.

PR #5849, opened 2024-12-24
Document required quoting of booleans in php.ini

brettmc:
Update PHP configuration documentation to mention that boolean values need to be protected by double-quotes
if provided via `php.ini`.

Closes: https://github.com/open-telemetry/opentelemetry-php/issues/1442


cartermp (approved):
Thanks!

PR #5857, opened 2024-12-29
Small updates to the java agent zero code docs

jaydeluca:
* Some small changes to the wording of some sections so they flow a little better
* Added emphasis on the note around the complexity of disabling instrumentation, as discussed in a recent SIG meeting
* Moved some gradle/maven code into a tab pane

One other thing I tried to fix was the [rendering of this section](https://opentelemetry.io/docs/zero-code/java/agent/instrumentation/#peer-service-name), particularly [this line](https://github.com/open-telemetry/opentelemetry.io/blob/main/content/en/docs/zero-code/java/agent/instrumentation/_index.md?plain=1#L22) which doesn't seem to be rendering properly:

image

I wasn't able to figure out a fix, in case anyone has any tips on how I can address that

### DIFF
--- a/content/en/blog/2024/fuzzing-audit-results.md
+++ b/content/en/blog/2024/fuzzing-audit-results.md
@@ -5,7 +5,7 @@ date: 2024-12-20
 author: '[Adam Korczynski](https://github.com/AdamKorcz)'
 issue: 5798
 sig: GC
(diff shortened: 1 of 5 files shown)


cartermp (approved):
Thanks! I think this is good to take as-is, since it doesn't introduce formatting issues that weren't already there.

Issue #5873, opened 2025-01-04
Create 2025 blog section and adjust redirects


chalin:
- [ ] Create `2025` blog section
- [ ] Adjust RSS redirects to 2025
- [ ] Remove RSS feed from 2024 pages

austinlparker:
As long as we're working on the RSS stuff, is there a way to make the rss feed discoverable from the base URL?

chalin:
It is already for `/blog`. Is that what you meant @austinlparker?

austinlparker:
I meant you should be able to get the blog feed from the bare URL as well -- idk, right now it feels less-discoverable to me? Isn't there a way to make a feed-of-feeds type of deal?

austinlparker:
i think what i'm thinking of is autodiscovery - https://www.rssboard.org/rss-autodiscovery

chalin:
> As long as we're working on the RSS stuff, is there a way to make the rss feed discoverable from the base URL?

This should be fixed by:

- #6002

Issue #5017, opened 2024-08-09
[CI] External link checking improvements for 24H2


chalin:
Successor to #2429. Looking forward, the plan is to:

- [ ] Resolve #3493 - show an "obsolete" banner for old blog pages, in particular indicating that links might no longer be valid
- [ ] Adjust our external link checking to ignore obsolete blog pages
  - [ ] Maybe we still want to do internal link checking for the obsolete blog pages too? I don't see an way to do this efficiently, so maybe the answer is "no" for this.
- [ ] Set up a cron job to refresh the refcache's oldest X entries every Y days, and report an issue and/or submit a PR with the cache updates

Here are items carried over from #2429:

- [ ] Clean up links that shouldn't be external:
  - [ ] Are Docsy-generated breadcrumb links still external?
- [ ] Check more GitHub links
- [ ] Normalize links to avoid checking a(n external) link (to the same destination) more than once, and creating duplicate entries in the ref cache:
   - [ ] external links: consistently end with a slash or not (for domain-only URLs). 
   - [ ] external links: use apex domain (w/ or w/o www), as appropriate

chalin:
Most items done. Will track remaining items via:

- #7953

PR #5874, opened 2025-01-04
Blog section for 2025

chalin:
- Contributes to:
  - #5017
  - #5873
- Introduces "2025" blog section, including an RSS feed
- Adds link-checker ignore rules for blog sections older than a year
- I'm delaying updating the redirect rules to point to the 2025 section until we have our first 2025 blog post. Otherwise it just looks like the blog section is broken, because no posts are shown.

**Preview and redirect tests**:

- https://deploy-preview-5874--opentelemetry.netlify.app/blog/ --> ~redirect to 2025 section~ still redirects to 2024
- https://deploy-preview-5874--opentelemetry.netlify.app/blog/index.xml --> ~redirect to 2025 RSS feed~ still redirects to 2024 RSS feed

### DIFF
--- a/.htmltest.yml
+++ b/.htmltest.yml
@@ -10,6 +10,8 @@ IgnoreDirs:
   # DO NOT EDIT! IgnoreDirs list is auto-generated from markdown file front matter.
   # Ignore blog index pages for all locales and in all blog sections (top-level and years)
   - ^(../)?blog/(\d+/)?page/\d+
+  # Ignore old blog posts
+  - ^(../)?blog/20(19|21|22|23)/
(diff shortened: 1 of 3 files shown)


cartermp (approved):

cartermp:
Ah, did not see this was a draft

chalin:
@cartermp - this is ready for review

**IMPORTANT note:** this PR introduces link-checker ignore rules for blog posts from 2023 and earlier. That is, we won't be checking any links (external or internal) for those posts. I'd prefer only excluding external link-checking, but it isn't worth the extra setup and computation time to check internal links for all pages (at least not for PR builds). We can always adjust this later if it becomes an issue.

PR #5886, opened 2025-01-06
[CI] Add PR action for refcache refresh and refresh some entries

chalin:
- Contributes to #5017
- Adds support for `/fix` command `refcache:refresh` that will prune the oldest 128 cached external references and then refresh the refcache.
- Refreshes 200+ refcache entries.


tiffany76 (approved):
RSLGTM.

chalin:
The workflow update in this PR isn't being used, so I'll go ahead and merge, then run PR-action tests after.

PR #5887, opened 2025-01-06
[CI] When updating submodules, also fetch tags

chalin:
Ensures that `update:submodule` also fetches tags. This command caters to two use cases where developers have submodules with:

- Only one remote (most likely named `origin`). This is likely the case for most devs.
- Two remotes, named `origin` and `upstream` (if you're a dev who also sometimes submit PRs to the submodule). In this case, the given `foreach` command will fetch tags from `upstream`. (The command is a bit brittle, but I don't foresee any likely cases for which this might cause an issue -- and if it does we can adjust at that time.)


tiffany76 (approved):
RSLGTM.

PR #5901, opened 2025-01-08
Fix data type in yaml

tapiau:
fix: transform should be array item, not map


cartermp:
@tapiau could you sign the CLA? Then we can review. Thanks!

tapiau:
gif: i don't want to live on this planet anymore

This is hideous. I understand to sign bunch of papers donating kidney, but dash and five spaces?!

cartermp (approved):
Yes, unfortunately this is the policy with CNCF assets. It's a legal protection thing.

PR #5907, opened 2025-01-10
Fix TS example for recording exceptions

### LINKED ISSUE #5805: Recording Exceptions Example for Typescript Doesn't Work Out of the Box
The example for "Recording exceptions" for Typescript doesn't work as-is. You have to do something to convince the compiler that you're allowed to pass the ```ex``` (typed as ```unknown``` by default since I think JS lets you throw anything) into the recordException call.

https://opentelemetry.io/docs/languages/js/instrumentation/#recording-exceptions

Typing it as ```any``` in the catch block or explicitly casting it as ```Error``` in the recordException call both seem to work.

Not sure if this is the right code but it seems like ```recordException``` is not making any assumptions about the input type anyways, so it seems safe to do these sorts of things.

https://github.com/open-telemetry/opentelemetry-js/blob/84cce753617a1195138cab3254e1b22f80982150/packages/opentelemetry-sdk-trace-base/src/Span.ts#L317

Grunet:
Resolves #5805 

I tested in the Typescript playground that this fixes the type checking issue.

I have not checked that this fully functionally works (is there an existing otel-js playground for that? If not I can probably find an online editor playground to setup in. Just lmk)

### DIFF
--- a/content/en/docs/languages/js/instrumentation.md
+++ b/content/en/docs/languages/js/instrumentation.md
@@ -1085,7 +1085,9 @@ import opentelemetry, { SpanStatusCode } from '@opentelemetry/api';
 try {
   doWork();
 } catch (ex) {
-  span.recordException(ex);
+  if (ex instanceof Error) {
+    span.recordException(ex);
+  }
   span.setStatus({ code: SpanStatusCode.ERROR });
 }
 ```
@@ -1100,7 +1102,9 @@ const opentelemetry = require('@opentelemetry/api');
 try {
   doWork();
 } catch (ex) {
-  span.recordException(ex);
+  if (ex instanceof Error) {
+    span.recordException(ex);
+  }
   span.setStatus({ code: opentelemetry.SpanStatusCode.ERROR });
 }
 ```


theletterf:
@open-telemetry/javascript-approvers please have a look — thanks!

pichlermarc (approved):
This looks good, thanks 🙂 

@Grunet I'm not aware of a playground to try things like this, but I checked that locally when I suggested it over at #5085. That's also the way we do it in some OTel instrumentations so I'm confident that this is the correct way to do it. 👍 

Functionally, there's a difference to what it did before as it won't record anything that's not an error (`recordSpan` would technically allow for custom types that roughly match `Error`, and `string`s too), but I think this check covers the vast majority of cases. If users need to record something other than `Error` then it's still possible to check up on the type-defintion of [`Span#recordException()`](https://open-telemetry.github.io/opentelemetry-js/interfaces/_opentelemetry_api.Span.html#recordException) but I don't think it needs to be spelled out in the docs :)

svrnm:
thanks @Grunet ! For testing/making sure that the code works we want to roll out a "code excerpt" solution long term, until then we appreciate if people test their work locally and provide it afterwards.

cartermp (approved):

cartermp:
I tested this and it looks good. Thanks!

PR #5911, opened 2025-01-10
Fix survey-on-merged-pr.yml

svrnm:
This workflow still doesn't work. This time I try to fix id by adding permissions to the job. I restructured it as well because I want to have it similar to the PR actions workflow.


chalin (approved):
RSLGTM.

How long does it take to fill out the survey? Maybe give an indication of that in the text: e.g., "... by taking a few minutes to fill out ..."

Issue #5951, opened 2025-01-16
[CI] Check links with references into GitHub


chalin:
- Parent issue: #5017 
- This is going to take a while since I'm currently finding a lot of invalid links. I'm creating this issue separate from the parent issue so that we can layout the work detail and share the link updates across repos. Details TBC

### Tasks

- [x] #5952 - general setup of checking of `github.com` links, with `IgnoreURLs` temporarily set

---

We'll need volunteers to address each of the ignored URLs organizations, projects, and folders currently being ignored via these regexs:

```regex
  - ^https?://github\.com/(OpenObservability|opentracing|openzipkin|[pP]rometheus|prometheus-community|prometheus-operator|avillela|census-instrumentation|cloudevents|dotnet|googleapis|in-toto|w3c)/
  - ^https?://github\.com/open-telemetry/(build-tools|community|opentelemetry-collector-contrib|opentelemetry-collector|opentelemetry-dotnet-instrumentation|opentelemetry-helm-charts|opentelemetry-java-instrumentation|opentelemetry-java|opentelemetry-operator|opentelemetry-proto|opentelemetry-python-contrib|opentelemetry-specification|oteps|semantic-conventions|sig-end-user|weaver)/
  - ^https?://github\.com/open-telemetry/opentelemetry-demo/blob/main/src/(imageprovider|loadgenerator|otelcollector|.*service)
```

#### GitHub organizations to un-ignore:

- [x] `OpenObservability` - #5954
- [x] `opentracing` - #5954
- [x] `openzipkin` - #5954
- [x] `[pP]rometheus` - #5954
- [x] `prometheus-community` - #5954
- [x] `prometheus-operator` - #5953, fixes some broken links
- [x] `avillela` - #5953
- [x] `census-instrumentation` - #5953
- [x] `cloudevents` - #5953
- [x] `dotnet` - #5953
- [x] `googleapis` - #5953
- [x] `in-toto`
  - [x] #5953
  - [x] ~Update semconv once the above is merged and published~
         #5974
- [x] `w3c` - #5953

#### `open-telemetry` projects to un-ignore:

- [x] `build-tools` - #5955
- [x] `community` - #5955
- [x] `opentelemetry-collector-contrib`
    - [x] #5955
    - [x] #5969
    - [x] #5970
- [x] `opentelemetry-collector` - #5970
- [x] `opentelemetry-dotnet-instrumentation` - #5973
- [x] `opentelemetry-helm-charts` - #5973
[...]

PR #5952, opened 2025-01-16
[CI] Check links into github.com, excluding 4XXs for now

chalin:
- This lays a foundation for, and contributes to #5951
- Updates refcache with many links into `github.com`
- Ignores URLs to `github.com` orgs and users that result in 400 statuses


tiffany76 (approved):
RSLGTM.

PR #5953, opened 2025-01-16
[CI] Fix some external links into GitHub, and clean up IgnoreURL rules

chalin:
- Contributes to #5951
- Drops unnecessary IgnoreURL regex alternatives (htmltest must have gotten confused or overwhelmed when it hit too many 429 statuses).
- Fixes some prometheus-operator links


tiffany76 (approved):
RSLGTM.

PR #5954, opened 2025-01-16
[CI] Resolve and drop all github.com IgnoreURL rules for non-OTel orgs

chalin:
- Contributes to #5951
- Resolves all non-OTel orgs


tiffany76 (approved):
RSLGTM.

PR #5962, opened 2025-01-17
[Operator] Add docs for instrumenting Deno

lucacasonato:
Also see https://github.com/open-telemetry/opentelemetry-operator/issues/3613

This adds documentation for using the operator to instrument Deno applications by setting `OTEL_DENO=true` and then injecting with `instrumentation.opentelemetry.io/inject-sdk: "true"`.

cc @jaronoff97

### DIFF
--- a/content/en/docs/kubernetes/operator/automatic.md
+++ b/content/en/docs/kubernetes/operator/automatic.md
@@ -180,6 +180,67 @@ spec:
 
 For more details, see [.NET Auto Instrumentation docs](/docs/zero-code/net/).
 
+### Deno
+
+The following command creates a basic Instrumentation resource that is
+configured for instrumenting Deno services.
+
+```bash
+kubectl apply -f - <<EOF
+apiVersion: opentelemetry.io/v1alpha1
+kind: Instrumentation
+metadata:
+  name: demo-instrumentation
+spec:
+  env:
+    - name: OTEL_DENO
+      value: 'true'
+  exporter:
+    endpoint: http://demo-collector:4318
+  propagators:
+    - tracecontext
+    - baggage
+  sampler:
+    type: parentbased_traceidratio
+    argument: '1'
+EOF
+```
+
+Deno processes automatically export telemetry data to the configured endpoint
+when they are started with the `OTEL_DENO=true` environment variable. Therefore,
+the example specifies this environment variable in the `env` field of the
+Instrumentation resource, so it is set for all services that have env vars
+injected with this Instrumentation resource.
+
+By default, the Instrumentation resource that auto-instruments Deno services
+uses `otlp` with the `http/proto` protocol. This means that the configured
+endpoint must be able to receive OTLP over `http/proto`. Therefore, the example
+uses `http://demo-collector:4318`, which connects to the `http/proto` port of
+the `otlpreceiver` of the Collector created in the previous step.
+
+{{% alert title="Note" color="info" %}}
+
+Deno's OpenTelemetry integration is not yet stable. As a result all workloads
+that want to be instrumented with Deno must have the `--unstable-otel` flag set
+when starting the Deno process.
+
+{{% /alert %}}
+
+#### Configuring console.log capturing in Deno
+
+By default, the Deno OpenTelemetry integration captures `console.log` output as
+logs, while still printing the logs to stdout / stderr. There are two other
+behaviours that can be configured:
+
+- `OTEL_DENO_CONSOLE=replace`: capture `console.log` output as logs and do not
+  print to stdout / stderr.
+- `OTEL_DENO_CONSOLE=ignore`: do not capture `console.log` output as logs.
+
+#### Learn more {#deno-learn-more}
+
+For more details, see
+[Deno's documentation on the OpenTelemetry integration](https://docs.deno.com/runtime/fundamentals/open_telemetry/).
+
 ### Java
 
 The following command creates a basic Instrumentation resource that is
@@ -483,6 +544,7 @@ done by updating your service’s `spec.template.metadata.annotations` to includ
 a language-specific annotation:
 
 - .NET: `instrumentation.opentelemetry.io/inject-dotnet: "true"`
+- Deno: `instrumentation.opentelemetry.io/inject-sdk: "true"`
 - Go: `instrumentation.opentelemetry.io/inject-go: "true"`
 - Java: `instrumentation.opentelemetry.io/inject-java: "true"`
 - Node.js: `instrumentation.opentelemetry.io/inject-nodejs: "true"`
@@ -684,12 +746,20 @@ Here are a few things to check for:
 - **Is the auto-instrumentation for the right language?** For example, when
   instrumenting a Python application, make sure that the annotation doesn't
   incorrectly say `instrumentation.opentelemetry.io/inject-java: "true"`
-  instead.
+    instead.
 - **Is the auto-instrumentation annotation in the correct location?** When
   defining a `Deployment`, annotations can be added in one of two locations:
   `spec.metadata.annotations`, and `spec.template.metadata.annotations`. The
   auto-instrumentation annotation needs to be added to
   `spec.template.metadata.annotations`, otherwise it won’t work.
+
+{{% alert title="Note" color="info" %}}
+
+Note that unlike other auto-instrumentation, the Deno auto-instrumentation uses
+the `instrumentation.opentelemetry.io/inject-sdk: "true"` annotation, not an
+annotation that contains the string `deno`.
+
+{{% /alert %}}
 
 ### Was the auto-instrumentation endpoint configured correctly?
 


svrnm:
TIL that deno comes with otel support, wow! 

We normally default to not allow documentation for non official otel projects, e.g. we have https://opentelemetry.io/docs/languages/other/ to list 3rd party implementations and language interest (which needs some love and attention from what I just see), but having a language provide out of the box support is very much in line with otel's mission & vision, so we should address that accordingly.

@lucacasonato thank you for bringing this to our attention, we need some time to think about that and address this properly.

swiatekm (approved):
From the operator perspective, the content looks good to me. I'm also very happy with the runtime having first-party Otel support.

jaronoff97 (approved):
🫡

content/en/docs/kubernetes/operator/automatic.md
```diff
@@ -180,6 +180,67 @@ spec:
 
 For more details, see [.NET Auto Instrumentation docs](/docs/zero-code/net/).
 
+### Deno
+
+The following command creates a basic Instrumentation resource that is
+configured for instrumenting Deno services.
```
chalin, line 186:
```suggestion
configured for instrumenting [Deno](https://deno.com/) services.
```

content/en/docs/kubernetes/operator/automatic.md
```diff
@@ -180,6 +180,67 @@ spec:
...
+
+By default, the Instrumentation resource that auto-instruments Deno services
+uses `otlp` with the `http/proto` protocol. This means that the configured
+endpoint must be able to receive OTLP over `http/proto`. Therefore, the example
+uses `http://demo-collector:4318`, which connects to the `http/proto` port of
+the `otlpreceiver` of the Collector created in the previous step.
+
+{{% alert title="Note" color="info" %}}
+
+Deno's OpenTelemetry integration is not yet stable. As a result all workloads
```
chalin, line 223:
```suggestion
[Deno's OpenTelemetry integration][deno-otel-docs] is not yet stable. As a result all workloads
```

Note that I define `[deno-otel-docs]` further below.

content/en/docs/kubernetes/operator/automatic.md
```diff
@@ -180,6 +180,67 @@ spec:
...
+that want to be instrumented with Deno must have the `--unstable-otel` flag set
+when starting the Deno process.
+
+{{% /alert %}}
+
+#### Configuring console.log capturing in Deno
+
+By default, the Deno OpenTelemetry integration captures `console.log` output as
+logs, while still printing the logs to stdout / stderr. There are two other
+behaviours that can be configured:
```
chalin, line 233:
```suggestion
By default, the Deno OpenTelemetry integration exports `console.log()` output as
[logs](/docs/concepts/signals/logs/), while still printing the logs to stdout / stderr. You can configure these alternative behaviors:
```

content/en/docs/kubernetes/operator/automatic.md
```diff
@@ -180,6 +180,67 @@ spec:
...
+
+{{% alert title="Note" color="info" %}}
+
+Deno's OpenTelemetry integration is not yet stable. As a result all workloads
+that want to be instrumented with Deno must have the `--unstable-otel` flag set
+when starting the Deno process.
+
+{{% /alert %}}
+
+#### Configuring console.log capturing in Deno
```
chalin, line 229:
```suggestion
#### Configuration options
```

content/en/docs/kubernetes/operator/automatic.md
```diff
@@ -180,6 +180,67 @@ spec:
...
+
+By default, the Deno OpenTelemetry integration captures `console.log` output as
+logs, while still printing the logs to stdout / stderr. There are two other
+behaviours that can be configured:
+
+- `OTEL_DENO_CONSOLE=replace`: capture `console.log` output as logs and do not
+  print to stdout / stderr.
```
chalin, line 236:
```suggestion
- `OTEL_DENO_CONSOLE=replace`: only export `console.log()` output as logs; do not
  print to stdout / stderr.
```

content/en/docs/kubernetes/operator/automatic.md
```diff
@@ -180,6 +180,67 @@ spec:
...
+- `OTEL_DENO_CONSOLE=ignore`: do not capture `console.log` output as logs.
```
chalin, line 237:
```suggestion
- `OTEL_DENO_CONSOLE=ignore`: do not export `console.log()` output as logs.
```

Question: does `ignore` imply that the output is completely ignored, and hence output isn't printed either? If so then the statement should probably be:

```suggestion
- `OTEL_DENO_CONSOLE=ignore`: neither export as logs nor print `console.log()` output.
```

content/en/docs/kubernetes/operator/automatic.md
```diff
@@ -180,6 +180,67 @@ spec:
...
+
+#### Learn more {#deno-learn-more}
+
+For more details, see
+[Deno's documentation on the OpenTelemetry integration](https://docs.deno.com/runtime/fundamentals/open_telemetry/).
```
chalin, line 242:
```suggestion
For more details, see Deno's [OpenTelemetry integration][deno-otel-docs] documentation.

[deno-otel-docs]: https://docs.deno.com/runtime/fundamentals/open_telemetry/
```

content/en/docs/kubernetes/operator/automatic.md
```diff
@@ -684,13 +746,21 @@ Here are a few things to check for:
 - **Is the auto-instrumentation for the right language?** For example, when
   instrumenting a Python application, make sure that the annotation doesn't
   incorrectly say `instrumentation.opentelemetry.io/inject-java: "true"`
-  instead.
+    instead.
```
chalin, line 749: Why this whitespace change? It should be undone, though the formatter should have fixed / reported it.

content/en/docs/kubernetes/operator/automatic.md
```diff
@@ -684,13 +746,21 @@ Here are a few things to check for:
...
   `spec.template.metadata.annotations`, otherwise it won’t work.
 
+{{% alert title="Note" color="info" %}}
+
+Note that unlike other auto-instrumentation, the Deno auto-instrumentation uses
+the `instrumentation.opentelemetry.io/inject-sdk: "true"` annotation, not an
+annotation that contains the string `deno`.
+
+{{% /alert %}}
+
```
chalin, line 763: I'd rather drop this alert and fold this alert text into a sub-bullet of **Is the auto-instrumentation for the right language?**, which already mentions exceptions (currently for Python). For example, consider replacing the bullet at line 746 of this PR (line 684 in the original file), with:

```markdown
- **Is the auto-instrumentation for the right language?** For example, when instrumenting an application in:
  - Deno, which uses the `instrumentation.opentelemetry.io/inject-sdk: "true"` annotation, rather than an annotation containing the string `deno`.
  - Python: ensure that the annotation doesn't incorrectly say `instrumentation.opentelemetry.io/inject-java: "true"` instead.
```

@open-telemetry/docs-approvers WDYT?

chalin (changes requested):
Copyedit pass. See inline suggestions.

chalin:
Let's keep the sections in alphabetical order: @lucacasonato - could you please move the Go section to be right after the new Deno section?

content/en/docs/kubernetes/operator/automatic.md
lucacasonato: Updated and clarified the behaviour.

content/en/docs/kubernetes/operator/automatic.md
lucacasonato: I added it to that bullet, but not quite like you suggested - the Python comments seems more general "for example, if you use python, don't use the java annotation" - it isn't: "specifically if you use python, do not use the java annotation".

LMK if the new wording works for you.

cartermp (approved):
LGTM pending @chalin approval

content/en/docs/kubernetes/operator/automatic.md
chalin: Yes, that is good. The only adjustment that I'd like to ask is that you make those two sentences subbullets so that the Deno statement is easier to see -- I might even suggest that you put **Deno** in bold there, since it is in effect an important "Note".

lucacasonato: adjusted.

chalin (approved):
LGTM

lucacasonato:
Thanks folks.

It looks like the static site generator does not support links inside of "notes":

image

chalin:
Oops. It does, but the link definition needs to be inside the note :-/.

lucacasonato:
https://github.com/open-telemetry/opentelemetry.io/pull/5993

PR #5972, opened 2025-01-17
Fix example yaml file indent

tammy-baylis-swi:
Fixes indent in example yaml config file for "Kubernetes Objects Receiver". `config` is root level, while `exporters` and `service` are its children.


tiffany76 (approved):
Thanks for the fix, @tammy-baylis-swi! LGTM.

@open-telemetry/helm-approvers, PTAL, thanks.

cartermp (approved):

svrnm (approved):

PR #5974, opened 2025-01-17
[CI] Address semconv link to in-toto

chalin:
- Contributes to #5951
- We don't actually need to wait for the semconv update because the in-toto link issue was fixed via #5969


tiffany76 (approved):
RSLGTM.

PR #6001, opened 2025-01-21
update text color to black architecture.md

sonalgaud12:
This PR fixes #5999 

Flowchart is not visible due to unspecified text color,  Added text color to black

image

image

PR #6013, opened 2025-01-21
Make collector mermaid diagrams darkmode compatible

### LINKED ISSUE #5999: Flowchart for collector pipeline architecture not displaying
**What needs to be changed?**  The flowchart text is not visible in the dark theme

Image

Image

Image

The flowchart text is not visible 

**What is the name + path of the page that needs changed?** The relative path
and page title where you found a problem.
 https://opentelemetry.io/docs/collector/architecture

**Additional context**: 

Solution: set text color to black in classDef

chalin:
- **Best reviewed by ignoring whitespace diffs**
- Fixes #5999
- Eliminates color overrides for most Mermaid diagrams, falling back to defaults that are dark-mode compatible
- Drops named anchors in "Running as ..." headings, since those prevented the corresponding TOC entries from working
- Closes #6001, as an alternative solution, which also fixes _all_ diagrams, and the TOC heading links

### Previews

- https://deploy-preview-6013--opentelemetry.netlify.app/docs/collector/architecture/#pipelines
- https://deploy-preview-6013--opentelemetry.netlify.app/docs/collector/architecture/#running-as-an-agent

### Screenshots of darkmode fixes

| Before | After |
|--------|--------|
| image | image |
| image | image | 

The only caveat is that in light mode, the last two diagrams have a yellow background (all other diagrams are the same in light mode before and after), something like this:

> image

Fixing this has proven tricky, and would require a change to Docsy, so that the non-dark theme chosen is `base` rather than `default`. But even then I couldn't get all of the colors for all of the diagrams to work. I'm submitting this PR because it is a good step forward, despite the yellow background.

### DIFF
--- a/content/en/docs/collector/architecture.md
+++ b/content/en/docs/collector/architecture.md
@@ -44,21 +44,18 @@ The following diagram represents a typical pipeline:
 title: Pipeline
 ---
 flowchart LR
-    R1(Receiver 1) --> P1[Processor 1]
-    R2(Receiver 2) --> P1
-    RM(...) ~~~ P1
-    RN(Receiver N) --> P1
-    P1 --> P2[Processor 2]
-    P2 --> PM[...]
-    PM --> PN[Processor N]
-    PN --> FO((fan-out))
-    FO --> E1[[Exporter 1]]
-    FO --> E2[[Exporter 2]]
-    FO ~~~ EM[[...]]
-    FO --> EN[[Exporter N]]
-
-    %% The stroke color matches the website header.
-    classDef default fill:#e3e8fc,stroke:#4f62ad
+  R1(Receiver 1) --> P1[Processor 1]
+  R2(Receiver 2) --> P1
+  RM(...) ~~~ P1
+  RN(Receiver N) --> P1
+  P1 --> P2[Processor 2]
+  P2 --> PM[...]
+  PM --> PN[Processor N]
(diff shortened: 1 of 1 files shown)


content/en/docs/collector/architecture.md
```diff
@@ -44,21 +44,18 @@ The following diagram represents a typical pipeline:
...
-    P2 --> PM[...]
-    PM --> PN[Processor N]
-    PN --> FO((fan-out))
-    FO --> E1[[Exporter 1]]
-    FO --> E2[[Exporter 2]]
-    FO ~~~ EM[[...]]
-    FO --> EN[[Exporter N]]
-
-    %% The stroke color matches the website header.
-    classDef default fill:#e3e8fc,stroke:#4f62ad
```
chalin, line 61: /cc @tiffany76, since you chose these original color overrides if I'm not mistaken.

tiffany76 (approved):

content/en/docs/collector/architecture.md
chalin: FYI, we're still quite close to brand colors for most of the diagrams. (I might have a solution to get rid of the yellow, but it'll take more work.)

chalin:
Given that these are CSS and editorial changes, I'll merge now. Thanks for the review and approval @tiffany76 🙌🏻

PR #6034, opened 2025-01-22
Fix daily semconv build workflow

trask:
[Secrets are not automatically passed to reusable workflows](https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions#using-secrets-in-a-workflow)

cc @open-telemetry/specs-semconv-maintainers

### DIFF
--- a/.github/workflows/build-semconv-daily.yml
+++ b/.github/workflows/build-semconv-daily.yml
@@ -21,4 +21,5 @@ jobs:
       success: ${{ needs.build-dev.result == 'success' }}
       repo: open-telemetry/semantic-conventions
     secrets:
-      opentelemetrybot_github_token: ${{ secrets.OPENTELEMETRYBOT_GITHUB_TOKEN }}
+      opentelemetrybot_github_token:
+        ${{ secrets.OPENTELEMETRYBOT_GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


chalin (commented):
If ever workflow arguments are visible (e.g., logged), is GH smart enough to not show the value when it is meant to be a secret?

trask:
> If ever workflow arguments are visible (e.g., logged), is GH smart enough to not show the value when it is meant to be a secret?

I tried to `echo $GITHUB_TOKEN` recently, and GH thwarted me, but I wouldn't trust it completely, why?

chalin:
> do you have a specific concern about this workflow?

That `secrets.OPENTELEMETRYBOT_GITHUB_TOKEN` is being passed as an argument value to a workflow, so my concern was to question whether that puts us at risk of exposing the token via GH workflow run logs. WDYT?

trask:
it's not passed as a regular `input`, instead it's explicitly passed as a `secret`. I believe that's the reason for having a separate way to pass secrets that's different from passing regular inputs

chalin (approved):
Right, of course, now that I'm paying closer attention. Ok, LGTM.

Btw, we might want to see how we can tweak this so that it doesn't fail if the refcache is updated, only if there are broken links. That might be a project for another time.

PR #6056, opened 2025-01-23
[CI] Handle missing tocstop tag in spec pages

chalin:
- Reports a warning if a spec page is missing a `tocstop` directive, and gracefully recovers
- For example, the script now reports issues like https://github.com/open-telemetry/semantic-conventions/issues/1799:
  ```
  WARN tmp/semconv/docs/exceptions/exceptions-spans.md:14: missing '<!-- tocstop -->' directive? Aborting toc scan at line:
    14: ## Exception event
  ```

/cc @trask


tiffany76 (approved):
RSLGTM.

PR #6060, opened 2025-01-23
[WIP] Update semantic-conventions to 1.30.0-dev

chalin:
It occurred to me that we could just submit a WIP PR that we progressively update to semconv at HEAD, and resolve any OTel.io page issues as we go along. This also nicely gives us a preview.

**Preview**: https://deploy-preview-6060--opentelemetry.netlify.app/docs/specs/semconv/

/cc @trask


chalin:
Closing since we're now working on the real deal:

- #6077

PR #6081, opened 2025-01-25
Update (or create) `opentelemetrybot/semantic-conventions-v*` branch daily

trask:
Attempting to automate parts of https://github.com/open-telemetry/semantic-conventions/issues/1817

This daily action will check if there's an existing `opentelemetrybot/semantic-conventions-v*` branch, and update that branch with the latest from `main` and update the semconv submodule with the latest from that repo's `main`.

If there's no existing `opentelemetrybot/semantic-conventions-v*` branch, it will create one and open a draft PR against it.

cc @open-telemetry/specs-semconv-maintainers


.github/workflows/build-dev.yml
```diff
@@ -7,14 +7,6 @@ on:
...
         default: content-modules
         type: string
-  workflow_call:
-    inputs:
-      submodule_path_regex:
-        type: string
-        required: true
-      skip_ref_cache_check:
-        type: boolean
-        default: false
```
trask, line 17: rolled back changes to this file since we're going a different direction now

.github/workflows/reusable-workflow-notification.yml
trask, line 1: removed since going in a different direction now

svrnm (approved):

chalin (commented):
I would prefer that the scripts work on a PR branch rather than a dedicated and "permanent" repo branch. We don't need to preserve a continuous history of semconv integration as a part of this repo. A PR branch seems more appropriate IMHO.

Thoughts?

trask:
would naming it `integration/semantic-conventions-1.31.0` be better, to be clear that we merge and delete it with each release?

I could have the script automatically create the branch (and associated PR) if it doesn't find one (meaning we've rolled over to the next release)

chalin:
Yes, I'd like that: a branch per release that can be closed once the release has landed (hence it becomes a PR branch). Thanks!

trask:
> Yes, I'd like that: a branch per release that can be closed once the release has landed (hence it becomes a PR branch). Thanks!

done, ptal, thanks!

.github/workflows/update-semconv-integration-branch.yml
```diff
@@ -0,0 +1,97 @@
...
+          else
+            git checkout $BRANCH
+          fi
+
+      - name: Use CLA approved github bot
+        run: |
+          git config user.name opentelemetrybot
+          git config user.email 107717825+opentelemetrybot@users.noreply.github.com
+
+      - name: Rebase
```
chalin, line 61: Hmm, I'm not _sure_ that I'd want to automatically rebase the otel.io repo (that _is_ what this is doing, right?). My gut feeling is that we might want to do this by hand. Maybe introduce a script bool argument to control this? Or just omit it and we can do it fully manually.

.github/workflows/update-semconv-integration-branch.yml
```diff
@@ -0,0 +1,97 @@
...
+          previous=$(git rev-parse HEAD)
+          git fetch origin
+          git merge origin/main
+          if [ "$(git rev-parse HEAD)" != "$previous" ]; then
+            git push
+          fi
+
+      - name: Update submodule
+        run: |
+          git submodule update --init content-modules/semantic-conventions
```
chalin, line 72: Ideally, we should be running `npm run get:submodule`, but that would require that in this step we not only update the submodule commit, but that we change the value of `semconv-pin` to correspond to the value returned by `git describe --tags`. Can we do that further below in this step?

.github/workflows/update-semconv-integration-branch.yml
```diff
@@ -0,0 +1,97 @@
...
+          if git ls-remote --exit-code --tags origin $VERSION; then
+            git reset --hard $VERSION
+          else
+            git reset --hard origin/main
+          fi
+
+          cd ../..
+
+          if ! git diff-index --quiet HEAD; then
+            git commit -am "Update submodule"
```
chalin, line 84: It would be nice to set `COMMIT_DESC` (or some other variable name) to the value of `git describe --tags` and use it as part of the commit message:

```suggestion
            git commit -am "Update semconv submodule to $COMMIT_DESC"
```

chalin (commented):
Thanks for working on this. See inline comments.

.github/workflows/update-semconv-integration-branch.yml
```diff
@@ -0,0 +1,97 @@
...
+
+jobs:
+  update-semconv-integration-branch:
+    runs-on: ubuntu-latest
+    if: github.repository == 'open-telemetry/opentelemetry.io'
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          # this is needed in order to do the rebase below
+          fetch-depth: 0
```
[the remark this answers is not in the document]
trask: yes

.github/workflows/update-semconv-integration-branch.yml
```diff
@@ -0,0 +1,97 @@
...
+
+          if [[ -z "$versions" ]]; then
+            latest_version=$(gh release view \
+                          --repo open-telemetry/semantic-conventions \
+                          --json tagName \
+                          --jq .tagName)
+            if [[ $latest_version =~ ^v([0-9]+)\.([0-9]+)\.[0-9]+ ]]; then
+              major="${BASH_REMATCH[1]}"
+              minor="${BASH_REMATCH[2]}"
+              version="v$major.$((minor + 1)).0"
```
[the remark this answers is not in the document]
trask: (I updated branch name instead though because I need the VERSION without `-dev` later)

.github/workflows/update-semconv-integration-branch.yml
trask: it's merging from main (no force pushing involved), I can remove it for now if you're prefer to test doing it by hand first

.github/workflows/update-semconv-integration-branch.yml
trask: I updated `semconv-pin` now, but I'm not sure what `npm run get:submodule` is for, or when to run it?

chalin: Actually, `npm run get:submodule` is mainly to help local devs sync their submodules to the pinned versions. Come to think of it, we won't need it here.

chalin (approved):
Looking forward to giving this a spin.

.github/workflows/update-semconv-integration-branch.yml
chalin: I'd prefer rebasing, to keep the PR branch clean, and make it easier to cherry pick if/once we want to create a final PR.

But let's leave this as is for now. We can revisit later if necessary.

PR #6087, opened 2025-01-27
Reset survey-on-merged-pr `on` to `pull_request`

svrnm:
if `pull_request_target` is used it seems that the `gh comment` is looking for the PR in the fork that's why this is breaking


tiffany76 (approved):
RSLGTM.

PR #6104, opened 2025-01-29
Add content/en/docs/kubernetes/operator/api.md

CharlieTLe:
Copying from opentelemetry-operator/docs/api.md so that it can accessible from the website.

Related to: https://github.com/open-telemetry/opentelemetry-operator/issues/3681


chalin (changes requested):
Hi. We can't just use a copy of that file, it'll quickly drift from its canonical source. @open-telemetry/operator-approvers should comment as to whether they'd actually like this to appear on the website, and if so, we'll see how we can setup a means of keeping this copy and the operator repo copy in sync.

tiffany76:
Closing this PR since the Operator SIG will handle in their repo. Thanks for being proactive, @CharlieTLe!

chalin:
@CharlieTLe - I've created the following issue so that we can address your very legitimate concern of wanting to be able to view the API docs :)

- #6200

PR #6123, opened 2025-01-30
Java metric guidelines link

nchen000:
Fix `metric guidelines` link under [Java API docs](https://opentelemetry.io/docs/languages/java/api/#meter) so it would stop points to `http://localhost:1313/docs/specs/semconv/general/metrics/#general-guidelines`.


chalin (approved):
Good catch @nchen000, and thanks for the fix.

@open-telemetry/docs-approvers - I'll follow up with:

- #6125

cartermp (approved):

PR #6124, opened 2025-01-30
Demo repo links: fix & check them all, update refcache

chalin:
- Contributes to #5017
- Removes temporary rule that had some external links into the demo repo igored
- Updates links to services etc that were recently renamed.
- Updates the refcache
- I decided to fix the broken links in non-`en` pages, and to tag the links with a `?i18n-patch` query parameter as an indicator for humans and tools that this is a link patch, but otherwise the rest of the page still matches it's `default_lang_commit`


julianocosta89 (approved):
LGTM, thx @chalin!

tiffany76 (approved):
RSLGTM.

PR #6128, opened 2025-01-30
Fix semconv integration branch workflow

trask:
It's failing on subsequent runs, this seems to fix it when I test on my fork...

### DIFF
--- a/.github/workflows/update-semconv-integration-branch.yml
+++ b/.github/workflows/update-semconv-integration-branch.yml
@@ -86,7 +86,7 @@ jobs:
 
           sed -i "s/^\tsemconv-pin = .*/\tsemconv-pin = $commit_desc/" .gitmodules
 
-          git add content-modules/semantic-conventions .gitmodules
+          git add -A
 
           if ! git diff-index --quiet --cached HEAD; then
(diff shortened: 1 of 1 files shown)


svrnm:
@chalin PTAL! You worked with @trask on that so I'd like you to review & merge

.github/workflows/update-semconv-integration-branch.yml
```diff
@@ -86,7 +86,9 @@ jobs:
 
           sed -i "s/^\tsemconv-pin = .*/\tsemconv-pin = $commit_desc/" .gitmodules
 
-          if ! git diff-index --quiet HEAD; then
+          git add content-modules/semantic-conventions .gitmodules
```
chalin, line 89: How about:

```suggestion
          git add -A
```

chalin: Hey @trask - you marked this as resolved but I'm not seeing the change yet. Is the update still pending?

trask: I thought I hit the commit suggestion button, but clearly not 🤔

chalin (approved):
Let's give this a try. ✨

PR #6162, opened 2025-01-31
[ja] translate demo/forking into ja

Msksgm:
translated demo/forking page.


Msksgm:
I closed this PR because I noticed that the same page had been in progress and it was merged.

PR #6165, opened 2025-02-01
[CI] Page-local dictionary normalization: drop unused words and duplicates (now case-insensitive compare)

chalin:
- Updates the page-local dictionary (cSpell:ignore) list, excluding words that don't appear in the page.
- Drops duplicates from the words list, ignoring case.
-  Ensures that any remaining words that were duplicates are in lower case (so as to match all capitalizations)


dmathieu (approved):

tiffany76 (approved):
RSLGTM.

PR #6168, opened 2025-02-02
Add ADOT instrumentations to distribution list

garysassano:

### DIFF
--- a/data/ecosystem/distributions.yaml
+++ b/data/ecosystem/distributions.yaml
@@ -7,6 +7,14 @@
   url: https://aws.amazon.com/otel/
   docsUrl: https://aws-otel.github.io/docs/getting-started/collector
   components: [Collector]
+- name: AWS AWS Distro for OpenTelemetry - Instrumentation for JavaScript
+  url: https://github.com/aws-observability/aws-otel-js
+  docsUrl: https://github.com/aws-observability/aws-otel-js
+  components: [Node]
+- name: AWS AWS Distro for OpenTelemetry - Instrumentation for Python
+  url: https://github.com/aws-observability/aws-otel-python
+  docsUrl: https://github.com/aws-observability/aws-otel-python
+  components: [Python]
 - name: Grafana Alloy
   url: https://github.com/grafana/alloy
   docsUrl: https://grafana.com/docs/alloy/latest/
--- a/static/refcache.json
+++ b/static/refcache.json
@@ -2891,6 +2891,14 @@
     "StatusCode": 206,
     "LastSeen": "2025-01-16T13:05:48.070495-05:00"
   },
+  "https://github.com/aws-observability/aws-otel-js": {
+    "StatusCode": 206,
+    "LastSeen": "2025-02-03T10:28:17.890526733Z"
+  },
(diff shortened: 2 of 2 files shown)


cartermp (changes requested):
Additions look good, but can you undo the deletion of the devex survey?

data/ecosystem/distributions.yaml
```diff
@@ -7,6 +7,14 @@
   url: https://aws.amazon.com/otel/
   docsUrl: https://aws-otel.github.io/docs/getting-started/collector
   components: [Collector]
+- name: AWS AWS Distro for OpenTelemetry - Instrumentation for JavaScript
+  url: https://github.com/aws-observability/aws-otel-js
+  docsUrl: https://github.com/aws-observability/aws-otel-js
+  components: [Node]
+- name: AWS AWS Distro for OpenTelemetry - Instrumentation for Python
+  url: https://github.com/aws-observability/aws-otel-python
+  docsUrl: https://github.com/aws-observability/aws-otel-python
```
chalin, line 16: @svrnm - what's acceptable here in terms of a link to docs. I'd rather this field be empty than point just to the repo. I don't see any docs there, unless you consider the README as being the docs @garysassano?

garysassano: I noticed that many elements in the list followed that approach, so I did the same. If you'd rather remove the `docsUrl`, I'm okay with that.

chalin: Right, thanks. I'm waiting on @svrnm's feedback.

svrnm: We might need the docsUrl, or the validation fails, so let's leave it for now, we need to review the registry anyways.

chalin: Can we fix the validation, 😄 (not in the scope of this PR, of course) rather than have to supply a bogus value? 

> we need to review the registry anyways.

Is there an issue open so that we can track specific changes (like this one regarding validation of the `docsUrl`)?

svrnm: > Can we fix the validation, 😄 (not in the scope of this PR, of course) rather than have to supply a bogus value?

yes

> > we need to review the registry anyways.
> 
> Is there an issue open so that we can track specific changes (like this one regarding validation of the `docsUrl`)?

no, a lot of the issues with the registry are not tracked

svrnm: ignore what I said, the distributions.yaml is not verified. Tbh let's leave it like that, eventually we should move those entries to the registry as well. 

I think we can resolve this conversation, wdyt @chalin ?

content/en/announcements/devex-survey.md
svrnm, line 1: why is this file deleted?

garysassano: I'm not sure, that was done automatically when running npm test.

data/ecosystem/distributions.yaml
chalin: Sure, let's mark this resolved for now.

chalin (approved):

chalin:
@cartermp - this good to go?

PR #6170, opened 2025-02-02
[ja] translate demo/tests into ja

Msksgm:


Msksgm:
I closed this PR because I noticed that the same page had been in progress and it was merged.

PR #6180, opened 2025-02-03
docs(languages/java/intro): current section links BOM

guspan-tanadi:

### DIFF
--- a/content/en/docs/languages/java/intro.md
+++ b/content/en/docs/languages/java/intro.md
@@ -106,7 +106,7 @@ dependency structure:
 ## Dependencies and BOMs
 
 A
-[bill of materials](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#bill-of-materials-bom-poms),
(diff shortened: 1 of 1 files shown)


steverao (approved):

trask (approved):
@open-telemetry/docs-maintainers I thought anchors were being verified somehow in this repo? (at least I think that's how we've found issues in the semconv repo before through this repo)

chalin:
> @open-telemetry/docs-maintainers I thought anchors were being verified somehow in this repo? (at least I think that's how we've found issues in the semconv repo before through this repo)

Links into the spec pages are "localized" -- they become local paths rather than external URLs -- when published this repo, which is why anchor IDs are being checked. It seems like [htmltest](https://github.com/wjdp/htmltest) doesn't check anchor IDs on external links after all :-/ 🤷🏼‍♂️

chalin:
> Links into the spec pages are "localized" -- they become local paths rather than external URLs -- when published this repo, which is why anchor IDs are being checked. It seems like [htmltest](https://github.com/wjdp/htmltest) doesn't check anchor IDs on external links after all :-/ 🤷🏼‍♂️

I've just modified our test scripts to check anchors for external links too.

chalin (approved):

Issue #6040, opened 2025-01-23
[CI] Ensure that fix:dict etc get run as part of checks


chalin:
We should be catching this during build checks, but apparently not, e.g.:

- #6039

Fix scripts to enable:

- [x] fix:dict - `check:dict` is run as a part of the spelling-check workflow
- [x] fix:filenames - is checked as a part of the file-check workflow
- [x] fix:expired - #6189
- [ ] fix:submodule
- [ ] Any more?

pratik-mahalle:
Can I work on this? @chalin

pratik-mahalle:
So, most of the work is actually done, I need to fix the sbmodule script only , right @chalin ?

Issue #2120, opened 2022-12-19
Site build error


chalin:
_Originally posted by @Chinwendu20 in https://github.com/open-telemetry/opentelemetry.io/issues/2094#issuecomment-1354495951_:

>   Trying to submit a PR, after cloning the repo and running an npm install, I get this error:
> ![image](https://user-images.githubusercontent.com/59079323/208074164-2761da40-1e8b-4859-b825-3ca3b8bd1a34.png)
> I have posted this on the slack group, would be grateful for any help that I get

chalin:
@svrnm added:

> https://cloud-native.slack.com/archives/C02UN96HZH6/p1671184942392839 leaving the discussion on the `npm install` here for reference.

chalin:
AFAIK, this is an issue under Windows only. The site builds fine under, e.g., GitPod.

cartermp:
FWIW I did notice this in gitpod as well several months ago, but I think it had to do with something on their end.

chalin:
> Note to self: consider fixing this on the Docsy / docsy-example side first.

svrnm:
During Outreachy we figured out some improvements for developing on windows, we need to add that to the documentation still, cc @theletterf

snehaio:
Hi @svrnm ,I’d like to contribute to opentelemetry.io, but I’m having trouble setting up the repository locally on Windows. I also tried using WSL, but the issue persists. Could you please assist me with this?
![{467029CC-83EB-4D94-842F-E26381126E53}](https://github.com/user-attachments/assets/cc9302ca-eb0d-4d7b-b340-be435bf4ce3f)

svrnm:
> Hi @svrnm ,I’d like to contribute to opentelemetry.io, but I’m having trouble setting up the repository locally on Windows. I also tried using WSL, but the issue persists. Could you please assist me with this?

linking to @theletterf's recommendation:

https://cloud-native.slack.com/archives/C02UN96HZH6/p1727881787891549?thread_ts=1727805453.507049&cid=C02UN96HZH6

> Folks, most script won't work when running npm on Windows. That's because they use Bash.
If you're on Windows, I strongly recommend that you use Ubuntu through WSL2 as explained in this guide: https://documentation.ubuntu.com/wsl/en/latest/guides/install-ubuntu-wsl2/

I am not a windows user myself, so I can not confirm if this helps or not, but please give it a try.

chalin:
[...]

PR #6189, opened 2025-02-03
[CI] GH check: add file-expired check, and more

chalin:
- Contributes to #6040
- Renames `.github/workflows/check-format.yml` to `.github/workflows/check-file.yml` since the checks are over files, and more than just formatting related
- Adds `check:expired` as a PR-build check in `.github/workflows/check-file.yml`
- Simplifies the `check-formatting`:
  - Eliminates the use of `PRETTIER_AT_VERS`
  - Adds `scripts/npx-helper.sh` script to be used instead of setting env var like `PRETTIER_AT_VERS`
  - Contributes to #2120
    - Adjusts the `__check:format` to be more Windows friendsly.
      @trask can you give this a try under Windows and let me know if this help?


tiffany76 (approved):
RSLGTM.

chalin:
`CSPELL:IGNORE check` seems stuck, maybe because it was renamed? Merging anyway, since that the only thing pending.

chalin:
> `CSPELL:IGNORE check` seems stuck, maybe because it was renamed? Merging anyway, since that the only thing pending.

Hmm, not sure that'll work, but we'll see.

.github/workflows/check-spelling.yml
```diff
@@ -26,16 +26,19 @@ jobs:
...
       - name: Any changed files?
         run: |
           CHANGES=`git status --porcelain`
           if [[ $CHANGES ]]; then
-            echo "Locally run `npm run fix:dict` and commit the changes:"
+            echo "Add comment '/fix:${{ env.FIX_CMD }}' to your PR in GitHub,"
+            echo "or locally run 'npm run ${{ env.FIX_CMD }}' and commit the changes:"
             echo "$CHANGES"
             exit 1
           else
```
chalin, line 44: Note to self: might want to define a script so that this can be reused.

PR #6195, opened 2025-02-04
[docs/collector] Clarify stance with regards to third-party instrumentation

mx-psi:
Clarifies maturity of third-party libraries for Collector internal telemetry.

This is particularly relevant for telemetry generated by [otelhttp][1] or [otelgrpc][2] (the former of which should be stable this year, but the latter needs a lot of work yet).

[1]: https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp
[2]: https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc


dmathieu (commented):
This assumes folks know exactly which metric is emitted by the collector and which one is emitted by otelhttp. Should that we provide a list of metrics otelhttp emits?

We could also mention that includes every metric defined by the [http semantic conventions](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-metrics.md)?

mx-psi:
> This assumes folks know exactly which metric is emitted by the collector and which one is emitted by otelhttp. Should that we provide a list of metrics otelhttp emits?

Users could know this by the `otelcol` prefix. It is mentioned, but maybe it's not as obvious as it should be. Should I reword to make this more explicit?

cartermp (approved):
I don't think this needs any further clarification.

tiffany76:
Merging as is. We can add clarification later, if needed.

PR #6208, opened 2025-02-04
Updates Embrace distros

davidlawrencer:
Add Embrace React Native distribution, condenses duplicated Embrace Android distribution information. Android previously had two entries, made this into one entry with two components.


chalin (approved):
LGTM once all checks are green.

PR #6214, opened 2025-02-04
[CI] Upgrade Docsy, and Hugo to 0.143.0

chalin:
- Upgrades to Hugo 0.143.0
- Upgrades Docsy, bringing with it the `comment` shortcode that Hugo has deprecated


tiffany76 (approved):
RSLGTM.

PR #6223, opened 2025-02-05
[IA] New platforms section with Android placeholder, and more

chalin:
- Contributes to #6222
- Creates a **Platforms** section
- Moves the following `/docs` section into Platforms:
  - FaaS
  - Kubernetes
- Copyedits (minor) Kubernetes landing page
- Adds a placeholder page for **Andriod**

**Preview**:

- https://deploy-preview-6223--opentelemetry.netlify.app/docs/platforms/

**Redirect tests**:

- https://deploy-preview-6223--opentelemetry.netlify.app/docs/faas/
- https://deploy-preview-6223--opentelemetry.netlify.app/docs/kubernetes/

Issue #6222, opened 2025-02-05
[IA] Create a Platforms section, add FaaS & Kubernetes, etc


chalin:
_Originally posted by @svrnm in [#6136](https://github.com/open-telemetry/opentelemetry.io/issues/6136#issuecomment-2636897572)_

> We talked about this during the SIG call yesterday, and it seems that the "Platforms"-idea is what we want to implement, i.e. we create "Platforms" as a new top level category, and it will contain:
> 
> * Android
> * Browser
> * Kubernetes
> * FaaS
> * Other
> 
> This is similar to the grpc-page as @chalin linked to it, and seems to be commonly understood. 
> 
> @open-telemetry/operator-approvers @open-telemetry/lambda-extension-approvers @open-telemetry/android-approvers @open-telemetry/sandbox-web-js-maintainers @open-telemetry/javascript-approvers PTAL! 

### Sections to be moved and/or reworked

- [x] FaaS
- [x] Kubernetes
- [ ] ...

### Tasks

- [x] #6223 
  - [x] #6248
- [x] #6232
  - [x] #6233
- [ ] ...

chalin:
With stubs in place (#7478), I'll consider this issue done.

PR #6232, opened 2025-02-06
[IA] Link normalization step of

chalin:
- Contributes to #6222
- Normalizes paths in response to #6223
- Drops temporary `htmltest` ignore rules


svrnm (approved):

chalin:
Since these are only path updates, no content changes, I'll move ahead with merging given @svrnm's approval even if other SIGs were solicited.

PR #6248, opened 2025-02-07
[IA] Delete stray page copy already now in platforms

chalin:
- Another followup to #6232
- Deletes stray `/docs/kubernetes` file (probably a side effect of VS Code saving copies of deleted files if you happen to have an editor open and make a search-and-replace change in it 🤷)
- Here's the diff between that stray copy and the one in the proper place:
  ```console
  $ diff content/en/docs/{platforms/,}kubernetes/operator/troubleshooting/target-allocator.md
  333c333
  < [`ServiceMonitor`](https://observability.thomasriley.co.uk/prometheus/configuring-prometheus/using-service-monitors/#:~:text=The%20ServiceMonitor%20is%20used%20to,build%20the%20required%20Prometheus%20configuration.)
  ---
  > [`ServiceMonitor`](https://observability.thomasriley.co.uk/prometheus/configuring-prometheus/using-service-monitors/)
  ```
- I've cleaned out the URL above with the `#:~:` fragment


tiffany76 (approved):
RSLGTM.

PR #6257, opened 2025-02-09
fix: removing deplulicated words of `file file`

Msksgm:
I found the a duplicated words in contributing/style-guide page.
I fixed it because  I think that it is a mistake.


chalin (approved):
✨

PR #6260, opened 2025-02-09
Update example config for handling tail-sampling and span metric generation when horizontally scaling collectors

swar8080:
Hello, we used this page's documentation when setting-up tail-sampling and span metric generation. The guide was helpful but there were a couple things it could've mentioned to make our implementation smoother:
1. The guide currently suggests having two separate deployments of collectors. One just for load-balancing spans and another for processing the load-balanced spans. We started with the two deployments but then found it easier to maintain a single deployment of collectors responsible for both tasks. However, it wasn't obvious (to us at least) that this was possible and I'm guessing others would prefer this approach too.
2. There wasn't any example configuration of how the load-balancing exporter, span metric connector, and tail sampling processor all fit together, so it took some jumping between READMEs and unintentional violations of single-writer assumption (which we weren't aware of) to get it right.

So this documentation suggestion shows how all the components work together in a single collector deployment.

### DIFF
--- a/content/en/docs/collector/scaling.md
+++ b/content/en/docs/collector/scaling.md
@@ -325,60 +325,132 @@ providing a custom `http_sd_config` per collector instance (pod).
 ### Scaling Stateful Collectors
 
 Certain components might hold data in memory, yielding different results when
-scaled up. It is the case for the tail-sampling processor, which holds spans in
-memory for a given period, evaluating the sampling decision only when the trace
-is considered complete. Scaling a Collector cluster by adding more replicas
-means that different collectors will receive spans for a given trace, causing
-each collector to evaluate whether that trace should be sampled, potentially
-coming to different answers. This behavior results in traces missing spans,
-misrepresenting what happened in that transaction.
+scaled up. It is the case for the
+[tail-sampling](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/tailsamplingprocessor/README.md)
+processor, which holds spans in memory for a given period, evaluating the
+sampling decision only when the trace is considered complete. Scaling a
+Collector cluster by adding more replicas means that different collectors will
+receive spans for a given trace, causing each collector to evaluate whether that
+trace should be sampled, potentially coming to different answers. This behavior
+results in traces missing spans, misrepresenting what happened in that
+transaction.
 
-A similar situation happens when using the span-to-metrics processor to generate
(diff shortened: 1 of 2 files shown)


svrnm:
thanks @swar8080 ! @open-telemetry/collector-approvers PTAL

content/en/docs/collector/scaling.md
```diff
@@ -325,60 +325,132 @@ providing a custom `http_sd_config` per collector instance (pod).
...
+      receivers:
+        - otlp/for_tail_sampling
+      processors:
+        - tail_sampling
+      exporters:
+        - otlp/vendor
+
+    traces/span_metrics:
+      receivers:
+        - otlp/for_tail_sampling
```
jpkrohling, line 445:
```suggestion
        - otlp/for_span_metrics
```

jpkrohling (commented):
I like the change, but I'd really prefer to have three collector instances (three config files) instead of one with three pipelines. The reason is that each pipeline (load-balancer, tail-sampler, span metrics) has a different load profile and would scale differently.

You absolutely CAN do it like depicted here, but without understanding the nuances, I'd prefer the official documentation to have each pipeline to be its own deployment.

swar8080:
Hi @jpkrohling, thanks for reviewing

Separating span metrics and tail sampling to different collectors is a good callout. A spike it span volume or span metric cardinality explosion would likely only cause problems for one deployment but not the other. I'll go ahead and change this into three collector configurations, or two if the below persuades you otherwise :).

Maybe we missed the benefits of load balancing being its own deployment. Both set-ups have to receive incoming spans, and load balancing to the same deployment has twice as many receives and exports. But it seems like the extra memory for load balancing is dwarfed by memory needed for tail-sampling and span metrics, which also grows with span volume. For CPU, load balancing exporter used a lot before this [optimization](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/32032), but now our pprof shows it as a small percent of total cpu time. So for us it didn't seem worth the effort of maintaining another deployment, which is why we ended-up consolidating it. That saved us another deployment to monitor, and also one less file to jump between when working on our collector config, since we did some filtering/edits before load balancing. So maybe two separate deployments, each load balancing to themselves, would be a good set-up for a lot of users?

svrnm:
@swar8080 @jpkrohling please take another look as this PR has been stalled for a while now

PR #6278, opened 2025-02-10
[infra] Use Docsy feature to display blog out-of-date banner

chalin:
- Upgrades Docsy
- Drops custom template overrides in favor of using Docsy's
- Generates the same site pages (modulo whitespace):
  ```console
  $ (cd public && git diff -bw) | grep ^diff
  ```


tiffany76 (approved):
RSLGTM.

PR #6279, opened 2025-02-10
Update protocol defined for internal collection

mjuarez:
Per https://github.com/open-telemetry/opentelemetry-collector/issues/12332, a breaking change for the protocol used was introduced


chalin (approved):
LGTM. Awaiting final approval of @open-telemetry/collector-approvers.

tiffany76:
Thanks for staying on top of the changes, @mjuarez! I see that the changelog update isn't quite merged yet, so let's ask the @open-telemetry/collector-approvers to take a look. 

One additional question. Do we need to add a comment about versions? Something like, "if you're using Collector v0.118.0 or older, `grpc/protobuf` or `grpc` can be used. From v0.119.0, `grpc/protobuf` is no longer valid." I don't know if this is useful or relevant information to users but figured I'd ask.

mjuarez:
I know the changelog update isn't in, but had a peer on my team reference the existing doc while using the latest published version (0.119.0) and it was broken for them. So anyone new to otel-collector wouldn't be set up for success.

I'm unsure about distinguishing config options by version. Has that been used elsewhere in the docs I can use as an example? Happy to update.

mx-psi (approved):

mx-psi:
Ah, good catch, yes! This needs to be removed as well

tiffany76:
@mjuarez, if you could remove the `/protobuf` from [line 172](https://github.com/open-telemetry/opentelemetry.io/blob/02c97e5f08cc85465964bd3b7f4cd4c5b8d2b29b/content/en/docs/collector/internal-telemetry.md?plain=1#L172) as well, I think we can call this one done!

codeboten:
Should be good to go now 👍🏻

PR #6281, opened 2025-02-11
fix: incorrect check:i18n option in a localization page

Msksgm:
I found the a mistake in contributing/localization page.
There is no `-u` option in `scripts/check-i18n.sh`.

I think that `-c` is correct.

https://github.com/open-telemetry/opentelemetry.io/blob/44059882dcc7aaeab26911ebccc4e09c2a0ba26a/scripts/check-i18n.sh#L39-L57


tiffany76:
@chalin, I think this one requires your expertise.

chalin (approved):
Thank you for catching that 👍🏻

PR #6303, opened 2025-02-13
Cleanup 2023 blog post's in-page CSS

chalin:
- Cleans up in-page styles of a 2023 blog. While we don't generally update old blog posts, they sometimes end up being used as templates or examples.
- Removes unused table-specific styling. For some reason the styles were inactive, maybe because an early run of Prettier put a space between the table and the markdown class attributes (which renders the attributes inactive).

**Preview**, e.g.: https://deploy-preview-6303--opentelemetry.netlify.app/blog/2023/http-conventions-declared-stable/#summary-of-changes


tiffany76 (approved):
RSLGTM.

PR #6311, opened 2025-02-13
[CI] PR-actions workflow: ensure error in cmd name can be reported

chalin:
- For context see https://github.com/open-telemetry/opentelemetry.io/actions/runs/13314886857/job/37186341674. We want the workflow to be able to successfully report an error, so the checkout step has to happen before any step that can fail.
- Reorders steps so that checkout happens earlier.


tiffany76 (approved):
RSLGTM.

PR #6324, opened 2025-02-14
fix: remove duplucated `)`in content/en/docs/concepts/instrumentation/libraries page

Msksgm:
I fixed a tidy mistake in content/en/docs/concepts/instrumentation/libraries page.


chalin (approved):
👍🏻

PR #6336, opened 2025-02-15
[i18n] Move registry search-form alert into registry page

chalin:
- Contributes to #4467
- Moves the alert from the registry search-form shortcode into the page where it is used. This will make it possible for the alert text to be localized more easily.
- There is no change to the generated site files.

/cc @vitorvasc


chalin:
Will merge this first. You can then rebase and update your PR. Thx

PR #6341, opened 2025-02-15
Fix typos in `distributions.yaml`

garysassano:


chalin (approved):
Thanks for submitting a fix!

PR #6364, opened 2025-02-18
[i18n] Make `instrumentation-intro` shortcode a simple include

chalin:
- Contributes to #4467
- Converts the `instrumentation-intro.md` shortcode into a regular markdown file, included via the new includes shortcode. This way we can enable all the regular text, markdown, and i18n checks (for drift etc.). This is done for the following languages that had a `instrumentation-intro.md` shortcode:
  - `en`
  - `es`
  - `pt`
- There is no change in the generated site files where we switched to using `include` instead of a shortcode.
- We do have the interesting side-effect that English fallback pages with an `include` of `instrumentation-intro.md` will be populated with the locale's version of the include. Is this ok with you @open-telemetry/docs-es-approvers @open-telemetry/docs-pt-approvers? It might be a bit strange (and we might need to modify the page-not-translated banner text), but I think that this is be best way moving forward.
  - ~(If we agree to edit the page-not-translated text, I'll do that in a followup PR. I'd probably just add a qualifier like: "... not yet been _fully_ translated ...")~ (I've added the qualifier "fully" to the banner text in this PR.)
- Updates `/xx/_includes` with:
  - A section index so that we can ensure that no page is generated for any includes, including the section itself

**Preview**:

- https://deploy-preview-6364--opentelemetry.netlify.app/docs/languages/go/instrumentation/ - page where the shortcode was replaced by an include. No change in the page content.
- https://deploy-preview-6364--opentelemetry.netlify.app/es/docs/languages/go/instrumentation/ - page where the Spanish include is used in the English page 🤷🏼‍♂️ 
- https://deploy-preview-6364--opentelemetry.netlify.app/pt/docs/languages/go/instrumentation/ - page where the shortcode was replaced by an include. No change in the page content.

### DIFF
--- a/content/en/_includes/_index.md
+++ b/content/en/_includes/_index.md
@@ -1,0 +1,5 @@
+---
+title: # bogus for markdownlint
+cascade:
+  build: { list: never, render: never }
+---
--- a/content/en/_includes/page-not-translated-msg.md
+++ b/content/en/_includes/page-not-translated-msg.md
@@ -3,5 +3,5 @@ title: # bogus entry for markdownlint
 ---
 
 <i class="fa-solid fa-circle-info" style="margin-left: -1.5rem"></i> You are
-viewing the **English version** of this page because it has not yet been
+viewing the **English version** of this page because it has not yet been fully
 translated. Interested in helping out? See [Contributing](/docs/contributing/).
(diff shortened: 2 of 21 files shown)


cartermp (approved):

chalin:
Thanks for the review @cartermp.
Since this is a simple "refactoring", I'll merge now. SIG comments are still welcome.

PR #6369, opened 2025-02-18
Fix typo traceProvider -> tracerProvider

pellared:
Related to https://github.com/open-telemetry/opentelemetry-go-contrib/issues/6802

I guess for Swift it might have even been a bug: https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-swift+registerTracerProvider&type=code

### DIFF
--- a/content/en/blog/2022/go-web-app-instrumentation/index.md
+++ b/content/en/blog/2022/go-web-app-instrumentation/index.md
@@ -183,7 +183,7 @@ Here’s what the setup looks like:
        semconv "go.opentelemetry.io/otel/semconv/v1.4.0"
    )
 
-   func JaegerTraceProvider()(*sdktrace.TracerProvider, error) {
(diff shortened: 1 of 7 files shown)


Kielek (approved):
.NET part - LGTM

chalin (changes requested):
Generally speaking, we let non-`en` maintainers update their pages (which, btw, are associated with specific version of the corresponding `en` page), unless it breaks the build. So please revert the changes to non-`en` pages.

pellared:
> Generally speaking, we let non-`en` maintainers update their pages (which, btw, are associated with specific version of the corresponding `en` page), unless it breaks the build. So please revert the changes to non-`en` pages.

Makes sense. Done

dmathieu (approved):

dashpole (approved):

MrAlias (approved):

PR #6372, opened 2025-02-18
[CI] Support relative links to oteps from OTel spec

chalin:
- In preparation for https://github.com/open-telemetry/opentelemetry-specification/pull/4410
- Ensures that path-relative links to OTEPS are translated into external links (into the spec repo)
- Other cleanup and rule consolidation
- There is no change to the generated site files, but it fixes https://github.com/open-telemetry/opentelemetry.io/actions/runs/13394526898/job/37410079743

/cc @carlosalberto


tiffany76 (approved):
RSLGTM.

PR #6380, opened 2025-02-19
Added Parseable as a vendor

Debanitrkl:
No code changes, added [Parseable](https://github.com/parseablehq/parseable) as a vendor.

Here's the docs on OTel support: https://www.parseable.com/docs/server/opentelemetry


data/ecosystem/vendors.yaml
```diff
@@ -306,6 +306,12 @@
   contact:
   oss: false
   commercial: true
+- name: Parseable
+  nativeOTLP: true
+  url: https://www.parseable.com/docs/server/opentelemetry
+  contact:
```
cartermp, line 312: @Debanitrkl is there a contact field we can list here?

Debanitrkl: Yeah sure will add a contact field

tiffany76 (approved):

PR #6391, opened 2025-02-20
Change wording of "size" to "number of spans".

breedx-splk:
The current wording around 2 config settings in the Java BatchSpanProcessor leaves room for misinterpretation. The use of "size" could suggest bytes or some other data unit, but the actual implementation is in "number of spans"...so this change clarifies that. 

Hopefully this makes things clearer for readers.


trask (approved):
can you make the same change to the "blrp" properties below?

jack-berg (approved):

tiffany76 (approved):

PR #6397, opened 2025-02-21
update protocol to http for internal collector metrics config

codeboten:
This follows the project's recommendation of using http/protobuf instead of grpc.


TylerHelmuth (approved):
Long live http

codeboten:
> Long live http

only when keep alive is on

tiffany76:
Thanks! Could you also update the introductory sentence on line 162?

codeboten:
good catch @tiffany76, updated

cartermp (approved):

PR #6398, opened 2025-02-21
[CI/i18n] Streamline includes and fix pt Aviso

chalin:
- Contributes to #4467
- Drops markdownlint check for a title so that we don't need to include one in included files. I doubt we've every had someone include a regular doc page without a title. (In any case, I'll look into creating a custom rule later if ever this becomes an issue).
- Greatly streamlines the "includes" features by making use of targeted cascades across (multilingual) sites
  - Deletes all `**/_includes/_index.md` files since we no longer need them
  - Drops bogus titles from all includes files
- Moves `content/en/docs/languages/js/_browser-instrumentation-warning.md` back into an includes folder since this will be required moving forward.
- Because of recent past changes to the includes, the `pt` aviso for JS pages wasn't being picked up, instead the English version was displayed, this is now fixed.

**Previews**:

- https://deploy-preview-6398--opentelemetry.netlify.app/fr/docs/
- https://deploy-preview-6398--opentelemetry.netlify.app/pt/docs/languages/js/getting-started/browser/ as an example of the fixed browser Aviso
- https://deploy-preview-6398--opentelemetry.netlify.app/docs/languages/js/getting-started/browser/

Proof that no `_includes` resources/pages are getting generated:

```console
$ find public -name "_includes*" | wc -l
       0
```

Also, https://deploy-preview-6398--opentelemetry.netlify.app/_includes/ is a 404.

### DIFF
--- a/.markdownlint.json
+++ b/.markdownlint.json
@@ -1,5 +1,6 @@
 {
   "emphasis-style": false,
+  "first-line-h1": false,
   "line-length": false,
   "link-fragments": false,
   "list-marker-space": false,
--- a/content/en/_includes/_index.md
+++ b/content/en/_includes/_index.md
@@ -1,5 +1,0 @@
----
-title: # bogus for markdownlint
-cascade:
-  build: { list: never, render: never }
----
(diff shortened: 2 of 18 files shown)


tiffany76 (approved):
RSLGTM.

PR #6421, opened 2025-02-26
[registry] Add otelgoyek instrumenation library

pellared:
As an author of https://github.com/goyek/goyek, I want to add the https://pkg.go.dev/github.com/goyek/x/otelgoyek (also maintained by myself) to the registry.


data/registry/instrumentation-go-goyek.yml
```diff
@@ -0,0 +1,19 @@
+# cSpell:ignore goyek, otelgoyek, Pająk
+title: goyek instrumentation
```
chalin, line 2:
```suggestion
title: Goyek instrumentation
```

pellared: Even if the library is named "lower case"?

chalin: Right, I understood that the package ("brand") name is `goyek` but this is a title.

Also, from your logo, it looks like GOyek to me :)
Or maybe it's GOYEK. All caps would make more sense to me.

Regardless, I don't feel strongly about any of this, but my _preference_ would be for the first letter of the title to be capitalized.

data/registry/instrumentation-go-goyek.yml
```diff
@@ -0,0 +1,19 @@
...
+description: Package otelgoyek provides OpenTelemetry instrumentation for goyek.
+authors:
+  - name: Robert Pająk
+    url: https://github.com/pellared
+urls:
+  repo: https://github.com/goyek/x/tree/main/otelgoyek
+  docs: https://pkg.go.dev/github.com/goyek/x/otelgoyek
+createdAt: 2025-02-26
+isNative: false
+isFirstParty: true
```
chalin, line 19: I don't think that this should be marked as first-party. I've created an issue to track the discussion and possible followup fixed:

- #6428

data/registry/instrumentation-go-goyek.yml
pellared: I respect your preference :wink: Applying suggestion

data/registry/instrumentation-go-goyek.yml
pellared: I did this in accordance with 

https://github.com/open-telemetry/opentelemetry.io/blob/61efc4685844e84b0b6995aff7e7eb0c27f3bd7f/templates/registry-entry.yml#L18

chalin (approved):

PR #6427, opened 2025-02-26
Specify correct annotation for span attributes

sergimola:
FIx small typo


chalin (approved):
Thanks. RSLGTM, and waiting on Java approvers before merging.

trask (approved):

PR #6440, opened 2025-02-28
Automated GitHub Action to Prevent PRs from Main Branch

shalini-saloni:
**Problem:** Sometimes, contributors accidentally submit PRs from the main branch instead of a separate branch.
### Why is this a problem?
If a PR is created from main, maintainers can't easily make edits or use automated /fix commands.
It can cause conflicts if multiple PRs are based on the same main branch.
It can lead to issues when merging, since main is usually the production or stable branch.
**Proposed Solution:**
I implemented a GitHub Actions check that automatically fails a PR if it originates from the main branch. This ensures contributors receive an immediate warning, preventing maintainers from discovering the issue later in the review process. 🚀

### DIFF
--- a/.github/workflows/check-branch.yml
+++ b/.github/workflows/check-branch.yml
@@ -1,0 +1,19 @@
+name: Enforce PR Branch Naming
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize]
+
+jobs:
+  check-branch-name:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Fail if PR is from 'main' branch
+        run: |
+          if [[ "${{ github.head_ref }}" == "main" ]]; then
+            echo "❌ Pull requests must not be created from the 'main' branch."
+            echo "❗ Please create a new branch and submit your PR again."
+            exit 1
+          else
+            echo "✅ Branch name is valid."
+          fi


.github/workflows/check-branch.yml
```diff
@@ -0,0 +1,19 @@
...
+  pull_request:
+    types: [opened, reopened, synchronize]
+
+jobs:
+  check-branch-name:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Fail if PR is from 'main' branch
+        run: |
+          if [[ "${{ github.head_ref }}" == "main" ]]; then
```
pichlermarc, line 13: I'm not a security expert, but I think there may be a risk of template injection here. 
`github.head_ref` is user-controllable input IIUC. I can't come up with something that would do harm right of the bat, but I can break the workflow by using a branch name like `foo";hostname`

Might be safer to have it be put in an env variable (outside of `run`) as to not influence script generation.

pichlermarc (changes requested):

opentelemetry.io
svrnm, line 1: remove this submodule

svrnm (changes requested):
While it is best practice indeed to not use the `main` branch for PRs, I am not worried that much about it:

> If a PR is created from main, maintainers can't easily make edits or use automated /fix commands.

That's not correct. the edits are allowed and I can run fix commands, this is only problematic if the PR comes from an ORG and not a USER. (This PR itself is from your main branch and I can see that github allows me to edit it)

>  It can cause conflicts if multiple PRs are based on the same main branch.

True, but that's then the problem of the contributor to figure out :-) 

>  It can lead to issues when merging, since main is usually the production or stable branch.

That statement makes no sense. If someone forks our repo and creates a PR from their main branch, nothing of that is production/stable. 

While I appreciate your time and effort, I don't think adding this workflow is worth the maintanance it will add to the repo

cartermp:
Closing old PR

Issue #6443, opened 2025-02-28
[i18n] Add banner to non-en pages that have drifted


chalin:
- Subtask of #4467 

### Tasks

- [ ] Enable means for scripts to know when pages have drifted
- [ ] ...

chalin:
This has been done for a while. Moving the remaining support item to another issue.

PR #6444, opened 2025-02-28
[CI] check-i18n: add support for drifted_from_default field

chalin:
- A first contribution to #6443
- Adds support to `scripts/check-i18n.sh` of a new flag `-D`, which will add a (~) boolean page-has-drifted front-matter field so that it can eventually be used to post a banner.
- No change to generated site files, nor any net change to the GH actions (no new checks enabled yet)


tiffany76 (approved):
RSLGTM.

PR #6446, opened 2025-02-28
[i18n] Add `drifted_from_default` status to drifted pages, adjust script, add "Drift status" section to l10n page

chalin:
- Followup to #6444
- Contributes to #6443
- Adds a section about page "Drift status" to the Localization page
- Adds `drifted_from_default: (true|file not found)` to non-English pages that are currently out-of-date


tiffany76 (approved):
RSLGTM.

PR #6253, opened 2025-02-08
fix: Unsupported markdown: list due to missing parsed period in contributing/pull-requests page

Msksgm:
Mermaid does not show proper graphs due to missing parsed period.

Fixes #6249

PR #6454, opened 2025-03-02
Fix flowchart syntax

lehmannro:
The Mermaid chart under *Discovery* currently renders with an error:
```
Unsupported markdown: list
```

Quotation marks are used to escape [characters that break Mermaid's own syntax](https://mermaid.js.org/syntax/flowchart.html#special-characters-that-break-syntax).  For Markdown escaping, [entity codes](https://mermaid.js.org/syntax/flowchart.html#entity-codes-to-escape-characters) have to be used.

### DIFF
--- a/content/en/blog/2024/prom-and-otel/index.md
+++ b/content/en/blog/2024/prom-and-otel/index.md
@@ -271,9 +271,9 @@ flowchart RL
   oc1 --> ta
   oc2 --> ta
   oc3 --> ta
-  sm ~~~|"1. Discover Prometheus Operator CRs"| sm
-  ta ~~~|"2. Add job to TA scrape configuration"| ta
-  oc3 ~~~|"3. Add job to OTel Collector scrape configuration"| oc3
+  sm ~~~|1#46; Discover Prometheus Operator CRs| sm
+  ta ~~~|2#46; Add job to TA scrape configuration| ta
+  oc3 ~~~|3#46; Add job to OTel Collector scrape configuration| oc3
 ```
 
 Even though Prometheus is not required to be installed in your Kubernetes


chalin (changes requested):
Thanks for submitting a fix @lehmannro 🙌🏻 

Let's switch to using a simple escape character rather than using entity codes. For an example of how we did this before, see https://github.com/open-telemetry/opentelemetry.io/pull/6253/files.

svrnm:
@lehmannro @chalin please take another look if we can make this PR move forward

content/en/blog/2024/prom-and-otel/index.md
```diff
@@ -271,9 +271,9 @@ flowchart RL
   oc1 --> ta
   oc2 --> ta
   oc3 --> ta
-  sm ~~~|"1. Discover Prometheus Operator CRs"| sm
-  ta ~~~|"2. Add job to TA scrape configuration"| ta
-  oc3 ~~~|"3. Add job to OTel Collector scrape configuration"| oc3
+  sm ~~~|1#46; Discover Prometheus Operator CRs| sm
+  ta ~~~|2#46; Add job to TA scrape configuration| ta
+  oc3 ~~~|3#46; Add job to OTel Collector scrape configuration| oc3
```
chalin, line 276:
```suggestion
  sm ~~~|1\. Discover Prometheus Operator CRs| sm
  ta ~~~|2\. Add job to TA scrape configuration| ta
  oc3 ~~~|3\. Add job to OTel Collector scrape configuration| oc3
```

chalin (approved):
LGTM now :)

PR #6475, opened 2025-03-05
[i18n] Display page-outdated banner for non-en pages that have drifted

chalin:
- Contributes to #6443
- Adds support for a "this page is outdated" banner for pages that have `drifted_from_default` set to a value other than false.
- I've not yet setup the banner so that it can be easily localized. That will come later.

### Previews

- https://deploy-preview-6475--opentelemetry.netlify.app/fr/docs/what-is-opentelemetry/ - page outdated because the English page has changed
- https://deploy-preview-6475--opentelemetry.netlify.app/es/docs/kubernetes/ - page outdated because the English page has moved

### Screenshots (corresponding to the pages above)

> [!NOTE]
> The banner illustrated next has a [`<details>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) element, which is show in its **expanded** form.

> image

> image

### DIFF
--- a/content/ja/docs/collector/_index.md
+++ b/content/ja/docs/collector/_index.md
@@ -5,6 +5,7 @@ cascade:
   vers: 0.120.0
 weight: 270
 default_lang_commit: cd90ab77550fb2e92ca37cb3c753ec2d8bb7d8dc
+drifted_from_default: true
 ---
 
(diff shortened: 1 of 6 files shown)


tiffany76 (approved):
RSLGTM.

Issue #6460, opened 2025-03-03
[i18n] Factor prose out of shortcodes like `docs/latest-release`


chalin:
This idea is to **separate concerns**, and put prose in includes files and code in shortcodes.

Context:

- #6337 and other non-en languages
- https://deploy-preview-6337--opentelemetry.netlify.app/ja/docs/collector/distributions/
  > <img width="633" alt="Image" src="https://github.com/user-attachments/assets/ceef853d-13bc-4e02-b2a5-f00062b2448f" />
    Which is solved by #6506

### Tasks

- [x] #6476 
- [x] #6494 
- [x] #6506 
- [x] #6581
- [x] #6583
- [x] `languages/index-intro`:
  - [x] `en` - #6750 
  - [x] `pt` - #6751
  - [x] `ja` - #6753
- [x] #6814
- [x] Cleanup & prep before making further changes:
  - [x] #6821
  - [x] #6822
- [ ] More cleanup:
  - [x] #6829
  - [x] #6836
  - [x] #6941
- [ ] `languages/resources-intro.md`
  - [x] `en`
    - [x] #6816
    - [x] #6824
  - [x] `pt` - #6829 by @vitorvasc 
  - [x] `ja` - #6840
- [ ] Consider moving `content/en/docs/_includes/exporters/intro.md` to be under `docs/languages/_includes`, and using an absolution site-specific path in the shortcode.
- [ ] ...

Related:

- [ ] Eliminate `no-duplicate-heading` exception from pages like exporters

Msksgm:
@chalin 

I think that the below issue is solved in https://github.com/open-telemetry/opentelemetry.io/pull/6506.
Then, I mentioned this issue in the PR.

> https://deploy-preview-6337--opentelemetry.netlify.app/ja/docs/collector/distributions/
>  ![Image](https://github.com/user-attachments/assets/861e2db7-e1fb-4ea7-a262-b4fd6bf1cd90)

svrnm:
@chalin how's the status on that, most TODOs are checked, can this be closed or is there much more missing?

PR #6494, opened 2025-03-07
[infra] Factor out include partial from shortcode

chalin:
- Refactoring in prep for #6460 and other work
- Why this change: we can't call shortcodes from other shortcodes or partials (without resorting to `.RenderString` magic, which doesn't work when we're trying to keep things all in markdown).
- Factors out the core functionality of the `includes.html` shortcode and makes it available as a partial of the same name
- In prep for 6460, this refactors the `docs/languages/index-intro.md` shortcode for better **separate concerns**: have the natural language part in an include file, while keeping the code in a shortcode
  - An include-file version of the `docs/languages/index-intro.md` shortcode
  - A new version of the shortcode (suffixed with "2"), which isn't currently used

The only change in the generated site files are for pages with includes. Any included text with an apostrophe (`'`) is now encoded as `&rsquo;`:

```console
$ (cd public && g diff -bw --ignore-blank-lines -I "'|&rsquo;") | grep ^diff | head
diff --git a/sitemap.xml b/sitemap.xml
```

**Preview**: https://deploy-preview-6494--opentelemetry.netlify.app/docs/languages/cpp/instrumentation/

Here's what the diff looks like:

> image

### DIFF
--- a/content/en/docs/languages/_includes/index-intro.md
+++ b/content/en/docs/languages/_includes/index-intro.md
@@ -1,0 +1,24 @@
+---
+---
+
+This is the OpenTelemetry {{ $name }} documentation. OpenTelemetry is an
+observability framework -- an API, SDK, and tools that are designed to aid in
+the generation and collection of application telemetry data such as metrics,
+logs, and traces. This documentation is designed to help you understand how to
+get started using OpenTelemetry {{ $name }}.
+
+## Status and Releases
+
+The current status of the major functional components for OpenTelemetry
+{{ $name }} is as follows:
+
+| Traces              | Metrics              | Logs              |
+| ------------------- | -------------------- | ----------------- |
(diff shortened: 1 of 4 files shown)


content/en/docs/languages/_includes/index-intro.md
```diff
@@ -0,0 +1,24 @@
+---
+---
+
+This is the OpenTelemetry {{ $name }} documentation. OpenTelemetry is an
```
tiffany76, line 4: I know this introduction is how it's been done for ages, but I find "This is" a bit weak, like it points out the obvious. Maybe we could say something like "Welcome to the OpenTelemetry (name) documentation...."?

tiffany76 (approved):
RSLGTM, with one non-blocking question.

content/en/docs/languages/_includes/index-intro.md
chalin: Agreed. Let's tackle that change in a followup PR.

PR #6495, opened 2025-03-07
[CI] Warn about spec pages without level 1 headings

chalin:
- Adjusts the `adjust-pages.pl` script to emit a warning when a spec page doesn't have a level-1 heading, as was the case for https://github.com/open-telemetry/semantic-conventions/issues/1971


tiffany76 (approved):
RSLGTM.

PR #6543, opened 2025-03-13
Add docs for @WithSpan params.

breedx-splk:
The documentation for Java's `@WithSpan` annotation currently omits the `kind` (SpanKind control) parameter. With the upcoming 2.14.0 release, we're also gaining another parameter (`inheritContext`)...so this seemed like a good opportunity to document them both.


theletterf (approved):
Long time no see, hermano! :)

LGTM. Approving.

jaydeluca (approved):

PR #6546, opened 2025-03-14
Updated description/ details for Purview.Telemetry.SourceGenerator in Registry

kieronlanning:
Updated the description for Purview.Telemetry.SourceGenerator based on a recent major change.

Also updated:

- The tags, to match the ones in NuGet
- The links, to include a link to the wiki providing consumers detailed documentation.
-


theletterf (approved):
Thanks!

PR #6572, opened 2025-03-19
[en]chore: refine Metric Gauges usage in Go SDK

zhengkezhou1:
We cannot directly compile the content at https://opentelemetry.io/docs/languages/go/instrumentation/#using-gauges, the issue is that the speedGauge is not declared before using

### DIFF
--- a/content/en/docs/languages/go/instrumentation.md
+++ b/content/en/docs/languages/go/instrumentation.md
@@ -551,6 +551,8 @@ import (
 
 var fanSpeedSubscription chan int64
 
+var speedGauge metric.Int64Gauge
+
 func init() {
 	speedGauge, err := meter.Int64Gauge(
 		"cpu.fan.speed",
--- a/content/en/docs/languages/go/instrumentation.md
+++ b/content/en/docs/languages/go/instrumentation.md
@@ -512,7 +512,8 @@ import (
 var itemsCounter metric.Int64UpDownCounter
 
 func init() {
-	itemsCounter, err := meter.Int64UpDownCounter(
+	var err error
+	itemsCounter, err = meter.Int64UpDownCounter(
 		"items.counter",
 		metric.WithDescription("Number of items."),
 		metric.WithUnit("{item}"),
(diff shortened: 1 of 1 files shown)


tiffany76:
Thanks, @zhengkezhou1! 

@open-telemetry/go-instrumentation-approvers, PTAL. Thanks!

content/en/docs/languages/go/instrumentation.md
```diff
@@ -551,6 +551,8 @@ import (
 
 var fanSpeedSubscription chan int64
 
+var speedGauge metric.Int64Gauge
```
dmathieu, line 554: Readability nit:

```suggestion
var (
  fanSpeedSubscription chan int64
  speedGauge metric.Int64Gauge
)
```

dmathieu (approved):

content/en/docs/languages/go/instrumentation.md
```diff
@@ -551,6 +551,8 @@ import (
...
+
 func init() {
 	speedGauge, err := meter.Int64Gauge(
```
MrAlias, line 557:
```suggestion
	var err error
	speedGauge, err = meter.Int64Gauge(
```

MrAlias: Similar for `getCPUFanSpeed`.

zhengkezhou1: I think this should not be changed, since `err` is only used once.

MrAlias: That's not how assignment works in Go.

content/en/docs/languages/go/instrumentation.md
```diff
@@ -512,8 +512,7 @@ import (
 var itemsCounter metric.Int64UpDownCounter
 
 func init() {
-	var err error
-	itemsCounter, err = meter.Int64UpDownCounter(
+	itemsCounter, err := meter.Int64UpDownCounter(
```
MrAlias, line 515: This is wrong. [Please see how Go makes assignments.](https://go.dev/ref/spec#Assignment_statements)

MrAlias (changes requested):

content/en/docs/languages/go/instrumentation.md
zhengkezhou1: thanks for that, i'm newbie for Golang.

MrAlias (approved):

tiffany76 (approved):

PR #6574, opened 2025-03-19
[blog] Announce JS SDK 2.0

JamieDanielson:
OpenTelemetry JavaScript recently released SDK 2.0. There have been many improvements and breaking changes and we want to make this easily findable.

---

**Preview**: https://deploy-preview-6574--opentelemetry.netlify.app/blog/2025/otel-js-sdk-2-0/

PR #6578, opened 2025-03-20
Add nudge to PR actions

theletterf:
This PR adds a /nudge action that relies on `gh run rerun`: https://cli.github.com/manual/gh_run_rerun

The idea is addressing the need for rerunning checks after a /fix command. I wonder if adding this command at the end of all fix commands could also solve the situation.

Might potentially fix https://github.com/open-telemetry/opentelemetry.io/issues/4339

### DIFF
--- a/.github/workflows/pr-actions.yml
+++ b/.github/workflows/pr-actions.yml
@@ -118,3 +118,61 @@ jobs:
           gh pr comment $PR_NUM -b "fix:${PR_ACTION} failed or was cancelled. For details, see $GITHUB_SERVER_URL/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID."
         env:
           GH_TOKEN: ${{ secrets.OPENTELEMETRYBOT_GITHUB_TOKEN }}
+
+  nudge-action:
+    name: Rerun PR checks
+    runs-on: ubuntu-latest
+
+    if: |
+      github.event.issue.pull_request &&
+      github.event.comment.body == '/nudge'
+    permissions:
+      contents: read
+      pull-requests: write
+
+    steps:
+      - name: Context info
+        run: |
+          echo $PR_NUM
+          echo $COMMENT
+        env:
+          PR_NUM: ${{ github.event.issue.number }}
+          COMMENT: ${{ github.event.comment.body }}
+
+      - name: Write start comment
+        run: |
+          gh pr comment $PR_NUM -b "You triggered /nudge action run at $GITHUB_SERVER_URL/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID"
+        env:
+          GH_TOKEN: ${{ secrets.OPENTELEMETRYBOT_GITHUB_TOKEN }}
+          PR_NUM: ${{ github.event.issue.number }}
+
+      - name: Rerun PR checks
+        run: |
+          # Get the latest check run to re-run all checks
+          gh pr checks $PR_NUM --repo $GITHUB_REPOSITORY --watch
+          # This doesn't actually re-run them, but shows the status
+          # Now re-run them
+          gh run rerun $(gh run list --branch $(gh pr view $PR_NUM --json headRefName -q .headRefName) --limit 1 --json databaseId -q '.[0].databaseId')
+        env:
+          GH_TOKEN: ${{ secrets.OPENTELEMETRYBOT_GITHUB_TOKEN }}
+          PR_NUM: ${{ github.event.issue.number }}
+
+      - name: Report success
+        if: ${{ !failure() && !cancelled() }}
+        run: |
+          gh pr comment $PR_NUM --body "$(cat <<EOF
+          \`/nudge\` check re-run was [triggered]($GITHUB_SERVER_URL/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID).
+          You can check the status of your PR checks in the PR's Checks tab.
+          EOF
+          )"
+        env:
+          GH_TOKEN: ${{ secrets.OPENTELEMETRYBOT_GITHUB_TOKEN }}
+          PR_NUM: ${{ github.event.issue.number }}
+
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-actions.yml
```diff
@@ -118,3 +118,61 @@ jobs:
...
+          GH_TOKEN: ${{ secrets.OPENTELEMETRYBOT_GITHUB_TOKEN }}
+          PR_NUM: ${{ github.event.issue.number }}
+
+      - name: Rerun PR checks
+        run: |
+          # Get the latest check run to re-run all checks
+          gh pr checks $PR_NUM --repo $GITHUB_REPOSITORY --watch
+          # This doesn't actually re-run them, but shows the status
+          # Now re-run them
+          gh run rerun $(gh run list --branch $(gh pr view $PR_NUM --json headRefName -q .headRefName) --limit 1 --json databaseId -q '.[0].databaseId')
```
chalin, line 155: There's a lot going on in this line. Could you factor out the fetching of argument values, possibly putting them in an `env:` variable for this step?

E.g., I did this locally or the (current) stuck PR #6574:

```console
$ PR_NUM=6574 
$ BRANCH=$(gh pr view $PR_NUM --json headRefName -q .headRefName)
$ RUNID=$(gh run list --branch $(gh pr view $PR_NUM --json headRefName -q .headRefName) --limit 1 --json databaseId -q '.[0].databaseId')
$ gh run rerun $RUNID                                                                                                                    
✓ Requested rerun of run 13966456806
```

But nothing has been happening to the PR on GH.

chalin (commented):
Cool! ✨ 

Have you tested this locally? I did, and the rerun request doesn't seem to be doing anything in GH (see inline comment for details):

```console
$ gh run rerun $RUNID                                                                                                                    
✓ Requested rerun of run 13966456806
```

.github/workflows/pr-actions.yml
chalin: Btw, I've rebased #6574 (because I wanted to see a clean check status), so it isn't stuck anymore.

theletterf:
@chalin I reviewed the action and went for a simpler approach, that is, creating and pushing an empty commit on demand. As we squash commits when merging PRs, this adds no noise to the git history of `main` and avoids complex checks-checking logic.

chalin:
As we talked about out-of-band, I'd tried that approach before, but GH would fail to rerun checks for empty commits. But if you can get it to work, that would be great!

theletterf:
@chalin PTAL

.github/workflows/pr-actions.yml
```diff
@@ -118,3 +118,13 @@ jobs:
           gh pr comment $PR_NUM -b "fix:${PR_ACTION} failed or was cancelled. For details, see $GITHUB_SERVER_URL/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID."
         env:
           GH_TOKEN: ${{ secrets.OPENTELEMETRYBOT_GITHUB_TOKEN }}
+
+  rerun_pr_tests:
+    name: rerun_pr_tests
+    if: ${{ github.event.issue.pull_request }}
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: estroz/rerun-actions@main
```
svrnm, line 127: can you pin this to a version or even better to a hash? Pulling from main from a repo that has not been updated 4 years sounds risky

.github/workflows/pr-actions.yml
```diff
@@ -118,3 +118,13 @@ jobs:
           gh pr comment $PR_NUM -b "fix:${PR_ACTION} failed or was cancelled. For details, see $GITHUB_SERVER_URL/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID."
         env:
           GH_TOKEN: ${{ secrets.OPENTELEMETRYBOT_GITHUB_TOKEN }}
+
+  rerun_pr_tests:
+    name: rerun_pr_tests
+    if: ${{ github.event.issue.pull_request }}
+    runs-on: ubuntu-20.04
```
svrnm, line 125: can we run this on a more recent runner as well or does this only work on the 20.04 runner?

svrnm (changes requested):
great work, some comments around picking the right runners + versions

.github/workflows/pr-actions.yml
theletterf: Thought about that, though the logic seems simple enough. Another option would be to rebuild the action ourselves, I guess. In any case, It's pinned now.

.github/workflows/pr-actions.yml
theletterf: Updated to latest!

theletterf:
Does this one still makes sense, @svrnm ?

.github/workflows/pr-actions.yml
```diff
@@ -118,3 +118,12 @@ jobs:
           gh pr comment $PR_NUM -b "fix:${PR_ACTION} failed or was cancelled. For details, see $GITHUB_SERVER_URL/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID."
         env:
           GH_TOKEN: ${{ secrets.OPENTELEMETRYBOT_GITHUB_TOKEN }}
+
+  rerun_pr_tests:
+    name: rerun_pr_tests
+    if: ${{ github.event.issue.pull_request }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: estroz/rerun-actions@v0.3.0 with:
```
svrnm, line 127:
```suggestion
      - uses: estroz/rerun-actions@6da343594fa5cc0869523be9b43ed9256c68d39b # v0.3.0 
         with:
```

Pin the version, fix the missing line break

svrnm:
Is https://github.com/open-telemetry/opentelemetry.io/issues/4339 still an issue? I was under the assumption that the way we have solved it today it "works as expected?"

theletterf:
Agreed! Let's close this one for now.

Issue #6592, opened 2025-03-21
[CI] Cron job workflow to run `test-and-fix` and commit results


chalin:
### Tasks

- [x] Prep: #10320
- [x] #10332

---

#### 2025/06

Here's what I think we need two reusable actions (such as [these actions from our repo](https://github.com/open-telemetry/opentelemetry.io/tree/main/.github/actions)) to do the each of the following (as separate actions):

- Apply any `npm run XXX` script (un trusted), and creates a patch files
- A reusable action (trusted), that can create a PR from a patch file

Create another action, or just a script to run `test-and-fix`, using the two actions previously mentioned.

#### Original comment

We should consider setting up a workflow to execute `npm run test-and-fix` and commit results, so as to minimize the likelihood of our contributors being told to keep their PRs focused to their own changes, like https://github.com/open-telemetry/opentelemetry.io/pull/6585#issuecomment-2743700862. /cc @svrnm

chalin:
Actually, as [discussed on Slack](https://cloud-native.slack.com/archives/C06EDFPQ5EH/p1742572953060709?thread_ts=1742427770.338439&cid=C06EDFPQ5EH), let's run the `i18n` checks on every PR instead.

svrnm:
> Actually, as [discussed on Slack](https://cloud-native.slack.com/archives/C06EDFPQ5EH/p1742572953060709?thread_ts=1742427770.338439&cid=C06EDFPQ5EH), let's run the `i18n` checks on every PR instead.

agreed, that's the one that makes the most issues

PR #6593, opened 2025-03-21
Test-and-fix results

chalin:
- Context: #6592
- Results from running `npm run test-and-fix` 
- Adds `pnpm-lock.yaml` to `.gitignore`, just in case folks are using `pnpm` instead of `npm` -- as was done in #6592


tiffany76 (approved):
RSLGTM.

PR #6597, opened 2025-03-21
Docsy upgrade to bring in new heading render hook

chalin:
- Upgrades Docsy to bring in new heading-render-hook functionality, in support of #6595 and use of `lang="en"` attributes
- Spell-check fixes


tiffany76 (approved):
RSLGTM.

PR #6641, opened 2025-04-03
Document java explicit span name with @WithSpan annotation

SylvainJuge:
Documents how to set a custom span name when using `@WithSpan` annotation.

### DIFF
--- a/content/en/docs/zero-code/java/agent/annotations.md
+++ b/content/en/docs/zero-code/java/agent/annotations.md
@@ -60,8 +60,8 @@ public class MyClass {
 
 Each time the application invokes the annotated method, it creates a span that
 denotes its duration and provides any thrown exceptions. By default, the span
-name will be `<className>.<methodName>`, unless a name is provided as an
-argument to the annotation.
+name will be `<className>.<methodName>`, unless a name is provided through the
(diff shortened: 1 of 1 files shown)


content/en/docs/zero-code/java/agent/annotations.md
```diff
@@ -87,16 +87,22 @@ customization of spans:
 
 | name             | type              | default    | description                                                                                                                                  |
 | ---------------- | ----------------- | ---------- | -------------------------------------------------------------------------------------------------------------------------------------------- |
+| `value`          | `String`          | `""`       | The span name, when empty (default) `<className>.<methodName>` will be used.                                                                 |
```
tiffany76, line 90:
```suggestion
| `value`          | `String`          | `""`       | The span name. If not specified, the default `<className>.<methodName>` is used.                                                                 |
```

tiffany76 (approved):
Thanks, @SylvainJuge! I made a suggestion for clarity and to avoid future tense. Hopefully I didn't change the meaning. Otherwise, LGTM!

Let's also get a review from another @open-telemetry/java-instrumentation-approvers.

jaydeluca (approved):

PR #6651, opened 2025-04-06
Highlight a gotcha when sending OTLP data to hostPort

temach:
As a newcomer to OTEL it took me quite a while to figure out the root cause why pod metrics were not enriched with k8s attributes (k8s.pod.name, k8s.namespace.name, etc.) after I followed the best practices for deploying otel. 

I installed otel in `mode: daemonset` via helm chart https://open-telemetry.github.io/opentelemetry-helm-charts and enabled multiple presets, in particular the kubernetesAttributes preset https://opentelemetry.io/docs/kubernetes/collector/components/#kubernetes-attributes-processor and followed the kubernetes best practices to use `hostPort` as written here - https://opentelemetry.io/docs/security/config-best-practices/#kubernetes

However sending to hostPort means that more configuration is necessary for the k8sattributes processor to discover the pod metadata, as connection info is lost (the node's ip address appears as ip source to the collector).  More configuration can be e.g. setting a resource attribute (k8s.pod.ip or k8s.pod.uid) on the pod via an ENV var https://opentelemetry.io/docs/languages/sdk-configuration/general/#otel_resource_attributes and then this attribute can be matched in the  k8sattributes processor config.

With this change I want to highlight this gotcha, so it does not waste so much time for others.

### DIFF
--- a/content/en/docs/security/config-best-practices.md
+++ b/content/en/docs/security/config-best-practices.md
@@ -239,6 +239,15 @@ env:
         fieldPath: status.hostIP
 ```
 
+However when sending OTLP data to `hostPort` this way, the default configuration of 
+[`k8sattributes` processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/k8sattributesprocessor) 
+prevents enriching pod OTLP data with k8s metadata (i.e. will not attach k8s.pod.name, 
+k8s.namespace.name, etc.). If Pod association rules are not configured for `k8sattributes`
+processor, resources are associated with metadata only by connection's IP Address. 
+Unfortunatelly sending via `hostPort` means that the Collector sees this connection
+as coming from the node and is unable to associate the pod with incoming OTLP data
+in its default configuration.
+
 ### Scrub sensitive data
 
 [Processors](/docs/collector/configuration/#processors) are the Collector


content/en/docs/security/config-best-practices.md
```diff
@@ -239,6 +239,15 @@ env:
...
 ```
 
+However when sending OTLP data to `hostPort` this way, the default configuration of 
+[`k8sattributes` processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/k8sattributesprocessor) 
+prevents enriching pod OTLP data with k8s metadata (i.e. will not attach k8s.pod.name, 
+k8s.namespace.name, etc.). If Pod association rules are not configured for `k8sattributes`
+processor, resources are associated with metadata only by connection's IP Address. 
+Unfortunatelly sending via `hostPort` means that the Collector sees this connection
+as coming from the node and is unable to associate the pod with incoming OTLP data
+in its default configuration.
```
theletterf, line 249:
```suggestion
When sending OTLP data to `hostPort` using this method, the default configuration of
the [`k8sattributes` processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/k8sattributesprocessor) 
prevents enriching pod OTLP data with k8s metadata. For example, the k8s.pod.name and
the k8s.namespace.name attributes won't be included. If pod association rules aren't 
configured for the `k8sattributes` processor, resources are associated with metadata only
by connection's IP Address.  This happens because the Collector sees this connection
as coming from the node and is unable to associate the pod with incoming OTLP data.
```

theletterf (changes requested):

theletterf:
Thanks for your patience, @temach. Yeah, that would be best!

Issue #6650, opened 2025-04-05
List of 404 Links


B-Mahdj:
**What happened?** Describe the problem that occurred.

I ran a Lychee scan on the website files on local and here are the results for the links that return 404 errors code : 

# Summary

| Status        | Count |
|---------------|-------|
| 🔍 Total      | 5191  |
| ✅ Successful | 3687  |
| ⏳ Timeouts   | 3     |
| 🔀 Redirected | 0     |
| 👻 Excluded   | 1460  |
| ❓ Unknown    | 17    |
| 🚫 Errors     | 24    |

### Errors in ./content/en/blog/2023/perf-testing/index.md

* [404] [https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-sdk/tests/performance/benchmarks/trace/test_benchmark_trace.py](https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-sdk/tests/performance/benchmarks/trace/test_benchmark_trace.py) | Network error: Not Found

### Errors in ./content/en/blog/2023/end-user-discussions-02.md

* [404] [https://github.com/open-telemetry/opentelemetry-collector/tree/main/extension/experimental/storage](https://github.com/open-telemetry/opentelemetry-collector/tree/main/extension/experimental/storage) | Network error: Not Found

### Errors in ./content/en/blog/2022/end-user-discussion-group-apac/index.md

* [404] [https://lu.ma/1w129wgu](https://lu.ma/1w129wgu) | Network error: Not Found

### Errors in ./content/en/docs/demo/services/shipping.md

* [404] [https://docs.rs/opentelemetry/latest/opentelemetry/struct.ContextGuard.html](https://docs.rs/opentelemetry/latest/opentelemetry/struct.ContextGuard.html) | Network error: Not Found

### Errors in ./content/en/blog/2022/otel-demo-app-nomad/index.md

* [ERROR] [https://rubiksqube.com/#/](https://rubiksqube.com/#/) | Network error: error sending request for url (https://rubiksqube.com/#/) Maybe a certificate error?

### Errors in ./content/en/blog/2022/new-end-user-resources.md

* [404] [https://github.com/open-telemetry/community/tree/main/working-groups/end-user](https://github.com/open-telemetry/community/tree/main/working-groups/end-user) | Network error: Not Found

### Errors in ./content/en/blog/2023/ecs-otel-semconv-convergence.md

[...]

PR #6653, opened 2025-04-06
Update link to jaeger client libraries

B-Mahdj:
Related to issue https://github.com/open-telemetry/opentelemetry.io/issues/6650 

The old link was giving a 404 this link should make the job for now. Maybe this link should be deleted in all the content files as it will probably be also deleted from the jager website like the old one ?


content/en/docs/migration/_index.md
```diff
@@ -18,7 +18,7 @@ both [OpenTracing](opentracing/) and [OpenCensus](opencensus/)
 ## Jaeger Client
 
 The [Jaeger community](https://www.jaegertracing.io/) deprecated their
-[Client Libraries](https://www.jaegertracing.io/docs/latest/client-libraries/)
+[Client Libraries](https://www.jaegertracing.io/docs/1.17/client-libraries/)
 and recommends using the OpenTelemetry APIs, SDKs and instrumentations.
```
cartermp, line 22:
```suggestion
client libraries and recommends [migrating](https://www.jaegertracing.io/docs/latest/migration/)
to the OpenTelemetry APIs, SDKs and instrumentations.
```

cartermp (changes requested):
I think we should change these to link to the migration doc for Jaeger instead now.

content/en/docs/migration/_index.md
B-Mahdj: Thanks for your comment. 
After a 5-min thinking I think you're right but on the link since it is simply a page with a link to a Google Doc why not put directly the link to the Google Doc ?

cartermp: I think this link is better because the migration guide could get codified into the docs page at some point and then updated, or the migration doc could move to another doc.

B-Mahdj: I've made the change on the EN version

content/zh/docs/migration/_index.md
```diff
@@ -17,7 +17,7 @@ OpenTelemetry 是 OpenTracing 和 OpenCensus 的合并版本。
 
 ## Jaeger 客户端
 
-[Jaeger 社区](https://www.jaegertracing.io/)弃用了他们的[客户端库](https://www.jaegertracing.io/docs/latest/client-libraries/)并建议使用
+[Jaeger 社区](https://www.jaegertracing.io/)弃用了他们的[客户端库](https://www.jaegertracing.io/docs/1.17/client-libraries/)并建议使用
 OpenTelemetry 的 API、SDK 和插桩器。
```
windsonsea, line 21:
```suggestion
[Jaeger 社区](https://www.jaegertracing.io/)弃用了他们的客户端库并建议[迁移为使用](https://www.jaegertracing.io/docs/latest/migration/)
OpenTelemetry 的 API、SDK 和插桩器。
```

windsonsea (commented):
Please only change one language like English, and leave other for localization teams if your are not familar with.

content/zh/docs/migration/_index.md
B-Mahdj: Thanks for the modification, I've changed the other language because the PR was only supposed to be a link change but it became a change of link and of content.

cartermp (approved):
Thanks!

reyang (approved):

tiffany76 (approved):

PR #6665, opened 2025-04-08
spring starter: use the alpha bom

zeitlinger:
Fixes https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/13667

### DIFF
--- a/content/en/docs/zero-code/java/spring-boot-starter/getting-started.md
+++ b/content/en/docs/zero-code/java/spring-boot-starter/getting-started.md
@@ -26,7 +26,7 @@ A Bill of Material
 ensures that versions of dependencies (including transitive ones) are aligned.
 
 To ensure version alignment across all OpenTelemetry dependencies, you must
-import the `opentelemetry-instrumentation-bom` BOM when using the OpenTelemetry
+import the `opentelemetry-instrumentation-bom-alpha` BOM when using the OpenTelemetry
 starter.
 
 {{% alert title="Note" color="info" %}}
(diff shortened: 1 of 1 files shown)


content/en/docs/zero-code/java/spring-boot-starter/getting-started.md
```diff
@@ -48,7 +48,7 @@ The following example shows how to import the OpenTelemetry BOMs using Maven:
     <dependencies>
         <dependency>
             <groupId>io.opentelemetry.instrumentation</groupId>
-            <artifactId>opentelemetry-instrumentation-bom</artifactId>
+            <artifactId>opentelemetry-instrumentation-bom-alpha</artifactId>
             <version>{{% param vers.instrumentation %}}</version>
```
laurit, line 52: the version should probably also be changed to include the `-alpha` suffix

jeanbisutti:
An alternative: https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/13667#issuecomment-2787120178

tiffany76:
> /fix:format

@zeitlinger We had to disable the fix: GitHub Actions temporarily. You can run the same locally with `npm run fix:format`.

content/en/docs/zero-code/java/spring-boot-starter/getting-started.md
```diff
@@ -26,8 +26,8 @@ A Bill of Material
 ensures that versions of dependencies (including transitive ones) are aligned.
 
 To ensure version alignment across all OpenTelemetry dependencies, you must
-import the `opentelemetry-instrumentation-bom` BOM when using the OpenTelemetry
-starter.
+import the `opentelemetry-instrumentation-bom-alpha` BOM when using the
+OpenTelemetry starter.
```
trask, line 30: this seems like an unfortunate workaround for a conflict between two libraries that shouldn't be used together anyways (otel spring boot starter and micrometer tracing bridge)

if we're going to go with this change, I think we should add some kind of explanation why the `-alpha` bom is needed for the stable otel spring boot starter

svrnm:
@zeitlinger @trask what's the status of this PR?

zeitlinger:
I think we are not sure if an edge case really justifies this - I'm closing it.

PR #6666, opened 2025-04-08
Remove text duplication in dotnet instrumentation

jdom:
The sentence being removed already exists in the layouts/shortcodes/docs/languages/span-status-preamble.md file, so it appears as duplicate in the output docs: https://opentelemetry.io/docs/languages/dotnet/instrumentation/#set-activity-status


tiffany76 (approved):
Good catch! Thanks, @jdom!

PR #6667, opened 2025-04-08
[.NET] Use AddException instead of RecordException

jdom:
activity.RecordException() does not exist


Kielek (approved):
There is a RecordException, but it was marked as obsolete some time ago in favor of native .NET SDK API. Good to merge

Ref. https://github.com/open-telemetry/opentelemetry-dotnet/blob/aa73ce3aafae2a581967fbc0600204169b3f9d08/src/OpenTelemetry.Api/Trace/ActivityExtensions.cs#L102

PR #6668, opened 2025-04-08
Update vendors.yaml for ControlTheory Org

j0nr33v3:
Adding ControlTheory as a vendor


cartermp (approved):
Thanks! We'll handle making sure CI passes as well.

tiffany76 (approved):

PR #6683, opened 2025-04-11
Add Otel demo link to Getting Started Dev

B-Mahdj:
## Goal 
Goal is to align the two Getting Started guides : 

https://opentelemetry.io/docs/getting-started/dev/
https://opentelemetry.io/docs/getting-started/ops/

The Getting Started Ops page have a link to the Otel demo but the Dev page doesn't have one. This should fix it. 

## Additional Notes

Also, I saw that the link goes to this page : 
https://opentelemetry.io/ecosystem/demo/
Which is not the demo page is this supposed to be this way ? 

Here is the direct demo page : 
https://opentelemetry.io/docs/demo/

### DIFF
--- a/content/en/docs/getting-started/dev.md
+++ b/content/en/docs/getting-started/dev.md
@@ -22,6 +22,9 @@ natively:
 
 - [How can I add native instrumentation to my library?](../../concepts/instrumentation/libraries/)
 
+If you are looking for a set of applications to try things out, you will find
+our official [OpenTelemetry demo](/ecosystem/demo/) useful.
+
 Next, you can deep dive into the documentations for the
 [language](../../languages/) you are using:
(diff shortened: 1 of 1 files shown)


theletterf (approved):
Makes sense. Thanks!

B-Mahdj:
Thanks for the fast reply @theletterf 

Do you know anything about this other point I made  ?
> ## Additional Notes
> 
> Also, I saw that the link goes to this page : https://opentelemetry.io/ecosystem/demo/ Which is not the demo page is this supposed to be this way ?
> 
> Here is the direct demo page : https://opentelemetry.io/docs/demo/

theletterf:
I think so, @B-Mahdj.

PR #6719, opened 2025-04-17
Add Tranzzo to OpenTelemetry adopters

illenko:
Tranzzo adopted OpenTelemetry ~ 1 year ago and actively uses it to collect traces and metrics.

OpenTelemetry Collector is a core component where all signals are sent, then data is processed by Prometheus (or VictoriaMetrics for other environment) and Jaeger.

We have 2 main stacks for our backend: Java + Spring and Scala + Play.

For Java + Spring services, we use Micrometer as a Vendor-neutral application observability facade with OTLP exporter.
Scala + Play services instrumented using Java OpenTelemetry SDK.

The metrics and traces gathered through OpenTelemetry play a crucial role in Tranzzo's release processes, ongoing monitoring efforts, and performance optimization initiatives.

We are happy with OpenTelemetry, which allows us to switch between observability backends easily.

For now we do not have blog post or something to add as a reference, but after we create it - I will add it ASAP.


tiffany76 (approved):
Thanks very much!

PR #6747, opened 2025-04-24
Docsy upgrade to use new render-heading partial

chalin:
- In prep for Hugo upgrade.
- Followup to #6743
- Upgrades Docsy to use new render-heading hook partial
- No change to generated site files other than CSS file updates due to Bootstrap upgrade


tiffany76 (approved):
RSLGTM.

Issue #6658, opened 2025-04-07
Add "Ask AI" widget to docs website


dvdksn:
**Desired feature or idea**: What would you like to see happen?

tl;dr - we would love to offer an "Ask AI" widget for your docs as part of our open source support program.

Hello OTel folks 👋🏻 I'm from kapa.ai. We build an Ask AI widget that helps users get answers to technical questions directly from documentation. We have an open source support program where we offer this tool at no cost.

The widget would allow users to ask natural language questions about OpenTelemetry and get accurate answers sourced directly from your documentation. It could help reduce the learning curve for new users and potentially decrease support burden on maintainers.

**Additional context**: anything else you can think about that would be helpful.

This is completely optional - just wanted to put it on your radar in case it's of interest. We've implemented similar solutions for [Polars](https://docs.pola.rs/api/python/stable/reference/index.html), [Nuxt](https://nuxt.com/docs/getting-started/introduction), [Apache Arrow](https://arrow.apache.org/docs/dev/developers/index.html#developers), and feedback has been that it helps users find information faster.

The implementation is lightweight (a small JavaScript snippet) and wouldn't interfere with your existing docs structure. All training would be done exclusively on your public documentation.

Happy to answer any questions or concerns!

cartermp:
I would wholly endorse AI-assisted search/generation of answers from a docs corpus.

svrnm:
@dvdksn thanks for raising, can you share some of the terms and conditions under which an open source project enrolls into your offering?

dvdksn:
Spun up a quick demo deployment, using the docs website and stackoverflow questions as a source. Feel free to [try it out here](https://demo.kapa.ai/widget/opentelemetry) to get a feel for how it works/performs.

austinlparker:
> The Kapa Open Source Program provides qualifying open source projects with a generous number of free questions on the Kapa platform (up to 10,000 questions per month). This allows you to offer AI-powered support and onboarding experiences to your project's users, enhancing their overall experience.

What happens if we go over the limit?

dvdksn:
@austinlparker Nothing! 10,000 is a generic (soft) limit. For OTel we won't restrict the # of questions.

moxious:
[...]

PR #6769, opened 2025-04-28
Add Kapa.ai integration with Ask AI button in navbar

austinlparker:
For #6658, adding the kapa.ai widget to opentelemetry.io.


cartermp:
@austinlparker Do we get the ability to look at the response data from "All feedback is reviewed by the team."?

chalin (changes requested):
Thanks for the preview of the feature.

I'd like to suggest that the nav title be a single icon / SVG (with a hover title that can be "Ask AI"). Otherwise the nav becomes too tight and the search box too narrow in some situations:

image

Implementation-wise, let's not bring this in via a layout override. (I can have something to propose later in the week.)

layouts/partials/hooks/head-end.html
```diff
@@ -25,3 +25,23 @@
...
+  data-project-logo="https://opentelemetry.io/img/logos/opentelemetry-logo-nav.png"
+  data-button-hide="true"
+  data-modal-override-open-id="kapa-ai-trigger"
+  data-modal-border-radius="8px"
+  data-modal-header-bg-color="#4f62ad"
+  data-modal-title="OpenTelemetry Docs AI"
+  data-modal-title-color="#ffffff"
+  data-modal-body-bg-color="#ffffff"
+  data-query-input-border-color="#f5a800"
+  data-modal-disclaimer="This is an experimental feature that uses AI to answer questions using the documentation, OpenTelemetry GitHub, StackOverflow, and other resources. Please give us feedback in [this GitHub issue](https://github.com/open-telemetry/opentelemetry.io/issues/6658)."
```
chalin, line 45:
```suggestion
  data-modal-disclaimer="This is an experimental feature that uses AI to answer questions based on this site's documentation, OpenTelemetry GitHub, StackOverflow, and other resources. Please give us feedback in [this GitHub issue](https://github.com/open-telemetry/opentelemetry.io/issues/6658)."
```

chalin: Also, we might want folks to open new issues rather than comment of the existing one?

austinlparker: I was thinking for now it'd be good to try and consolidate the feedback in one place.

austinlparker:
> Thanks for the preview of the feature.
> 
> I'd like to suggest that the nav title be a single icon / SVG (with a hover title that can be "Ask AI"). Otherwise the nav becomes too tight and the search box too narrow in some situations:
> 
> image
> Implementation-wise, let's not bring this in via a layout override. (I can have something to propose later in the week.)

If you'd like to propose an alternate implementation, that'd be fine with me - I do think it'd be good to have a visible label rather than just an icon.

layouts/partials/hooks/head-end.html
svrnm: I am fine with both, as long as we give people the option to provide feedback

tiffany76:
FWIW, I agree that we should use text to label the AI search box, with or without an icon.

chalin:
I've pushed some changes that:

- Switch to using a Bootstrap popover. The advantage of this solution is that it works under all display widths, without taking up any extra top-nav real estate. 
- Enables Command-K to invoke the Ask AI dialog
- Configured the title and dialog text via i18n keys (that might have been over engineering, but it's done; we can always pull it out).

### Screenshots

Wide:

> image

Pad/tablet:

> image

Mobile:

> image

---

> image

On mobile:

> image

chalin (approved):
I'm ok with this first trial implementation. PTAL

austinlparker:
lgtm, let's ship it!

PR #6777, opened 2025-04-29
Add username to survey message

maryliag:
While sending the same message on the JS sdk repo, we got the feedback that we should add the username to the message, so they get properly pinged, otherwise people can miss it.
Adding the same change here after being tested on the other repo for awhile.


tiffany76 (approved):
RSLGTM. @svrnm, do you want to take a look?

svrnm (approved):

PR #6788, opened 2025-04-30
Results from fix:all, and dropping of obsolete spec patches

chalin:
- Results from running `fix:all`
- Drops obsolete spec patches

Site diff is as expected:

```console
$ (cd public && git diff -- ':(exclude)*.xml') | grep ^diff 
diff --git a/ja/docs/what-is-opentelemetry/index.html b/ja/docs/what-is-opentelemetry/index.html
```


tiffany76 (approved):
RSLGTM.

PR #6820, opened 2025-05-05
[Getting Started] Add reference app specification

svrnm:
Contributes to the effort to rework our getting started guides

### DIFF
--- a/content/en/docs/getting-started/reference-application-specification.md
+++ b/content/en/docs/getting-started/reference-application-specification.md
@@ -5,20 +5,25 @@ description:
   This document contains the specification for the reference application that is
   used across the Getting Started guide
 toc_hide: true
+cspell:ignore: uninstrumented rolldice
 ---
 
-# General requirements
+The purpose of the reference application is to have a standardized sample
+application that can be implemented in all languages for which an OpenTelemetry
+SDK exists.
+
+## General requirements
 
 - Implementations of the Reference App are owned by language SIGs implementing
   the OpenTelemetry APIs and SDKs.
 - There must be an uninstrumented and instrumented version of the application.
-- There must be a language-specific CI action that verifies the
-  application builds and runs in both versions.
+- There must be a language-specific CI action that verifies the application
+  builds and runs in both versions.
 - The application must be runnable from a command line interface.
 - There must be a Dockerfile to run the application within a containerized
   environment.
-- The application must run standing alone. In other words, it cannot take hard dependencies
-  on any other content of the repository enclosing it.
+- The application must run standing alone. In other words, it cannot take hard
+  dependencies on any other content of the repository enclosing it.
 
 ## Service requirements
 
@@ -27,45 +32,76 @@ toc_hide: true
 - For handling the HTTP requests, a library should be used for which an
   instrumentation library is available. The application must provide the
   endpoint `/rolldice?rolls=<n>` via `GET` (and optionally `POST`) and return
-  the following HTTP status codes and JSON results: - if `rolls` is not set or
-  has a valid input (positive integer > 0): status code `200` and either a
-  single number between 1 and 6 if `rolls` is not set or 1, or an array of `n`
-  numbers between 1 and 6 where `n` is the value of `rolls` - if `rolls` is set
-  to an invalid input (not a number): status code `400` and
-  `{"status": "error", "message": "Parameter rolls must be a positive integer"}` -
-  if `rolls` is set to a negative integer: status code `500` and no JSON output.
+  the following HTTP status codes and JSON results:
+
+  - if `rolls` is not set or has a valid input (positive integer > 0): status
+    code `200` and either a single number between 1 and 6 if `rolls` is not set
+    or 1, or an array of `n` numbers between 1 and 6 where `n` is the value of
+    `rolls`
+  - if `rolls` is set to an invalid input (not a number): status code `400` and
+    `{"status": "error", "message": "Parameter rolls must be a positive integer"}`
+  - if `rolls` is set to a `0` or a negative integer: status code `500` and no
+    JSON output.
+
+- There may be an optional attribute `player=name` to the `/rolldice` endpoint,
+  where
+
 - The application must output the following log lines using a language-specific
   common logging framework:
-  - an INFO-level message for each HTTP request with a status code <400
-  - a WARN-level message for each HTTP request with a status code between 400
-    and 400, including the message which will be sent in the JSON result
-  - a ERROR-level message for each HTTP request with a status code above 499
+  - an INFO-level message for each HTTP request with a status code `<400`
+  - a WARN-level message for each HTTP request with a status code between `400`
+    and `499`, including the message which will be sent in the JSON result
+  - a ERROR-level message for each HTTP request with a status code above `499`
   - a DEBUG-level message for each dice roll containing the rolled number
 - The code of the application must be split into two files:
-  - an "app" file that contains the handling of the HTTP requests
-  - a "lib" file that contains the implementation of the roll dice function.
+  - an `app` file that contains the handling of the HTTP requests
+  - a `lib` file that contains the implementation of the roll dice function.
+- Error handling for `rolls` should be split as follows:
+  - the `app` checks if `rolls` is defined, and if not, sets it to `1`.
+  - the `app` only checks if `rolls` is a number. If yes, it calls the function
+    for rolling the dice in `lib`. If no, it does the error handling with a
+    `400` error.
+  - the `lib` checks if `rolls` is a positive number. If no, it throws an
+    exception. `app` is catching the error and sending back the `500` error.
+- The `lib` should have an outer function that does the error handling as
+  described above. The outer function then does the following depending on the
+  value of `rolls`:
+  - `rolls == 1`: Run the inner function once and return the value.
+  - `rolls > 1`: Run a loop which calls the inner function `rolls` time and
+    return the results in an array.
+- The inner function of `lib` is creating the random number between 1 and 6 and
+  returns that value.
 
 ## Instrumentation requirements
 
 - If possible, the initialization for the OpenTelemetry SDK should be contained
   in a separate file and imported within the "app" file. Otherwise, it should be
   part of the "app" file.
+- Common resource detectors should be loaded in the initialization, e.g. for
+  `process`, `container`, `os`, ...
 - The "lib" file must depend only on the OpenTelemetry API.
 - The `service.*` attributes should be added via environment variables
   (`OTEL_SERVICE_NAME`, `OTEL_RESOURCE_ATTRIBUTES`).
-- The exporters for all signals should be set via environment variables. At a
-  minimum, a console output and OTLP export must be supported that way.
+- For exporting telemetry the application should use an exporter for
+  `stdout`/`console` and `otlp`.
 - There should be an option to enable diagnostic logging for the OpenTelemetry
   components, ideally via `OTEL_LOG_LEVEL`.
 - There should be a way to add the instrumentation library for the used HTTP
   library. This instrumentation library should use the stable semantic
-  conventions for HTTP.
+  conventions for HTTP. Prefer a library with a coverage for most signals,
+  ideally they have traces and metrics.
+- If no instrumentation library is available, only then the function in `app`
+  which handles the `/rolldice` endpoint is instrumented by the user by adding a
+  span and metrics.
 - There should be a log bridge for the used logging mechanism such that all logs
   are automatically collected and exported.
-
-## TODO
-
-- Add spans
-- Add metric measurements
-- Add resource detectors
-- Look into context propagation (+ baggage)
+- There should be a span created for the outer function in `lib`. The span
+  tracks the time of the function. It records the exception if thrown. It adds
+  attributes to the span like the value of `rolls`, `code.*`, ...
+- There should be a span created for the inner function in `lib`. The span
+  tracks the time of the function. It adds the random number it generated as an
+  attribue to the span.
+- In the `lib` file the following metrics should be created:
+  - a counter for the calls of the outer function
+  - a histogram for the distribution of the outcomes (1-6)
+  - a gauge for the last value of `rolls`
--- a/content/en/docs/getting-started/reference-application-specification.md
+++ b/content/en/docs/getting-started/reference-application-specification.md
@@ -1,0 +1,71 @@
+---
+title: Specification for the reference application
+linkTitle: Reference App Spec
+description:
+  This document contains the specification for the reference application that is
+  used across the Getting Started guide
+toc_hide: true
+---
+
+# General requirements
+
+- Implementations of the Reference App are owned by language SIGs implementing
+  the OpenTelemetry APIs & SDKs
+- There must be an uninstrumented and instrumented version of the application
+- There must be a language specific CI action, that verifies that the
+  application builds and runs in both versions
+- The application must be runnable from a commandline interface
+- There must be a Dockerfile to run the application within a containerized
+  environment
+- The application must run stand alone, i.e. it can not take hard dependencies
+  on any other content of the repository enclosing it
+
+## Service requirements
+
+- The application must listent at port 8080 for HTTP requests by default. The
+  port should be configurable via the environment variable APPLICATION_PORT.
+- For handling the HTTP requests a library should be used for which an
+  instrumentation library is available. The application must provide the
+  endpoint `/rolldice?rolls=<n>` via `GET` (and optionally `POST`) and return
+  the following HTTP status codes and JSON results: - if `rolls` is not set or
+  has a valid input (positive integer > 0): status code `200` and either a
+  single number between 1 and 6 if `rolls` is not set or 1, or an array of `n`
+  numbers between 1 and 6 where `n` is the value of `rolls` - if `rolls` is set
+  to an invalid input (not a number): status code `400` and
+  `{"status": "error", "message": "Parameter rolls must be a positive integer"}` -
+  if `rolls` is set to a negative integer: status code `500` and no JSON output.
+- The application must output the following log lines using a language specific
+  common logging framework:
+  - a INFO-level message for each HTTP request with a status code <400
+  - a WARN-level message for each HTTP request with a status code between 400
+    and 400, including the message which will be sent in the JSON result
+  - a ERROR-level message for each HTTP request with a status code above 499
+  - a DEBUG-level message for each dice roll containing the rolled number
+- The code of the application must be split into two files:
+  - an "app" file that contains the handling of the HTTP requests
+  - a "lib" file that contains the implementation of the roll dice function.
+
+## Instrumentation requirements
+
+- If possible the initialization for the OpenTelemetry SDK should be contained
+  in a separate file and imported within the "app" file. Otherwise it should be
+  part of the "app" file.
+- The "lib" file must only depend on the OpenTelemetry API.
+- The `service.*` attributes should be added via environment variables
+  (`OTEL_SERVICE_NAME`, `OTEL_RESOURCE_ATTRIBUTES`).
+- The exporters for all signals should be set via environment variables. At a
+  minimum a console output and OTLP export have to be supported that way.
+- There should be an option to enable diagnostic logging for the OpenTelemetry
+  components, ideally via `OTEL_LOG_LEVEL`.
+- There should be a way to add the instrumentation library for the used HTTP
+  library. This instrumentation library should use the stable semantic
+  conventions for HTTP.
+- There should be a log bridge for the used logging mechanism such that all logs
+  are automatically collected and exported.
+
+## TODO
+
+- Add spans
+- Add metric measurements
+- Add resource detectors
+- Look into context propagation (+ baggage)
--- a/content/en/docs/getting-started/reference-application-specification.md
+++ b/content/en/docs/getting-started/reference-application-specification.md
@@ -15,15 +15,21 @@ SDK exists.
 ## General requirements
 
 - Implementations of the Reference App are owned by language SIGs implementing
-  the OpenTelemetry APIs and SDKs.
+  the OpenTelemetry APIs and SDKs. This ensures that the application follows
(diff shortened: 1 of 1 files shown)


content/en/docs/getting-started/reference-application-specification.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Instrumentation requirements
+
+- If possible the initialization for the OpenTelemetry SDK should be contained
+  in a separate file and imported within the "app" file. Otherwise it should be
+  part of the "app" file.
+- The "lib" file must only depend on the OpenTelemetry API.
+- The `service.*` attributes should be added via environment variables
+  (`OTEL_SERVICE_NAME`, `OTEL_RESOURCE_ATTRIBUTES`).
+- The exporters for all signals should be set via environment variables. At a
```
cijothomas, line 56: does all language support such environment variables?

content/en/docs/getting-started/reference-application-specification.md
```diff
@@ -0,0 +1,71 @@
...
+  minimum a console output and OTLP export have to be supported that way.
+- There should be an option to enable diagnostic logging for the OpenTelemetry
+  components, ideally via `OTEL_LOG_LEVEL`.
+- There should be a way to add the instrumentation library for the used HTTP
+  library. This instrumentation library should use the stable semantic
```
cijothomas, line 61: call out that both metrics and spans should be covered by the instrumentation library?

content/en/docs/getting-started/reference-application-specification.md
```diff
@@ -0,0 +1,71 @@
...
+  application builds and runs in both versions
+- The application must be runnable from a commandline interface
+- There must be a Dockerfile to run the application within a containerized
+  environment
+- The application must run stand alone, i.e. it can not take hard dependencies
+  on any other content of the repository enclosing it
+
+## Service requirements
+
+- The application must listent at port 8080 for HTTP requests by default. The
```
kaylareopelle, line 25:
```suggestion
- The application must listen at port 8080 for HTTP requests by default. The
```

content/en/docs/getting-started/reference-application-specification.md
```diff
@@ -0,0 +1,71 @@
...
+- For handling the HTTP requests a library should be used for which an
+  instrumentation library is available. The application must provide the
+  endpoint `/rolldice?rolls=<n>` via `GET` (and optionally `POST`) and return
+  the following HTTP status codes and JSON results: - if `rolls` is not set or
+  has a valid input (positive integer > 0): status code `200` and either a
+  single number between 1 and 6 if `rolls` is not set or 1, or an array of `n`
+  numbers between 1 and 6 where `n` is the value of `rolls` - if `rolls` is set
+  to an invalid input (not a number): status code `400` and
+  `{"status": "error", "message": "Parameter rolls must be a positive integer"}` -
+  if `rolls` is set to a negative integer: status code `500` and no JSON output.
```
kaylareopelle, line 36: Some indentation around the `-if ` bullet points might be helpful here for readability.

kaylareopelle (commented):
Great start! Thank you!

content/en/docs/getting-started/reference-application-specification.md
svrnm: No they don't, yet I want to keep it that way for the following reasons:

* if we do not control exporters (and other configuration) via environment variables (and maybe later down with declarative configuration?) we add a layer of complexity that makes maintaining the code much harder, since instead of having 2 copies of the code (uinstrumentred, instrumented) we need 3 (uninstrumented, instrumented with console exporter, instrumented with otlp exporter). 
* I am hoping that we can trigger some development of those features for languages while we add more and more of them to the getting started. We will for sure start with only a few of them and then slowly add the other ones, and as needed discuss if we need to add something to the implementation to simplify the getting started flow.
* Last but not least, although I didn't write this in normative language, I picked "should" over "must" since eventually we need to be open for exceptions.

content/en/docs/getting-started/reference-application-specification.md
svrnm: that's a great addition, thanks!

content/en/docs/getting-started/reference-application-specification.md
cijothomas: I am totally okay with using ENV variables for configuration - but exporter selection is not like any other variables - the package has to bundle all exporters with it, so that it can support exporter selection via env variable. This is quite different from using ENV variable for setting an endpoint etc.

> we need 3 (uninstrumented, instrumented with console exporter, instrumented with otlp exporter).

That is a reality, and in my observation, changing from stdout to otlp exporter is <5 lines of code + adding the right package!.

content/en/docs/getting-started/reference-application-specification.md
```diff
@@ -0,0 +1,71 @@
...
+- The application must output the following log lines using a language specific
+  common logging framework:
+  - a INFO-level message for each HTTP request with a status code <400
+  - a WARN-level message for each HTTP request with a status code between 400
+    and 400, including the message which will be sent in the JSON result
+  - a ERROR-level message for each HTTP request with a status code above 499
+  - a DEBUG-level message for each dice roll containing the rolled number
+- The code of the application must be split into two files:
+  - an "app" file that contains the handling of the HTTP requests
+  - a "lib" file that contains the implementation of the roll dice function.
```
kaylareopelle, line 46: I don't think this will be the best approach for Ruby. The current rolldice app is written in Rails, and Rails has some very specific directory naming guidelines. A lib directory exists, but putting the roll dice function there would be a little odd.

Do you think the location of HTTP requests/roll dice function can be idiomatic for the language?

content/en/docs/getting-started/reference-application-specification.md
svrnm: I will drop this requirement for now, we have to see how far we get in practical terms and then later can reconsider

content/en/docs/getting-started/reference-application-specification.md
svrnm: The goal of splitting out the HTTP request piece and the "business logic" (roll the dice) is for demonstrating that a library does only need to take dependency on the API. So the goal is to split them, how those files are called and were they live does not matter. Would that work with ruby somehow?

kaylareopelle: Yes, that works!

content/en/docs/getting-started/reference-application-specification.md
```diff
@@ -0,0 +1,71 @@
...
+- Implementations of the Reference App are owned by language SIGs implementing
+  the OpenTelemetry APIs & SDKs
+- There must be an uninstrumented and instrumented version of the application
+- There must be a language specific CI action, that verifies that the
+  application builds and runs in both versions
+- The application must be runnable from a commandline interface
+- There must be a Dockerfile to run the application within a containerized
+  environment
+- The application must run stand alone, i.e. it can not take hard dependencies
+  on any other content of the repository enclosing it
```
tiffany76, line 21:
```suggestion
- Implementations of the Reference App are owned by language SIGs implementing
  the OpenTelemetry APIs and SDKs.
- There must be an uninstrumented and instrumented version of the application.
- There must be a language-specific CI action that verifies the
  application builds and runs in both versions.
- The application must be runnable from a command line interface.
- There must be a Dockerfile to run the application within a containerized
  environment.
- The application must run standing alone. In other words, it cannot take hard dependencies
  on any other content of the repository enclosing it.
```

content/en/docs/getting-started/reference-application-specification.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Service requirements
+
+- The application must listen at port 8080 for HTTP requests by default. The
+  port should be configurable via the environment variable APPLICATION_PORT.
+- For handling the HTTP requests a library should be used for which an
```
tiffany76, line 27:
```suggestion
- For handling the HTTP requests, a library should be used for which an
```

content/en/docs/getting-started/reference-application-specification.md
```diff
@@ -0,0 +1,71 @@
...
+  instrumentation library is available. The application must provide the
+  endpoint `/rolldice?rolls=<n>` via `GET` (and optionally `POST`) and return
+  the following HTTP status codes and JSON results: - if `rolls` is not set or
+  has a valid input (positive integer > 0): status code `200` and either a
+  single number between 1 and 6 if `rolls` is not set or 1, or an array of `n`
+  numbers between 1 and 6 where `n` is the value of `rolls` - if `rolls` is set
+  to an invalid input (not a number): status code `400` and
+  `{"status": "error", "message": "Parameter rolls must be a positive integer"}` -
+  if `rolls` is set to a negative integer: status code `500` and no JSON output.
+- The application must output the following log lines using a language specific
```
tiffany76, line 37:
```suggestion
- The application must output the following log lines using a language-specific
```

content/en/docs/getting-started/reference-application-specification.md
```diff
@@ -0,0 +1,71 @@
...
+  common logging framework:
+  - a INFO-level message for each HTTP request with a status code <400
```
tiffany76, line 39:
```suggestion
  - an INFO-level message for each HTTP request with a status code <400
```

content/en/docs/getting-started/reference-application-specification.md
```diff
@@ -0,0 +1,71 @@
...
+  - a WARN-level message for each HTTP request with a status code between 400
+    and 400, including the message which will be sent in the JSON result
```
tiffany76, line 41: Should this say "between 400 and 499"?

content/en/docs/getting-started/reference-application-specification.md
```diff
@@ -0,0 +1,71 @@
...
+  - a DEBUG-level message for each dice roll containing the rolled number
+- The code of the application must be split into two files:
+  - an "app" file that contains the handling of the HTTP requests
+  - a "lib" file that contains the implementation of the roll dice function.
+
+## Instrumentation requirements
+
+- If possible the initialization for the OpenTelemetry SDK should be contained
+  in a separate file and imported within the "app" file. Otherwise it should be
+  part of the "app" file.
```
tiffany76, line 52:
```suggestion
- If possible, the initialization for the OpenTelemetry SDK should be contained
  in a separate file and imported within the "app" file. Otherwise, it should be
  part of the "app" file.
```

content/en/docs/getting-started/reference-application-specification.md
```diff
@@ -0,0 +1,71 @@
...
+- The "lib" file must only depend on the OpenTelemetry API.
```
tiffany76, line 53:
```suggestion
- The "lib" file must depend only on the OpenTelemetry API.
```

content/en/docs/getting-started/reference-application-specification.md
```diff
@@ -0,0 +1,71 @@
...
+- The `service.*` attributes should be added via environment variables
+  (`OTEL_SERVICE_NAME`, `OTEL_RESOURCE_ATTRIBUTES`).
+- The exporters for all signals should be set via environment variables. At a
+  minimum a console output and OTLP export have to be supported that way.
```
tiffany76, line 57:
```suggestion
  minimum, a console output and OTLP export must be supported that way.
```

tiffany76 (commented):
Took a quick copyedit first pass. Thanks!

svrnm:
Apologies for the delay! I took some time today to incorporate your reviews and expand on the details, please take a look!

content/en/docs/getting-started/reference-application-specification.md
```diff
@@ -0,0 +1,107 @@
...
+    code `200` and either a single number between 1 and 6 if `rolls` is not set
+    or 1, or an array of `n` numbers between 1 and 6 where `n` is the value of
+    `rolls`
+  - if `rolls` is set to an invalid input (not a number): status code `400` and
+    `{"status": "error", "message": "Parameter rolls must be a positive integer"}`
+  - if `rolls` is set to a `0` or a negative integer: status code `500` and no
+    JSON output.
+
+- There may be an optional attribute `player=name` to the `/rolldice` endpoint,
+  where
```
tiffany76: Did we lose the end of this sentence somewhere?

svrnm: ah, yes, I was writing that and forgot about it... lol, let me fix that.

svrnm: fixed

content/en/docs/getting-started/reference-application-specification.md
```diff
@@ -0,0 +1,107 @@
...
+  - a WARN-level message for each HTTP request with a status code between `400`
+    and `499`, including the message which will be sent in the JSON result
+  - a ERROR-level message for each HTTP request with a status code above `499`
+  - if the optional `player` attribute is set, a DEBUG-level message that outputs
+    the value of `player` and the rolled number
+  - if the optional `player` attribute is not set, a DEBUG-level message that outputs
+    the static value `anonymous player` and the rolled number.
+- The code of the application must be split into two files:
+  - an `app` file that contains the handling of the HTTP requests
+  - a `lib` file that contains the implementation of the roll dice function.
```
martincostello: Is this a hard requirement (`must` implies to me it is)? I get the argument for standardisation, but this is likely not idiomatic for a number of programming languages/frameworks (it's not for .NET/C#).

svrnm: So the thinking about this, is that we can demonstrate much better that if you have a library you only leverage the API and not the SDK. How is it not idiomatic for C# to put a functionality like that into a separate file? I would assume it works similar to java that you would create a Class that provides that functionality as service?

martincostello: By not idiomatic I mean files called `app` and `lib`. This reads as if it's mandating the file names.

If I were to write a simple idiomatic .NET application today to do a simple `rolldice` endpoint, I could do it like this in a single `Program.cs` file:

```csharp
using System.Globalization;

var builder = WebApplication.CreateBuilder(args);

// Configure telemetry here, then...

var app = builder.Build();

app.MapGet("/rolldice/{player?}", (string? player, ILogger<Program> logger) =>
{
    var result = Random.Shared.Next(1, 7);

    if (string.IsNullOrEmpty(player))
    {
        logger.LogInformation("Anonymous player is rolling the dice: {result}", result);
    }
    else
    {
        logger.LogInformation("{player} is rolling the dice: {result}", player, result);
    }

    return result.ToString(CultureInfo.InvariantCulture);
});

app.Run();
```

This is essentially how we implemented `rolldice` in the .NET example in our [grafana/docker-otel-lgtm](https://github.com/grafana/docker-otel-lgtm/tree/main/examples/dotnet) repository.

If you still wanted the additional structure, then to be mostly-idiomatic you'd have something like this in .NET 6+:

```
- <App Directory>
|
|--> Program.cs - The entry-point to the application, needs to call and setup `App`.
|--> App.cs - The `App` class that sets up the webservice and defines the endpoints.
|--> Lib.cs - A weirdly-named (IMHO) class that provides the logic of the endpoint to roll the dice.
|--> Project.csproj - The project definition file.
```

With .NET 10, I could remove the need for the project file ([_Announcing dotnet run app.cs – A simpler way to start with C# and .NET 10_](https://devblogs.microsoft.com/dotnet/announcing-dotnet-run-app/)) and at least change the name to `app.cs`, but I would still find `lib.cs` strange to find in a .NET project.

As the document is written right now, it seems overly-prescriptive to me. Rather, it would just talk about separating the HTTP handler code from the "business logic" as appropriate for the programming language/framework in question, rather than directing specific file names.

svrnm: It does not have to be "App.cs" or "Lib.cs" it can be something more meaningful in the context of your language. 

The key point is that we separate the "library" that provides the dice rolling functionality from the "application" that consumes it. The business logic is super simple, that makes it awkward in most/all languages ...  but it's here to demonstrate some concepts (separation of API & SDK, etc.)

So, if it makes more sense to call it "DiceRollService.cs" or "RollerMiddleware.cs" or whatever makes sense, and if "App.cs" is not necessary and it's functionality can be offloaded to "Program.cs" that's totally fine

martincostello: Cool, that basically fits with what I was getting at. I think the text as-written sounds like it's suggesting the files must be called that, rather than explaining what's wanted at a higher-level architectural/layout level.

svrnm: let me rewrite that, I also think I need to put more "why are we doing X" at certain points of the document

content/en/docs/getting-started/reference-application-specification.md
```diff
@@ -0,0 +1,107 @@
...
+---
+
+The purpose of the reference application is to have a standardized sample
+application that can be implemented in all languages for which an OpenTelemetry
+SDK exists.
+
+## General requirements
+
+- Implementations of the Reference App are owned by language SIGs implementing
+  the OpenTelemetry APIs and SDKs.
```
LisaHJung: Listed in order of popularity, the most commonly used programming languages among OTel users are:

1. JavaScript
2. Java
3. Go
4. Python

Can we prioritize standardizing apps for these languages and consult their respective SIGs?

cijothomas: nit:
Curious, where ca we get data about Otel language popularity etc.?

Going with just [Github stars](https://github.com/orgs/open-telemetry/repositories?q=sort%3Astars), Go, DotNet, JS, Java/Rust is the order.

content/en/docs/getting-started/reference-application-specification.md
```diff
@@ -0,0 +1,107 @@
...
+- Common resource detectors should be loaded in the initialization, e.g. for
+  `process`, `container`, `os`, ...
+- The "lib" file must depend only on the OpenTelemetry API.
+- The `service.*` attributes should be added via environment variables
+  (`OTEL_SERVICE_NAME`, `OTEL_RESOURCE_ATTRIBUTES`).
+- Other `resource detectors` should be added to the initialization of the SDK.
+- For exporting telemetry the application should use an exporter for
+  `stdout`/`console` and `otlp`.
+- There should be an option to enable diagnostic logging for the OpenTelemetry
+  components, ideally via `OTEL_LOG_LEVEL`.
```
cijothomas: nit: not sure if `OTEL_LOG_LEVEL` is something that all languages support

content/en/docs/getting-started/reference-application-specification.md
LisaHJung: Hi @cijothomas ! The OTel End User SIG conducted a [survey](https://opentelemetry.io/blog/2024/otel-get-started-survey/) on the language popularity in addition to other helpful insights about the getting-started experience of the OTel users.

content/en/docs/getting-started/reference-application-specification.md
svrnm: that's why I wrote "ideally" but I can add a "if supported"

content/en/docs/getting-started/reference-application-specification.md
svrnm: We can prioritize them and make it a goal to have them as early as possible. 

But eventually it comes down to the individuals who implement it :-)

content/en/docs/getting-started/reference-application-specification.md
svrnm: fixed!

content/en/docs/getting-started/reference-application-specification.md
```diff
@@ -0,0 +1,119 @@
...
+    outputs the static value `anonymous player` and the rolled number. The log
+    lines will be used to add a log bridge during instrumentation to demonstrate
+    how OpenTelemetry can connect to existing logging frameworks.
+- The code of the application must be split into two files:
+  - an `app` file that contains the handling of the HTTP requests
+  - a `library` file that contains the implementation of the roll dice function.
+    The names of those files should be idiomatic within the implementing
+    language, like `app.js` and `roll-the-dice.js`. The important point is that
+    by separating the two, the code demonstrates that the `library` only takes
+    dependency on the API and all the SDK code is initalized in the `app` code.
```
svrnm, line 69: @martincostello @cijothomas is this better, I picked a `.js` example but if you provide me with what a good naming for .NET is, I can update that accordingly.

content/en/docs/getting-started/reference-application-specification.md
```diff
@@ -0,0 +1,119 @@
...
+- Common resource detectors should be loaded in the initialization, e.g. for
+  `process`, `container`, `os`, ...
+- The "lib" file must depend only on the OpenTelemetry API.
+- The `service.*` attributes should be added via environment variables
+  (`OTEL_SERVICE_NAME`, `OTEL_RESOURCE_ATTRIBUTES`).
+- Other `resource detectors` should be added to the initialization of the SDK.
+- For exporting telemetry the application should use an exporter for
+  `stdout`/`console` and `otlp`.
+- There should be an option to enable diagnostic logging for the OpenTelemetry
+  components, if implemented ideally via `OTEL_LOG_LEVEL`.
```
svrnm, line 100: @cijothomas PTAL!

svrnm:
I am going to merge this PR now, as it had a lot of time to build up, it will definitely be iterated on, also the page will be hidden for now

content/en/docs/getting-started/reference-application-specification.md
```diff
@@ -0,0 +1,119 @@
+---
+title: Specification for the reference application
+linkTitle: Reference App Spec
+description:
+  This document contains the specification for the reference application that is
+  used across the Getting Started guide
+toc_hide: true
```
chalin, line 7: @svrnm - why hide this page from the sidenav? Is this page linked to anywhere else?

svrnm: this is very much in progress and serves no value to end users (yet), I didn't have a better place to store it for now, but I also do not want to confuse people with it.

chalin: Makes sense.

PR #6835, opened 2025-05-07
Add the OTTL Playground to the registry

edmocosta:
Adding the OTTL Playground (https://ottl.run) to the registry.


tiffany76 (approved):
Thanks, @edmocosta

PR #6829, opened 2025-05-06
[pt] Fix shortcode path references on Portuguese pages

vitorvasc:
## Description

Tracked on #6812 

1. Updates and fixes the shortcode references by removing the `pt/` prefix for the following pages:
  - content/pt/docs/languages/go/exporters.md
  - content/pt/docs/concepts/components.md
  - content/pt/docs/concepts/resources/index.md
2. Updates the drifted content for the snippets:
  - content/pt/docs/_includes/exporters/intro.md
  - content/pt/docs/_includes/resources-intro.md

---

### Preview

- https://deploy-preview-6829--opentelemetry.netlify.app/pt/docs/concepts/resources/
- https://deploy-preview-6829--opentelemetry.netlify.app/pt/docs/languages/erlang/resources/
- https://deploy-preview-6829--opentelemetry.netlify.app/pt/docs/languages/go/exporters/ -- @chalin: this PR actually fixes the start of this page. The first page heading was missing and the following text wrong.

PR #6581, opened 2025-03-20
[i18n] Convert docs/exporter shortcodes into includes

chalin:
- Contributes to #6460
- Reworks all `docs/languages/exporters` shortcodes except for the intro that I'll handle separately because it is much more involved
- Replaced the HTML-encoded callouts by calls to the alert shortcode

### Previews

- https://deploy-preview-6581--opentelemetry.netlify.app/docs/languages/cpp/exporters/, specifically
- https://deploy-preview-6581--opentelemetry.netlify.app/docs/languages/cpp/exporters/#jaeger
- https://deploy-preview-6581--opentelemetry.netlify.app/docs/languages/cpp/exporters/#prometheus
- https://deploy-preview-6581--opentelemetry.netlify.app/docs/languages/cpp/exporters/#zipkin
- https://deploy-preview-6581--opentelemetry.netlify.app/docs/languages/cpp/exporters/#custom-exporters

PR #6836, opened 2025-05-07
[pt] Drop obsolete pt shortcodes

chalin:
- Followup to #6829
- Contributes to #6460
- Drops obsolete `pt` shortcodes, now that content for these shortcodes is found in include files

There is no change in generated site files, other than the build timestamp:

```console
$ (cd public && g diff) | grep ^diff       
```


vitorvasc:
Thanks, it was one of my TODOs to tackle this and delete the deprecated files 😛 

But I was checking on the preview and found that the content is back in English: https://deploy-preview-6836--opentelemetry.netlify.app/pt/docs/languages/python/exporters/#jaeger-backend-setup

I didn't copy these files to the new directory on the previous PR, could that be the reason?

tiffany76 (approved):

chalin:
> I didn't copy these files to the new directory on the previous PR, could that be the reason?

Ah, yes. I didn't check the generated pages, only that there was no diff in the site output. If you prefer, you can submit a new PR that supersedes (and closes) this one @vitorvasc.

chalin:
Actually, given that this PR introduces no changes relative to `main`, I'm going to merge now and let you (@vitorvasc) submit a followup PR to add the Portuguese from those shortcodes as include files. The problem actually dates back to when I refactored via #6581. :-/

PR #6894, opened 2025-05-15
Bring back pr actions

svrnm:
Reverts #6663 with an improved workflow to run fix commands via GitHub comment:

- add hardened runners by StepSecurity
- split the creation of the changes & the application of them into two separate steps to disable the possibility to inject something malicious into the workflow run

@open-telemetry/docs-approvers @trask PTAL!

@theletterf I remember you had some code that would also check for the handle if the person is allowed to trigger the workflow (e.g. only org members, maintainers, etc.), we could add this as well.

NOTE: As always this github action has not been yet tested in full details, so unfortunately we will need to merge it at some point and then iterate on it until it works as expected.

### DIFF
--- a/.github/workflows/pr-actions.yml
+++ b/.github/workflows/pr-actions.yml
@@ -1,0 +1,160 @@
+name: PR actions
+
+on:
+  issue_comment:
+    types: [created]
+
+permissions: read-all
+
+env:
+  COMMENT: ${{ github.event.comment.body }}
+  PR_NUM: ${{ github.event.issue.number }}
+  USER_EMAIL: 107717825+opentelemetrybot@users.noreply.github.com
+  USER_NAME: opentelemetrybot
+  MAX_PATCH_SIZE_KB: 1024
+
+jobs:
+  generate-patch:
+    name: Run fixer and generate patch (untrusted)
+    runs-on: ubuntu-latest
+    if: |
+      github.event.issue.pull_request &&
+      startsWith(github.event.comment.body, '/fix:')
+    permissions:
+      contents: read
+      pull-requests: write
+
+    outputs:
+      action_name: ${{ steps.extract.outputs.action_name }}
+      patch_name: pr-fix-${{ github.run_id }}
+      patch_skipped: ${{ steps.check_patch.outputs.skipped }}
+
+    steps:
+      - uses: step-security/harden-runner@0634a2670c59f64b4a01f0f96f84700a4088b9f0 # v2.12.0
+        with:
+          egress-policy: audit
+
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          ref: refs/pull/${{ github.event.issue.number }}/head
+          fetch-depth: 999
+
+      - name: Extract action name
+        id: extract
+        run: |
+          PR_ACTION=$(echo "$COMMENT" | grep -oP '/fix:\K[:-_0-9a-z]+')
+          echo "action_name=$PR_ACTION" >> "$GITHUB_OUTPUT"
+
+      - name: Setup Node
+        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
+        with:
+          node-version-file: .nvmrc
+          cache: npm
+
+      - name: Install deps & run fixer
+        run: |
+          npm ci
+          case ${{ steps.extract.outputs.action_name }} in
+            all|refcache*|text)
+              npm install --omit=optional
+              ;&
+            *)
+              npm run fix:${{ steps.extract.outputs.action_name }}
+              ;;
+          esac
+
+      - name: Generate and validate patch
+        id: check_patch
+        run: |
+          git diff > pr-fix.patch
+
+          if [ ! -s pr-fix.patch ]; then
+            echo "No changes detected. Skipping patch."
+            echo "skipped=true" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+
+          actual_size_kb=$(du -k pr-fix.patch | cut -f1)
+          if (( actual_size_kb > MAX_PATCH_SIZE_KB )); then
+            echo "Patch too large: ${actual_size_kb} KB (limit: ${MAX_PATCH_SIZE_KB} KB)"
+            exit 1
+          fi
+
+          echo "skipped=false" >> "$GITHUB_OUTPUT"
+
+      - name: Upload patch artifact
+        if: steps.check_patch.outputs.skipped != 'true'
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
+        with:
+          name: pr-fix-${{ github.run_id }}
+          path: pr-fix.patch
+          retention-days: 1
+
+  apply-patch:
+    name: Apply and push patch (trusted)
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-actions.yml
```diff
@@ -0,0 +1,160 @@
...
+jobs:
+  generate-patch:
+    name: Run fixer and generate patch (untrusted)
+    runs-on: ubuntu-latest
+    if: |
+      github.event.issue.pull_request &&
+      startsWith(github.event.comment.body, '/fix:')
+    permissions:
+      contents: read
+      pull-requests: write
```
trask, line 25: this is the only nit I can find, which could give someone the ability to add labels/milestones/projects/etc to their (own) PR, but I think that's probably an ok compromise

the GitHub App requested in https://github.com/open-telemetry/community/issues/2746 does something similar, but doesn't require this permission since it uploads the diff to their service

but it is an external GitHub App, which brings its own security considerations

theletterf: I'd love to see a label spamming attack :D

trask: the only concern I think is if a repo has some label-driven automation or bypasses

theletterf: True that. I don't think that's our case though, right @chalin ?

svrnm: I wonder if the permission is even necessary in this step (sorry, I rendered some of it with AI + copy&paste from the existing one). We can drop it and run without first

chalin (approved):
Great! Thanks so much for this @svrnm!

I don't have the time today to review this in detail (I'll be mostly AFK), so I'll let @trask et all work through the details. I'm glad to RSLGTM, get this merged and give it a spin! (I'm also looking forward to study this more closely later.)

.github/workflows/pr-actions.yml
```diff
@@ -0,0 +1,160 @@
+name: PR actions
+
+on:
+  issue_comment:
+    types: [created]
+
+permissions: read-all
```
trask, line 7:
```suggestion
permissions:
  contents: read
```

.github/workflows/pr-actions.yml
trask, line 25: I agree with @svrnm's https://github.com/open-telemetry/opentelemetry.io/pull/6894/files#r2091365013 that this isn't needed (e.g. PRs can upload artifacts, and PRs don't have any trusted permissions)

```suggestion
```

trask (approved):
approving with the two permission updates above

trask:
@svrnm now that you have a safe way to do this, what do you think of running this (or at least the most common ones?) on every PR build and automatically applying it? (not waiting for someone to add the comment)

theletterf:
@trask The [rerun action](https://github.com/open-telemetry/opentelemetry.io/pull/6578) one would likely come after this one!

svrnm:
> > @svrnm now that you have a safe way to do this, what do you think of running this (or at least the most common ones?) on every PR build and automatically applying it? (not waiting for someone to add the comment)
> 
> just saw your [slack comment](https://cloud-native.slack.com/archives/C07BPU981PV/p1747319165079489?thread_ts=1747317974.476759&cid=C07BPU981PV), which makes sense why you don't want to do this:
> 
> > [applying] updates without any user intervention, which may be OK in a repo that is 90% contributed by the same folks, but we have way to many drive by contributors which will struggle with pushing and pulling and sometimes even don't know have to do git properly

Yeah, and thinking about it, there is actually another reason why we don't want to do that: some people update their PRs with lots of small commits, so I am not sure how this would behave, i.e. would we have a flooded action pipeline?

chalin:
> @svrnm now that you have a safe way to do this, what do you think of running this (or at least the most common ones?) on every PR build and automatically applying it? (not waiting for someone to add the comment)

If we're aiming to streamline the PR process for 80% the cases, I'd be willing to give this a try. Often, drive-by PRs need either `fix:format` or a refcache update. If we find that it doesn't make it easier for contributors and us maintainers, we can always pull it back.

Anyhow, let's get this merged first and see if the old `/fix:...` commands are working again. WDYT?

.github/workflows/pr-actions.yml
```diff
@@ -0,0 +1,158 @@
...
+
+      - name: Apply patch and push
+        run: |
+          git apply --check pr-fix.patch && git apply pr-fix.patch || {
+            echo "Patch failed to apply"
+            exit 1
+          }
+
+          git config user.name "$USER_NAME"
+          git config user.email "$USER_EMAIL"
```
trask, line 123: do you want to use opentelemetrybot here?

svrnm: yes, that's why in line 132 the token is set to the otel bot token, should we not do that and use the generic github token instead?

.github/workflows/pr-actions.yml
```diff
@@ -0,0 +1,158 @@
...
+      pull-requests: write
+
+    steps:
+      - uses: step-security/harden-runner@0634a2670c59f64b4a01f0f96f84700a4088b9f0 # v2.12.0
+        with:
+          egress-policy: audit
+
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          ref: refs/pull/${{ github.event.issue.number }}/head
```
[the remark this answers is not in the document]
svrnm: good find! Previously we had a separate step checking out the branch, maybe we need to bring that back

.github/workflows/pr-actions.yml
trask: oh sorry I see that `USER_NAME` _is_ `opentelemetrybot`, ignore me!

trask:
heads up, CodeQL is complaining security flaws in my PR where I'm testing out this approach: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/13872

I'm working through them now to understand and hopefully address...

svrnm:
I tested the command in my own repo, it worked for `/fix:format`, it tripped with `/fix:all` but that might be related to the drift detection, I also had some issues with npm dependencies&caching, might be related to some things missing in my repo what we have here, so we can revisit that.

chalin:
@trask - anything left to do before we can merge and test?

.github/workflows/pr-actions.yml
```diff
@@ -0,0 +1,166 @@
...
+    steps:
+      - uses: step-security/harden-runner@0634a2670c59f64b4a01f0f96f84700a4088b9f0 # v2.12.0
+        with:
+          egress-policy: audit
+
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          fetch-depth: 999
+
+      - run: gh pr checkout $PR_NUM -b "pr-action-${RANDOM}"
```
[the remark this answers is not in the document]
svrnm: If I remember correctly we introduced this back in the day because we had some issues with branch names that were used on the fork and upstream (think: `patch-1` or similar), I could dig up the issue for that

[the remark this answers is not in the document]
trask: this simplified version is working in the (new) Java instrumentation auto apply action: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/.github/workflows/auto-update-pull-request.yml

chalin:
Merging. @svrnm et al. will follow through with the pending comments via another PR. Thanks all!

PR #6896, opened 2025-05-15
[CI] Adjust fix:all and test-and-fix to omit fix:i18n

chalin:
- Followup on our Comms meeting decision from 2025/05/14
- Drops `fix:i18n` from `fix:all` and `test-and-fix`
- As a compromise, power uers/contributors can run the internal `fix` script to really run all individual fixes -- in our contrib docs we only refer to `fix:all`, not `fix`.


theletterf (approved):
Thank you!

PR #6900, opened 2025-05-15
Update nvm install command in development.md doc

isaacaflores2:
Rhe first command under the development local setup `nvm install` does not correctly install the latest version. Changing to `nvm install --lts`. 

I validated the change after running ` npm run serve`.

image

### DIFF
--- a/content/en/docs/contributing/development.md
+++ b/content/en/docs/contributing/development.md
@@ -60,7 +60,7 @@ Your development environment will be initialized via the
     specified in the .nvmrc file:
 
     ```sh
-    nvm install
+    nvm install --lts
     ```
 
     To [install under Windows][nodejs-win], use [nvm-windows]:


chalin (changes requested):
Hi. We rely on https://github.com/open-telemetry/opentelemetry.io/blob/main/.nvmrc instead, so `--lts` isn't required.

isaacaflores2:
Hi @chalin , okay I think my issue was I ran `nvm install` outside of the repository.  So I missed the second step `Go to the repository directory.`

theletterf:
@isaacaflores2 Perhaps you could reword the instructions slightly to clarify that:

- You need to run nvm install from the repo's root folder.
- We could turn the Windows command indication into an admonition or an inline code snippet. It's a bit distracting right there and I know some people skipped right to that step.

isaacaflores2:
@chalin @theletterf I added an explicit `cd` command to the 2nd step similar to the instructions [here](https://opentelemetry.io/docs/contributing/pull-requests/#clone-and-set-upstream). Please let me know what you think. 


I attempted to add the Windows specific instructions within an admonition/alert-md but that broke the links.

content/en/docs/contributing/development.md
```diff
@@ -54,6 +54,11 @@ Your development environment will be initialized via the
 1.  [Fork] and then [clone] the website repository at
     <{{% param github_repo %}}>.
 2.  Go to the repository directory.
```
chalin, line 56:
```suggestion
2.  Go to the repository directory:
```

chalin (commented):
Let's avoid a Windows alert. See inline for one nit. Otherwise, I'm ok with this change.

chalin (approved):
Works for me if it works for you @theletterf 🙌🏻

PR #6925, opened 2025-05-19
[k8s operator] Provide link to cert-manager installation

johannaojeling:
Updates the OpenTelemetry Operator [Getting started](https://opentelemetry.io/docs/platforms/kubernetes/operator/#getting-started) guide with a link the to `cert-manager` installation docs, so that users more easily know what software they need to install.

The same can be found in the open-telemetry/opentelemetry-operator [README](https://github.com/open-telemetry/opentelemetry-operator/blob/2bae42922d067e8978ec534e959a808a1a87acbd/README.md?plain=1#L24).

### DIFF
--- a/content/en/docs/platforms/kubernetes/operator/_index.md
+++ b/content/en/docs/platforms/kubernetes/operator/_index.md
@@ -30,7 +30,7 @@ The operator manages:
 ## Getting started
 
 To install the operator in an existing cluster, make sure you have
-[`cert-manager` installed](https://cert-manager.io/docs/installation/) and run:
+[`cert-manager`](https://cert-manager.io/docs/installation/) installed and run:
 
 ```bash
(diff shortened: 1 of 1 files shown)


content/en/docs/platforms/kubernetes/operator/_index.md
```diff
@@ -29,8 +29,8 @@ The operator manages:
 
 ## Getting started
 
-To install the operator in an existing cluster, make sure you have cert-manager
-installed and run:
+To install the operator in an existing cluster, make sure you have
+[`cert-manager` installed](https://cert-manager.io/docs/installation/) and run:
```
tiffany76, line 33: With the monospace formatting on cert-manager, the link has a weird break in it in the rendered version. So I think we can leave "installed" out of the link text in this case.

```suggestion
[`cert-manager`](https://cert-manager.io/docs/installation/) installed and run:
```

tiffany76 (approved):
One tiny nit, but it LGTM!

PR #6958, opened 2025-05-22
Add fix:refcache to semconv update automation

trask:


.github/workflows/update-semconv-integration-branch.yml
```diff
@@ -86,6 +86,8 @@ jobs:
 
           sed -i "s/^\tsemconv-pin = .*/\tsemconv-pin = $commit_desc/" .gitmodules
 
+          npm run fix:refcache
+
```
trask, line 90: I didn't test this, is it (at least in theory) correct?

chalin (commented):
You'll need to run something like `npm install --omit=optional` first before you can run the fix script, probably something like this:

```yaml
      - uses: actions/setup-node@v4
        with:
          node-version-file: .nvmrc
          cache: npm

      - run: npm install --omit=optional
```

For a (more involved) example, see https://github.com/open-telemetry/opentelemetry.io/blob/main/.github/workflows/check-links.yml.

This will slow down the execution, but that isn't an issue for a cron job IMHO.

chalin (approved):
LGTM. Let's give this a spin!

PR #6959, opened 2025-05-22
Fix semconv automation

trask:
I think this is the issue here: https://github.com/open-telemetry/opentelemetry.io/actions/runs/15190161284/job/42720612269

### DIFF
--- a/.github/workflows/update-semconv-integration-branch.yml
+++ b/.github/workflows/update-semconv-integration-branch.yml
@@ -90,6 +90,7 @@ jobs:
         with:
           node-version-file: .nvmrc
           cache: npm
+          cache-dependency-path: tmp/package-ci.json
 
       - name: Fix refcache
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/update-semconv-integration-branch.yml
```diff
@@ -90,6 +90,7 @@ jobs:
         with:
           node-version-file: .nvmrc
           cache: npm
+          cache-dependency-path: tmp/package-ci.json
```
chalin, line 93: We don't create a temporary package spec so let's use the main one:

```suggestion
          cache-dependency-path: package.json
```

chalin (approved):

.github/workflows/update-semconv-integration-branch.yml
chalin: Or we could just turn off caching?

chalin (approved):

PR #6960, opened 2025-05-22
Fix semconv automation issue

trask:
the submodule wasn't getting updated anymore, so rolling back and just tacking fix:refcache on to the end of the (previously working) workflow


chalin (approved):
Oh right, the submodule update needs to happen first, ... and be committed! 🙃 🙌🏻

PR #6969, opened 2025-05-25
Update vendors.yaml - Added base14 Scout

rnjn:
base14 Scout is an observability and reliability system that uses OpenTelemetry agents for collecting observability signals.


theletterf (approved):
Thanks!

pr-fix.patch
chalin, line 1: @theletterf - we shouldn't have committed the patch file here.

Issue #6880, opened 2025-05-13
[CI] Upgrade markdownlint packages and fix related build issues


chalin:
... and address failures when `check:markdown` is run.

chalin:
Fixed by:

- #6984

PR #6053, opened 2025-01-23
[chore] Accessible links 5

jsoref:
The web isn't limited to people who read content with their eyes. Some users use screen readers and other accessibility agents and some people have other reasons for appreciating descriptive links. To that end, [check-spelling/spell-check-this](https://github.com/check-spelling/spell-check-this) includes a rule:

#### [Do not use `(click) here` links](https://github.com/check-spelling/spell-check-this/blob/4a57c3641b1c1603e70804208eb1bc644294bdf6/.github/actions/spelling/line_forbidden.patterns#L57-L63)

For more information, see:
* https://www.w3.org/QA/Tips/noClickHere
* https://webaim.org/techniques/hypertext/link_text
* https://granicus.com/blog/why-click-here-links-are-bad/
* https://heyoka.medium.com/dont-use-click-here-f32f445d1021
```
(?i)(?:>|\[)(?:(?:click |)here|link|(?:read |)more)(?:</|\]\()
```

The changes included here are designed to help users who benefit from descriptive links

PR #6984, opened 2025-05-27
Upgrade markdownlint-cli2 and address issues

chalin:
- Contributes to #6880
- Updates `markdownlint-cli2` to `@latest`
- Addresses issues raised by linter in Glossary, which detects and reports links with link-text of the form "more" and "link", much like what was done by @jsoref in #6053 etc.
- **Preview**:

### DIFF
--- a/content/en/docs/concepts/glossary.md
+++ b/content/en/docs/concepts/glossary.md
@@ -370,25 +370,25 @@ on web pages when requested. See [zPages].
 [baggage]: /docs/specs/otel/baggage/api/
 [context propagation]: /docs/specs/otel/overview#context-propagation
 [dag]: https://en.wikipedia.org/wiki/Directed_acyclic_graph
-[distributed tracing]: /docs/concepts/signals/traces/
-[distributions]: /docs/concepts/distributions/
+[distributed tracing]: ../signals/traces/
+[distributions]: ../distributions/
 [field]: /docs/specs/otel/logs/data-model#field-kinds
 [http]: https://en.wikipedia.org/wiki/Hypertext_Transfer_Protocol
 [instrumented library]: /docs/specs/otel/glossary/#instrumented-library
 [json]: https://en.wikipedia.org/wiki/JSON
 [log record]: /docs/specs/otel/glossary#log-record
 [log]: /docs/specs/otel/glossary#log
-[metric]: /docs/concepts/signals/metrics/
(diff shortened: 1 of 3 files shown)


content/en/docs/concepts/glossary.md
```diff
@@ -177,13 +177,13 @@ A language-specific collection of behavior invoked by an interface.
 Sometimes used to refer to a collection of [Log records](#log-record). Can be
 ambiguous since people also sometimes use [Log](#log) to refer to a single
 [Log record](#log-record). Where ambiguity is possible, use additional
-qualifiers, for example, `Log record`. See [more][log]
+qualifiers, for example, `Log record`. See [Log].
```
tiffany76, line 180: Non-blocking, but I just want to make sure we want to link to the spec glossary definition here and not the [Logs](https://opentelemetry.io/docs/concepts/signals/logs/) concept page as we do for ### Metrics.

content/en/docs/concepts/glossary.md
```diff
@@ -266,17 +266,17 @@ Short for [OpenTelemetry Protocol](/docs/specs/otlp/).
 ### Propagators
 
 Used to serialize and deserialize specific parts of telemetry data such as span
-context and [Baggage](#baggage) in [Spans](#span). See [more][propagators].
+context and [Baggage](#baggage) in [Spans](#span). See [Propagators].
```
tiffany76, line 269: This links to the Go instrumentation page, but I think Propagators can be used in other languages? Maybe the [Propagators API](https://opentelemetry.io/docs/specs/otel/context/api-propagators/) page is better.

tiffany76 (approved):
Two non-blocking suggestions. Otherwise, LGTM.

content/en/docs/concepts/glossary.md
chalin: Ah, right. There are actually other problems with this entry. I'll create a separate tracker issue for this.

content/en/docs/concepts/glossary.md
chalin: Will track via:

- #6985

Reviewers, please comment in that issue.

content/en/docs/concepts/glossary.md
chalin: Will track via:

- #6985

Reviewers, please comment in that issue.

PR #6948, opened 2025-05-21
DRAFT Update semantic conventions to unreleased v1.35.0-dev

opentelemetrybot:
This PR updates the semantic conventions to v1.35.0-dev.

PR #6986, opened 2025-05-28
[CI] Update pr-actions.yml: delete pr-fix.patch file after it is applied

chalin:
- Contributes to 
- Context: example of us having to delete the file manually, https://github.com/open-telemetry/opentelemetry.io/pull/6948/commits/feaf2dea5d2fba98ba5072b885e7de233fe791ee
- Ensure that `pr-fix.patch` is deleted after the patch is applied so that the file isn't committed.

/cc @svrnm

### DIFF
--- a/.github/workflows/pr-actions.yml
+++ b/.github/workflows/pr-actions.yml
@@ -116,6 +116,7 @@ jobs:
             echo "Patch failed to apply"
             exit 1
           }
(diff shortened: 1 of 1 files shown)


chalin:
@open-telemetry/docs-approvers - let me know if you give the `/fix` command a try, and whether it works now.

PR #6991, opened 2025-05-28
Update `deployment.environment` to `deployment.environment.name`

adriangb:
I believe the former is deprecated.


lmolkova (approved):
Thank you!

PR #6998, opened 2025-05-30
Bump opentelemetry-go submodule to latest (v1.36.0)

svrnm:
Updating opentelemetry-go to the most recent version, hoping that this fixes a bug with the CI which struggles to find the pinned version


chalin (commented):
Did you run code-excerpt refresh to see if that is impacted? If so, then the code and associated docs need to be updated.

If no one is available to perform such an update, then we can find another way to resolve the original issue (of the commit hash not being found).

svrnm:
> Did you run code-excerpt refresh to see if that is impacted? If so, then the code and associated docs need to be updated.
> 
> If no one is available to perform such an update, then we can find another way to resolve the original issue (of the commit hash not being found).

No I didn't run that, this was the reason why we kept it on 1.27, right? I think I remember the conversations now.

What alternatives do we have? Increase DEPTH?

chalin:
> What alternatives do we have? Increase DEPTH?

We could, but might run in to the problem again. I think I have a better solution, see #7000.

Issue #6947, opened 2025-05-21
[CI] PR-action script tuning


chalin:
I think it can be simplified, but have no issue with this being merged as-is

_Originally posted by @trask in https://github.com/open-telemetry/opentelemetry.io/pull/6894#pullrequestreview-2858062904_

svrnm:
Re simplifcation, let's continue the discussion here: we introduced that rather complex way of renaming the branches etc because we had multiple instances were people used branch names that duplicated existing branches (e.g. patch-1 or patch-2, etc.), this may no longer be an issue, especially since GitHub started to add your github handle into the name.

For the other 2 issues, let me see that I provide a fix!

chalin:
Script fails when `/fix:refcache` makes no changes to the refcache. For example, see:

- https://github.com/open-telemetry/opentelemetry.io/actions/runs/15562738548

svrnm:
yeah, there seems to be a need to still checkout the repo such that `gh` can comment to the right destiniation

PR #7004, opened 2025-05-30
PR-actions: use verbose flag when applying patch

chalin:
- Contributes to #6947
- Provides a bit better diagnostics when running `git apply` by using the `--verbose` flag
- Also lets the reviewer know that they can download the patch and try to apply it themselves

### DIFF
--- a/.github/workflows/pr-actions.yml
+++ b/.github/workflows/pr-actions.yml
@@ -112,10 +112,11 @@ jobs:
 
       - name: Apply patch
         run: |
-          git apply --check pr-fix.patch && git apply pr-fix.patch || {
-            echo "Patch failed to apply"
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-actions.yml
```diff
@@ -132,7 +133,7 @@ jobs:
             echo remote_repo=$remote_repo
             remote_branch=$(git config --get branch.${current_branch}.merge)
             echo remote_branch=$remote_branch
-            git commit -m "Results from /fix:${PR_ACTION}"
+            git commit -m "Results from /fix directive"
```
chalin, line 136: We don't have access to `PR_ACTION` in this job, and I don't think that it's work re-parsing it, so I've simplified the message. /cc @svrnm

PR #6918, opened 2025-05-19
[chore] Split distributions list

julianocosta89:
- This PR enhances the documentation for Collector distributions by adding vendor-specific entries and updating the ecosystem tables. 
- I've introduced a new YAML files for OSS and another one for vendor distributions. 
- I've also updated the distributions table shortcode to support dynamic data sources.
- Fixes #6888

**Previews**:

- https://deploy-preview-6918--opentelemetry.netlify.app/ecosystem/distributions/
- https://deploy-preview-6918--opentelemetry.netlify.app/docs/collector/distributions/

PR #7008, opened 2025-05-30
Ecosystem distributions page copyedits

chalin:
- A few copyedits of the ecosystem distro page. Things I noticed during review of #6918.
- These changes shouldn't conflict with #6918
- /cc @julianocosta89 @tiffany76 et all


content/en/ecosystem/distributions.md
```diff
@@ -1,24 +1,25 @@
...
+use OpenTelemetry components, though distributions might facilitate usage under
 certain circumstances, such as specific vendor requirements.
 
 The following list contains a sample of OpenTelemetry distributions and the
 component they customize.
 
 {{% alert title="Note" color="warning" %}} OpenTelemetry **does not validate or
-endorse** the third-party distributions listed in the following table. The list
-is provided as a convenience for the community. {{% /alert %}}
+endorse** the third-party distributions listed below. This list is provided as a
```
theletterf, line 21: Oh, I like to avoid positional language, hence the "following table".

content/en/ecosystem/distributions.md
```diff
@@ -1,24 +1,25 @@
...
+[components] so that they're easier to deploy and use with specific
+observability backends.
 
-Any third-party can customize OpenTelemetry components with backend, vendor, or
-end-user specific changes. You don't have to use a distributions in order to use
-OpenTelemetry components, though distributions might facilitate usage under
+Any third-party can customize OpenTelemetry components with backend, [vendor],
+or end-user specific changes. You don't have to use a distributions in order to
+use OpenTelemetry components, though distributions might facilitate usage under
 certain circumstances, such as specific vendor requirements.
```
tiffany76: Rewrote to avoid the negative and use plainer language.

```suggestion
or end-user specific changes. You can use OpenTelemetry components without a distribution, but a distribution might make things easier in some cases, like when a vendor has specific requirements.
```

content/en/ecosystem/distributions.md
chalin: But "following" is positional too, ... no? :)

chalin:
Thanks for the reviews and feedback all.

content/en/ecosystem/distributions.md
chalin: Jokes aside, I understand the subtle difference, and I generally prefer "the following" too. In this case, the wording seemed simpler when using "below". Since you approved, I'm assuming you're ok with this going in as is.

PR #5749, opened 2024-12-09
remove suggestion to process internal telemetry through collector

codeboten:
The following PR removes the suggestion in the documentation to send the Collector's internal telemetry through the Collector's pipelines. As the warning suggests, doing this is risky and we likely don't want to recommend users do this.

PR #7035, opened 2025-06-03
Update Internal Collector Telemetry Docs

avillela:
This PR contains updates to the documentation on Internal Collector Telemetry to help clarify some of the approaches for exporting internal Collector metrics. It also includes an explanation on why self-ingesting telemetry is not advisable.

### DIFF
--- a/content/en/docs/collector/internal-telemetry.md
+++ b/content/en/docs/collector/internal-telemetry.md
@@ -133,8 +133,10 @@ There are a couple of ways to export internal Collector metrics.
                    endpoint: https://${OTLP_ENDPOINT}:4318
    ```
 
-   If you're using gRPC, then `protocol` would be `grpc/protobuf`, and
-   `endpoint` would be `https://${OTLP_ENDPOINT}:4317`.
+   {{% alert title="Notes" color="info" %}}
+   - If you're using gRPC instead of HTTP, then `protocol` would be `grpc`.
+   - Be sure to include a port number for your endpoint as needed.
+     {{% /alert %}}
 
    You can also send send telemetry directly to an OTLP backend for analysis. In
    this case, you will also need to include a `headers` configuration:
@@ -154,12 +156,15 @@ There are a couple of ways to export internal Collector metrics.
                        value: '${API_TOKEN}'
    ```
 
-   {{% alert title="WARNING" color="warning" %}} Although the above approach is
-   possible, it is not recommended, as each Collector in your fleet would send
-   its own internal telemetry directly to the observability backend, causing
-   traffic saturation at the ingest point. Instead, you should consider sending
-   telemetry to a central Collector dedicated to ingesting telemetry from the
-   Collectors in your fleet. {{% /alert %}}
+   {{% alert title="WARNING" color="warning" %}}
+
+   Although the above approach is possible, it is not recommended, as each
+   Collector in your fleet would send its own internal telemetry directly to the
+   observability backend, causing traffic saturation at the ingest point.
+   Instead, you should consider sending telemetry to a central Collector
+   dedicated to ingesting telemetry from the Collectors in your fleet.
+
+   {{% /alert %}}
 
 {{% alert title="Internal telemetry configuration changes" color="info" %}}
 
@@ -277,11 +282,13 @@ service:
                 endpoint: https://${OTLP_ENDPOINT}
 ```
 
+{{% alert title="Notes" color="info" %}}
+
+- If you're using gRPC instead of HTTP, then `protocol` would be `grpc`.
+- Be sure to include a port number for your endpoint as needed. {{% /alert %}}
+
 Again, it is recommended to send internal Collector telemetry to a Collector
-dedicated to processing only internal telemetry from other Collectors. If using
-a dedicated Collector for processing internal telemetry, you also need to
-include the port numbers in the `endpoint` configuration - `4317` for `grpc`,
-and `4318` for `http/protobuf`.
+dedicated to processing only internal telemetry from other Collectors.
 
 See the [example configuration][kitchen-sink-config] for additional options.
 Note that the `logger_provider` section there corresponds to `logs` here.
@@ -312,11 +319,13 @@ service:
                 endpoint: https://${OTLP_ENDPOINT}
 ```
 
+{{% alert title="Notes" color="info" %}}
+
+- If you're using gRPC instead of HTTP, then `protocol` would be `grpc`.
+- Be sure to include a port number for your endpoint as needed. {{% /alert %}}
+
 Again, it is recommended to send internal Collector telemetry to a Collector
-dedicated to processing only internal telemetry from other Collectors. If using
-a dedicated Collector for processing internal telemetry, you also need to
-include the port numbers in the `endpoint` configuration - `4317` for
-`grpc/protobuf`, and `4318` for `http/protobuf`.
+dedicated to processing only internal telemetry from other Collectors.
 
 See the [example configuration][kitchen-sink-config] for additional options.
 Note that the `tracer_provider` section there corresponds to `traces` here.


content/en/docs/collector/internal-telemetry.md
```diff
@@ -38,38 +54,80 @@ exposes them using the OpenTelemetry Go
...
-        - pull:
-            exporter:
-              prometheus:
-                host: '0.0.0.0'
-                port: 8888
-```
+There are three ways to export internal Collector metrics.
+
+1. Self-ingesting, exporting internal metrics via the
+   [Prometheus exporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/prometheusexporter).
```
jmichalek132: Afaik this is not using the prometheus exporter, rather https://github.com/open-telemetry/opentelemetry-go/tree/main/exporters/prometheus

content/en/docs/collector/internal-telemetry.md
```diff
@@ -38,38 +54,80 @@ exposes them using the OpenTelemetry Go
...
+   ```yaml
+   service:
+     telemetry:
+       metrics:
+         readers:
+           - pull:
+               exporter:
+                 prometheus:
+                   host: '0.0.0.0'
+                   port: 8888
```
jmichalek132, line 79: might be also nice to provide example how to get the original names back
https://github.com/open-telemetry/opentelemetry-collector/blob/e1f670844604a5b119d8560bc079ceca4c92bf72/CHANGELOG.md?plain=1#L347

avillela: @jmichalek132 happy to do that. Can you elaborate on what is meant by the line [`Users who do not customize the Prometheus reader should not be impacted.`](https://github.com/open-telemetry/opentelemetry-collector/blob/e1f670844604a5b119d8560bc079ceca4c92bf72/CHANGELOG.md?plain=1#L344-L345) in the changelog? Is the "Prometheus reader" the same as the "Prometheus receiver"?

Do the Prometheus receiver (`receivers::prometheus`) and/or Prometheus exporter (`exporters::prometheus`) have to be configured when the internal metrics exporter is `prometheus`?

content/en/docs/collector/internal-telemetry.md
avillela: @jmichalek132 thanks for the clarification. How does the Go Prometheus exporter differ from one in the Collector?? I was under the impression that the Collector's one was based on the Go one??

content/en/docs/collector/internal-telemetry.md
```diff
@@ -24,11 +24,27 @@ For details and to track progress see
...
 
-- Internal [metrics](#configure-internal-metrics) are exposed using a Prometheus
-  interface which defaults to port `8888`.
-- [Logs](#configure-internal-logs) are emitted to `stderr` by default.
+- [Metrics](#configure-internal-metrics)
+- [Logs](#configure-internal-logs)
+- [Traces](#configure-internal-traces)
+
+{{% alert title="Who monitors the monitor?" color="info" %}} Internal Collector
+metrics can be exported directly to a backend for analysis, or to the Collector
```
kallangerard: I agree that a collector shouldn't self monitor it's own telemetry, I would be wary of suggesting the telemetry should be sent directly to a backend.

For example imagine a common agent and gateway pattern on Kubernetes. We could have hundreds or thousands node agents, batching and shipping application telemetry to a gateway layer that could be comprised of only a few instances.

If someone tried to make connections from thousands of node agents to a vendors backend for otelcol telemetry, there may be a lot of scaling issues from that. The internal-telemetry would also never have a chance to be enriched with Kubernetes metadata.

I've been thinking of using a dedicated internal telemetry gateway of otelcol instances for this purpose, so every otelcol instance regardless of whether it's a node agent, a gateway or a load balancing exporter layer, would send to the same collector instances dedicated to otelcol telemetry. I'm not sure what to call this pattern, but maybe we can suggest it here?

content/en/docs/collector/internal-telemetry.md
```diff
@@ -24,11 +24,27 @@ For details and to track progress see
...
+
+As a matter of best practice, the subject being monitored shouldn't also be the
+monitor.
+
+When a Collector is responsible for handling its own telemetry through a traces,
+metrics, or logs pipeline and encounters an issue (e.g. memory limiter blocking
+data), its internal telemetry won't be sent to its intended destination. This
+makes it difficult to detect problems with the Collector itself. Likewise, if
+the Collector generates additional telemetry related to the above issue, such
+error logs, and those logs are sent into the same collector, it can create a
```
kallangerard: Do we need an example of excluding otelcol's own logs from log tailing. I know Spunk's otelcol helm had some examples of this. I believe it was using custom exclude annotations if I recall correctly.

content/en/docs/collector/internal-telemetry.md
```diff
@@ -38,38 +54,80 @@ exposes them using the OpenTelemetry Go
...
+
+   ```yaml
+   service:
+     telemetry:
+       metrics:
+         readers:
+           - pull:
+               exporter:
+                 prometheus:
+                   host: '0.0.0.0'
```
kallangerard, line 78: Is `0.0.0.0` right here?

- I'm not sure if this will work for IPv6 only stacks
- The default behaviour has been changed in endpoints and such from `0.0.0.0` to `localhost`. Should we just be using `localhost` if we are intending to expose for self scraping, or be more explicit for public interfaces?

Probably not something we need to tackle here, but I'd love a chime in from anyone with better container/Linux networking knowledge than me.

content/en/docs/collector/internal-telemetry.md
avillela: @kallangerard I like that idea! I'll make the revisions.

content/en/docs/collector/internal-telemetry.md
avillela: @kallangerard can you point me in the direction of this documentation?

content/en/docs/collector/internal-telemetry.md
avillela: @kallangerard this was in the original docs, so I can't speak for it (and haven't used it). Hope someone else will chime in with more info. 😁

content/en/docs/collector/internal-telemetry.md
jmichalek132: @dashpole can answer that nicely.

dashpole: There are two reasons:

First, the exporters implement different interfaces.  There is a [Reader](https://github.com/open-telemetry/opentelemetry-go/blob/1c983778cb51a799ffd13105a0a7e4f6b58fd870/sdk/metric/reader.go#L44), and the [exporter.Metrics](https://github.com/open-telemetry/opentelemetry-collector/blob/fd4d3c44650723d4362732b59b0d3f501c02cd1a/exporter/exporter.go#L22) for the collector.

Second, the collector's exporter is designed to aggregate metrics from multiple resources/targets together, similar to how the Prometheus server's [/federate](https://prometheus.io/docs/prometheus/latest/federation/) endpoint works.  The Go SDK exporter is designed to only handle metrics from a single resource, more like prometheus [client_golang](https://github.com/prometheus/client_golang).

content/en/docs/collector/internal-telemetry.md
jmichalek132: On the same not we just use prometheus to scrape otel collector metrics directly, might be worth calling it out as an option.

avillela: @jmichalek132 sorry...I'm confused. Isn't that what setting the `exporter` to prometheus does in the first config??

content/en/docs/collector/internal-telemetry.md
avillela: @dashpole so when `exporter` is configured to use prometheus to export internal metrics, then it's using the Go Prometheus exporter behind the scenes then?

dashpole: Yes, that is correct.  You can link to go.opentelemetry.io/otel/exporters/prometheus

content/en/docs/collector/internal-telemetry.md
kallangerard: I just had a look, I was wrong, Splunk used path filtering in the filelog receiver to exclude their self-logs, while otlp logs will filter out logs with an exclude annotation on the pod. They seem be excluding filelog capability entirely in their latest examples though.

See  this old version   https://github.com/signalfx/splunk-otel-collector-chart/blob/0fa56adc9c55728094da367fd71a51655b1da40a/examples/only-logs-with-extra-file-logs/rendered_manifests/configmap-agent.yaml#L133-L136

I think if someone is using the filelog receiver, they're likely to have already come across this issue and are handling it in their own way.

For internal telemetry logs to an otlp endpoint, I don't think there's any way to do it safely by sending to its own otlp receiver endpoint. I'm not 100% sure on a safe alternative though. I believe there's some internal rate limiting in the otelcol's internal logs, but I haven't tested it with a self-exporting broken logging pipeline. I've been scraping otelcol logs with Datadog for a while and haven't seen any runaway log volumes, but I guess that's not self consumption. 🤷😅

Outside of this PR I'll try and write up some examples of a dedicated internal telemetry collector.

content/en/docs/collector/internal-telemetry.md
tiffany76: For reference, we used to have a similar warning on this page, but it was [removed](https://github.com/open-telemetry/opentelemetry.io/pull/5749) when the self-monitoring section was removed. I'll leave it up to @codeboten if we want to add the warning again.

tiffany76:
@open-telemetry/collector-approvers, PTAL. Thanks!

content/en/docs/collector/internal-telemetry.md
avillela: @jmichalek132 friendly reminder for clarification on above item 😁

content/en/docs/collector/internal-telemetry.md
```diff
@@ -38,40 +53,120 @@ exposes them using the OpenTelemetry Go
...
-            exporter:
-              prometheus:
-                host: '0.0.0.0'
-                port: 8888
-```
-
-{{% alert title="Internal telemetry configuration changes" %}}
+There are three ways to export internal Collector metrics.
+
+1. Self-ingesting, exporting internal metrics via the
```
jade-guiton-dd: What does "self-ingesting" mean here? It doesn't look like this section has the Collector ingest its own metrics.

content/en/docs/collector/internal-telemetry.md
```diff
@@ -38,40 +53,120 @@ exposes them using the OpenTelemetry Go
...
+         readers:
+           - pull:
+               exporter:
+                 prometheus:
+                   host: '0.0.0.0'
+                   port: 8888
+   ```
+
+2. Self-ingesting and exporting, scraping metrics via the Collector's own
+   [Prometheus Receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/prometheusreceiver).
```
jade-guiton-dd: I'm not sure it makes sense to have a section about self-ingesting Prometheus-exported metrics, for two reasons:

1. We used to have a section about self-ingesting OTLP-exported metrics in the past, but it was removed, because we want to discourage users from doing self-ingestion, as it can introduce reliability and data loss issues (if the Collector is unhealthy, who's going to export the metrics showing that it is?). See #5749 for precedent.

3. If users are going to be doing self-ingestion anyways, I think it makes more sense to do it through OTLP rather than Prometheus. This will preserve all of the telemetry and its semantics as-is, without unintended conversions (eg. metric names with the default config) or data loss (eg. scope attributes until recently) that could occur due to the Prometheus exporter/receiver combo. Moreover, it's possible with the Core Collector distribution, not just Contrib.

content/en/docs/collector/internal-telemetry.md
```diff
@@ -38,40 +53,120 @@ exposes them using the OpenTelemetry Go
...
+                 otlp:
+                   protocol: http/protobuf
+                   endpoint: https://${OTLP_ENDPOINT}
+                   headers:
+                     - name: ${HEADER}
+                       value: '${API_TOKEN}'
+   ```
+
+   {{% alert title="WARNING" color="warning" %}} Although the above approach is
+   possible, it is not recommended, as it can introduce scaling issues.
```
jade-guiton-dd: What kind of scaling issues do you expect when exporting metrics through OTLP? Is this comment meant to be about self-ingestion? I'd rather we didn't discourage people from emitting internal telemetry with our own protocol.

content/en/docs/collector/internal-telemetry.md
```diff
@@ -161,9 +256,12 @@ service:
             exporter:
               otlp:
                 protocol: http/protobuf
-                endpoint: https://backend:4318
+                endpoint: https://${OTLP_ENDPOINT}
```
jade-guiton-dd, line 282: Nitpick: People often mix up OTLP/gRPC and OTLP/HTTP, so I think keeping a reminder of the standard port would be good. Maybe add a line like "This will load the endpoint from the OTLP_ENDPOINT environment variable, which should look something like `backend:4318`"

content/en/docs/collector/internal-telemetry.md
```diff
@@ -24,11 +24,26 @@ For details and to track progress see
...
-- Internal [metrics](#configure-internal-metrics) are exposed using a Prometheus
-  interface which defaults to port `8888`.
-- [Logs](#configure-internal-logs) are emitted to `stderr` by default.
+By default, the Collector exposes its own telemetry in three ways:
+
+- [Metrics](#configure-internal-metrics)
+- [Logs](#configure-internal-logs)
+- [Traces](#configure-internal-traces)
+
+{{% alert title="Who monitors the monitor?" color="info" %}} As a matter of best
```
tiffany76:
```suggestion
{{% alert title="Who monitors the monitor?" %}} As a matter of best
```

content/en/docs/collector/internal-telemetry.md
```diff
@@ -24,11 +24,26 @@ For details and to track progress see
...
+practice, the subject being monitored shouldn't also be the monitor. Internal
+Collector telemetry should ideally be exported to another OpenTelemetry
+Collector dedicated to processing telemetry from other Collectors, which then
+exports the telemetry to an OTLP backend for analysis.
+
+When a Collector is responsible for handling its own telemetry through a traces,
+metrics, or logs pipeline and encounters an issue (e.g. memory limiter blocking
```
tiffany76:
```suggestion
metrics, or logs pipeline and encounters an issue (for example, memory limiter blocking
```

content/en/docs/collector/internal-telemetry.md
```diff
@@ -24,11 +24,26 @@ For details and to track progress see
...
+data), its internal telemetry won't be sent to its intended destination. This
+makes it difficult to detect problems with the Collector itself. Likewise, if
+the Collector generates additional telemetry related to the above issue, such
```
tiffany76:
```suggestion
the Collector generates additional telemetry related to the above issue, such as
```

content/en/docs/collector/internal-telemetry.md
tiffany76: @avillela We just added a note about this further down the page. You could link to it from here, if you want, and I think that should handle this suggestion.

https://opentelemetry.io/docs/collector/internal-telemetry/#_total-suffix

tiffany76 (commented):
I left a few suggestions, but I'll do a more thorough copy edit after the content is finalized based on the other suggestions. Thanks, @avillela!

svrnm:
@avillela can you take a look at the reviews provided? thanks

avillela:
> @avillela can you take a look at the reviews provided? thanks

@svrnm Taking a look!

content/en/docs/collector/internal-telemetry.md
avillela: @jade-guiton-dd excellent points. We want to discourage bad patterns, so I'll just remove point 2.

content/en/docs/collector/internal-telemetry.md
avillela: @jade-guiton-dd I was attempting to clarify the existing docs for this piece, but I'm afraid I may have conflated things. The current explanation of "expose the Prometheus endpoint to one specific or all network interfaces when needed" doesn't make much sense to me. Is this just configuration to send internal collector telemetry directly to Prometheus?

content/en/docs/collector/internal-telemetry.md
avillela: @jade-guiton-dd I was referring to [this earlier comment](https://github.com/open-telemetry/opentelemetry.io/pull/7035#discussion_r2125599263) from @kallangerard. I have added more clarification to the note.

content/en/docs/collector/internal-telemetry.md
avillela: @jade-guiton-dd I feel like that is a bit misleading. For example, if I set the endpoint to be VendorX Observability backend (i.e. NOT the OTel Collector), then my backend's URL does not have the 4317 or 4318 port number. I can add that as a clarification for an OTLP destination that IS an OTel Collector.

content/en/docs/collector/internal-telemetry.md
jade-guiton-dd: More or less, yes. Note that Prometheus is a pull-based protocol; Prometheus will scrape the Prometheus endpoint the Collector exposes to get the metrics, rather than the Collector proactively sending them, but it's the same idea.

content/en/docs/collector/internal-telemetry.md
jade-guiton-dd: (For context, the note now reads:)
> Although the above approach is possible, it is not recommended, as each Collector in your fleet would send its own internal telemetry directly to the observability backend, causing traffic saturation at the ingest point. Instead, you should consider sending telemetry to a central Collector dedicated to ingesting telemetry from the Collectors in your fleet.

I think what @kallangerard was describing is the "gateway" deployment pattern. I think we can definitely call it out as a possibility compared to sending directly to an endpoint. Adding a link to [the docs page describing it](https://opentelemetry.io/docs/collector/deployment/gateway/) may be a good idea as well.

However:
- I think backend endpoint saturation would only become a problem past a certain scale.
- If your gateway only has a single "central Collector", you'll probably end up with even worse saturation issues than before. You need multiple Collectors and a load balancer to make an effective Gateway layer.
- Moreover, your backend vendor probably already has all of that. So the way I understand it, the gateway pattern is mostly useful if:
    - you want to enrich the telemetry further with Collector components, or
    - you want to export to a non-OTLP backend or
    - you want to filter or aggregate the telemetry to reduce the load.

So I don't think it's reasonable to so directly recommend a gateway architecture vs directly sending the data to the backend. I think it will depend on your scale and your needs, and whether you're willing to handle the additional complexity.

content/en/docs/collector/internal-telemetry.md
jade-guiton-dd: That's a good point, it's true that the endpoints can be on any port, and OTLP/HTTP endpoints are probably often on the default HTTP port. I think the warning calling out the difference between the two protocols is enough.

content/en/docs/collector/internal-telemetry.md
```diff
@@ -38,59 +53,115 @@ exposes them using the OpenTelemetry Go
...
+       metrics:
+         readers:
+           - periodic:
+               exporter:
+                 otlp:
+                   protocol: http/protobuf
+                   endpoint: https://${OTLP_ENDPOINT}:4318
+   ```
+
+   If you're using gRPC, then `protocol` would be `grpc/protobuf`, and
```
jade-guiton-dd: `protocol` would [actually be](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/caee80916a50f168c7152967dabaefd0c3cd17c0/otelconf/v0.3.0/config.go#L31) just `grpc` for OTLP over gRPC. (Yes, it's a bit inconsistent.)

avillela: Ah, thanks @jade-guiton-dd

content/en/docs/collector/internal-telemetry.md
```diff
@@ -229,9 +309,15 @@ service:
...
                 protocol: http/protobuf
-                endpoint: https://backend:4318
+                endpoint: https://${OTLP_ENDPOINT}
 ```
 
+Again, it is recommended to send internal Collector telemetry to a Collector
+dedicated to processing only internal telemetry from other Collectors. If using
+a dedicated Collector for processing internal telemetry, you also need to
+include the port numbers in the `endpoint` configuration - `4317` for
+`grpc/protobuf`, and `4318` for `http/protobuf`.
```
jade-guiton-dd: This part needs to be updated to `grpc` as well.

I think this explanation regarding ports is misleading: there may be backend endpoints that require a port number; whether it's a Collector or not is unrelated. (And in fact, you can configure a Collector to listen on port 80/443 if you want.) Maybe we should just warn users to not forget to specify what port the endpoint is on if necessary.

avillela: My bad - you're totally right. I've made the updates.

content/en/docs/collector/internal-telemetry.md
mx-psi: I agree with Jade, I don't think we should add the warning as is, the situation is more nuanced than that

svrnm:
@open-telemetry/collector-approvers PTAL!

jade-guiton-dd:
In my opinion, [the thread](https://github.com/open-telemetry/opentelemetry.io/pull/7035#discussion_r2174787031) about "you should consider sending telemetry to a central Collector dedicated to ingesting telemetry from the Collectors in your fleet" has not been addressed.

svrnm:
> In my opinion, [the thread](https://github.com/open-telemetry/opentelemetry.io/pull/7035#discussion_r2174787031) about "you should consider sending telemetry to a central Collector dedicated to ingesting telemetry from the Collectors in your fleet" has not been addressed.

@avillela can you take another look?

vitorvasc:
Hi! Just checking in, is this still valid? 

/cc @avillela @svrnm @tiffany76

svrnm:
This has lots of conflicts with other changes that have been applied, so I will close this now, @avillela please revisit the updated page and see if and where your changes are applicable, thank you!

PR #7044, opened 2025-06-04
add lang:bn to component label map

svrnm:
I think we missed this one.

### DIFF
--- a/.github/component-label-map.yml
+++ b/.github/component-label-map.yml
@@ -26,6 +26,10 @@ lang:fr:
   - changed-files:
       - any-glob-to-any-file:
           - content/fr/**
+lang:bn:
+  - changed-files:
+      - any-glob-to-any-file:
+          - content/bn/**
 sig:cpp:
   - changed-files:
       - any-glob-to-any-file:


tiffany76 (approved):
Weird that I see `lang:bn` in the label list, but you're right, it's not in this file. 
Screenshot 2025-06-04 at 2 09 58 PM

chalin:
It is my understanding that if one adds it to the file, it'll show up in the web i/f, but not the other way around.

tiffany76:
> It is my understanding that if one adds it to the file, it'll show up in the web i/f, but not the other way around.

For my own understanding, why do you think I could already see it in the GitHub UI before this PR was merged?

chalin:
Because someone must have added the label from the web UI (without first updating the file). In such a case, the web UI change doesn't propagate to the file (since that would require a repo commit / PR).

The implication is this behavior is: if changes are made via the web UI, and later the file is update, then web UI changes will be lost unless they are also encoded in the file.

tiffany76:
Got it. Thanks for explaining!

PR #7060, opened 2025-06-05
Invalid process runtime resource provider class name in Java configuration

youribonnaffe:
The name of the class for the process runtime resource provider was incorrect (missing `Resource`).

Change is touching multiple lines to properly format the table.


jaydeluca (approved):

youribonnaffe:
> /fix:format

Thank you, looks like I missed an extra space for the table.

PR #6952, opened 2025-05-21
Update opentelemetry-specification version to v1.46.0

opentelemetrybot:
- This is a ~draft~ PR used for identifying issues integrating the latest (unreleased) specification.
- Closes #7098 by superseding it. This PR contains all the necessary fixes.

---

**Preview**: https://deploy-preview-6952--opentelemetry.netlify.app/docs/specs/otel/

PR #7092, opened 2025-06-11
Fix semconv and spec PR workflows

trask:
I looked at the workflow and I _think_ it was supposed to do what @chalin is asking for in https://github.com/open-telemetry/opentelemetry.io/pull/6952#issuecomment-2963227719, but there was just a bug 🤞


chalin (approved):
Good catch! (It's easy to forget the -E.)  Looking forward to see if this helps streamline the update to a stable version from the working PR. Thanks for working on this 🙌🏻.

PR #7098, opened 2025-06-12
Update opentelemetry-specification version to v1.46.0, last step

opentelemetrybot:
Update opentelemetry-specification version to `v1.46.0`.

See https://github.com/open-telemetry/opentelemetry-specification/releases/tag/v1.46.0.

PR #7101, opened 2025-06-12
Update adjust-pages.pl also during spec and semconv automation

trask:
See https://github.com/open-telemetry/opentelemetry.io/pull/7098#issuecomment-2968156636

### DIFF
--- a/.github/workflows/update-semconv-integration-branch.yml
+++ b/.github/workflows/update-semconv-integration-branch.yml
@@ -85,6 +85,7 @@ jobs:
           cd ../..
 
           sed -i "s/^\tsemconv-pin = .*/\tsemconv-pin = $commit_desc/" .gitmodules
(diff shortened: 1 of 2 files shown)


chalin (approved):
Thanks for the update! I suspect that we might want to update the version in adjust-pages only when it's an official (non-dev) release. Although, if the PR isn't merged until then, it might not matter 🤔. Ok, let's give this a try.

trask:
> Thanks for the update! I suspect that we might want to update the version in adjust-pages only when it's an official (non-dev) release. Although, if the PR isn't merged until then, it might not matter 🤔. Ok, let's give this a try.

ya, I wasn't sure about that, it should be easy to only do it on official (non-dev) release, we know that a few lines above this change via this conditional:

> `if git ls-remote --exit-code --tags origin $VERSION; then`

PR #7109, opened 2025-06-14
Fix propogation link to express instrumentation

olaals:
In the "Automatic context propogation" section for JS SDK, there are two links, one that states "@opentelemetry/instrumentation-http" and one that states "@opentelemetry/instrumentation-express". 
However both these links point to the instrumentation-http package. 

This fixes the express link to point to express npm package.


martincostello (approved):

cartermp (approved):
Thanks for the catch!

PR #7110, opened 2025-06-14
mark SigNoz as oss and add contact email

makeavish:
Mark SigNoz OSS as it's opensource with 2 licenses - https://github.com/SigNoz/signoz/blob/main/LICENSE
Add contact email, can be verified here - https://signoz.io/support/

### DIFF
--- a/data/ecosystem/vendors.yaml
+++ b/data/ecosystem/vendors.yaml
@@ -369,8 +369,8 @@
 - name: SigNoz
   nativeOTLP: true
   url: https://signoz.io
-  contact:
-  oss: false
+  contact: support@signoz.io
+  oss: true
   commercial: true
 - name: SolarWinds
   nativeOTLP: true


chalin (changes requested):
"MIT Expat" isn't a license. The license is "MIT". Your project should have the full copy of the MIT license file. 

@svrnm - do we have other instances where oss is true and commercial is true as well?

(I might drop LF legal a question on this one. WDYT?)]

---

**Edit**. regarding naming, it seems that "MIT Expat" is acceptable as a way of naming the license, thought "MIT License" is preferable, unless you have a specific reason to refer to the name "MIT Expat".

makeavish:
> ~"MIT Expat" isn't a license. The license is "MIT".~ Your project should have the full copy of the MIT license file.
> 
> @svrnm - do we have other instances where oss is true and commercial is true as well?
> 
> (I might drop LF legal a question on this one. WDYT?)]
> 
> **Edit**. regarding naming, it seems that "MIT Expat" is acceptable as a way of naming the license, thought "MIT License" is preferable, unless you have a specific reason to refer to the name "MIT Expat".

Actually MIT Expat is more clear, MIT License is misleading and GNU doesn't recommend it: https://www.gnu.org/licenses/license-list.en.html#Expat

Also there are other instances where both oss and commercial is true, example: https://github.com/open-telemetry/opentelemetry.io/blob/main/data/ecosystem/vendors.yaml#L74

chalin:
Ok, that works for me. @svrnm, thoughts?

chalin:
> image

You'll need to resubmit your PR using a branch name other than `main`; see https://opentelemetry.io/docs/contributing/pull-requests/#pr-guidelines. Link to this PR when you create the new one.

svrnm:
yes, we have other instances of OSS/commercial, e.g. grafana.

makeavish:
> > image
> 
> You'll need to resubmit your PR using a branch name other than `main`; see https://opentelemetry.io/docs/contributing/pull-requests/#pr-guidelines. Link to this PR when you create the new one.

Apologies for missing this, created another PR here: https://github.com/open-telemetry/opentelemetry.io/pull/7127

chalin:
No worries, it's easy to miss. Thanks for the new PR.

PR #7127, opened 2025-06-16
mark SigNoz as oss and add contact email

makeavish:
Mark SigNoz OSS as it's opensource with 2 licenses - https://github.com/SigNoz/signoz/blob/main/LICENSE
Add contact email, can be verified here - https://signoz.io/support/

---

Previously:

- #7110


chalin (commented):
Based on previous discussions (#7110), this seems good to me.

@tiffany76 @svnm - PTAL

Issue #4745, opened 2024-06-25
[i18n] Tooling to detect inconsistent translation and formatting Japanese text


svrnm:
_Originally posted by @ymotongpoo in https://github.com/open-telemetry/opentelemetry.io/discussions/4302#discussioncomment-9870096_:

> @svrnm @chalin Thank you for supporting each language localization! At least, the localization for Japanese could make a smooth start thanks to the great automation tools. 
> 
> Now I started feeling like I need some automation tool to detect inconsistent translation and Japanese formatting to have more contributions from many people. How can I add that feature to the repo? Tricky parts are:
> 
> - Only `lang:ja` labeled PRs need the 
> - The tool needs to have a set of mappings between correct and incorrect wordings (in some file)
> 
> What do you think about this idea?

svrnm:
hey @ymotongpoo, I copied your comment from the discussion to a dedicated issue because this is a scoped task.

We have tooling for that for English content already, but turned it off for translations, since it requires additional language-specific configuration. So here is what we need to do:

- for each tool create a configuration for Japanese that is only applied to files in `content/ja`
- enable them locally and test those tools if they work properly
- enable them in the GitHub actions workflows 

The tools are:

- [ ] cspell for spell checking
- [ ] prettier for formatting
- [ ] textlint (especially "terminology") for consistent translations

We should address them 1-by-1

ymotongpoo:
@svrnm thank you for routing the topic to the issue.

- cspell doesn't work well for Japanese and for this purpose. 
- it is said prettier 3.0.0+ works well with Japanese (ref: https://github.com/prettier/prettier/pull/11597)  
- textlint is the most fitting and common tool for this kind of Japanese document. (ref: an example for Japanese technical magazine https://gist.github.com/inao/f55e8232e150aee918b9)

svrnm:
>     * cspell doesn't work well for Japanese and for this purpose.

I was not aware of that, I did a quick check and indeed it only supports English so far, that's interesting. Do you have any recommendations for a spell checker that works well for Japanese?

>     * it is said prettier 3.0.0+ works well with Japanese (ref: [Markdown[next branch]: Do not insert spaces between Chinese/Japanese & latin letters prettier/prettier#11597](https://github.com/prettier/prettier/pull/11597))

If it works well, then let's have a PR to introduce it for Japanese. If you provide the language-specific config (if needed/where needed), @chalin or I can handle the rest

[...]

PR #7160, opened 2025-06-21
[wip] [ja] style: add prh/ja.yml to improve translation activities

Msksgm:
This PR is caused by https://github.com/open-telemetry/opentelemetry.io/issues/4745 to improve Japanese translation activities more efficiently.

I added `prh/ja.yml` and `textlint-rule-prh` dependency.
Other Markdown files are fixed by `npm run fix:text`.

A bellow image is result of ci, if I do not fix Japanese docs which have wrong words.

Screenshot 2025-06-21 at 15 14 23


Msksgm:
I'm not sure the name and file path of `prh/ja.yml`.
Please confirm it or suggest more appropriate a directory or a file name.

svrnm:
> I'm not sure the name and file path of `prh/ja.yml`. Please confirm it or suggest more appropriate a directory or a file name.

Not sure either, adding new top level folders always adds clutter, but I also don't see a better place for now. let's keep it and we can move it around later.

PR #7210, opened 2025-06-27
fix(docs): add links to stable, beta, development in Language API

### LINKED ISSUE #7172: Language APIs - Links to definition of `Stable`, `Beta` and `Development`
Looking at https://opentelemetry.io/docs/languages/#status-and-releases

seems like the definitions used above are from:
https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/0232-maturity-of-otel.md#beta

right? can we please easily hop from that page to the definitions

For example https://opentelemetry.io/docs/languages/rust/ seems to have good links to the terminology definitions

starlightknown:
closes #7172


theletterf (approved):
Excellent work, thank you!

PR #7217, opened 2025-06-27
Fix error message on failed check spell

maryliag:
Currently when a check spell CI test fails, it shows the message `Add comment '/fix:fix:format'` instead of the correct `Add comment '/fix:format'`


tiffany76 (approved):
Thanks for catching that! ✨

Issue #7223, opened 2025-06-29
[bug] - Register page search entries form do not work


kei-ichi:
Q: **What happened?** Describe the problem that occurred.
A: No matter whatever I input in the "search" field, after I click to the "Submit" button, **NOTHING** changed.
Even "filter" feature like "Language", "Type", "Flags" do not work either. Again, no matter what I choice to filter out, it will "always" show "Search 882 entries"

Q: **What did you expect would happen?** Describe the expected result/output.
A: Be able to "search" for specific entry or be able to filter out whatever I want

Q: **What is the name + path of the page where you encountered the issue?** The
relative path and page title where you found a problem.
A: [https://opentelemetry.io/ecosystem/registry/](https://opentelemetry.io/ecosystem/registry/)

Q: **What browser, OS, and platform were you using when you encountered the
problem?** Include the browser version if you know it as well.
A: Tested on macOS 15.5 (24F74), Chrome, Firefox, Safari, Brave. iPadOS (latest stable version): Safari, Firefox, Chrome

Below is the screenshot when I tried to search for "Prometheus":
Search URL: https://opentelemetry.io/ecosystem/registry/?s=Prometheus&component=&language=&flag=all
Search Result - Nothing changed, "Cloud Foundry" still in the "1st" result

![Image](https://github.com/user-attachments/assets/5f5b3309-7851-48a1-b358-7c68310c2909)

tiffany76:
Thanks for reporting your issue, @kei-ichi. 

Possibly related to #7100. @svrnm, FYI.

svrnm:
@kei-ichi the "882 entries" is static and will not change upon entering anything into the searchbox. However you are right that when you search for Prometheus you should see a different result list. I suspect that this is related to #7100 indeed. Can you share screenshots from the developer toolbar?

kei-ichi:
@svrnm Below are the screenshot I take from my machine (Mac Studio) with different browsers:

1. From Firefox

![Image](https://github.com/user-attachments/assets/22e7a633-b9e6-473f-888f-fce647acf091)

2. From Google Chrome

![Image](https://github.com/user-attachments/assets/5c29fe11-86ba-48ee-a0ee-54d01f294e93)

3. From Apple Safari

![Image](https://github.com/user-attachments/assets/a039f79e-5850-480e-9de0-f9c7916934fa)

4. From Brave

[...]

Issue #7100, opened 2025-06-12
otel registry search seems broken


breedx-splk:
The [registry](https://opentelemetry.io/ecosystem/registry/) on the website has a nifty search feature, but it looks to be broken. For example, entering the word "spring" and pressing enter yields this URL in the browser
https://opentelemetry.io/ecosystem/registry/?s=spring&component=&language=&flag=all but the results don't match (and don't seem to have changed).

I'm not sure what the root cause might be, but I think it used to work.

tiffany76:
Hi @breedx-splk, thanks for reporting this issue. I tried reproducing, but when I search for "spring" I get two results: 

<img width="1277" alt="Image" src="https://github.com/user-attachments/assets/67b1fa87-54ee-41eb-b786-df41c75f08f2" />

Is this not what you see? Or if it is, were you expecting to see something different?

breedx-splk:
Thanks @tiffany76 no that is not what I see at all! I enter "spring" in the search field, and it doesn't matter if I hit enter or if I click the submit button...the window kinda flashes and then just displays the same default list...

<img width="1241" alt="Image" src="https://github.com/user-attachments/assets/bc028744-b58a-4a5e-808b-edf13a6eedee" />

I use Chrome and also tried Safari and got the same results.

breedx-splk:
Hmmm, just tried it from a different machine and it worked fine. Might be some kind of local problem then. Sorry for the distraction.

tiffany76:
No worries! Thanks for reporting back.

mtwo:
I was having the same issue that @breedx-splk in Firefox, but it was working fine in Chrome. I cleared my Firefox cache and then it worked fine. The weird thing is that I'm pretty sure that I'd cleared my cache more recently than I'd visited the registry (Home Assistant sometimes has issues with the browser cache and I had to clear it a few weeks ago to fix it).

![Image](https://github.com/user-attachments/assets/a83ea924-6eb3-4565-a7b6-5ee80d639209)

This might still be worth looking into? Probably not super high priority though. Feel free to re-close if you disagree, I don't have a strong opinion!

breedx-splk:
I have this problem in both Chrome and Safari on MacOS. I cleared all site data in Safari, confirmed that it worked because light mode replaced dark mode in the website/registry, but search still didn't work for me. 😿

tiffany76:
Thanks for the updates! I'm going to pull in @chalin to see if he has time to look into this.

[...]

PR #7227, opened 2025-06-30
Add minisearch as dependency, remove CDN link

svrnm:
Re: #7100 #7223 

Not exactly sure what the problem is, but it seems there is a caching issue with the minisearch from the CDN. 

This PR works around that by adding minisearch as a dependency to package.json and importing it into the registrySearch.js directly.


tiffany76 (approved):
RSLGTM.

PR #7269, opened 2025-07-08
[zh] Add signals/baggage.md and update two pages

windsonsea:
Add zh text to:

```
added content/zh/docs/concepts/signals/baggage.md
modified  content/zh/docs/concepts/signals/_index.md
modified  content/zh/docs/concepts/signals/traces.md
```

### DIFF
--- a/content/zh/docs/concepts/signals/_index.md
+++ b/content/zh/docs/concepts/signals/_index.md
@@ -1,28 +1,31 @@
 ---
 title: 信号
 description: 了解 OpenTelemetry 支持的遥测类别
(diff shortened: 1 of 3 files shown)


content/zh/docs/concepts/signals/otel-baggage-2.svg
svrnm, line 1: you can leave the image(s) out since they are not edited

windsonsea: Fine, thanks, those images have been removed.

PR #7271, opened 2025-07-08
Update faas nodejs supported runtime version

wpessers:
This information is outdated. We dropped support for node 16 when we upgraded to the v2 js sdk. (https://github.com/open-telemetry/opentelemetry-js?tab=readme-ov-file#supported-runtimes)

nodejs16.x lambda runtimes are also deprecated already and will no longer be supported starting late 2025: https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html#runtimes-deprecated


tiffany76 (approved):
Thank you!

PR #7279, opened 2025-07-09
Deprecate the JS `@opentelemetry/instrumentation-redis-4` package

trentm:
It was merged into the `@opentelemetry/instrumentation-redis` package
(as of version 0.50.0), which already has an entry here.
See the deprecation notice at https://www.npmjs.com/package/@opentelemetry/instrumentation-redis-4

Refs: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2915


svrnm:
Instead of removing it we also can deprecate it, that way we can let people know about it, both options are valid, just wanted to call out that option

trentm:
> Instead of removing it we also can deprecate it

@svrnm Ah, that seems better. I'll update my PR. I hadn't realized taht was an option.

PR #7316, opened 2025-07-14
Updating Heroku adopters and vendor information

alexmarnell:
With the launch of the [Heroku Fir  Platform](https://www.heroku.com/blog/heroku-fir-generally-available-new-platform-capabilities/) which includes OpenTelemetry integration, we want to update the vendors list to include Heroku.

This PR also updates the adopters list to include updated information about the OpenTelemetry libraries and components we have adopted for internal use.

### DIFF
--- a/data/ecosystem/adopters.yaml
+++ b/data/ecosystem/adopters.yaml
@@ -67,8 +67,8 @@
   contact: ''
 - name: Heroku
   url: https://heroku.com
-  components: [Ruby]
-  reference: ''
+  components: [Collector, Go, Ruby, JavaScript, Operator]
+  reference: 'https://www.heroku.com/blog/opentelemetry-basics-on-heroku-fir/'
   contact: ''
 - name: MercadoLibre
   url: https://www.mercadolibre.com/
--- a/data/ecosystem/vendors.yaml
+++ b/data/ecosystem/vendors.yaml
@@ -156,6 +156,12 @@
   contact:
   oss: false
   commercial: true
+- name: Heroku
+  nativeOTLP: true
+  url: https://devcenter.heroku.com/articles/heroku-telemetry
+  contact:
+  oss: false
+  commercial: true
 - name: Highlight
   nativeOTLP: true
   url: https://www.highlight.io/docs/general/company/open-source/contributing/adding-an-sdk/


vitorvasc:
Hi @alexmarnell, thanks for your contribution! 😄 

Could you please sign the CLA? Even though this PR is still in draft mode, I noticed this requirement is missing.

alexmarnell:
> Could you please sign the CLA? Even though this PR is still in draft mode, I noticed this requirement is missing.

I will circle back once I've jumped through the corp legal bits on my end.

alexmarnell:
@vitorvasc I have signed the CLA.

data/ecosystem/vendors.yaml
```diff
@@ -156,6 +156,12 @@
   contact:
   oss: false
   commercial: true
+- name: Heroku
+  nativeOTLP: true
+  url: https://devcenter.heroku.com/articles/heroku-telemetry
+  contact:
```
tiffany76: Hi @alexmarnell, thanks for the update! Is there a contact point you could add here? It can be a GitHub profile or generic email address. We'd just like to have someone to contact if, for example, the link goes dead. Thanks!

alexmarnell: ~Sure, would something as generic like `https://github.com/heroku` work, or perhaps `https://www.heroku.com/policy/support/`?~

@tiffany76 I've updated the PR with a recommendation from our Marketing team.

svrnm (approved):

data/ecosystem/adopters.yaml
```diff
@@ -67,9 +66,10 @@
   contact: ''
 - name: Heroku
   url: https://heroku.com
-  components: [Ruby]
-  reference: ''
-  contact: ''
+  components: [Collector, Go, Ruby, JavaScript, Operator]
+  reference: 'https://www.heroku.com/blog/opentelemetry-basics-on-heroku-fir/'
+  referenceTitle: blog post
+  contact: 'https://www.heroku.com/contact/'
```
svrnm: Reading this, I would argue this is an integration and not adoption:

Adoption would mean "here is how heroku uses OpenTelemetry internally to ensure that services are up and running"

Integration means "here is how heroku provides OpenTelemetry to end users"

svrnm (changes requested):
@alexmarnell PTAL at my comment around adoption vs integration, I know it's confusing but can you move it to integarations?

data/ecosystem/adopters.yaml
alexmarnell: @svrnm That makes total sense. I've update the PR to add Heroku to the integration registry. I also tweaked the adopters to link off to the talk I presented at KubeCon NA24 on that topic.

Please let me know if there are any other tweaks you would like to see.

svrnm (approved):

PR #7329, opened 2025-07-16
fix auto update versions workflow

svrnm:
Intermediate fix to make the auto update script work, later we need to revisit this and other workflows to work with the otelbot


tiffany76 (approved):
RSLGTM.

PR #7354, opened 2025-07-21
Add information regarding BatchSpanProcessor's actual queue size

DavidAkaFunky:
When running tests to understand how `BatchSpanProcessor`'s properties affected its behaviour, I noticed the `otel.bsp.max.queue.size` I supplied was being rounded up to the nearest power of 2.

After doing some digging, I found out this class calls `JcTools.newFixedSizeQueue()` to create its span queue. When available, this method returns a new `MpscArrayQueue`, which inherits from `ConcurrentCircularArrayQueue`, whose size is calculated [by calling `Pow2.roundToPowerOfTwo()`](https://github.com/JCTools/JCTools/blob/a17b56f09cc9ec7d773707578c89f3ca9c11f968/jctools-core/src/main/java/org/jctools/queues/ConcurrentCircularArrayQueue.java#L49-L60).

This is not explicitly documented, so I felt the need to open a PR in case other people run into the same issue.


tiffany76 (approved):
Thanks so much, @DavidAkaFunky! We really appreciate your initiative!

It looks good to me from a copy edit perspective. @open-telemetry/java-instrumentation-approvers, PTAL!

vitorvasc:
@open-telemetry/java-approvers @open-telemetry/java-instrumentation-approvers PTAL!

laurit:
The queue implementation used by `BatchSpanProcessor` has changed from `MpscArrayQueue` to `MpscAtomicArrayQueue` that does not extend `ConcurrentCircularArrayQueue`, but it sill rounds the queue size to the next power of 2. If you wish to proceed with this I'd advise you to use language that suggests that a different size may be used if the provided value is unsuitable for the underlying queue implementation e.g. provided queue size may be rounded to the next power of 2. Linking to specific queue implementations may not be a good idea as it is an implementation detail and may change.

content/en/docs/languages/java/configuration.md
```diff
@@ -193,12 +193,12 @@ artifact coordinates.
...
+| -------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------- |
+| `otel.bsp.schedule.delay`        | The interval, in milliseconds, between two consecutive exports.                                                                                                                                                                                                                              | `5000`  |
+| `otel.bsp.max.queue.size`        | The maximum number of spans that can be queued before batching, rounded to the nearest power of two ([reference for the underlying queue's implementation](https://github.com/JCTools/JCTools/blob/master/jctools-core/src/main/java/org/jctools/queues/ConcurrentCircularArrayQueue.java)). | `2048`  |
```
trask, line 199: based on @laurit's https://github.com/open-telemetry/opentelemetry.io/pull/7354#issuecomment-3596078959, I'd suggest something like

```suggestion
| `otel.bsp.max.queue.size`        | The maximum number of spans that can be queued before batching, may be rounded to the next power of two. | `2048`  |
```

cartermp:
Agreed with @laurit and @trask's suggestions.

svrnm:
@DavidAkaFunky PTAL!

vitorvasc:
Hi @DavidAkaFunky!

Due to the lack of a response and inactivity, I'm closing this PR for now. If you have a chance to take a look at the previous comments, just let us know and we can reopen it!

PR #6910, opened 2025-05-16
Adding Ukrainian homepage

Andygol:
- Contributes to #6909 
- **Preview**: https://deploy-preview-6910--opentelemetry.netlify.app/uk/

Issue #6909, opened 2025-05-16
Adding Ukrainian to the OTel Documentation


Andygol:
- [x] Language info:
  - ISO 639-1 language code: `uk`
  - Language name: Ukrainian 
- [x] Locale team info:
  - [x] Locale mentor: @Andygol
  - [x] Contributors: @Andygol
- [x] Read through
      [Localization](https://opentelemetry.io/docs/contributing/localization/)
      and all other pages in the Contributing section
- [x] Localize site homepage to Ukrainian
- [ ] OTel maintainers:
  - [ ] Update `hugo.yaml`
  - [ ] Configure cSpell and other tooling support
  - [ ] Create an issue label for `lang:LANG_ID`
  - [ ] Create org-level group for `LANG_ID` approvers
  - [ ] Update components owners for `content/LANG_ID`

re: PR https://github.com/open-telemetry/opentelemetry.io/pull/5891#pullrequestreview-2534177793

kostiantynkovalchuk:
Hi, I am here to help as well.

svrnm:
working on the group etc, but that should not stop folks from contirubing!

Andygol:
@svrnm 🛎️
Is there any progress?

svrnm:
@Andygol progress on what?

Andygol:
@svrnm 
> https://github.com/open-telemetry/opentelemetry.io/issues/6909#issuecomment-3159408101
>
> working on the group etc, but that should not stop folks from contirubing!

and some checkboxes on the top are still not flagged

svrnm:
ah yes, they are all completed, I forgot to check them off and close this onboarding issue, thanks, done + closing now!

PR #7428, opened 2025-07-29
[uk] Reinstate Prettier

chalin:
- Contributes to #6909
- Followup to #6910
- Prettier seems to be playing nice with Ukrainian, so this PR reinstates it.

> [!NOTE]
> If ever we see an issue down the line, I propose that we keep Prettier enabled, but set prose wrap to preserve.

/cc @Andygol @idvoretskyi


tiffany76 (approved):
RSLGTM.

Andygol:
@chalin I'd love to ignore prettier for Ukrainian if you don't mind 🙏

chalin:
> @chalin I'd love to ignore prettier for Ukrainian if you don't mind 🙏

@Andygol We use Prettier for the entire site, in particular for locales in the capacity that it supports each locale. It supports Ukrainian, so we're enabling it. We can adjust if/when semantic issues arise down the line.

Andygol:
Prettier really helps with `code` formatting, but with **text** it brings unnecessary distraction (hanging end of lines). Modern IDE has feature to format text according to users' window width, using Prettier spoilt the final result. 👇 

Monosnap Image 2025-07-31 14-26-22

So that's why I'd love to ignore Prettier for Ukrainian.

chalin:
Got it, thanks. Let's move this discussion to 

- #7444

PR #7433, opened 2025-07-30
[experimental] enable pt/ja maintainers to merge PRs

svrnm:
With the localization teams becoming more and more self-reliable, enabling them to merge PRs on their own is the next logical step. 

Based on my research this is the easiest option to do so, all other options are much more complex and require additional setup of bots, repositories, etc.

I'd like to run this as an experiment, which means that we enable 2 localization teams (`ja` and `pt`) for a limited time (until end of september) to merge their PRs independently, after that time we will review the outcome and evaluate if we continue with it, or need to put another solution in place.

If this works as expected I suggest that we also not have maintainers per language, but one `docs-localization-maintainers` group which will step to take responsibility for the localization efforts.

@open-telemetry/docs-maintainers PTAL

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -13,4 +13,9 @@
 #
 
 # Global owners, will be the owners for everything in the repo.
-* @open-telemetry/docs-approvers
+*             @open-telemetry/docs-approvers
+
+# Localization maintainers should be able to approve & merge PRs on their own
+# This is experimental
+content/pt    @open-telemetry/docs-pt-maintainers
+content/ja    @open-telemetry/docs-ja-maintainers


tiffany76 (approved):
I assume this setup will still require PRs to have a review and approval from someone who is not the author? I know that several of the language approvers are actively localizing in addition to reviewing others' PRs. In the case where an approver is the author, it seems like a good idea to require another set of eyes.

But otherwise, I agree with the idea and the experiment. Thanks!

maryliag:
> I suggest that we also not have maintainers per language, but one docs-localization-maintainers group which will step to take responsibility for the localization efforts.

I think we should keep groups per language, because we use that to tag on PRs for review, and it would be very annoying for one language reviewers to keep getting notification of all other languages they don't know anything about

svrnm:
> > I suggest that we also not have maintainers per language, but one docs-localization-maintainers group which will step to take responsibility for the localization efforts.
> 
> I think we should keep groups per language, because we use that to tag on PRs for review, and it would be very annoying for one language reviewers to keep getting notification of all other languages they don't know anything about

The idea is that language maintainers take on the duties of what the docs maintainers are doing right now, i.e. they not only ensure that the content is correct (which of course they only can do for the languages they can read), but also that the structure is correct and all rules are applied (e.g. having the commit ID updated etc.)

So even if you are an approver for `pt` you can review a `ja` or `zh` PR after an approver has taken care of the language piece and then merge it.

PR #7445, opened 2025-07-31
[uk] Don't prosewrap uk pages

### LINKED ISSUE #7444: [uk] Adjust prosewrap config
_Originally posted by @Andygol in https://github.com/open-telemetry/opentelemetry.io/issues/7428#issuecomment-3139619937_:

> Prettier really helps with `code` formatting, but with **text** it brings unnecessary distraction (hanging end of lines). Modern IDE has feature to format text according to users' window width, using Prettier spoilt the final result. 👇 
> 
> Monosnap Image 2025-07-31 14-26-22
> 
> So that's why I'd love to ignore Prettier for Ukrainian.

The solution is to disable forced prose wrap.

chalin:
- Contributes to #6909
- Fixes #7444
- Reverts #7428
- While this addresses the stylistic preference mentioned by @Andygol in 7444, I'm ok with this config change given that not too long ago Prettier had prose-wrap related bugs for uk.

/cc @Andygol PTAL


package.json
```diff
@@ -8,7 +8,7 @@
     "_build": "npm run _hugo -- -e dev --buildDrafts --buildFuture --baseURL \"${DEPLOY_PRIME_URL:-http://localhost}\"",
     "__check:format": "./scripts/npx-helper.sh prettier",
     "_check:format:any": "npm run __check:format -- --check --ignore-path ''",
-    "_check:format:ja+zh": "npm run _check:format:nowrap -- content/ja content/zh",
+    "_check:format:ja+zh": "npm run _check:format:nowrap -- content/ja content/uk content/zh",
```
vitorvasc: Shouldn't the command also be updated?

```suggestion
    "_check:format:ja+uk+zh": "npm run _check:format:nowrap -- content/ja content/uk content/zh",
```

Most likely this isn't the only reference, there are definitely other places where `npm run _check:format:ja+zh` must be used

Andygol:
👍  LGTM

package.json
chalin: Good catch. I'll change the script name in a next round so that we don't have to keep on adding language codes :).

package.json
```diff
@@ -8,7 +8,7 @@
     "_build": "npm run _hugo -- -e dev --buildDrafts --buildFuture --baseURL \"${DEPLOY_PRIME_URL:-http://localhost}\"",
     "__check:format": "./scripts/npx-helper.sh prettier",
     "_check:format:any": "npm run __check:format -- --check --ignore-path ''",
-    "_check:format:ja+zh": "npm run _check:format:nowrap -- content/ja content/zh",
+    "_check:format:ja+uk+zh": "npm run _check:format:nowrap -- content/ja content/uk content/zh",
```
chalin: Oops, also need to fix the script name further below where the script is called. So I might as well just update the name too.

PR #7452, opened 2025-08-01
[i18n] Move alert prose from `native-libraries` partial into an include file

chalin:
- Contributes to #6460
- Factors out the prose part of the `native-libraries` partial, putting it into an include file
  - Replaces the inline alert-shortcode HTML by a call to the alert shortcode
- Adds a `func` partial used to determine the language that a page is a part of when under `docs/languages`
- There was not change to the generated site code as a result of the refactoring. Changes were introduced by using the alert shortcode instead of hard-coded (obsolete) HTML.

### Preview

- https://deploy-preview-7452--opentelemetry.netlify.app/docs/languages/cpp/library/#use-natively-instrumented-libraries - shows "Help wanted" alert
- https://deploy-preview-7452--opentelemetry.netlify.app/docs/languages/java/instrumentation/#native-instrumentation - shows an info alert

### DIFF
--- a/content/en/docs/languages/_includes/native-lib-alert.md
+++ b/content/en/docs/languages/_includes/native-lib-alert.md
@@ -1,0 +1,29 @@
+---
+---
+
+{{ if $noIntegrations }}
+
+<div class="alert alert-secondary" role="alert">
+<div class="h4 alert-title">Help wanted!</div>
+
+As of today, we don't know about any {{ $name }} library that has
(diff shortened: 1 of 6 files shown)


tiffany76 (approved):
RSLGTM.

chalin:
The Java page changes are editorial only, dropping a shortcode argument that is no longer needed, so I'll merge now.

Issue #5227, opened 2024-09-21
[es] Spanish folder localization docs/kubernetes


krol3:
This issue is to track the translation pages for the folder `docs/kubernetes`

Below, you can check the list of the pages we recommend translate.

- [x] opentelemetry.io/content/en/docs/kubernetes/_index.md - https://github.com/open-telemetry/opentelemetry.io/pull/5055
- [ ] opentelemetry.io/content/en/docs/kubernetes/collector/_index.md
- [ ] opentelemetry.io/content/en/docs/kubernetes/collector/components.md
- [ ] opentelemetry.io/content/en/docs/kubernetes/getting-started.md

opentelemetry.io/content/en/docs/kubernetes/helm/_index.md
opentelemetry.io/content/en/docs/kubernetes/helm/collector.md
opentelemetry.io/content/en/docs/kubernetes/helm/demo.md
opentelemetry.io/content/en/docs/kubernetes/helm/operator.md
opentelemetry.io/content/en/docs/kubernetes/operator/_index.md
opentelemetry.io/content/en/docs/kubernetes/operator/automatic.md
opentelemetry.io/content/en/docs/kubernetes/operator/target-allocator.md
opentelemetry.io/content/en/docs/kubernetes/operator/troubleshooting/_index.md
opentelemetry.io/content/en/docs/kubernetes/operator/troubleshooting/automatic.md
opentelemetry.io/content/en/docs/kubernetes/operator/troubleshooting/prometheus-alerts-runbooks.md
opentelemetry.io/content/en/docs/kubernetes/operator/troubleshooting/target-allocator.md

mlpaucara:
Voy a traducir la siguiente página opentelemetry.io/content/**en**/docs/kubernetes/collector/_index.md que significa crear la nueva página en **opentelemetry.io/content/es/docs/kubernetes/collector/_index.md**

mlpaucara:
@theletterf I was using an incorrect global settings for user.name and user.email in my closed PR

falconcr:
Voy a traducir la siguiente pagina opentelemetry.io/content/en/docs/kubernetes/collector/components.md que significa crear la nueva página en **opentelemetry.io/content/es/docs/kubernetes/collector/components.md**

theletterf:
Sure! Pick one and open a PR referencing this issue

Victorsesan:
@krol3 @theletterf I have translated pages in issue 4,5 & 6 PR pending for review thanks

[...]

PR #7477, opened 2025-08-05
feat: [es] platforms folder

krol3:
Creating the platforms folder, and putting the kubernetes folder translations in the correct new folder structure.
related to https://github.com/open-telemetry/opentelemetry.io/issues/5227

### DIFF
--- a/content/es/docs/platforms/_index.md
+++ b/content/es/docs/platforms/_index.md
@@ -1,0 +1,13 @@
+---
+title: Plataformas y entornos
+linkTitle: Plataformas
+description:
+  Plataformas y entornos de hardware y software que admiten OpenTelemetry
+weight: 260
+default_lang_commit: f9a0439ac56dba1515283e1a1cb6d6a90634a20f
+---
+
+OpenTelemetry está disponible para una variedad de plataformas y entornos, lo
+que garantiza una observabilidad sin interrupciones en sistemas híbridos. Esta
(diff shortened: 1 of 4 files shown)


krol3:
cc: @vitorvasc please review it :)

vitorvasc:
> cc: @vitorvasc please review it :)

Hi @krol3 😄 

The page `content/es/docs/platforms/kubernetes/collector/components.md` has some links pointing to the content in the Helm directory using the relative paths, example: `../../helm/collector/` 

Since the Helm directory hasn't been localized in Spanish yet, this check will fail. I'm not sure if that's the expected behavior, let me check with the @open-telemetry/docs-approvers to conffirm how relative paths should be handled and whether we should fix this in the English pages before moving forward with this PR.

That said, if you already have translations for the files under the Helm directory in another branch, adding them could solve this issue in the meantime.

vitorvasc:
Also, I noticed that the files in this PR have an invalid value in the `drifted_from_default` field in the front matter: `drifted_from_default: file not found`

vitorvasc:
@krol3 good news: since #7525 has already been merged, the URLs should no longer block this PR.

The only update needed is on the `update the content/es/docs/platforms/kubernetes/collector/components.md` file to use the actual absolute path for the Helm directory references, instead of the relative path.

tiffany76:
@theletterf - Is this good from an `es` perspective?

PR #7484, opened 2025-08-06
Add cardinality definition to Portuguese glossary

jpkrohling:
This PR adds a definition for 'cardinalidade' to the Portuguese glossary.

The definition explains:
- What cardinality means (número de valores únicos para atributos)
- The impact of high vs low cardinality on telemetry backends
- Examples of high cardinality (user_id) vs low cardinality (status_code) attributes

This is part of a series of improvements to help users understand key OpenTelemetry concepts.

### DIFF
--- a/content/pt/docs/concepts/glossary.md
+++ b/content/pt/docs/concepts/glossary.md
@@ -104,7 +104,7 @@ rastreamento. Consulte a [especificação de campos][field].
 
 O número de valores únicos para um determinado [Atributo](#attribute) ou
 conjunto de atributos. Alta cardinalidade significa muitos valores únicos, o que
-pode impactar o desempenho e os requisitos de armazenamento dos backends de
+pode impactar o desempenho e os requisitos de armazenamento dos _backends_ de
 telemetria. Por exemplo, um atributo `user_id` teria alta cardinalidade,
 enquanto um atributo `status_code` com valores como "200", "404", "500" teria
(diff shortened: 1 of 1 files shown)


vitorvasc:
@jpkrohling would you mind adding a `# patched` after the hash on `default_lang_commit`?

We (@open-telemetry/docs-approvers) were just discussing cases like this, when targeted content needs to be added to a localized page that's drifted, but we don't want to update the rest of the content. 

We're also tracking this in #7485.

jpkrohling:
@vitorvasc, I added it to the last commit. Could you please double-check?

content/pt/docs/concepts/glossary.md
```diff
@@ -100,6 +100,15 @@ incluindo [Atributos](#attribute) e [Recursos](#resource). Outros campos também
 podem ser considerados `Metadados`, incluindo severidade e informações de
 rastreamento. Consulte a [especificação de campos][field].
 
+### Cardinalidade {#cardinality}
+
+O número de valores únicos para um determinado [Atributo](#attribute) ou
+conjunto de atributos. Alta cardinalidade significa muitos valores únicos, o que
+pode impactar o desempenho e os requisitos de armazenamento dos backends de
```
vitorvasc: Oops, faltou isso 😬

```suggestion
pode impactar o desempenho e os requisitos de armazenamento dos _backends_ de
```

Issue #4876, opened 2024-07-24
[i18n] Consistently implement the handling of the OTel specs across localizations, and document the solution


chalin:
Propagate the `ja` solution to other localizations, and document it under https://opentelemetry.io/docs/contributing/localization.

## Tasks

- [x] Prep: #6755 
- [ ] ...

ymotongpoo:
reviving this topic. because the spec docs are located in the different repository (https://github.com/open-telemetry/opentelemetry-specification), how do we localize the spec documents? do you already have any ideas for the procedure?

chalin:
@ymotongpoo: this issue is about enabling all localizations to refer to the English specification via absolute and relative paths to `/docs/specs`. It is not about localization of the spec itself, which is beyond the scope of this repo.

vitorvasc:
@chalin I came across this issue while updating some references in #7525.

We have a note at [Localization > Links](https://opentelemetry.io/docs/contributing/localization/#links) that mentions absolute paths, but I'm not sure if that's the documentation originally intended to be created here. Should we add a separate heading specifically mentioning the usage of absolute paths?

Also, there might be a few pages still using relative paths, should we consider updating those pages as well?

chalin:
> [@chalin](https://github.com/chalin) I came across this issue while updating some references in [#7525](https://github.com/open-telemetry/opentelemetry.io/pull/7525).

@vitorvasc - Thanks for pointing out #7525, but I'm unsure why you say that 7525 illustrates an instance of the problem reported in this issue.

vitorvasc:
@chalin sorry - I came across this while looking into other issues or PRs related to the use of relative paths in general, not specifically for the specs 😅

chalin:
Thanks for clarifying @vitorvasc, I wanted to ensure that I wasn't missing anything.

PR #7525, opened 2025-08-11
[i18n] Update files under /docs/platforms/kubernetes to use absolute path

vitorvasc:
- Contributes to #4876
- Unblocks #7477


tiffany76 (approved):

chalin:
Got to here from @vitorvasc's comment in https://github.com/open-telemetry/opentelemetry.io/issues/4876#issuecomment-3174813628. On first inspection, I don't think that the links should have been updated, but I'll take a closer look as soon as I can.

PR #7537, opened 2025-08-12
Fix link for collector internal-telemetry config

jackgopack4:
Config file moved for service otelconf telemetry in collector


content/en/docs/collector/internal-telemetry.md
```diff
@@ -156,7 +156,7 @@ OpenTelemetry Configuration schema
 
 Log output is found in `stderr`. You can configure logs in the config
 `service::telemetry::logs`. The
-[configuration options](https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/telemetry/config.go)
+[configuration options](https://github.com/open-telemetry/opentelemetry-collector/blob/ed09fb912b2ff7cb3998cdb0b54027471ed65eaa/service%2Ftelemetry%2Fotelconftelemetry%2Fconfig.go)
```
vitorvasc: Hi @jackgopack4! Thanks for bringing this update 😄 

Quick question that came up, is this URL correct? I noticed it points to a specific commit, however this file can be found using the same path on the main branch as well: https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/telemetry/otelconftelemetry/config.go

```suggestion
[configuration options](https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/telemetry/otelconftelemetry/config.go)
```

vitorvasc:
Hi @jackgopack4! Just following up, did you get a chance to check out my previous comment? 😄

content/en/docs/collector/internal-telemetry.md
jackgopack4: Hello, main link should be fine. I used commit hash in case it changes in the future, but maybe we should have it break if the file moves or changes so that someone will submit an update PR to fix it

jackgopack4:
Yes, sorry for the delay. Committed your changes.

vitorvasc (approved):
@jackgopack4 thanks!

LGTM!

PR #7549, opened 2025-08-13
Update Node Getting Started to use tsx over ts-node

Ecksters:
Just swapping out `ts-node` for `tsx` in the examples, `ts-node` runs into errors around the ESM syntax and the `.ts` extension, making the guide simply not followable. I'm mostly just following the recommendation from here: https://stackoverflow.com/a/76343394

Express instrumentation still is not working as expected, ideally the "Example outputs" would also be updated once that's resolved, as they've changed quite a bit since older versions.

### DIFF
--- a/content/en/docs/languages/js/getting-started/nodejs.md
+++ b/content/en/docs/languages/js/getting-started/nodejs.md
@@ -48,7 +48,6 @@ Next, install Express dependencies.
 
 ```sh
 npm install typescript \
-  ts-node \
   @types/node \
   express \
   @types/express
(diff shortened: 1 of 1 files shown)


tiffany76:
Thanks for submitting a fix, @Ecksters! I'm sure others will benefit from your efforts. ✨ 

@open-telemetry/javascript-approvers, PTAL!

maryliag:
Thank you for opening the issue and for your contribution @Ecksters ! Keep them coming 😄 
I will close this PR, because the one Trent created is more complete #7606

PR #7552, opened 2025-08-14
[zh] Update languages/_includes/ files

windsonsea:
Update files in:

```
content/zh/docs/languages/_includes/
```


svrnm:
@open-telemetry/docs-zh-approvers PTAL!

PR #7566, opened 2025-08-14
Comment out cross-ref link to OBI K8s helm chart setup

tiffany76:
We have hidden the "deploy OBI in K8s with Helm chart" page from the left nav until the Helm chart work is complete. But a cross-reference link on the manual Kubernetes setup page made the the Helm page still accessible. This PR comments out the cross-ref link.

Helm chart PR for reference: https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1704


marctc (approved):

tiffany76:
Actually looks like the page is appearing in other places because of Hugo automation. We might have to change how the page is hidden. I'll do that in a follow-up PR.

PR #7581, opened 2025-08-15
Add installation instructions for Windows

vjofusha:
Added Windows installation instructions for MSI and .tar.gz packages.


tiffany76:
Thanks, @vjofusha!

@open-telemetry/collector-approvers, PTAL.

content/en/docs/collector/installation.md
```diff
@@ -256,9 +256,22 @@ unpacking.
 
 ## Windows
 
-Windows [releases][] are packaged as gzipped tarballs (`.tar.gz`). Every
-Collector release includes an `otelcol.exe` executable that you can run after
-unpacking.
+Windows [releases][] are available as MSI installers and gzipped tarballs
+(`.tar.gz`). Every Collector release includes an `otelcol.exe` executable that
```
pjanotti: I think we want to make a finer point here: the MSI installs the collector as a Windows service, named according to the distribution itself. The point of using the MSI is to get it correctly installed as a Windows service with proper Event Log source registration.

pjanotti (commented):
Thanks @vjofusha - good addition.

content/en/docs/collector/installation.md
pjanotti: 🤔 perhaps here to keep things simple we could just mention the Windows service display name "OpenTelemetry Collector", the exact service name used for command-line and so on it is the one that varies according to the distribution.

pjanotti: nit: on the tarball the name of the executable is the name of the distro, i.e.: it will be `otelcol.exe` in `otelcol_0.132.4_windows_amd64.tar.gz` and `otelcol-contrib.exe` in `otelcol-contrib_0.132.4_windows_amd64.tar.gz`. Instead of exposing this complexity I suggest to not name the executable here, people will see it when they extract the contents of the tarball.

vjofusha: Makes sense. I’ll update the documentation with these suggestions and re-test the commands on a Windows machine. Thank you!

pjanotti (approved):

tiffany76:
Looking good! @open-telemetry/collector-approvers, could we get a stamp? Thanks!

PR #7598, opened 2025-08-19
[Infra] Add fix:refcache to auto-update-community-members.yml workflow

vitorvasc:
## Description

Closes #6966.

Adds the `fix:refcache` command to the execution of the `auto-update-community-members.yml` workflow, since we often need to manually trigger refcache update before merging the PRs.

FYI @open-telemetry/docs-maintainers

Issue #6966, opened 2025-05-23
Add refcache-update to update-community-members and other workflows


chalin:
_Originally posted by @chalin in https://github.com/open-telemetry/opentelemetry.io/issues/6964#issuecomment-2904217815_:

> We often have to update the refcache. @svrnm @theletterf @tiffany76 - might one of you be available to update the workflow in a manner similar to what @trask did in:
>
> - #6963

It might be a useful addition to a few workflows. WDYT?

theletterf:
Agreed! Just a matter of lifting and shifting that code to the other workflows?

PR #7599, opened 2025-08-19
[Infra] add fix:refcache to workflows: auto-update-versions.yml, auto-update-registry.yml

vitorvasc:
Contributes to #6966.

Complements the #7598 and adds the `fix:refcache` command to the workflows `auto-update-versions.yml` and `auto-update-registry.yml`

### DIFF
--- a/.github/workflows/scripts/update-registry-versions.sh
+++ b/.github/workflows/scripts/update-registry-versions.sh
@@ -127,6 +127,10 @@ if [[ -n $(git status --porcelain) ]]; then
 
     $GIT checkout -b "$branch"
     $GIT commit -a -m "$message"
+
+    $NPM run fix:refcache
+    $GIT commit -a -m "Fix refcache"
+
     $GIT push --set-upstream origin "$branch"
 
     body_file=$(mktemp)
--- a/scripts/auto-update/version-in-file.sh
+++ b/scripts/auto-update/version-in-file.sh
@@ -4,6 +4,7 @@
 
 GH=gh
 GIT=git
+NPM=npm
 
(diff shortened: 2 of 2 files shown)


chalin (changes requested):
It's great that you're working on this! @vitorvasc 🙌🏻 

Have you tested these changes locally?

I don't know enough about `.github/workflows/scripts/update-registry-versions.sh`, so I'll let @svrnm review the changes made to that file.

Regarding `version-in-file.sh`: take a look at `scripts/auto-update/all-versions.sh`, and you'll see that the `version-in-file.sh` script can be run many times. We'll want to fix the refcache only once at the end when all versions will have been checked, and only if there have been some version changes.

Thanks!

vitorvasc:
> Have you tested these changes locally?

I tried to test using my fork but it didn't work, so I only tested by running the bash scripts directly. The `update-registry-versions.sh` completed successfully in dry-run mode. 

The `version-in-file.sh` script didn't show any updates, but I realize I might have executed it incorrectly, since it should have been triggered by the `all-versions.sh`.

I've committed the updates so that `fix:refcache` runs once. Now I've got a better picture of how the whole update process works 😄

vitorvasc:
@chalin it's been a while since I've seen refcache errors in these workflows. 

I think this migh've been fixed by changes introduced in other issues / PRs. I'd say we can close it, WDYT?

svrnm:
> @chalin it's been a while since I've seen refcache errors in these workflows.
> 
> I think this migh've been fixed by changes introduced in other issues / PRs. I'd say we can close it, WDYT?

I am in favor of closing before keeping it around, closing it now and if anything is left to do open a new one!

PR #7603, opened 2025-08-20
Fix/collector diagram

### LINKED ISSUE #5356: Add "connectors" to the collector visualization
The image demonstrating the inner workings of the OpenTelemetry collector (https://opentelemetry.io/docs/collector/img/otel-collector.svg) is not showing connectors, which are now part of the collector. Having them added would be great. In general the image should remain as it is.

leandrocaracciolo:
This PR adds the processing loop to the collector diagram. The change was made to improve the accuracy and clarity of the diagram.

Fixes #5356. 

---
Preview: https://deploy-preview-7603--opentelemetry.netlify.app/docs/collector/building/connector/#the-old-architecture


leandrocaracciolo:
This is my suggestion for the new image. Do you think it works?

tiffany76:
Hi @leandrocaracciolo, thanks for taking this on! We should have the @open-telemetry/collector-approvers take a look (and @jpkrohling, if he has time).

Here are some thoughts based on an initial review.
- I think the orange line might not be necessary. I think if we can adjust the green line to have noticeable arrows that point in the direction of traces connector >>>> metrics connector, that might solve the (pared-down) problem as described in the issue. We'll ask the Collector folks to weigh in.
- Also, I noticed that the padding around the image and the size/zoom is a little off (as shown in the screenshot below). I wouldn't normally be so nitpicky, but with the two images (old and new) right next to each other, the difference is a bit distracting.
- Finally, I'm wondering why the second image (from the docs homepage) is included in this PR. Did you make changes to it?

Screenshot 2025-08-20 at 3 54 36 PM

jpkrohling:
> I think the orange line might not be necessary. I think if we can adjust the green line to have noticeable arrows that point in the direction of traces connector >>>> metrics connector, that might solve the (pared-down) problem as described in the issue. We'll ask the Collector folks to weigh in.

I agree!

leandrocaracciolo:
Hi @tiffany76 ,

Thanks for the review and the great feedback! I have just updated the PR to address all your comments.

    Regarding the image sizing, you are right. The diagram was increased in size to accommodate the orange line previously, but I have now adjusted the width and height to match the original image.

    The second image was indeed included by mistake from a previous task. I apologize for the confusion; it has now been removed from this PR.

    I have also removed the orange line and added arrows to the green line as you and @jpkrohling suggested.

The PR is ready for another review. Thanks again for your time and guidance!

tiffany76:
Hi @leandrocaracciolo! Thanks for updating the PR. I think we're almost there. Just two more changes that I can see.

First, it looks like you inadvertently deleted the "old architecture" diagram. It shows up as a broken image in the preview (see below). Could you add it back? Thanks!

Second, the arrows on the green line should only point in one direction. The arrow should point _toward_ the metrics pipeline (which is on the lower part of the diagram. Hopefully this screenshot helps make it clearer!

Screenshot 2025-08-26 at 1 44 43 PM

leandrocaracciolo:
Hi @tiffany76 ,

Just a quick update to let you know that I've pushed the final requested changes.

I've restored the otel-collector-before-connector.svg diagram and corrected the direction of the arrow on the otel-collector-after-connector.svg diagram.

The PR should now be ready for re-review. 

Thank you again for all the guidance!

jpkrohling:
I like it!

PR #7606, opened 2025-08-20
Fix JS/TS "getting started" code to work

trentm:
- This switches from ts-node to tsx to do the on-the-fly transpilation.
  'npx tsx-node app.ts' fails with the default "tsconfig.json" created
  via 'tsc --init' now. tsx is better maintained and simplifies this
  example somewhat.
- Switch from using `--require ...` to `--import ...` for pre-loading
  the OpenTelemetry bootstrap code. `--import` is supported now in all
  supported Node.js versions and it helps avoid some breakage and
  confusion with Node.js 22 and 24 that implement type-stripping
  of .ts files. With type-stripping the .ts file (using `import ...`
  ESM syntax) is run directly. `--require` doesn't support loading
  ESM, but `--import` supports loading ESM (and CJS as a fallback).

Fixes: https://github.com/open-telemetry/opentelemetry.io/issues/7547

PR #7607, opened 2025-08-20
ci: improve the error message when the 'REFCACHE updates?' check fails

trentm:
In this PR I hit a "Links / REFCACHE updates?" check failure in CI:
https://github.com/open-telemetry/opentelemetry.io/pull/7606#issuecomment-3208176252

The error message in the GitHub Actions log was:

```
> _diff:check
> git diff --name-only --exit-code

static/refcache.json

ERROR: the files above have changed. Locally rerun `npm run test-and-fix` and commit changes
```

which is misleading, because (I'm told) the correct action to fix is `npm run fix:refcache`.  This attempts to improve that error message.

### DIFF
--- a/.github/workflows/check-links.yml
+++ b/.github/workflows/check-links.yml
@@ -115,7 +115,10 @@ jobs:
             exit 1
           fi
       - name: Does the refcache need updating?
-        run: npm run diff:fail
+        run: |
+          echo "If the diff check below fails, then update static/refcache.json by"
+          echo "running 'npm run fix:refcache' locally and commiting the changes."
+          npm run _diff:fail
 
   check-build-log-for-issues:
     name: WARNINGS in build log?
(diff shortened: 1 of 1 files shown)


.github/workflows/check-links.yml
```diff
@@ -115,7 +115,10 @@ jobs:
             exit 1
           fi
       - name: Does the refcache need updating?
-        run: npm run diff:fail
+        run: |
+          echo "If the diff check below fails, then update static/refcache.json by"
+          echo "running 'npm run fix:refcache' locally and commiting the changes."
```
maryliag:
```suggestion
          echo "running 'npm run fix:refcache' locally and committing the changes"
          echo "or adding the comment '/fix:refcache' to your PR in GitHub."
```

PR #7624, opened 2025-08-24
Update configuration.md under Connector section

soodr:
Corrected emitting/consuming components


svrnm:
as confusing as it sounds that it "consumes as an exporter" and "emits as a receiver", the current wording is correct, since the pipe goes as follows:

```
pipe1 --> connector --> pipe2
```

So the connector consumes data from pipe1 (as exporter) and then emits data as a receiver into pipe2

vitorvasc:
Hi @soodr! I appreciate you taking the time to submit this contribution.

Since we haven't heard back after @svrnm's comment, and given that the current wording is correct, I'll go ahead and close this PR for now.

Thanks!

PR #7602, opened 2025-08-20
feat(blog): add 'How to Name Your Span Attributes' blog post

jpkrohling:
This is a continuation to the "How to Name Your Spans" blog post. 

Signed-off-by: Juraci Paixão Kröhling <juraci@kroehling.de>

PR #7630, opened 2025-08-25
[es] Translate 'How to Name Your Span Attributes' blog post

vitorvasc:
## Description

It adds the Spanish translation for the blog post 'How to Name your Span Attributes': #7602

/cc @open-telemetry/docs-es-approvers


vitorvasc:
> There seem to be refcache issues still, could you check?

@theletterf Oh, I see this is happening because the canonical URL doesn't exist yet. Also, #7602 hasn't been published, so the URL will probably go live at the same time. Well, I'll mark this PR as a draft until everything is published 😅

Issue #7623, opened 2025-08-24
[Page feedback]: Modern javascript code examples


danielpza:
### URL

https://opentelemetry.io/docs/languages/js/instrumentation/

### Description

Some examples of how to modernize javascript code examples:

- replace commonjs (require/module.exports) with esm (import/export)

```js
// commonjs
const express = require('express');
const { rollTheDice } = require('./dice.js');

// esm
import express from 'express';
import { rollTheDice } from './dice.js'
```

- modern Node.js support running typescript directly, `ts-node` might not be needed. Furthermore it seems `tsx` is suggested over `ts-node` for running typescript code directly (aside from node type stripping). See https://nodejs.org/api/typescript.html

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

svrnm:
@open-telemetry/javascript-approvers PTAL

trentm:
Indeed we did just switch https://opentelemetry.io/docs/languages/js/getting-started/nodejs/ over to using `tsx` just recently in https://github.com/open-telemetry/opentelemetry.io/pull/7606

We can do the same for the `/docs/languages/js/instrumentation/` page.

trentm:
> * replace commonjs (require/module.exports) with esm (import/export)

We might be able to do so in this case, but note that there are limitations with OTel JS instrumenting ESM code in some cases. See https://github.com/open-telemetry/opentelemetry-js/blob/main/doc/esm-support.md and https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1942

trentm:
> > * replace commonjs (require/module.exports) with esm (import/export)
> 
> We might be able to do so in this case

Another thing that is a bit of a pain for these docs if wanting to use ESM in .js files is that `npm init ...` only supports the `--init-type=module` switch in npm v11 and later. It isn't until Node.js v24 that npm v11 is the default.  If a user is using Node.js v22 or lower and run a suggested `npm init -y --init-type=module`, the result is the `--init-type=...` argument is silently ignored, package.json defaults to commonjs, and .js files are interpreted by `node` as a commonjs.

So, if a user is following along with a version that uses ESM, a Node.js 18 user will see a crash:

[...]

PR #7636, opened 2025-08-26
JS instrumentation/propagation docs: switch to tsx for on-the-fly TypeScript compilation

trentm:
Similar to https://github.com/open-telemetry/opentelemetry.io/pull/7606
this updates the remaining parts of the JavaScript tracing docs to use
'tsx' rather than the no-longer-maintained 'ts-node' for TypeScript
examples.

As well (as in #7606) this updates to '--import' usage over '--require'
for preloading OTel bootstrap code. This is in prep for future ESM
support, and helps with some confusion around automatic type-stripping
in recent versions of Node.js.

Refs: https://github.com/open-telemetry/opentelemetry.io/issues/7623

### DIFF
--- a/content/en/docs/languages/js/instrumentation.md
+++ b/content/en/docs/languages/js/instrumentation.md
@@ -46,11 +46,8 @@ Next, install Express dependencies.
 {{< tabpane text=true >}} {{% tab TypeScript %}}
 
 ```sh
-npm install typescript \
-  ts-node \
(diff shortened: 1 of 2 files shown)


theletterf (approved):
LGTM from the code side.

@open-telemetry/javascript-approvers PTAL

maryliag (approved):

PR #7638, opened 2025-08-26
Update semantic-conventions version to v1.37.0

opentelemetrybot:
Update semantic-conventions version to `v1.37.0`.

See https://github.com/open-telemetry/semantic-conventions/releases/tag/v1.37.0.


lmolkova:
it seems internal links to internal json files are a problem - the link to json is rewritten, but json file itself is not published. It makes sense to use absolute url to github for these. 
 
Anything special we can do on the semconv side going forward in cases like this?

chalin:
@lmolkova - yes, among other things, the JSON files are causing build failures. Let's continue the discussion via:

- https://github.com/open-telemetry/semantic-conventions/issues/2690

chalin:
@lmolkova @trask et all: I've patched this PR so that we can get the spec published. All checks are passing now. Please reapprove this PR to signify that you agree with the temporary patches.

IMPORTANT: After this PR is merged (if you agree to do so), we'll need to make further changes in this repo and the semconv repo. I'll follow up with the necessary changes.

For example, the patches to this PR have as a consequence that the `.ipynb` file is served raw. That isn't very useful. I'll followup with a change that will link into the semconv repo. GitHub provides a visual preview of such files, see: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/non-normative/models.ipynb

scripts/content-modules/adjust-pages.pl
```diff
@@ -95,12 +95,22 @@ ($$$)
...
-#   return unless $ARGV =~ /^tmp\/otel\/specification\//
-#     && applyPatchOrPrintMsgIf('2025-05-13-attribute-registry-refactoring', 'spec', '1.45.0-dev');
+sub patchSpec_because_of_SemConv_MetricReqLevelHashDNE() {
+  return unless $ARGV =~ /^tmp\/semconv\/docs\//
+    && applyPatchOrPrintMsgIf('2025-08-28-metric-request-level-hash-dne', 'semconv', '1.37.0-dev');
 
-#   s|/attributes-registry/|/registry/attributes/|g;
-# }
+  # See https://github.com/open-telemetry/semantic-conventions/issues/2690#issuecomment-3235079573
+  s|/docs/general/metrics.md#metric-requirement-levels|/docs/general/attribute-requirement-level.md#recommended|g;
```
lmolkova: this one should be 
```suggestion
  s|/docs/general/metrics.md#metric-requirement-levels|/docs/general/metric-requirement-level.md#recommended|g;
```

https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/metric-requirement-level.md#recommended

lmolkova:
> For example, the patches to this PR have as a consequence that the .ipynb file is served raw. That isn't very useful. I'll followup with a change that will link into the semconv repo. GitHub provides a visual preview of such files, see: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/non-normative/models.ipynb

I think it makes total sense not to publish ipynb on the otel.io and instead link to the github repo. It's non-normative. JSON files are normative, but I also don't see a problem if they are not on the website, but we preserve links to them on github.
I don't mind them being published either, but I can see how having one or both of them on the website is problematic. 

Thanks a lot for looking into this!

chalin:
Thanks @lmolkova. Let's track followup changes via:

- #7666

chalin (approved):

chalin:
@lmolkova approved for semconv, and all checks are passing. Merging.

PR #7652, opened 2025-08-27
add sveltekit to registry

svrnm:
Via https://svelte.dev/blog/sveltekit-integrated-observability


data/registry/instrumentation-js-sveltekit.yml
```diff
@@ -0,0 +1,19 @@
+# cSpell:ignore nestjs vercel
```
svrnm:
```suggestion
# cSpell:ignore sveltekit vercel
```

PR #7694, opened 2025-09-03
Update go getting-started examples in english

dmathieu:
This upgrades the Go getting started code samples, as named return variables are not a recommended practice, and our linter prevents their use.

See https://github.com/open-telemetry/opentelemetry-go-contrib/pull/7852

### DIFF
--- a/content/en/docs/languages/go/getting-started.md
+++ b/content/en/docs/languages/go/getting-started.md
@@ -157,13 +157,14 @@ import (
 
 // setupOTelSDK bootstraps the OpenTelemetry pipeline.
 // If it does not return an error, make sure to call shutdown for proper cleanup.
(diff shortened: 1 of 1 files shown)


dmathieu:
I don't know how you folks usually handle updates like this for other languages.
Would you prefer I do every language in here? Should I not care about other languages?

pellared (approved):

vitorvasc:
> I don't know how you folks usually handle updates like this for other languages. Would you prefer I do every language in here? Should I not care about other languages?

@dmathieu for cases like this, it's fine to use the same PR and apply the changes to other locales too, it counts as a [purely editorial change](https://opentelemetry.io/docs/contributing/localization/#patch-locale-links). Instead of updating the `default_lang_commit`, we just add a `# patched` comment after the hash.

MrAlias (approved):

PR #7707, opened 2025-09-05
Fix depencency order for erlang apps

hoyon:
The opentelmetry_exporter dependency must be specified before the other opentelemetry dependencies to ensure clean startup.

With the current order, this warning is produced:

```
11:44:13.752 [debug] OTLP exporter failed to initialize: exception throw: {application_either_not_started_or_not_ready,
                     tls_certificate_check}
```

The opentelemtry-erlang repo also specifies this ordering in the [README](https://github.com/open-telemetry/opentelemetry-erlang?tab=readme-ov-file#design).


svrnm:
Thank you @hoyon 

@open-telemetry/erlang-approvers PTAL!

tsloughter:
@hoyon ah, is this the case when booting with `iex`? For a release the order should be specified in `releases`.

PR #7712, opened 2025-09-05
Add more detail to Java agent Tomcat instructions

trask:
Add a couple of details that some folks have found useful.


svrnm:
@open-telemetry/java-approvers PTAL!

PR #7781, opened 2025-09-16
fix: remove trailing backslash in helm install command

voldikss:
The example command for installing opentelemetry-collector with helm contained a trailing backslash on the last line, which is unnecessary and may confuse users. This commit removes it for correctness.


theletterf:
@voldikss Could you sign the CLA?

voldikss:
@theletterf signed

PR #7792, opened 2025-09-17
Add Oracle Database Metrics Exporter to the registry

anders-swanson:
Adds the Oracle Database Metrics Exporter (a go collector) to the registry.


theletterf:
@anders-swanson Thanks! Could you please add an email and URL?

data/registry/collector-exporter-oracle-database.yml
```diff
@@ -0,0 +1,18 @@
+title: Oracle Database Metrics Exporter
+registryType: exporter
+language: go
```
svrnm: this would mean that it is an exporter that go developers can use, but if I understand it correctly this is a standalone tool, right?

svrnm (changes requested):

data/registry/collector-exporter-oracle-database.yml
anders-swanson: Yes, it's a standalone tool. `collector` is the correct value?

svrnm:
```
registryType: utilities
language: go
```

in that case. Also rename the file to: `tools-oracle-database-metrics-exporter`

anders-swanson: Sorry, let me rephrase. The collector is a go service that runs standalone. The collector connects to one or more database servers, and serves OTEL metrics over HTTP.

anders-swanson:
Any further changes requested? Thanks :)

data/registry/collector-exporter-oracle-database.yml
svrnm: but it's not an OpenTelemetry collector, right? Only in that case the registryType can be "collector"

anders-swanson: Is there an explanation of the registry types? I am getting confused on what the appropriate value should be here.

svrnm: > Is there an explanation of the registry types? I am getting confused on what the appropriate value should be here.

Unfortunately no, but the rule of thumb is:
- language: whatever language this application is written in
- registryType, one of the following:
  - "application integration" -> anything that is natively integrated, e.g. if Oracle DB would come with otel ootb, see https://opentelemetry.io/ecosystem/integrations/
  - "core" -> otel core components, never applicable to non-project components
  - "exporter": either an exporter component of the collector, or an exporter within a language-specific SDK, not applicable for a 3rd party component that is exporting telemetry
  - "extension": collector or SDK extensions
  - "instrumentation": instrumentation libraries or native instrumentations for libraries
  - "log-bridge": implementation of log-bridges for languages
  - "processor": collecor component "processor"
  - "provider": collector compoent "provider"
  - "receiver": collector component "receiver", not applicable for 3rd party components that are receiving otel telemetry
  - "resource-detector": Resource Detectors for language specific SDKs
  - "utilities": any other tool that people can use to work with OpenTelemetry, this is commonly used for components like what you have here.

To be more specific on your use case (exporting metrics or other telemetry from an oracle database):
- if the database would be able to export telemetry without any additional tool, this would fall under "application integration", e.g. https://dev.mysql.com/doc/refman/9.4/en/telemetry.html
- if the OpenTelemetry Collector would receive/scrape/... the telemetry from the database it would be a "receiver", e.g. https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/mysqlreceiver
- if there is a standalone binary that pulls the metrics from the db and then pushes them out to an OTLP compatible backend, this falls under "utilities" independent of how you name it.

anders-swanson: OK, I'll list it as a utility 👍 It collects metrics from Oracle Database(s), and those metrics are scraped by some other service.

data/registry/tools-oracle-database-metrics-exporter.yml
svrnm, line 1: you dropped the file extension

anders-swanson: whoops, good catch. Fixed it

svrnm (approved):

PR #7803, opened 2025-09-19
add missing collector components to registry

svrnm:
Ran the registry scanner after a long while once again, these are some registry components that were missing.


data/registry/collector-receiver-netflow.yml
```diff
@@ -0,0 +1,24 @@
+# cSpell:ignore netflowreceiver netflow
```
svrnm:
```suggestion
# cSpell:ignore goflow ipfix netflowreceiver netflow
```

Issue #7844, opened 2025-09-23
[CI] Workflow update-semconv-integration-branch.yml is failing


chalin:
The following workflow has been failing for quite some time: https://github.com/open-telemetry/opentelemetry.io/actions/workflows/update-semconv-integration-branch.yml

chalin:
There seemed to be a few issues. Here's what I did:

- [x] Deleted old semconv-integration branches
- [x] Identified that the workflow will fail if there are link-check failures, which can happen during development. Link-check failures should allowed, and the integration branch still updated. The PR checks will flag link-check failures
  - [x] PR #7848
  - [x] Once merged, rerun https://github.com/open-telemetry/opentelemetry.io/actions/workflows/update-semconv-integration-branch.yml

chalin:
Workflow is running again. I have some concerns that it might be sensitive to refcache merge conflicts, but I'll close this issue for now, and we can revisit later if/when needed.

Closed by:

- #7848

PR #7848, opened 2025-09-24
[CI] Semconv integration workflow: allow link-checking to fail

chalin:
- Contributes to #7844
- Renamed the `Fix refcache` step to `Update refcache`, since we're actually updating it with new entries
- Adjusted the `Update refcache` step so that it ignores link-check failures so that the refcache can still be updated with any new (non-404) entries, and so as to not fail the job


tiffany76 (approved):
RSLGTM.

Issue #7865, opened 2025-09-25
[CI] Improve link-checking performance


chalin:
With the many locales (all having a full set of doc pages due to fallback pages), link checking is now taking about 9 minutes on GitHub. That's too slow. Let's see if we can split the checking across jobs via a matrix dimension.

My current plan is to run a test per locale.

### Tasks

- [x] Split build from link checking -- currently they are in the same job - #7866
- [x] Update branch protection rules to account for job name changes, dropping `BUILD and CHECK LINKS`, adding `BUILD` and `CHECK LINKS`:
  - [x] `main` branch protection rule updated
  - [x] Do any other rules need to be changed? [Doesn't seem like it]
- [x] #7879 
  - [x] Update contrib pages as necessary
- [x] Parallelize link checking, probably by locale group:
  - [x] #7881
  - [x] #7885
- [x] https://github.com/open-telemetry/admin/pull/279
- [x] Final housekeeping: #7895

svrnm:
Thanks for making some progress here, @chalin !

PR #7866, opened 2025-09-25
[CI] Create separate GH action jobs for build and link-checking

chalin:
- Contributes to #7865
- Splits `build-and-check-links` job into `build` and `check-links`. I might later merge the `check-refcache` into `check-links`. Let's see if this simple refactoring works first.


tiffany76 (approved):
RSLGTM.

chalin:
Thanks @tiffany76. I'll merge tomorrow so that I can immediately address any issues with the updated workflow. It seems to be working.

@svrnm : why is GH waiting (requiring) the old "BUILD and CHECK LINKS" jobs which no longer exists. Is this configured somewhere?

chalin:
> @svrnm : why is GH waiting (requiring) the old "BUILD and CHECK LINKS" jobs which no longer exists. Is this configured somewhere?

Ah, found it in the branch protection rules. Not sure how we should coordinate removing the old and adding the new names, given that there are other PRs in flight.

svrnm:
> > @svrnm : why is GH waiting (requiring) the old "BUILD and CHECK LINKS" jobs which no longer exists. Is this configured somewhere?
> 
> Ah, found it in the branch protection rules. Not sure how we should coordinate removing the old and adding the new names, given that there are other PRs in flight.

We would need to bypass branch protection for that one, and then remove the old ones.

vitorvasc (approved):

chalin:
All checks are passing except for the one that no longer exists, so I'll merge:

> image

PR #7938, opened 2025-09-30
feat: add otel_upgrade_helper to registry

wbollock:
Invited to add to the registry in https://github.com/wbollock/otel_upgrade_helper/issues/6

cc: @svrnm

### DIFF
--- a/data/registry/tools-otel-upgrade-helper.yml
+++ b/data/registry/tools-otel-upgrade-helper.yml
@@ -1,7 +1,7 @@
 # cSpell:ignore wbollock
 title: OpenTelemetry Collector Release Notes Comparator
 registryType: utilities
-language: collector
+language: go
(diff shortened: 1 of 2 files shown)


data/registry/tools-otel-upgrade-helper.yml
```diff
@@ -0,0 +1,19 @@
+# cSpell:ignore wbollock
+title: OpenTelemetry Collector Release Notes Comparator
+registryType: utilities
+language: collector
```
svrnm:
```suggestion
language: go
```

I know this is unexpected, but for tools we track the language they are written in.

wbollock: np! committed

svrnm (approved):

Issue #7436, opened 2025-07-30
Language menu is absent from non-doc pages on mobile


Andygol:
**What happened?** Describe the problem that occurred.

Open https://opentelemetry.io/ on your smartphone or use Responsive Design Mode in your web browser.

**What did you expect would happen?** Describe the expected result/output.

I expect to be able to use language picker

**Additional context**: anything else you can think about that would be helpful.

<img width="1197" height="1192" alt="Image" src="https://github.com/user-attachments/assets/b29dd9dc-ed17-4139-89cc-a2ffab8e1761" />

chalin:
Thanks, this is a known issue, and will be fixed in Docsy. The main tracker issue is:

- https://github.com/google/docsy/issues/2001

Layout design makes it a bit tricky to fix, but I have a solution or two in mind.

chalin:
Just to be clear: the language menu/picker is available from the hamburger in the docs section, as you can see from this screen shot:

> <img width="333" height="584" alt="Image" src="https://github.com/user-attachments/assets/0fd7a9e0-a3cf-4151-a5fb-aaa457a314ac" />

I've updated this issue's title accordingly.

Andygol:
Thanks for making that clear!

chalin:
I knew that we had an issue somewhere:

- https://github.com/google/docsy/issues/2035

Fixed via:

- https://github.com/google/docsy/pull/2303

(I'll be updating Docsy in this repo as soon as I can.)

PR #7943, opened 2025-09-30
fix: Use recent docsy theme

Andygol:
Closes: #7436


theletterf:
@Andygol The PR seems empty.

Andygol:
Hm 😢

It seems that I don't figure out how to update submodules. 

@theletterf, may I ask you for help?

chalin (changes requested):
@Andygol - there are more changes I need to bring in to Docsy and/or double check before we can update here. This is towards the top of my todo list. I'll do the Docsy update as soon as I can. Thanks for your patience.

chalin:
Closing. This will be done when I work on #7436.

PR #7954, opened 2025-10-01
[CI] Create refcache-refresh.yml workflow

chalin:
- Contributes to #2554
- Adds `refcache-refresh.yml` workflow, set to run daily. See file for description


chalin:
@trask - I used the semconv integration workflow as a base. Can you review the code to ensure that it makes sense in this context? Thanks!

chalin:
Maybe we can just land this and then test the workflow.

chalin:
@trask - I'm going to merge and test. Feedback is still welcome.

PR #7977, opened 2025-10-02
Upgrade to Hugo 0.150.1

chalin:
Changes to generated site files include:

- Syntax highlighting
- Missing robots noindex from some blog numbered pages such as `blog/2019/page/1/index.html`, which is fine since we've not set `enableRobotsTXT`.

But overall, the newly generated files seem ok.


tiffany76 (approved):
RSLGTM.

PR #7978, opened 2025-10-03
[CI] Upgrade all deps

chalin:
- Followup to #7977
- Upgrades all other dependencies
- Fixes a spelling error. Not sure why it wasn't caught before. Strange that a patch release upgrade of cSpell would have a spelling error surface.


tiffany76 (approved):
RSLGTM.

I actually saw that spelling issue when doing the KubeCon post but wasn't sure if I should commit it.

PR #7982, opened 2025-10-03
Update refcache-refresh.yml: leave 4xx entries in the cache

chalin:
- Contributes to #2554
- Adds code to validate inputs
- Keeps 4xx entries for further processing
- Improves PR creation code, and prevents the job from failing if there is nothing to commit.


tiffany76 (approved):
RSLGTM.

PR #7995, opened 2025-10-06
add lightstep receiver to registry

svrnm:
adds the lightstep receiver by zalando to the registry


data/registry/collector-receiver-lightstep.yml
```diff
@@ -0,0 +1,23 @@
+# cSpell:ignore loki lokireceiver
```
svrnm:
```suggestion
# cSpell:ignore lightstepreceiver zalando
```

PR #8001, opened 2025-10-06
Add pages for collector components

jaydeluca:
I am experimenting with automation that scrapes collector metadata and generates these pages. Github action jobs will run after each collector release and identify changes, and then will regenerate the pages and submit PRs accordingly.

This is a starting point, and likely to evolve as we expand the data we include for each component.

cc @tiffany76 

---

**PREVIEW**: https://deploy-preview-8001--opentelemetry.netlify.app/docs/collector/components/

### DIFF
--- a/content/en/announcements/gc-election.md
+++ b/content/en/announcements/gc-election.md
@@ -1,9 +1,0 @@
----
-title: 2025 OpenTelemetry Governance Committee Election - Voting is NOW OPEN!
-date: 2025-10-27
-expiryDate: 2025-10-30
----
-
-<i class="fas fa-person-booth"></i> OpenTelemetry Governance Committee
-Election -
-[Voting is NOW OPEN](https://vote.heliosvoting.org/helios/e/otel-gc-2025/vote)!
--- a/content/en/blog/2025/complex-attribute-types.md
+++ b/content/en/blog/2025/complex-attribute-types.md
@@ -1,0 +1,153 @@
+---
+title: Announcing Support for Complex Attribute Types in OTel
+linkTitle: Announcing Support for Complex Attribute Types in OTel
+date: 2025-11-05
+author: >-
+  [Liudmila Molkova](https://github.com/lmolkova) (Grafana Labs), [Robert
+  Pajak](https://github.com/pellared) (Splunk), [Trask
+  Stalnaker](https://github.com/trask) (Microsoft), [Austin
+  Parker](https://github.com/austinlparker) (honeycomb.io)
+sig: Specification, Logs
+issue: https://github.com/open-telemetry/opentelemetry-specification/pull/4485
+cSpell:ignore: Liudmila Molkova Pajak
+---
+
+It’s common to use simple key-value properties as attributes in telemetry. Most
+telemetry backends are optimized for this pattern, making it efficient to store,
+index, and query data.
+
+OpenTelemetry is designed with this in mind. Semantic conventions and
+instrumentations aim to provide useful attributes that can be easily filtered
+and aggregated.
+
+But what happens when the data itself is complex? OpenTelemetry also strives to
+capture observability for real-world systems, libraries, and applications whose
+observable properties are sometimes complex.
+
+Recently OpenTelemetry has announced upcoming support for capturing complex data
+across all OTel signals starting with OTLP
+[1.9.0](https://github.com/open-telemetry/opentelemetry-proto/releases/tag/v1.9.0),
+and in future OpenTelemetry API and SDK versions across the ecosystem.
+
+In this post, we’ll cover when and how to use complex data, when to avoid it,
+and how backends can start supporting it.
+
+## Upcoming support for complex attribute types in OpenTelemetry
+
+OpenTelemetry APIs and SDKs are adding support for the following attribute types
+on all signals:
+
+- Maps (with string keys and values of any supported type)
+- Heterogeneous arrays (containing elements of any supported type)
+- Byte arrays
+- Empty values
+
+Until recently, these types were supported only on logs - attributes on other
+signals were limited to primitives and arrays of primitives.
+
+Following
+[OTEP 4485: Extending attributes to support complex values](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.49.0/oteps/4485-extending-attributes-to-support-complex-values.md)
+and its implementation in OTLP and the specification, this support is being
+extended to all OTel signals.
+
+The new attribute types, especially maps and heterogeneous arrays, should be
+used with care. Many observability backends are not optimized to query, index,
+or aggregate complex attributes. Semantic conventions will assume complex
+attributes are not indexed and will avoid using them on metrics or in other
+scenarios where efficient querying is important.
+
+When possible, stick to primitive values.
+
+## Why are we doing this?
+
+As we work on semantic conventions and instrumentations, we increasingly
+encounter cases where flat attributes cannot reasonably capture the complexity
+of real-world scenarios.
+
+Examples include:
+
+- **[LLM operations](/docs/specs/semconv/gen-ai/non-normative/examples-llm-calls)**
+  — input parameters like tool definitions and input/output messages are
+  inherently structured
+- **GraphQL** — responses may include
+  [lists of structured errors](https://graphql.org/learn/response/#errors)
+- **[Database operations](/docs/specs/semconv/database/database-spans)** — batch
+  operations have properties that flat attributes cannot adequately capture
+
+Before extending support for complex attributes to all signals, we explored
+several alternatives:
+
+### Limiting support to logs (and spans)
+
+Having different attribute collection types for different signals affects API
+ergonomics, making it less convenient and efficient to work with attributes.
+
+### Flattening
+
+Flattening works well for maps of primitives but breaks down for arrays. For
+example, structured data like:
+
+```json
+{
+  "data": [
+    {
+      "foo": "bar",
+      "baz": 42
+    }
+  ]
+}
+```
+
+is reported as either:
+
+```text
+data.0.foo = "bar"
+data.0.baz = 42
+```
+
+or:
+
+```text
+data.foo = ["bar"]
+data.baz = [ 42 ]
+```
+
+Both approaches are limited and lead to a poor user experience.
+
+### String serialization
+
+Another option is requiring users or instrumentations to serialize complex data
+into strings. While workable, this risks inconsistencies and errors. It also
+limits post-processing and leads to poor truncation strategies. Ultimately, it
+should be the backend’s decision whether to preserve the structure or serialize
+the data. Handling serialization at the backend side provides greater
+consistency and convenience for end users.
+
+## How should backends support complex attributes?
+
+We encourage backends to build user experiences that leverage structured
+attribute values, allowing users to query data based on nested properties.
+
+In the meantime, we recommend that backends serialize complex attributes to JSON
+(or another suitable format) at ingestion time.
+
+The OTel Specification, Semantic Conventions, and API documentation will clearly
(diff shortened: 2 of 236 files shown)


content/en/docs/collector/components/connector.md
```diff
@@ -0,0 +1,39 @@
...
+## Core Distribution
+
+Components from the [OpenTelemetry Collector](https://github.com/open-telemetry/opentelemetry-collector) core distribution.
+
+The **Traces**, **Metrics**, and **Logs** columns show the stability level for each signal type.
+
+| Name | Traces | Metrics | Logs |
+|------|--------|---------|------|
+| [forwardconnector](https://github.com/open-telemetry/opentelemetry-collector/tree/main/connector/forwardconnector) | - | - | - |
+| [xconnector](https://github.com/open-telemetry/opentelemetry-collector/tree/main/connector/xconnector) | - | - | - |
```
jaydeluca: i wasnt sure whether it made sense to include the `x` components?

dmathieu:
This is nice. But looking at it, it could be read that those are the only components anyone can use, which isn't true (if you build your own distribution, you can use any component, including private ones).

Speaking of distributions, there are 5 different official collector distributions. How about adding which of them each component is included into?

tiffany76:
Thanks for taking a look at this POC, @dmathieu. We're just getting started and can use expert input!

We plan to update the language on the pages and will not only call out that users can use other components, but also cross-reference the registry, which includes many that are not part of core/contrib. 

We're starting with a light lift for this POC while we investigate exactly what should be included and how it should be displayed. I'll add the otelcol distribution to our list of what should be included.

Thanks again!

content/en/docs/collector/components/connector.md
mx-psi: I think we can exclude those

content/en/docs/collector/components/_index.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Stability Levels
+
+Each component has a stability level that indicates its maturity:
+
+- **stable** - Ready for production use
+- **beta** - Mostly stable, but may have minor changes
+- **alpha** - Early development, expect breaking changes
+- **development** - Experimental, subject to change or removal
+- **unmaintained** - No longer actively maintained
```
chalin: OTel defines official levels of maturity in the spec(s) and elsewhere, such as https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/component-stability.md (as you probably know :). Could we link to that page as the main source of truth?

tiffany76:
@mx-psi, thank you for the review!
 
> I wonder if it would make sense to style/order components differently depending on their stability level (since the stability level can vary per signal I guess the following can be done by taking the max stability level).

That's an interesting idea. We should consider how to offer filtered/sorted views when we iterate on how the data is presented, especially if we move away from tables. 👍 

> We are missing profiling support from the list! This is an experimental signal so maybe can be excluded

Noted!

> I think it would be great to have links to https://opentelemetry.io/docs/collector/configuration/#receivers and similar somewhere in these pages

I was thinking these pages will merge with that content to actually _become_ the /receivers, /processors, /exporters, etc. pages in the new IA. So the corresponding content on that monolithic configuration page will be broken out into separate pages, and each page will have the OTel-maintained components listed (right now, in a table). WDYT?
Screenshot 2025-10-07 at 9 05 07 AM

> It would also make sense to me to link to https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/component-stability.md somewhere

Noted! I think Patrice had the same idea. 👍

mx-psi:
> I was thinking these pages will merge with that content to actually become the /receivers, /processors, /exporters, etc. pages in the new IA. So the corresponding content on that monolithic configuration page will be broken out into separate pages, and each page will have the OTel-maintained components listed (right now, in a table). WDYT?

Ah, right, yeah, that makes sense

tiffany76:
Thanks for giving this a look, @chalin!
 
> Have you seen https://opentelemetry.io/status/#collector? The added page content seems to be mostly statuses for components. Might it make sense to put these pages (or at least their status content), under https://opentelemetry.io/status/#collector or a new subsection of https://opentelemetry.io/status/?

Although right now it looks like the new content is mostly adding statuses, the ultimate goal is to bridge the gap between component documentation and the official docs. As we iterate, we will hopefully be able to bring in more of the component metadata than just the stability level. So I don't think we should move this content to the /status page.

> Re. the automation: under which repo will the GH actions run (and where will be the associated scripts, if any)?

I think @jaydeluca can answer that better than me.

> There is https://opentelemetry.io/docs/collector/distributions/. Is there an overlap with what is being proposed here as content? If so, will it lead to confusion for readers?

No, there shouldn't be much overlap. The current Distributions page will be reworked in the new IA to help users focus on choosing a distribution. But I don't think we'll be going into component-level detail there. The only overlap I see (as of now, that could change, of course) is if we indicate which distributions each component is part of, as recommended by Damien.

svrnm:
Thanks, this is really interesting and going to be helpful. One comment/ask:

While we still have it, can this script populate registry entries first and then fill the pages with that information?

jaydeluca:
thank you everyone for all the valuable feedback thus far, I'm excited to see what we can do here. Apologies for the delay, I was traveling this week.

Here is my plan for next steps, as of now:

- Work on reworking the page copy, including taking out stability verbiage and link to existing resources for that information
- Include more details about the other distributions and which ones the components are included in ([comment](https://github.com/open-telemetry/opentelemetry.io/pull/8001#issuecomment-3373681734))
- Remove the x- components
- Explore integration with the existing registry as well
- Rework things a bit so the automation just manages the list contents on the pages instead of generating the entire page content

As for the tooling itself, I have everything running in https://github.com/jaydeluca/collector-watcher for now, which I think makes sense so I can iterate quickly until I get things in a good place and working - unless there are others who would like to collaborate and be included in code reviews, I'm happy to do whatever makes sense. I can also rewrite it in another language if that makes it easier for others to get involved, although it's not too complex.

My thinking is that I can get everything working end to end in an automated fashion from that repo and let it soak for a bit, monitor that it works as we want, and then we could move it either into this repo or if we want to keep it as a separate repo within the OTel org.

So the goal is that once these pages are finalized and merged, every night the automation will run, detect if there are updates, and will open PR's automatically.

Keep the feedback coming if you have other ideas or concerns. I will hopefully make some progress on this over the next week. Thanks all!

content/en/docs/collector/components/exporter.md
```diff
@@ -0,0 +1,78 @@
...
+| [azureblobexporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/azureblobexporter)                             | contrib                  | alpha       | alpha        | alpha       |
+| [azuredataexplorerexporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/azuredataexplorerexporter)             | contrib                  | beta        | beta         | beta        |
+| [azuremonitorexporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/azuremonitorexporter)                       | contrib                  | beta        | beta         | beta        |
+| [bmchelixexporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/bmchelixexporter)                               | contrib                  | -           | alpha        | -           |
+| [carbonexporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/carbonexporter)                                   | contrib                  | -           | unmaintained | -           |
```
jaydeluca: Should we exclude `unmaintained` components from being included?

content/en/docs/collector/components/connector.md
```diff
@@ -0,0 +1,38 @@
...
+
+Connectors connect two pipelines, acting as both exporter and receiver. For more
+information on how to configure connectors, see the
+[Collector configuration documentation](/docs/collector/configuration/#connectors).
+
+<!-- BEGIN GENERATED: connector-table -->
+
+| Name                                                                                                                                       | Distributions[^1]  | Traces[^2] | Metrics[^2] | Logs[^2] |
+| ------------------------------------------------------------------------------------------------------------------------------------------ | ------------------ | ---------- | ----------- | -------- |
+| [countconnector](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/connector/countconnector)                     | contrib, K8s       | -          | -           | -        |
```
jaydeluca: another thought: would there be value in including the latest release number, with a link to the release page or anything like that?

content/en/docs/collector/components/exporter.md
mx-psi: I think at the very least, they should be faded/clearly marked as something that you should not use

content/en/docs/collector/components/connector.md
mx-psi: The version is going to be the same for most components, so maybe it's not very useful?

content/en/docs/collector/components/exporter.md
jaydeluca: added a marker:

image

and note at the bottom of each page:

image

jaydeluca:
I have implemented most of the feedback. a couple notes:
- I had originally mentioned that I would have automation that runs nightly, but instead it makes more sense to trigger this when a new release is cut, so that's what we'll do instead
- I will work on an integration with the existing registry in a separate PR, since that will be a pretty large changeset

jaydeluca:
> If I look at this with half a screen, I am not able to tell that I need to do horizontal scrolling to see the stability for other signals

This seems to be an issue across the site (I tested with the java sdk page on my phone and see the same thing). When @tiffany76 and I first started discussing this, we figured that we'd eventually want to display more than just this basic stability information, so a table is not likely to be the best "final form" for this, as I think the accessibility will only get worse. So at some point we'll come up with a new layout all together, perhaps an actual section for each, with a bit more information included, and ditch the table all together.

In the meantime, maybe the @open-telemetry/docs-approvers might have some suggestions/ideas in case they've come across this before.

> think for a first version we can just remove these columns for connectors. I suspect the best way of handling this is to have a 'highest stability' column (but that can be for a later PR)

Done, removed the columns for now


> I feel like 'components' should be closer to the top of the list on the sidebar, I would expect to want to look at components when I am thinking about how to configure my Collector

I moved it to below the "Configuration" item, let me know if you want it elsewhere

image

mx-psi:
>So at some point we'll come up with a new layout all together, perhaps an actual section for each, with a bit more information included, and ditch the table all together.
>
> In the meantime, maybe the @open-telemetry/docs-approvers might have some suggestions/ideas in case they've come across this before.

Alright, I'll leave it to the docs approvers to decide whether this is something that needs to be tackled now or can be left for later :) 

> I moved it to below the "Configuration" item, let me know if you want it elsewhere

I think that's the closest to the proposed refactor (https://www.mindomo.com/mindmap/c6ececd0512d46edb8f5048d19f18de1) where it would land inside a new "Configure the Collector" section. I guess this will be shuffled around further when the refactor happens, but for now I feel like that is the right place

mx-psi (approved):
I think this looks good for a first version, my only concern is the table visualization in small screens (which may be something we can leave for later)

jaydeluca:
> has anyone weighed in on where exactly the watcher tool should live?

In my opinion, I think we should leave it where it is for now. After this is merged, I'm going to finalize the workflow to trigger it to kick off after each new collector release to automatically create PRs. I imagine I'm going to need to tweak and iterate that a bit. Once I have that in a good place, we can then move it to wherever makes sense. I'm happy to do whatever though

chalin:
All checks are green. @tiffany76 - so, are we waiting for a final approval from @open-telemetry/collector-approvers?

tiffany76:
We've got @mx-psi's approval, which works for me. And @dmathieu's feedback was partially addressed in this first pass. I think it's ready. 🎉

Now we merge...and then we iterate!

PR #8015, opened 2025-10-07
[CI] Update contrib does, and fix script name

chalin:
- Contributes to #2554
- Fixes the script name in the workflow
- Also updates the script name in the contrib docs

### DIFF
--- a/content/en/docs/contributing/pr-checks.md
+++ b/content/en/docs/contributing/pr-checks.md
@@ -125,13 +125,16 @@ have the link checker ignore it: `?no-link-check`. For example,
 {{% alert title="Maintainers tip" %}}
 
 Maintainers can run the following script immediately after having run the link
-checker to have Puppeteer attempt to validate links with non-success statuses
+checker to have Puppeteer attempt to validate links with non-ok statuses:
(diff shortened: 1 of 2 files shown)


content/en/docs/contributing/pr-checks.md
```diff
@@ -125,13 +125,16 @@ have the link checker ignore it: `?no-link-check`. For example,
...
 ```sh
-./scripts/double-check-refcache-400s.mjs -f --max-num-to-update 99
+./scripts/double-check-refcache-4XX.mjs
 ```
 
-This script also validates URL fragments, which the link checker doesn't do.
+Use the `-f` flag to also validate URL fragments (anchors) in external links,
+which `htmltest` doesn't do. We don't currently run this often, so you will
+probably want to limit the number of updated entries using the `-m N` flag. For
+usage info, run with `-h`.
```
chalin, line 137: @vitorvasc - FYI, this is how to check if external URL fragments are valid, and to have the script report any URLs with invalid fragments.

tiffany76 (approved):
RSLGTM.

PR #8017, opened 2025-10-08
[CI] refcache-refresh: fix chrome path

chalin:
- Contributes to #2554
- Explicitly sets an env var with the path to Chrome
- Uses the set env var in the appropriate helper script


tiffany76 (approved):
RSLGTM.

PR #8029, opened 2025-10-08
[CI] Generate sidenav active-page style dynamically

### LINKED ISSUE #5646: [CI/infra] Reduce site build time by 60% by setting `sidebar_cache_limit` to 1
Learning from our sister project [kubernetes/website](https://github.com/kubernetes/website), I'd like to propose that we set the following config parameter ([reference](https://github.com/kubernetes/website/blob/68ed21755dcbcabd50167943188d8b4c17470414/hugo.toml#L223-L224)):

```toml
[params.ui]
sidebar_cache_limit = 1 # https://github.com/gohugoio/hugo/issues/8918#issuecomment-903314696
```

In local tests, doing so reduces the site build time by 60%. The cost is that the active side-nav entry corresponding to the current page is dynamically marked as active client-side via the JS in https://github.com/google/docsy/blob/fed2d28d247f4d9f1190a5088f0fea4f09676b84/layouts/partials/sidebar.html (that JS should be moved, but that's another story).

 WDYT @open-telemetry/docs-maintainers?

This might be blocked on the following issue, but we don't have the language menu in the side nav anymore, so it isn't blocking:

- https://github.com/google/docsy/issues/2131

/cc @sftim @nate-double-u

chalin:
- Fixes #5646
- Reduces the build time by >30%, e.g., compare:
  - 1m41 = 101s https://github.com/open-telemetry/opentelemetry.io/actions/runs/18344515259/job/52247740051?pr=8029
  - 2m31 = 151s https://github.com/open-telemetry/opentelemetry.io/actions/runs/18343970670/job/52245934086?pr=8028

### DIFF
--- a/hugo.yaml
+++ b/hugo.yaml
@@ -138,6 +138,7 @@ params:
     sidebar_menu_compact: true
     sidebar_menu_foldable: true
     sidebar_search_disable: true
+    sidebar_cache_limit: 1
     feedback:
       enable: true
       'yes': >-


theletterf (approved):
Whoa!

PR #8040, opened 2025-10-09
[CI] Refcache-refresh: more crates.io processing fixes

chalin:
- Contributes to #2554
- Fixes processing of crates.io URLs

### DIFF
--- a/scripts/double-check-refcache-4XX.mjs
+++ b/scripts/double-check-refcache-4XX.mjs
@@ -2,7 +2,11 @@
 
 import fs from 'fs/promises';
 import { parseArgs } from 'node:util';
-import { getUrlStatus, isStatusNotFound, isHttp2XX } from './get-url-status.mjs';
+import {
+  getUrlStatus,
+  isStatusNotFound,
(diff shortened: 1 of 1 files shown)


scripts/get-url-status.mjs
```diff
@@ -297,6 +307,13 @@ function getChromePath() {
   );
 }
 
+// Returns true iff status is 404 and URL is assumed Not Found. See "Note about
```
tiffany76, line 310:
```suggestion
// Returns true if status is 404 and URL is assumed Not Found. See "Note about
```

chalin: Thanks for the careful review. This was indentionally an "iff` as in https://en.wikipedia.org/wiki/If_and_only_if :) ... but I can switch it back to just a plain "if".

tiffany76: Oh, no, that's my ignorance then! Sorry!

chalin: No worries! 🙌🏻

PR #8045, opened 2025-10-09
[i18n] Update drifted status of non-en pages

chalin:
- Adds drifted status to non-`en` pages


chalin:
Thanks @tiffany76. Housekeeping, for all locales, so will merge once checks pass.

PR #8078, opened 2025-10-13
[editorial] Fix lint errors and drift status after markdownlint upgrade

chalin:
- Followup to #8077
- Upgrades `markdownlint` to `@latest`
- Fixes newly reported lint issues, mainly heading levels starting at h3 instead of h2
- Disables `table-column-style`, a new rule, because it reports lots of issues with ja and zh pages, whose md tables aren't styled by Prettier 🤷🏼‍♂️. For the rule doc page, see https://github.com/DavidAnson/markdownlint/blob/main/doc/md060.md.
- Updated drift status for a few pages

### DIFF
--- a/.markdownlint.yaml
+++ b/.markdownlint.yaml
@@ -15,7 +15,7 @@ no-trailing-spaces:
   br_spaces: 0
   strict: true
 
-# table-column-style: false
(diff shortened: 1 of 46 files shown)


chalin:
Editorial changes only, so merging this pan-locale PR.

PR #8149, opened 2025-10-17
[fr] platforms/kubernetes/operator translation

clementduveau:
This is an initial proposal for the platforms/kubernetes/operator directory translation (in french). I did not find an issue tracking this page. Not a lot more to add !


scripts/generate-i18n-anchors.sh
```diff
@@ -0,0 +1,122 @@
+#!/bin/bash
+
+# Script to add anchors to titles in a translated file or directory based on the English version
```
dmathieu: I feel like this file should be in its own PR, since it's not related to the french translation.

dmathieu (approved):

scripts/generate-i18n-anchors.sh
clementduveau: True, I am removing it from this PR

Issue #7953, opened 2025-10-01
[CI] External link checking improvements for 25H2-26H1


chalin:
- [ ] Set up a cron job to refresh the refcache's oldest X entries every Y days, and report an issue and/or submit a PR with the cache updates - #2554
- [x] #6175
- [ ] Normalize links to avoid checking a(n external) link (to the same destination) more than once, and creating duplicate entries in the ref cache:
   - [ ] external links: consistently end with a slash or not (for domain-only URLs). 
   - [ ] external links: use apex domain (w/ or w/o www), as appropriate

Related issues:

- https://github.com/wjdp/htmltest/issues/239

Previous:

- #5017

chalin:
Closing: with the switch to Lychee (#10912), all substantive items are done, and the remaining normalization items are either resolved by construction or no longer worth infra effort:

- **Trailing slash (domain-only URLs)** — resolved by the switch: Lychee normalizes `https://host` → `https://host/` before checking and caching. The current `.lycheecache` has 0 slashless domain-only entries (the old refcache had 60, of which 17 were cached in both forms).
- **www vs apex** — reduced to noise: 12 www/apex pairs out of ~6040 cache entries, of which at least 3 are stale (no longer in any source; they age out via `max_cache_age` + the daily prune-and-refresh cycle). Lychee checks each distinct URL once per run, so the residual cost is a handful of redundant fetches per cache-expiry cycle. What remains is a content-style choice (one form per host), not a link-check infrastructure concern.

The htmltest issue referenced under "Related" is moot since htmltest was dropped (#10911).

PR #8153, opened 2025-10-18
[CI] Disable StripQueryString htmltest config, and update refcache

chalin:
- Contributes to #7953
- Disables StripQueryString htmltest config parameter so that query strings are saved to the cache
- Implements a workaround for https://github.com/wjdp/htmltest/issues/239, by simply reading in the JSON and writing it back out unescaped


tiffany76 (approved):
RSLGTM.

Issue #7939, opened 2025-09-30
spec_status shortcode doesn't work for languages other than english


Andygol:
The exception was previously reported in the PR #5891 

Job [Links / BUILD and CHECK LINKS (pull_request)](https://github.com/open-telemetry/opentelemetry.io/actions/runs/17342323498/job/49237862240?pr=5891) is constantly failing due to issue in spec_status shortcode that is used in https://github.com/Andygol/opentelemetry.io/blob/d9d8051ceb5355d3eff5958d2f6fccb6fdb9346c/content/uk/docs/specs/status.md#L56-L59

```log
WARN  spec_status shortcode error: page '/docs/specs/otlp' does not contain a line matching '/document-status.*for.*log'.
WARN  spec_status shortcode error: '/document-status.*for.*log' value did not match one of '(?i:development|experimental|stable|deprecated|removed|mixed|feature.freeze|alpha|beta)' in file '/docs/specs/otlp': matched line is ''
WARN  spec_status shortcode error: page '/docs/specs/otlp' does not contain a line matching '/document-status.*for.*profiles'.
WARN  spec_status shortcode error: '/document-status.*for.*profiles' value did not match one of '(?i:development|experimental|stable|deprecated|removed|mixed|feature.freeze|alpha|beta)' in file '/docs/specs/otlp': matched line is ''
```

https://github.com/open-telemetry/opentelemetry.io/actions/runs/17342323498/job/49237862240?pr=5891#step:8:106

_Originally posted by @Andygol in https://github.com/open-telemetry/opentelemetry.io/issues/5891#issuecomment-3239160934_

svrnm:
I think the problem is that localization of the spec is not something we support right now, as this comes with a few other challenges.

@Andygol if you can take a look and if it is easy to fix, feel free to provide a PR.

PR #8156, opened 2025-10-18
[CI/Infra] Always use 'en' for spec status

Andygol:
This PR contains a fix for the spec_status shortcode. This fix allows the status of OpenTelemetry APIs, SDKs, and protocols to be displayed on localized pages, not just English-language ones.

Closes: #7939

/cc @svrnm, @chalin

### DIFF
--- a/content/en/docs/specs/status.md
+++ b/content/en/docs/specs/status.md
@@ -57,8 +57,8 @@ same as the **Protocol** status.
 ### Tracing
 
 - [Specification][tracing]
-- {{% spec_status "API" "otel/trace/api" "Status" %}}
-- {{% spec_status "SDK" "otel/trace/sdk" "Status" %}}
+- {{% spec_status "API" "otel/trace/api" %}}
+- {{% spec_status "SDK" "otel/trace/sdk" %}}
 - {{% spec_status "Protocol" "otlp" "/document-status.*for.*trace" %}}
 - Notes:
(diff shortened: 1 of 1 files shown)


layouts/_shortcodes/spec_status.html
```diff
@@ -4,7 +4,15 @@
...
-{{ $page := .Site.GetPage $pageRef -}}
+
+{{/* Always use English version as the primary source for specs status. */ -}}
+{{ $page := "" }}
+{{- with where .Site.Sites "Language.Lang" "eq" "en" -}}
+  {{- with index . 0 -}}
+    {{ $page = .GetPage $pageRef -}}
+  {{ end -}}
+{{ end -}}
+
```
chalin:
```suggestion
{{ $enSite := index site.Sites 0 -}}
{{ $page := $enSite.GetPage $pageRef -}}
```

chalin (changes requested):
Thanks for the PR @Andygol.
See inline suggestion for a solution without lookup.

Also update `content/en/docs/specs/status.md` and remove the `"Status"` argument to the shortcode calls, since that value is the default.

Andygol:
@chalin Thanks for the code simplification 👍 

I've updated PR according to your recommendation. Looking forward to seeing these changes into main.

chalin (approved):
✨

PR #8224, opened 2025-10-24
Update Sentry URL in vendors.yaml

sfanahata:
Updating what page to direct to for Open Telemetry's vendor page.

Right now we're pointing to a dated product page. We want to point to a richer documentation page. New URL: https://docs.sentry.io/concepts/otlp/


data/ecosystem/vendors.yaml
```diff
@@ -356,7 +356,7 @@
   commercial: true
 - name: Sentry
   nativeOTLP: true
-  url: https://sentry.io/for/opentelemetry/
+  url: https://docs.sentry.io/concepts/otlp/
   contact:
```
svrnm: @sfanahata while you're fixing this, can you update this as well, preferred is a github handle (like yours) that if we need to change something we can tag you accordingly. email is fine as well, but might get lost somewhere

sfanahata: @svrnm done! We just updated the folks attached to this email address, and it's specifically for the otel team here, so should be good for folks getting a timely response from us.

maryliag (approved):
Thank you for adding the contact changes too

PR #8226, opened 2025-10-24
Add internal OTLP resource attribute documentation, formatting

n0coast:
Added configuration options for internal OTLP metric resource attributes, the only other reference for this I've found is in [github.com/open-telemetry/opentelemetry-collector/issues/12040
](https://github.com/open-telemetry/opentelemetry-collector/issues/12040)
Minor formatting to break up the "Configure internal metrics" section by topic.


svrnm:
@open-telemetry/collector-approvers PTAL

content/en/docs/collector/internal-telemetry.md
```diff
@@ -53,6 +54,20 @@ service:
                 endpoint: https://backend:4318
 ```
 
+If you want to add additional resource attributes to the OTLP metrics,
```
jade-guiton-dd: Nitpick: these resource attributes will be used for all signals, including logs and traces, not just metrics, and not just when exporting through OTLP.

n0coast: Good catch - so would "If you want to add additional resource attributes to service telemetry" be more appropriate? "service signals"? "self-monitoring metrics, traces, and logs"? I'm not sure how best to word this.

n0coast:
> Looks mostly good from a technical standpoint.
> 
> Maybe we could take the opportunity to document that the Collector automatically adds `service.name`, `service.version`, and `service.instance.id` (randomly generated) resource attributes, and that they can be disabled with `service.name: null`?

I can add that, I wasn't aware that was the case. Just to confirm adding `service.name: null` disables `service.name`, `service.version`, and `service.instance.id`? Or do you mean that adding a `null` value disables that specific `key:value` combination?

content/en/docs/collector/internal-telemetry.md
jade-guiton-dd: Maybe "to the Collector's internal telemetry"?

jade-guiton-dd:
> Just to confirm adding service.name: null disables service.name, service.version, and service.instance.id? Or do you mean that adding a null value disables that specific key:value combination?

Setting `service.name: null` only disables `service.name`, it's on a per-key basis. It's essentially a way of overriding the default values set by the service.

(And to clarify my previous explanation: only `service.instance.id` is randomly generated; `service.name` and `service.version` are based on the Collector distribution)

n0coast:
> > Just to confirm adding service.name: null disables service.name, service.version, and service.instance.id? Or do you mean that adding a null value disables that specific key:value combination?
> 
> Setting `service.name: null` only disables `service.name`, it's on a per-key basis. It's essentially a way of overriding the default values set by the service.
> 
> (And to clarify my previous explanation: only `service.instance.id` is randomly generated; `service.name` and `service.version` are based on the Collector distribution)

I incorporated your suggestions, let me know what you think.

vitorvasc (approved):

hugo.yaml
chalin, line 1: This PR should not have included changes to this (`hugo.yaml`) file.

PR #8263, opened 2025-10-28
PR Actions workflow: use GitHub App token for checkout

vitorvasc:
Updates the `pr-action.yml` workflow to retrieve and pass the GitHub App token when using `actions/checkout`. 

This change contributes to #4339 and attempts to fix workflows that get stuck when `/fix:*` commands are executed.

### DIFF
--- a/.github/workflows/pr-actions.yml
+++ b/.github/workflows/pr-actions.yml
@@ -34,16 +34,8 @@ jobs:
         with:
           egress-policy: audit
 
-      - uses: actions/create-github-app-token@67018539274d69449ef7c02e8e71183d1719ab42 # v2.1.4
-        id: otelbot-token
-        with:
-          app-id: ${{ vars.OTELBOT_DOCS_APP_ID }}
-          private-key: ${{ secrets.OTELBOT_DOCS_PRIVATE_KEY }}
-
       - uses: actions/checkout@v5
-        with:
-          fetch-depth: 0
-          token: ${{ steps.otelbot-token.outputs.token }}
+        with: { fetch-depth: 0 }
 
       - name: Extract action name
         id: extract


svrnm:
> Hmm, no, we can't do this because we purposefully split things up into an untrusted job and a trusted one. The untrusted one isn't supposed to have access to tokens. We might need to think about this a bit longer. /cc @svrnm

exactly, my assumption is that it should be enough to update the checkout of the trusted step

.github/workflows/pr-actions.yml
```diff
@@ -34,8 +34,16 @@ jobs:
...
+        id: otelbot-token
+        with:
+          app-id: ${{ vars.OTELBOT_DOCS_APP_ID }}
+          private-key: ${{ secrets.OTELBOT_DOCS_PRIVATE_KEY }}
+
       - uses: actions/checkout@v5
-        with: { fetch-depth: 0 }
+        with:
+          fetch-depth: 0
+          token: ${{ steps.otelbot-token.outputs.token }}
```
svrnm: undo this in the generation of the patch, we do not want to expose tokens here

.github/workflows/pr-actions.yml
```diff
@@ -109,17 +117,18 @@ jobs:
...
         with:
           # the higher privileged docs token is needed to push commits to an existing PR
           app-id: ${{ vars.OTELBOT_DOCS_APP_ID }}
           private-key: ${{ secrets.OTELBOT_DOCS_PRIVATE_KEY }}
 
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        with:
+          fetch-depth: 999
+          token: ${{ steps.otelbot-token.outputs.token }}
+
```
svrnm, line 123: keep only this

vitorvasc:
@chalin @svrnm done! I've reverted the token retrieval in the untrusted job.

.github/workflows/pr-actions.yml
chalin: @vitorvasc - ~could you move this step back to where it was,  so that it is more clear that the only change is the addition of the token?~ We can't move it back because we need the token from the previous step.

PR #8264, opened 2025-10-28
Rewrite java agent extensions page

### LINKED ISSUE #4910: Add more detail to the Java Agent Extension docs
**Desired feature or idea**:

As a user of the java agent, I want documentation that lays out everything I need to know in order to understand use cases for extensions, how to leverage an existing extension, and how to approach writing my own. The [existing page](https://opentelemetry.io/docs/zero-code/java/agent/extensions/) provides context for use cases but doesn't have any of the technical details in order to get started.

There is a lot of good information in the [example section of the instrumentation repo](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/examples/extension/README.md) and elsewhere that I think would be worth unifying and porting over.

**Additional context**:

Here is what I am thinking for the updated page structure:

```
## Introduction
all the current information around use cases will remain here.

## Loading an extension at runtime
detail using the `-Dotel.javaagent.extensions` startup command.

## Embedding extensions in the Agent
detail the `extendedAgent` gradle task approach from the example repo

## Writing an extension
provide example code for an extension, and the steps needed to use it with an agent

## Extension examples
keep this existing section with a link to the example section of the instrumentation library
```

I would be happy to work on contributing this.

### LINKED ISSUE #7793: [Page feedback]: Adding extensions is poorly documented
### URL

https://opentelemetry.io/docs/zero-code/java/agent/extensions/

### Description

It has been _extremely_ difficult to figure out how to wire up BaggageSpanProcessor extension.  We have pretty much given up.  

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

jaydeluca:
Fixes #4910
Fixes #7793

Deploy Preview: https://deploy-preview-8264--opentelemetry.netlify.app/docs/zero-code/java/agent/extensions/


content/en/docs/zero-code/java/agent/extensions.md
```diff
@@ -1,23 +1,41 @@
 ---
```
jaydeluca, line 1: [For reviewers] I took the approach of including code examples instead of just linking to the example in the java agent repo because I personally don't like the user experience of having to jump around and reference multiple sources when reading documentation. It also didn't seem like the code would change frequently, so upkeep shouldn't be too bad (hopefully). 

With that being said, If we think this is too verbose, I can remove the code example and any other sections and just link out to the example code.

svrnm (approved):

content/en/docs/zero-code/java/agent/extensions.md
SylvainJuge: I agree with you, having code examples directly into documentation without having to open a separate tab is definitely a better experience, one of the main downsides though is keeping this code relevant and maintained over time.

Maybe it's a dumb idea, but if we were able to put some markers in the example in instrumentation and then automatically extract and inline the relevant parts into this documentation it would be nice, but that sounds also quite brittle and could easily break over time.

zeitlinger: we actually have this mechanism - but it's not documented and doesn't run on a regular basis

@svrnm is this something you're planning to do or are looking for help?

content/en/docs/zero-code/java/agent/extensions.md
```diff
@@ -1,23 +1,41 @@
...
+
+Extensions are designed to override or customize the instrumentation provided by
+the upstream agent without modifying the agent code. Consider an instrumented
+database client that creates a span per database call and extracts data from the
+database connection. Here are some sample use cases that can be solved using
 extensions:
 
+- _"I want to customize instrumentation without modifying the instrumentation"_:
+
+  The `InstrumenterCustomizerProvider` extension point allows you to customize
```
laurit: is it worth mentioning that `InstrumenterCustomizerProvider` is an experimental api?

content/en/docs/zero-code/java/agent/extensions.md
```diff
@@ -49,8 +68,1083 @@ extensions:
   existing `Tracer` directly or extend it. As you have your own `Advice`, you
   can control which `Tracer` you use.
 
-## Extension examples
+## Loading an Extension at Runtime
+
+There are two ways to use extensions with the Java agent:
+
+1. **Load at runtime** using the `-Dotel.javaagent.extensions` option (described
```
laurit: `Load at runtime` is a bit weird, extensions added to agent are also loaded at runtime. Perhaps we shoudl use `As a separate jar file` or something like that.

content/en/docs/zero-code/java/agent/extensions.md
```diff
@@ -49,8 +68,1083 @@ extensions:
...
+   in this section)
+2. **Embed in the agent** to create a single JAR file (described in the
+   [next section](#embedding-extensions-in-the-agent))
+
+### Basic Usage
+
+Extensions can be loaded at runtime using the `otel.javaagent.extensions` system
```
laurit, line 157: could mention that `OTEL_JAVAAGENT_EXTENSIONS` env var also work

content/en/docs/zero-code/java/agent/extensions.md
```diff
@@ -49,8 +68,1083 @@ extensions:
...
+     -jar myapp.jar
+```
+
+### How Extension Loading Works
+
+When you load extensions at runtime, the agent:
+
+1. Creates an
+   [isolated class loader](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/docs/contributing/javaagent-structure.md#extension-class-loader)
+   for each extension JAR to prevent conflicts between extensions
```
laurit: This is an implementation detail. We could consider giving up on it and loading extensions together with the agent code. This would for example allow agent code that uses `ServiceLoader.load` without specifying a class loader to discover spis from extensions. Admittedly this is less of an issue now that the sdk seems to always use user specified class loader for looking up spis.

content/en/docs/zero-code/java/agent/extensions.md
```diff
@@ -49,8 +68,1083 @@ extensions:
...
+2. Makes OpenTelemetry APIs available to your extension (without needing to
+   package them in your extension JAR)
+3. Automatically applies shading to extension classes at runtime (extensions use
+   normal OpenTelemetry imports, and the agent handles compatibility
+   automatically)
```
laurit: I think the shading is kind of an advanced topic, hopefully most of the extension users don't even realize that something is shaded. If we are going to mention it then I think the goal should be to reduce confusion for the ones who try to debug the extension and notice that things aren't quite as they'd expect.

content/en/docs/zero-code/java/agent/extensions.md
```diff
@@ -49,8 +68,1083 @@ extensions:
...
+4. Discovers your extension's components using Java's ServiceLoader mechanism
+   (via `@AutoService` annotations in your code)
```
laurit: this sentence leaves an impression that using `@AutoService` is the only way to declare these components. Could also consider linking to somewhere that describes what `ServiceLoader` is and how to use it.

content/en/docs/zero-code/java/agent/extensions.md
```diff
@@ -49,8 +68,1083 @@ extensions:
...
+Creating an extension involves implementing one or more Service Provider
+Interface (SPI) classes and packaging them as a JAR file.
+
+### Quick Start
+
+Here's a minimal extension that adds a custom span processor:
+
+**1. Create a Gradle project:**
+
+```groovy
```
laurit: would it make sense to use gradle kotlin instead?

content/en/docs/zero-code/java/agent/extensions.md
```diff
@@ -49,8 +68,1083 @@ extensions:
...
+interfaces:
+
+| Extension Point                       | Package                                                       | Purpose                                |
+| ------------------------------------- | ------------------------------------------------------------- | -------------------------------------- |
+| `AutoConfigurationCustomizerProvider` | `io.opentelemetry.sdk.autoconfigure.spi`                      | Main entry point for SDK customization |
+| `ConfigurablePropagatorProvider`      | `io.opentelemetry.sdk.autoconfigure.spi`                      | Register custom propagators            |
+| `ConfigurableSamplerProvider`         | `io.opentelemetry.sdk.autoconfigure.spi.traces`               | Register custom samplers               |
+| `ResourceProvider`                    | `io.opentelemetry.sdk.autoconfigure.spi`                      | Add custom resource attributes         |
+| `InstrumenterCustomizerProvider`      | `io.opentelemetry.instrumentation.api.incubator.instrumenter` | Customize existing instrumentations    |
+| `InstrumentationModule`               | `io.opentelemetry.javaagent.extension.instrumentation`        | Create new instrumentations            |
```
laurit, line 408: as far as I can tell this isn't a complete list of available SPIs, but I guess it is fine for start

content/en/docs/zero-code/java/agent/extensions.md
```diff
@@ -49,8 +68,1083 @@ extensions:
...
+
+### Dependency Management
+
+Extensions must carefully manage their dependencies to avoid conflicts with the
+agent and application.
+
+#### Dependencies Provided by Agent (use `compileOnly`)
+
+These APIs are available at runtime from the agent class loader:
```
laurit: Not completely true, some of these are in boot loader. Could instead say that these apis are provided by the agent.

content/en/docs/zero-code/java/agent/extensions.md
```diff
@@ -49,8 +68,1083 @@ extensions:
...
+implementation("org.apache.commons:commons-lang3:3.19.0")
+implementation("com.google.guava:guava:33.0.0-jre")
+```
+
+**Critical:** Extensions cannot load dependencies from separate JAR files. All
+dependencies must be merged into a single shadow JAR.
+
+### Using @AutoService
+
+The `@AutoService` annotation automatically generates the required
```
laurit, line 471: for `@AutoService` to work you also need to configure the annotation processor

content/en/docs/zero-code/java/agent/extensions.md
```diff
@@ -49,8 +68,1083 @@ extensions:
...
+`META-INF/services/io.opentelemetry.sdk.autoconfigure.spi.AutoConfigurationCustomizerProvider`
+with your class name, but much more maintainable.
+
+---
+
+## Extension Point Reference
+
+### AutoConfigurationCustomizerProvider
+
+The main entry point for customizing SDK configuration. This allows you to:
```
laurit, line 505: Do we have sdk documentation where we could link for this? Should it be mentioned that these won't work with the declarative configuration?

content/en/docs/zero-code/java/agent/extensions.md
```diff
@@ -49,8 +68,1083 @@ extensions:
...
+    return System.getenv().getOrDefault("APP_VERSION", "unknown");
+  }
+}
+```
+
+## Application-Agent Communication
+
+### Accessing the Current Span
+
+In instrumentation advice classes, use `Java8BytecodeBridge` to access the
```
laurit: Should we explain why and when one needs to use `Java8BytecodeBridge` instead of `Span.current()` or `Context.current()`?
Jumping to advice while skipping `InstrumentationModule` and `TypeInstrumentation` feels a bit confusing.

content/en/docs/zero-code/java/agent/extensions.md
```diff
@@ -49,8 +68,1083 @@ extensions:
...
+
+    // Add events
+    span.addEvent("custom.event");
+  }
+}
+```
+
+### Sharing Data via Context
+
+Use the OpenTelemetry `Context` API to share data throughout the request
```
laurit: Imo this is also confusing. Could start with introduction to instrumentation api, context customization is really one of the less used aspects there.

content/en/docs/zero-code/java/agent/extensions.md
```diff
@@ -49,8 +68,1083 @@ extensions:
...
+directory inside the agent JAR file, so we can use a Gradle build task to:
+
+1. Download the official OpenTelemetry Java agent
+2. Extract its contents
+3. Add your extension JAR(s) into the `extensions/` directory
+4. Repackage everything into a single JAR
+
+### The `extendedAgent` Gradle Task
+
+Add the following to your extension project's `build.gradle` file:
```
sydneyosbornagility: this assumes people are using gradle - could we also have a maven example?

content/en/docs/zero-code/java/agent/extensions.md
```diff
@@ -49,8 +68,1083 @@ extensions:
...
+**2. Create the SpanProcessor:**
+
+```java
+package com.example.myextension;
+
+import io.opentelemetry.context.Context;
+import io.opentelemetry.sdk.trace.ReadWriteSpan;
+import io.opentelemetry.sdk.trace.ReadableSpan;
+import io.opentelemetry.sdk.trace.SpanProcessor;
+
```
sydneyosbornagility: How to handle logging might be worth covering.  I found the explanation of what logger to use and how the agent pipes logs to sfl4j to be very helpful in the ticket.  Might be worth discussing here and including an example as you'll want to start out with some minimal logging to ensure your extension is actually running.

content/en/docs/zero-code/java/agent/extensions.md
zeitlinger: yes, but I'd only do it whet the examples it the agent repo have been converted

vitorvasc:
Hi folks, since it's been a couple weeks since the last activity, I'm just checking in!

Do we have any updates on this? 🙂

jaydeluca:
Still a WIP, i will put it in draft in the meantime

content/en/docs/zero-code/java/agent/extensions.md
```diff
@@ -277,33 +286,451 @@ task extendedAgent(type: Jar) {
...
 
-- Verify the version number exists at
-  [Maven Central](https://central.sonatype.com/artifact/io.opentelemetry.javaagent/opentelemetry-javaagent)
-- If using a SNAPSHOT version, add the Sonatype snapshots repository:
+**1. Create a Gradle project:**
 
 ```groovy
+plugins {
+  id "java"
+  id "com.gradleup.shadow" version "9.2.2"
```
SylvainJuge: In the examples in the instrumentation repository we also have the muzzle plugins, does omitting them here is intentional for simplicity ? While it works for simple extensions, doing any complex instrumentation will require it when there are helper classes.

Also, as I painfully learned we also need a few extra dependencies in muzzle/codegen configurations, do you know if it would make sense to make the muzzle plugin automatically handle those ? (see https://cloud-native.slack.com/archives/C014L2KCTE3/p1767802448072319) I think that doing so could help to keep the extension gradle build simpler and prevent some version mismatches (the dependency versions would then be enforced by choosing the muzzle plugin version)

content/en/docs/zero-code/java/agent/extensions.md
SylvainJuge: I've opened https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/15793 in the agent repo for migration.

content/en/docs/zero-code/java/agent/extensions.md
jaydeluca: i am now using code snippets from the examples repo and the extensions example in the main repo, where possible

content/en/docs/zero-code/java/agent/extensions.md
jaydeluca: > does omitting them here is intentional for simplicity 

yes I was just trying to provide a minimal setup, but it sounds like it would be useful to cover that in some way

jaydeluca: i completely revamped the custom instrumentation sections, and included information about these plugins. I didn't add the other dependency stuff yet until we sort out some of the ideas we are throwing around in terms of updating those plugins or creating a new plugin to encompass the other two

chalin:
👋 Please rebase this PR over the latest `main`: it predates our link-checker [switch to Lychee][#10911] and can hit CI failures unrelated to your changes. This PR also changes a submodule pointer that `main` has since moved (`content-modules/opentelemetry-java-instrumentation`: this PR sets it to `351b697`; `main` is now at `9cf4fba`) — merging as-is would regress the submodule state on `main`. For update guidance, see [#10990][]. Thanks for your contribution!

[#10911]: https://github.com/open-telemetry/opentelemetry.io/pull/10911
[#10990]: https://github.com/open-telemetry/opentelemetry.io/issues/10990

PR #8268, opened 2025-10-29
Fix temporality preference environment variable name

snoopdouglas:
The [Python SDK docs](https://opentelemetry.io/docs/languages/python/exporters/) say that the env var to change the temporality preference is `OTEL_EXPORTER_METRICS_TEMPORALITY_PREFERENCE`, but this is incorrect - it should be `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE`.

See: https://github.com/open-telemetry/opentelemetry-python/blob/e0ac0a46cac4d5d9bc6b0562ce5a3c32329330cd/opentelemetry-sdk/src/opentelemetry/sdk/environment_variables/__init__.py#L781

### DIFF
--- a/.github/workflows/pr-actions.yml
+++ b/.github/workflows/pr-actions.yml
@@ -109,16 +109,17 @@ jobs:
         with:
           egress-policy: audit
 
-      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
-        with:
-          fetch-depth: 999
-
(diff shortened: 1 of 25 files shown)


snoopdouglas:
Have now signed the CLA.

tammy-baylis-swi (approved):
Thank you for this, lgtm!

>  could there be any other places in the docs where these environment variables aren't updated?

For this particular env var, I think that's it after a grep of this repo. ✔️

PR #8325, opened 2025-11-03
Added anchors to internal-telemetry.md

Andygol:
This PR added explicit anchors to avoid Prettier replacing `_` --> `\_` in the anchors on localized pages.


chalin (approved):
I confirm that the `_` causes problems when the heading ID is given explicitly. Thanks for this fix @Andygol.

chalin:
Editorial. Merging.

PR #8344, opened 2025-11-06
Make Collector docs a sidebar root

chalin:
- Kind of a followup to #8001
- Makes the Collector section of the docs its own sidebar root. See screenshots below
- **Preview**: https://deploy-preview-8344--opentelemetry.netlify.app/docs/collector/

WDYT @tiffany76 @jaydeluca et all?

### Screenshots

Before:

> image

After:

> image

### DIFF
--- a/content/en/docs/collector/_index.md
+++ b/content/en/docs/collector/_index.md
@@ -5,6 +5,7 @@ aliases: [./about]
 cascade:
   vers: 0.138.0
 weight: 270
+sidebar_root_for: self
 ---
 
 ![OpenTelemetry Collector diagram with Jaeger, OTLP and Prometheus integration](img/otel-collector.svg)


content/en/docs/collector/_index.md
```diff
@@ -5,6 +5,7 @@ aliases: [./about]
 cascade:
   vers: 0.138.0
 weight: 270
+sidebar_root_for: self
```
jaydeluca, line 8: I like that it declutters the navigation, especially when there are so many options now. 

I don't like that it requires leaving the current page in order to get back to a state where you can view the entire navigation tree (when you click on the "Collector ^" link). I don't see that as a blocker, but it was unexpected behavior from my perspective.

tiffany76:
I agree with both of @jaydeluca's comments. 

I like the idea! With the new information architecture, the number of top-level headings will shrink quite a bit (we'll go from 17 to 6), so I don't know how necessary it will be. But I do like it.

chalin:
Oh, if the section will be simplified, then I'd skip the section sidebar rooting. I was proposing this here because I though that the Collector section would get richer and potentially deeper. I'd vote to close this PR. WDYT?

tiffany76:
Yeah, here's the architecture plan: https://www.mindomo.com/mindmap/c6ececd0512d46edb8f5048d19f18de1

I'm fine closing this for now. Once we've made the IA changes, we can always revisit it.

PR #8350, opened 2025-11-07
Use full-qualified name for OpenTelemetry Collector role name

headcr4sh:
When using the built-in task `ansible.builtin.include_role`, a role from a collection must be referenced using its fully-qualified name.

(I just tried to follow the example in my test lab and figured out that this small change is probably needed here)

I assume, that the roles from the Ansible Galaxy collection `grafana.grafana` were installed in the default `roles` folder of the writer of this documentation so the "wrong" role name never was an issue..?


tiffany76:
Thanks for putting up this PR, @headcr4sh! We don't usually update old blog posts, but if this is a technical inaccuracy, we probably should. @ishanjainn, could you confirm that this update is correct? Thanks!

ishanjainn:
My bad for the late review, Looks good to me!

PR #8451, opened 2025-11-19
Add backstage to integrations

svrnm:
https://backstage.io/docs/tutorials/setup-opentelemetry


data/registry/application-integration-js-backstage.yml
```diff
@@ -0,0 +1,20 @@
+title: Backstage
+registryType: application integration
+language: js
+tags:
+  - backstage
+  - javascript
```
maryliag, line 6: their code seems to be majority in typescript. Do we add that as well or keep more generic with js only?

svrnm: we can add this as a tag, but not as a dedicated language

svrnm:
```suggestion
  - javascript
  - typescript
```

PR #8464, opened 2025-11-20
[ja] docs(i18n): Translate content/ja/docs/collector/components/connector.md

kohbis:
## English page

https://opentelemetry.io/docs/collector/components/connector/

### File with SHA
https://github.com/open-telemetry/opentelemetry.io/blob/1c2b0563e8e66ef0952c442e3662e4bec18a8762/content/en/docs/collector/components/connector.md

## Preview

https://deploy-preview-8464--opentelemetry.netlify.app/ja/docs/collector/components/connector/

### DIFF
--- a/content/ja/docs/collector/components/connector.md
+++ b/content/ja/docs/collector/components/connector.md
@@ -1,0 +1,34 @@
+---
+title: コネクター
+description: OpenTelemetry Collectorで利用可能なコネクターの一覧
+weight: 340
+default_lang_commit: 1c2b0563e8e66ef0952c442e3662e4bec18a8762
+# prettier-ignore
+cSpell:ignore: countconnector datadogconnector exceptionsconnector failoverconnector forwardconnector grafanacloudconnector otlpjsonconnector roundrobinconnector routingconnector servicegraphconnector signaltometricsconnector slowsqlconnector spanmetricsconnector sumconnector
+---
+
+コネクターは、2つのパイプラインを接続し、エクスポーターとレシーバーの両方として機能します。
+コネクターの詳細な設定方法については、[Collectorの設定ドキュメント](/docs/collector/configuration/#connectors)を参照してください。
+
+<!-- BEGIN GENERATED: connector-table -->
+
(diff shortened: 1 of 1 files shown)


ymotongpoo (approved):

content/ja/docs/collector/components/connector.md
```diff
@@ -0,0 +1,34 @@
...
+# prettier-ignore
+cSpell:ignore: countconnector datadogconnector exceptionsconnector failoverconnector forwardconnector grafanacloudconnector otlpjsonconnector roundrobinconnector routingconnector servicegraphconnector signaltometricsconnector slowsqlconnector spanmetricsconnector sumconnector
+---
+
+コネクターは、2つのパイプラインを接続し、エクスポーターとレシーバーの両方として機能します。
+コネクターの詳細な設定方法については、[Collectorの設定ドキュメント](/docs/collector/configuration/#connectors)を参照してください。
+
+<!-- BEGIN GENERATED: connector-table -->
+
+| 名前                                                                                                                                       | ディストリビューション[^1] |
```
vitorvasc, line 15: Hi @jaydeluca, quick question: the table headings were translated after the `<!-- BEGIN GENERATED: connector-table -->` mark. Could it somehow be affected when the next update is out?

In that case, we're still doing it manually, right? We don't have the watcher living in our repo so far — please correct me if I'm wrong. 😅

jaydeluca: it might break things right now, but I think it makes more sense to remove any of the heading/copy/footnotes etc from the generated code to allow for having that customized per language. 

So I think we should move in that direction, where within the `<!--` markers we should just keep the table contents, and then everything else should be maintained manually. 

Also, I hadn't considered keeping these translated pages up to date as well, but that should be a pretty simple update

vitorvasc: @jaydeluca thanks! I think we can merge this one for now. I've also opened #8644 to track the subsequent steps in the meantime.

Issue #8450, opened 2025-11-19
Setup i18n entries for alert tags


chalin:
_Originally posted by @chalin in https://github.com/open-telemetry/opentelemetry.io/issues/8445#issuecomment-3553279271_

> Possible followup action is to setup i18n entries for the [alert] tags if we don't have them already?

/cc @svrnm @vitorvasc

### Tasks

- [ ] #8469 - note that the new alert IDs aren't being picked up yet. To track the fix to this, see #8481
- [ ] ...

svrnm:
What needs to be done, see https://gohugo.io/render-hooks/blockquotes/#example

To override the label, create these entries in your i18n files:

```yml
caution: Caution
important: Important
note: Note
tip: Tip
warning: Warning
```

PR #8469, opened 2025-11-20
[pt][i18n] Add Portuguese translations for alert types

vitorvasc:
- Contributes to #8450 
- Creates the `pt.toml` file and adds i18n entries for the [alert] tags (reference: https://gohugo.io/render-hooks/blockquotes/#example)

### DIFF
--- a/i18n/pt.toml
+++ b/i18n/pt.toml
@@ -1,6 +1,9 @@
-# Docsy overrides
+# TODO: if/once Hugo is fixed, rename this file to pt-br.toml
 
-# Alert type
+# Blockquote alert types, see:
+# https://gohugo.io/render-hooks/blockquotes/#alerts
+# TODO: Upstream to Docsy
+
 caution = "Cuidado"
 important = "Importante"
 note = "Nota"


vitorvasc:
@chalin @svrnm there's something weird going on here and I'm not sure why/how. Let me try to break down what I'm seeing:

Just creating the `i18n/pt.toml` and adding the override for the alert tags doesn't seem to work. For example, see the Portuguese version of the Traces page: https://deploy-preview-8469--opentelemetry.netlify.app/pt/docs/concepts/signals/traces/

After adding the `i18n/pt.toml` file, locally building the project started showing these warnings:

```
WARN  Failed to get translated string for language "pt" and ID "ui_search": %!s(<nil>)
WARN  Failed to get translated string for language "pt" and ID "footer_all_rights_reserved": %!s(<nil>)
WARN  Failed to get translated string for language "pt" and ID "post_view_this": %!s(<nil>)
WARN  Failed to get translated string for language "pt" and ID "post_edit_this": %!s(<nil>)
WARN  Failed to get translated string for language "pt" and ID "post_create_child_page": %!s(<nil>)
WARN  Failed to get translated string for language "pt" and ID "post_create_issue": %!s(<nil>)
WARN  Failed to get translated string for language "pt" and ID "feedback_title": %!s(<nil>)
WARN  Failed to get translated string for language "pt" and ID "feedback_question": %!s(<nil>)
WARN  Failed to get translated string for language "pt" and ID "feedback_positive": %!s(<nil>)
WARN  Failed to get translated string for language "pt" and ID "feedback_negative": %!s(<nil>)
WARN  Failed to get translated string for language "pt" and ID "post_last_mod": %!s(<nil>)
```     

These warnings can also be seen in the PR checks: https://github.com/open-telemetry/opentelemetry.io/actions/runs/19542025476/job/55950872988?pr=8469

The strange part is that somehow these changes affected other components that weren't being translated before, like the "View page source" or "Edit this page" text on the right sidebar, and also the "Was this page helpful?" at the bottom.

Examples:

**Current content on the OTel.io website:** https://opentelemetry.io/pt/docs/concepts/signals/traces/

image

image

**Preview:** https://deploy-preview-8469--opentelemetry.netlify.app/pt/docs/concepts/signals/traces/

image

image

chalin:
Thank you for this PR and the careful analysis @vitorvasc. I'll dig deeper as soon as I have the time, later today or tomorrow. Btw, Docsy has https://github.com/google/docsy/blob/main/i18n/pt-br.toml, which I thought was already being used in our repo, but it seems like it isn't/wasn't until you added `pt.toml` -- which Docsy doesn't have. So I wonder if the presence of `pt.toml` has Hugo merge the two? 🤔

i18n/pt.toml
```diff
@@ -0,0 +1,8 @@
+# Docsy overrides
+
+# Alert type
```
chalin:
```suggestion
# TODO: if/once Hugo is fixed, rename this file to pt-br.toml

# Blockquote alert types, see
# https://gohugo.io/render-hooks/blockquotes/#alerts
# TODO: Upstream to Docsy
```

chalin:
- #8481

chalin:
@vitorvasc - I've pushed a commit with added entries to the warning skip list. That will allow the build to pass. As I mention in #8481, I'll need to dig deeper to see if we can address the source of the problem (a Hugo bug).

chalin:
Ok, so the changes in this PR enable the `pt-br.toml` translations provided by Docsy, but the alert IDs are **not** being picked up due to bug(s) in Hugo. What do you prefer doing @vitorvasc. I'd vote to merge this and we can address the alert ID issue separately.

vitorvasc:
> Ok, so the changes in this PR enable the `pt-br.toml` translations provided by Docsy, but the alert IDs are **not** being picked up due to bug(s) in Hugo. What do you prefer doing @vitorvasc. I'd vote to merge this and we can address the alert ID issue separately.

I'm also fine with merging it and handling the ID issue separately. Thanks for taking it!

PR #8504, opened 2025-11-24
Update Docsy to 0.13.0

chalin:
**Preview**: https://deploy-preview-8504--opentelemetry.netlify.app/

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2,7 +2,7 @@
 [submodule "themes/docsy"]
 	path = themes/docsy
 	url = https://github.com/google/docsy.git
-	docsy-pin = v0.12.0-80-g39ee7caf
+	docsy-pin = v0.13.0
 	docsy-note = "2024-04-01 Switching to google/docsy.git from cncf/docsy.git since we don't have any CNCF customizations."
 	docsy-reminder = "Ensure that any tag referenced by `docsy-pin` is present in the remote repo (url), otherwise add (push) the tags to the repo."
 [submodule "content-modules/opentelemetry-specification"]


tiffany76:
I have no idea if this behavior was introduced with this PR or not, but as I was checking the deploy preview, I noticed that the right side page nav isn't selecting the correct item. To recreate:
1. Go to docs > Collector > Install the Collector.
2. Click on Nomad in the right nav. For me, the nav selection moves all the way to RPM installation.
3. Click on macOS in the right nav. For me, the nav selection moves to MSI Installation.

vitorvasc:
I hadn't noticed this before, but I'm seeing the same behavior @tiffany76 mentioned.

I also managed to reproduce it on the Docsy website:

1. I opened the [Release 0.13.0](https://www.docsy.dev/blog/2025/0.13.0/) page.
2. Clicked the "most upvoted" link under the "Highlights" section.
3. It scrolled down to the `#active-toc-entry-tracking` section.
4. In the TOC list, the "Language menu visibility" item is highlighted. 

image

----

@tiffany76 good catch!

chalin:
Right, you're both hitting a known issue and currently limitation of ScrollSpy and how it has to be configured in Docsy.

Let me first confirm with you @tiffany76 and @vitorvasc that the active-TOC entry that is highlighted is actually visible in the viewport, right? And if you scroll just a tiny bit up, the active-TOC entry pops back to the top-most element (probably the one you would have expected to be hightlighted in the first place), right?

I've created an upstream issue to track the discussion and resolution of this surprising behavior:

- https://github.com/google/docsy/issues/2417

vitorvasc:
> the active-TOC entry that is highlighted is actually visible in the viewport, right?

That's correct! 

So this isn't something related to this update, then. Thanks, @chalin!

PR #8512, opened 2025-11-25
Write-up of context propagation techniques in OBI

grcevski:
- [X] I have read and followed the
      [contribution guidelines](https://opentelemetry.io/docs/contributing/),
      including the **First-time contributing?** note.
- [ ] This PR includes **AI generated content**, and I have read and conform to
      the
      [Generative AI Contribution Policy](https://github.com/open-telemetry/community/blob/main/policies/genai.md).

This PR adds OBI documentation on how the parent child correlation works for different technologies, for the purpose of context propagation.

Relates to https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/903

### DIFF
--- a/content/en/docs/zero-code/obi/context-propagation.md
+++ b/content/en/docs/zero-code/obi/context-propagation.md
@@ -1,0 +1,141 @@
+---
+title: Context propagation implementation details
+linkTitle: Context propagation
+description: Learn how OBI does context propagation for various languages and frameworks.
+weight: 24
+---
+
+## Introduction
+
+To support distributed tracing for HTTP and other protocols, OBI performs automatic context
+propagation, from incoming to outgoing requests. This context propagation works across protocols, 
+for example, correlation may occur between incoming HTTP request and an outgoing SQL request.
+
+In general, performing context propagation depends on how the application handles hand-off between
+worker threads. In some cases, the worker threads are not actual OS threads, but they can be an 
+abstraction on top of OS threads (e.g. `goroutines`, `coroutines`, `green threads`, `virtual threads`).
+Therefore, it's very difficult to implement a single correct context propagation approach that works
+for all programming languages and frameworks.
+
+OBI implements various approaches to tackle specific programming languages and frameworks, but
+they are by no means complete, and will fail to perform correct correlation in certain scenarios.
+This documents attempts to capture some of the details, and provide guidance on what works and
+what doesn't.
+
+## Implementation details
+
+In this section we'll go over each individual approach that is currently implemented and explain
+the scope.
+
+### Go programming language
+
+The Go threading model is based on `goroutines`, where many `goroutines` map to a single OS thread.
+Therefore, in order to perform correct correlation of incoming to outgoing requests, OBI tracks the
+`goroutines` lifecycle. We attach probes to `runtime.newproc1` and `runtime.goexit1`, to track the
+parent to child `goroutine` relationship, by keeping track the `goroutine` relationships in a 
+in-memory map.
+
+At a time of an outgoing client request, we use the parent to child relationship map, up to a depth of 3,
+to lookup any, still active, incoming request that has launched the outgoing request goroutine.
+
+Special consideration is taken for correct context propagation for `gRPC`, because the 
+outgoing requests are all handled by the so called `loopyWriter`, which is a single goroutine 
+handling the outgoing requests. In this specific scenario, we use the `HTTP2`/`gRPC` stream ID, 
+as a unique key to identify the original request, when it's appended to the write queue when
+`google.golang.org/grpc.(*ClientConn).NewStream` is invoked.
+
+### Node.js
+
+The `Node.js` runtime has an async request queue that is handling all incoming and outgoing requests.
+When an incoming request is handled, any outgoing requests, typically done in async fashion to
+avoid holding up the event loop, are added on the event queue loop as separate requests. To correctly
+propagate context between incoming and outgoing requests, we must know which async operation was 
+scheduled by what other async operation.
+
+Prior to `Node.js` 20 we used to track the async event loop correlation, by injecting probes into
+`EmitAsyncInit` and `AsyncReset`. However, from `Node.js` 20 and onward, more of the runtime 
+handling was rewritten in JavaScript, and the eBPF probe approach misses number of critical correlation
+steps.
+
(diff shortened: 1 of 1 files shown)


content/en/docs/zero-code/obi/context-propagation.md
```diff
@@ -0,0 +1,141 @@
...
+The Go threading model is based on `goroutines`, where many `goroutines` map to a single OS thread.
+Therefore, in order to perform correct correlation of incoming to outgoing requests, OBI tracks the
+`goroutines` lifecycle. We attach probes to `runtime.newproc1` and `runtime.goexit1`, to track the
+parent to child `goroutine` relationship, by keeping track the `goroutine` relationships in a 
+in-memory map.
+
+At a time of an outgoing client request, we use the parent to child relationship map, up to a depth of 3,
+to lookup any, still active, incoming request that has launched the outgoing request goroutine.
+
+Special consideration is taken for correct context propagation for `gRPC`, because the
```
dashpole, line 41: Did you already implement the "metadata injection" portion of context propagation for gRPC, or is this referring to go auto-instrumentation?  Out of curiosity, are languages below not supported because we haven't implemented it yet, or because the approach used in Go won't work in those languages?

content/en/docs/zero-code/obi/context-propagation.md
```diff
@@ -0,0 +1,141 @@
...
+This approach works well for single threaded programming languages and frameworks,
+for example `Python` applications that do not use `asyncio`. The OpenJDK internal
+HTTP client and the Apache HTTP client also work, because the outgoing connections
+are 'inspected' on the same thread that handles the incoming request.
+
+The generic correlation doesn't work for `HTTP2`/`gRPC`.
+
+## What doesn't work
+
+- Any reactive programming frameworks.
```
dashpole, line 124: I like the phrasing of "applications that do FOO, such as reactive frameworks" you use below, since it helps me understand what it is doing that OBI can't track.  Some popular examples might also be helpful.

content/en/docs/zero-code/obi/context-propagation.md
```diff
@@ -0,0 +1,141 @@
...
+- Application frameworks with thread pools that handle the full connection
+lifecycle on the outgoing worker threads. 
+- Application frameworks that handle all incoming traffic on the same thread,
+before handing over the request to be served by a worker thread. An example of this
+the `Puma reactor` used in `Ruby on Rails`.
+- `.net` `async`/`await` create number of complex thread pools that are unpredictable
+in how the work is scheduled.
+
+## Future work
```
dashpole, line 133: I would probably exclude this from the website, since it is speculative.

content/en/docs/zero-code/obi/context-propagation.md
```diff
@@ -0,0 +1,141 @@
...
+At the time `ngx_http_upstream_init` is invoked, we start tracking the connection information
+of the incoming request, while when `ngx_event_connect_peer` returns, we create a mapping
+of the incoming request to the outgoing file descriptor. At the time OBI handles the outgoing 
+request, it looks up the outgoing call file descriptor and fetches the connection information
+of the incoming request, managing to correlate both. In a sense, this is very similar to how
+we lookup the information for the `Node.js` event loop.
+
+`nginx` correlation doesn't for for `HTTP2`/`gRPC`.
+
+### Generic approach
```
dashpole, line 87: Maybe call this language-agnostic?

dashpole (commented):
This is really useful, and I think starting with this scope is good.

I'm imagining that users will probably come to this page with questions like "will this work with my java Spring Boot application?" (my google search suggests it might work with Spring MVC, but not with Spring WebFlux, which is reactive, or Java Virtual Threads...). With some digging I think people can use what you've written to answer their question here (e.g. is my framework reactive?).  But I also think that the more we can enumerate popular examples of things that will work, or popular examples of things that won't work, the fewer user questions we will get.

content/en/docs/zero-code/obi/context-propagation.md
```diff
@@ -0,0 +1,141 @@
...
+The fallback approach of incoming to outgoing request correlation, for the purpose of
+distributed trace context propagation, relies on thread ID correlation. The following
+section outlines all the techniques we currently employ to match the requests by
+thread ID:
+
+1. Same thread. OBI tries to detect if the incoming and outgoing requests are handled
+by the same thread. It detects if multiple current outgoing requests are handled by
+the same thread and marks the correlation information as invalid. This helps us
+prevent incorrect correlation, when the application framework handles multiple 
+connections on the same thread.
```
aabmass, line 98: 👍 that was my biggest question for asyncio, if OBI would generate incorrect parent-child relationships

content/en/docs/zero-code/obi/context-propagation.md
grcevski: We were just discussing this part at the SIG, so I'll summarise here. What we have for Go gRPC at the moment is using the `bpf_probe_write_user` helper, which is banned on locked kernels and it's Go specific. We have new design that will use the same approach as for HTTP and implement this for all languages. We need to change a few internal data structures to hold the streamID, since it's no longer a simple connection based key.

grcevski:
> This is really useful, and I think starting with this scope is good.
> 
> I'm imagining that users will probably come to this page with questions like "will this work with my java Spring Boot application?" (my google search suggests it might work with Spring MVC, but not with Spring WebFlux, which is reactive, or Java Virtual Threads...). With some digging I think people can use what you've written to answer their question here (e.g. is my framework reactive?). But I also think that the more we can enumerate popular examples of things that will work, or popular examples of things that won't work, the fewer user questions we will get.

Good point. I'll try to make a list of examples which work and which don't for the various programming languages.

svrnm:
@grcevski move this out of draft whenever you think @open-telemetry/docs-approvers should take a look

MrAlias (commented):
Overall, this looks good. I think it needs an update at this point:

- Missing the Java task-to-thread hand-off support.
- There is some Puma specific support now
- There is a Node.js async callback context-refresh mechanism now (as far as I can tell)

I think these are the main points to improve here, but could also see tightening the gRPC scope up a bit and adding a framework example matrix.

I'm happy to expand on this review, or help take over if you would find this useful @grcevski.

Again, great start. Thanks for jumping in on this. 😄

PR #8554, opened 2025-12-01
Sort CVEs

trask:
Sort the CVEs (desc) on https://opentelemetry.io/docs/security/cve/

(now that the data is being updated, see https://github.com/open-telemetry/sig-security/issues/188)


theletterf:
Thanks!

PR #8576, opened 2025-12-04
[Collector] Update unmaintained components message

vitorvasc:
- Updates the unmaintained components message to use GitHub markdown alert styling
- Moves the alert above the table (cc @jaydeluca @mx-psi WDYT about this?) 

----

**Before**

Current version: https://opentelemetry.io/docs/collector/components/receiver/

image

----

**After**

Preview: https://deploy-preview-8576--opentelemetry.netlify.app/docs/collector/components/receiver/

image


content/en/docs/collector/components/_includes/unmaintained-components-msg.md
```diff
@@ -0,0 +1,9 @@
+---
+---
+
+{{% alert title="Note" color="warning" %}}
+
+Components marked with :warning: are unmaintained and have no active codeowners.
+They may not receive regular updates or bug fixes.
```
jaydeluca, line 7: nice! i do think this looks nicer than the existing. I'm on the fence about whether I think it should be at the top or bottom of the page, but I think the top works 👍

vitorvasc: I don't have strong opinions on the position either. I tried to find some logic across similar alerts — like the "Important" one in [Language APIs & SDKs  > Status and Releases](https://opentelemetry.io/docs/languages/#status-and-releases), but even on that same page, the next section ("API references") shows an "Information" alert after the list, so... 😅

Issue #8112, opened 2025-10-16
[CI] Add bot action to auto-respond to first-timers asking "assign to me / can I work on this"


chalin:
For example, see  this comment _originally posted by @svrnm in [#8105](https://github.com/open-telemetry/opentelemetry.io/issues/8105#issuecomment-3409436369)_:

> ... sure, give it a try, note that we are currently not assigning issues, so please provide a draft PR early  

It would be nice to have a bot action to auto-respond to such inquiries. I think I've seen this sort of bot action in another CNCF-project repo.

A canned response might be something like:

> Thank you for your interest in [contributing](https://opentelemetry.io/docs/contributing/) to this project. If this is your [first time contributing](https://opentelemetry.io/docs/contributing/issues/#first-issue), then carefully read through [Fixing an existing issue](https://opentelemetry.io/docs/contributing/issues/#fixing-an-existing-issue). As noted there, we don't assign issues unless you're a regular contributor.

pratik-mahalle:
Hey @svrnm, I have created the [pr](https://github.com/open-telemetry/opentelemetry.io/pull/8733) for this issue. Please have a look on this and let me know if we can improve this workflow.
Thanks

Issue #8421, opened 2025-11-15
[infra] Drop git submodules, keep the git external deps


chalin:
The use of Git submodules has made for a bumpy ride for some contributors and headaches for maintainers. I've been giving this some thought for a while, and we have 80% of a non-gitsubmodule solution already in place.

Anyone object to dropping submodules from this repo? We'd essentially keep the current scripts and `.gitsubmodule` file as a **deps manifest**, but drop the git submodules from version control.

What will change for contributors & the build process? Nothing, the scripts we have in place (with small adjustments), will emulate git submodules by git cloning them manually rather than relying on `git submodule update` etc.

@open-telemetry/docs-maintainers - thoughts? Upvote if you agree.

cartermp:
I'd love that.

svrnm:
+1 for that one

YASHMAHAKAL:
@svrnm, i would like to take this one !! can i be assigned ??

svrnm:
@YASHMAHAKAL this is not an issue that can be assigned easily as it requires a lot of knowledge about the repository internals.

PR #8733, opened 2026-01-05
[CI] Add bot action to auto-respond to first-timers asking for issue assignment

pratik-mahalle:
- [x] I have read and followed the
      [contribution guidelines](https://opentelemetry.io/docs/contributing/),
      including the **First-time contributing?** note.
- [ ] This PR includes **AI generated content**, and I have read and conform to
      the
      [Generative AI Contribution Policy](https://github.com/open-telemetry/community/blob/main/policies/genai.md).

Adds a GitHub Actions workflow that automatically responds to issue comments
containing common phrases like 'assign to me', 'can I work on this', etc.

### Related Issue
Closes https://github.com/open-telemetry/opentelemetry.io/issues/8112

### DIFF
--- a/.github/workflows/first-timer-response.yml
+++ b/.github/workflows/first-timer-response.yml
@@ -6,13 +6,13 @@ on:
     types: [created]
 
 permissions:
-  contents: read
   issues: write
 
 env:
   ISSUE_NUM: ${{ github.event.issue.number }}
   COMMENT_AUTHOR: ${{ github.event.comment.user.login }}
-  CONTRIBUTING_URL: https://opentelemetry.io/docs/contributing/issues/#fixing-an-existing-issue
+  URL_FIRST_TIME_CONTRIBUTING: https://opentelemetry.io/docs/contributing/#first-time-contributing
+  URL_FIXING_EXISTING_ISSUE: https://opentelemetry.io/docs/contributing/issues/#fixing-an-existing-issue
 
 jobs:
   respond-to-assignment-request:
@@ -80,13 +80,14 @@ jobs:
         env:
           GH_TOKEN: ${{ steps.otelbot-token.outputs.token }}
           COMMENT_AUTHOR: ${{ github.event.comment.user.login }}
-          CONTRIBUTING_URL: ${{ env.CONTRIBUTING_URL }}
+          URL_FIRST_TIME_CONTRIBUTING: ${{ env.URL_FIRST_TIME_CONTRIBUTING }}
+          URL_FIXING_EXISTING_ISSUE: ${{ env.URL_FIXING_EXISTING_ISSUE }}
           ISSUE_NUM: ${{ env.ISSUE_NUM }}
           GITHUB_REPOSITORY: ${{ github.repository }}
         run: |
           RESPONSE_BODY="Thank you for your interest in contributing to this project, @$COMMENT_AUTHOR! 🙏
 
-          If this is your first time contributing, then please carefully read through [Fixing an existing issue]($CONTRIBUTING_URL).
+          If this is your first time contributing, please read [First time contributing]($URL_FIRST_TIME_CONTRIBUTING) and [Fixing an existing issue]($URL_FIXING_EXISTING_ISSUE).
 
           As noted there, **we don't assign issues** unless you're a regular contributor. If you don't see any linked PRs and no one else has said they are working on the issue, feel free to submit a draft PR early to signal that you're working on it."
 


.github/workflows/first-timer-response.yml
```diff
@@ -0,0 +1,77 @@
...
+          GH_TOKEN: ${{ steps.otelbot-token.outputs.token }}
+
+      - name: Post response to first-time contributor
+        if: steps.check-membership.outputs.is_member != 'true'
+        run: |
+          RESPONSE_BODY="Thank you for your interest in contributing to this project, @${COMMENT_AUTHOR}! 🙏
+
+          If this is your first time contributing, then please carefully read through [Fixing an existing issue](${CONTRIBUTING_URL}).
+
+          As noted there, **we don't assign issues** unless you're a regular contributor. Feel free to submit a draft PR early to signal that you're working on it."
```
tiffany76: I've not reviewed the GitHub Action part of this, but I wanted to add here that we should ask them to read preceding comments to see if someone else is already working on the issue. Something like the following, but I'm wide open to other wording:

```suggestion
          As noted there, **we don't assign issues** unless you're a regular contributor. If you don't see any linked PRs and no one else has said they are working on the issue, feel free to submit a draft PR early to signal that you're working on it." 
```

.github/workflows/first-timer-response.yml
vitorvasc, line 1: Should we also consider adding a condition to trigger this workflow only on issues tagged as `good first issue`?

For example, this might be useful when someone asks to work on something that isn't "newcomer-friendly" or requires deeper deeper knowledge of the repo, like in this case: https://github.com/open-telemetry/opentelemetry.io/issues/8421#issuecomment-3711766904

@pratik-mahalle @open-telemetry/docs-approvers WDYT?

.github/workflows/first-timer-response.yml
```diff
@@ -0,0 +1,77 @@
...
+
+jobs:
+  respond-to-assignment-request:
+    name: Respond to assignment requests
+    runs-on: ubuntu-latest
+    # Only trigger on issue comments (not PR comments) and when comment matches assignment patterns
+    if: |
+      !github.event.issue.pull_request &&
+      (
+        contains(github.event.comment.body, 'assign this to me') ||
```
jaydeluca: what if we added `lower()` to the body for all of these, and then made the assertions all lowercase to increase chances of matches and make it not case sensitive?

```suggestion
        contains(lower(github.event.comment.body), 'assign this to me') ||
```

.github/workflows/first-timer-response.yml
pratik-mahalle: That makes sense 👍

Adding a condition to trigger this only for issues labeled good first issue sounds like a solid improvement. It helps keep the response accurate and avoids encouraging newcomers to jump into issues that require deeper context or prior repo knowledge.

We could either gate the workflow strictly on the label, or keep the current behavior but add an extra check so the bot response is more cautious when the label isn’t present. Happy to adjust the workflow based on what the docs approvers feel is the right balance here. @vitorvasc

.github/workflows/first-timer-response.yml
pratik-mahalle: Updated this!

.github/workflows/first-timer-response.yml
pratik-mahalle: Great suggestion — making the checks case-insensitive will definitely improve match reliability, especially for first-time contributors who may not follow the exact phrasing.
I’ll update the conditions to wrap github.event.comment.body with lower() and normalize all string comparisons to lowercase. That should keep the intent the same while making the workflow more forgiving. Thanks for pointing this out!

pratik-mahalle: Added please review it once again @jaydeluca

jaydeluca: you still need to add `lower()` to all of these `github.event.comment.body`s

pratik-mahalle: added

pratik-mahalle: but it is failing as saying 
`Unrecognized function: 'lower'. Located at position 50 within expression `

jaydeluca: Close the parenthesis before the text you are matching against

pratik-mahalle: I am not sure, if `lower()` is supported in the github actions!

jaydeluca (approved):

.github/workflows/first-timer-response.yml
vitorvasc: Let's gate the workflow based on the labels for now, so we can monitor how often it's triggered. In the future, if we feel we can remove this rule and improve the bot's response instead, we can move in that direction!

For now, let's just make sure one the issue has one of the following labels:

- `triage:accepted:needs-pr`
- `good first issue`

pratik-mahalle: cool, updated!

chalin (commented):
@vitorvasc and @jaydeluca - LGTM if you're satisfied. We can always iterate. Have you considered asking Copilot for a review?

.github/workflows/first-timer-response.yml
```diff
@@ -0,0 +1,93 @@
+name: Auto-respond to first-time contributors
+# Automatically respond to issue comments asking for assignment
+
+on:
+  issue_comment:
+    types: [created]
+
+permissions:
+  contents: read
```
vitorvasc: Since we're not using a checkout action (for example), we can remove the `contents: read` permission.

```suggestion
```

.github/workflows/first-timer-response.yml
```diff
@@ -0,0 +1,93 @@
...
+  issues: write
+
+env:
+  ISSUE_NUM: ${{ github.event.issue.number }}
+  COMMENT_AUTHOR: ${{ github.event.comment.user.login }}
+  CONTRIBUTING_URL: https://opentelemetry.io/docs/contributing/issues/#fixing-an-existing-issue
```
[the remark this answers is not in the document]
vitorvasc:
```suggestion
  URL_FIRST_TIME_CONTRIBUTING: https://opentelemetry.io/docs/contributing/#first-time-contributing
  URL_FIXING_EXISTING_ISSUE: https://opentelemetry.io/docs/contributing/issues/#fixing-an-existing-issue
```

.github/workflows/first-timer-response.yml
```diff
@@ -0,0 +1,93 @@
...
+        env:
+          GH_TOKEN: ${{ steps.otelbot-token.outputs.token }}
+          COMMENT_AUTHOR: ${{ github.event.comment.user.login }}
+          CONTRIBUTING_URL: ${{ env.CONTRIBUTING_URL }}
+          ISSUE_NUM: ${{ env.ISSUE_NUM }}
+          GITHUB_REPOSITORY: ${{ github.repository }}
+        run: |
+          RESPONSE_BODY="Thank you for your interest in contributing to this project, @$COMMENT_AUTHOR! 🙏
+
+          If this is your first time contributing, then please carefully read through [Fixing an existing issue]($CONTRIBUTING_URL).
```
vitorvasc:
```suggestion
          If this is your first time contributing, please read [First time contributing]($URL_FIRST_TIME_CONTRIBUTING) and [Fixing an existing issue]($URL_FIXING_EXISTING_ISSUE).
```

.github/workflows/first-timer-response.yml
pratik-mahalle: cool

pratik-mahalle:
@vitorvasc Changes have been updated!

PR #8767, opened 2026-01-08
[i18n] Update drifted status for localized content

vitorvasc:
Updates the drifted status for localized content.

Current drift status after this PR:

| Locale | Drifted files | Total files |
| ------ | ------------ | --------- |
| bn | 8 | 17 |
| es | 45 | 85 |
| fr | 27 | 58 |
| ja | 80 | 193 |
| pt | 43 | 115 |
| ro | 2 | 7 |
| uk | 2 | 7 |
| zh | 74 | 160 |

### DIFF
--- a/content/zh/blog/_index.md
+++ b/content/zh/blog/_index.md
@@ -3,4 +3,5 @@ title: 博客
 menu: { main: { weight: 50 } }
 outputs: [HTML, RSS]
 default_lang_commit: c0a5eea5d720b0e075efa87f99dcf58c89106268
+drifted_from_default: true
 ---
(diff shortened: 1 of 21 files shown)


content/fr/docs/concepts/context-propagation.md
```diff
@@ -3,7 +3,7 @@ title: Propagation du contexte
 weight: 10
 description: Présentation du concept permettant le traçage distribué.
 default_lang_commit: 71833a5f8b84110dadf1e98604b87a900724ac33
-drifted_from_default: true
+drifted_from_default: file not found
```
dmathieu, line 6: It seems the file has been moved to a subfolder. Should we do that move? Or would you prefer it's done in another PR?

vitorvasc: We can handle that in another PR! There are other locales containing pages that were either moved or deleted.

I'll also open a meta-issue for each locale so they can easily track that and address those changes individually.

PR #8771, opened 2026-01-08
Extend collector configuration doc

dmathieu:
The `service` package in Collector Core has a pretty extensive documentation of how its configuration works.
https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/README.md

That documentation is rather missplaced, as the service package doesn't handle CLI flags, so this configuration doesn't even happen in there.

This PR aims to integrate the README from that package into the end-user documentation, so we can rewrite the package's README into something that better matches what it does.

### DIFF
--- a/content/en/docs/collector/configuration.md
+++ b/content/en/docs/collector/configuration.md
@@ -919,7 +919,7 @@ recommended.
 The `--set` option has the following limitations:
 
 1. Does not support setting a key that contains a dot `.`.
-2. Does not support setting a key that contains a equal sign `=`.
+2. Does not support setting a key that contains an equal sign `=`.
 3. The configuration key separator inside the value part of the property is
    "::". For example `--set "name={a::b: c}"` is equivalent with
    `--set name::a::b=c`.
@@ -995,7 +995,7 @@ Note that by default the configuration will only print when it is valid, and
 that sensitive information will be redacted. To print a potentially invalid
 configuration, use `--validate=false`.
 
-### How to view sensitive fields?
+### How to view sensitive fields
 
 Use `print-config` with `--mode=unredacted` and
 `--feature-gates=otelcol.printInitialConfig`:
@@ -1004,7 +1004,7 @@ Use `print-config` with `--mode=unredacted` and
 otelcol print-config --mode=unredacted --config=file:examples/local/otel-config.yaml
 ```
 
-### How to print the final configuration in JSON format?
+### How to print the final configuration in JSON format
 
 {{% alert title="Important" color="warning" %}}
 


content/en/docs/collector/configuration.md
```diff
@@ -31,6 +31,17 @@ example:
...
+The `--config` flag accepts either a file path or values in the form of a config
+URI `"<scheme>:<opaque_data>"`. Currently, the OpenTelemetry Collector supports
+the following providers `scheme`:
+
+- **file** - Reads configuration from a file. E.g. `file:path/to/config.yaml`.
+- **env** - Reads configuration from an environment variable. E.g.
+  `env:MY_CONFIG_IN_AN_ENVVAR`.
+- **yaml** - Reads configuration from yaml bytes. E.g.
+  `yaml:exporters::debug::verbosity: detailed`.
+- **http** - Reads configuration from a HTTP URI. E.g. `http://www.example.com`
```
dmathieu: The linter wants this to be `**HTTP**`. I think that would be wrong though, as we describe a config scheme. So having it lower cased is the appropriate thing to do.

How can I ignore this lint failure?

content/en/docs/collector/configuration.md
```diff
@@ -855,14 +866,146 @@ define with this method are merged into the final configuration after all
...
+
+### Complex nested keys
+
+Use dot (`.`) in the pair's name as key separator to reference nested map
+values. For example, `--set outer.inner=value` is translated into this:
+
+```yaml
+outer:
+  inner: value
+```
```
mx-psi, line 890: Is this right? I would expect that you have to use `::` instead

content/en/docs/collector/configuration.md
```diff
@@ -855,14 +866,146 @@ define with this method are merged into the final configuration after all
...
+
+### Map values
+
+Maps can be expressed by enclosing values in `{}`. For example,
+`"--set "key={a: c}"` translates to:
+
+```yaml
+key:
+  a: c
 ```
```
mx-psi: I think this can be summarized as "Use YAML to describe more complex data structures"

content/en/docs/collector/configuration.md
```diff
@@ -855,14 +866,146 @@ define with this method are merged into the final configuration after all
...
+  - memory_limiter
+exporters:
+  - otlp
+  - otlphttp
+  - debug
+extensions:
+  - zpages
+```
+
+## How to examine the final configuration
```
mx-psi, line 978: Let's add a warning that this is experimental functionality

content/en/docs/collector/configuration.md
```diff
@@ -855,14 +866,146 @@ define with this method are merged into the final configuration after all
...
+```
+
+Note that by default the configuration will only print when it is valid, and
+that sensitive information will be redacted. To print a potentially invalid
+configuration, use `--validate=false`.
+
+### How to view sensitive fields?
+
+Use `print-config` with `--mode=unredacted` and
+`--feature-gates=otelcol.printInitialConfig`:
```
mx-psi, line 1001: ditto

content/en/docs/collector/configuration.md
vitorvasc: That makes sense to have it lower cased. We can wrap this line with prettier-ignore instructions:

```suggestion
<!-- prettier-ignore-start -->
- **http** - Reads configuration from a HTTP URI. E.g. `http://www.example.com`
<!-- prettier-ignore-end -->
```

content/en/docs/collector/configuration.md
dmathieu: Both should work

https://github.com/open-telemetry/opentelemetry-collector/blob/4db91572129c14907ca5d939bcf9114629b85ce7/otelcol/flags.go#L48

I used the dot format, as it's what was in the service package readme. I can switch to `::` though.

mx-psi: Ah, alright. I would prefer to switch to `::`, I think it is a bit more consistent

content/en/docs/collector/configuration.md
```diff
@@ -31,6 +31,19 @@ example:
 otelcol --config=customconfig.yaml
 ```
 
+The `--config` flag accepts either a file path or values in the form of a config
+URI `"<scheme>:<opaque_data>"`. Currently, the OpenTelemetry Collector supports
+the following providers `scheme`:
```
evan-bradley:
```suggestion
the following providers for `scheme`:
```

content/en/docs/collector/configuration.md
```diff
@@ -31,6 +31,19 @@ example:
...
+
+- **file** - Reads configuration from a file. E.g. `file:path/to/config.yaml`.
+- **env** - Reads configuration from an environment variable. E.g.
+  `env:MY_CONFIG_IN_AN_ENVVAR`.
+- **yaml** - Reads configuration from yaml bytes. E.g.
```
evan-bradley:
```suggestion
- **yaml** - Reads configuration from a YAML string, with `::` delimiting subpaths. E.g.
```
Unless it's raw bytes, we should mention it's a string. Also, I think `::` is unique to the Collector, so we should call that out.

content/en/docs/collector/configuration.md
```diff
@@ -31,6 +31,19 @@ example:
...
+`yaml:exporters::debug::verbosity: detailed`.
+<!-- prettier-ignore-start -->
+- **http** - Reads configuration from a HTTP URI. E.g. `http://www.example.com`
```
evan-bradley:
```suggestion
- **http** - Reads configuration from an HTTP URI. E.g. `http://www.example.com`
- **https** - Reads configuration from an HTTPS URI. E.g. `https://www.example.com`
```
We should make sure we explicitly call out HTTPS here, just to be exhaustive. Also, I think most people pronounce the letters when saying "HTTP(S)", so "an" works better as an article.

content/en/docs/collector/configuration.md
```diff
@@ -855,14 +868,153 @@ define with this method are merged into the final configuration after all
...
+The `--set` option takes always one key/value pair, and it is used like this:
+`--set key=value`. The YAML equivalent of that is:
+
+```yaml
+key: value
+```
+
+### Complex nested keys
+
+Use dot (`.`) in the pair's name as key separator to reference nested map
```
evan-bradley:
```suggestion
Use two colons (`::`) in the pair's name as key separator to reference nested map
```
Assuming you copied this from the service readme, this is probably outdated from before it was changed to two colons.

evan-bradley (commented):
Overall looks good, thanks for moving this.

content/en/docs/collector/configuration.md
```diff
@@ -855,14 +870,153 @@ define with this method are merged into the final configuration after all
...
+recommended.
+
 {{% alert title="Important" color="warning" %}}
 
-The `--set` option doesn't support setting a key that contains a dot or an equal
-sign.
+The `--set` option has the following limitations:
+
+1. Does not support setting a key that contains a dot `.`.
+2. Does not support setting a key that contains a equal sign `=`.
```
tiffany76:
```suggestion
2. Does not support setting a key that contains an equal sign `=`.
```

content/en/docs/collector/configuration.md
```diff
@@ -855,14 +870,153 @@ define with this method are merged into the final configuration after all
...
+
+```bash
+otelcol print-config --config=file:examples/local/otel-config.yaml
+```
+
+Note that by default the configuration will only print when it is valid, and
+that sensitive information will be redacted. To print a potentially invalid
+configuration, use `--validate=false`.
+
+### How to view sensitive fields?
```
tiffany76:
```suggestion
### How to view sensitive fields
```

content/en/docs/collector/configuration.md
```diff
@@ -855,14 +870,153 @@ define with this method are merged into the final configuration after all
...
+
+Use `print-config` with `--mode=unredacted` and
+`--feature-gates=otelcol.printInitialConfig`:
+
+```bash
+otelcol print-config --mode=unredacted --config=file:examples/local/otel-config.yaml
+```
+
+### How to print the final configuration in JSON format?
```
tiffany76:
```suggestion
### How to print the final configuration in JSON format
```

tiffany76 (approved):
LGTM with a few tiny fixes.

Issue #8839, opened 2026-01-11
Upgrade Hugo to 0.153+, address issues


chalin:
### Prep and cleanup

It seems that the new handling of locales affects how the sidenave gets generated. Quirks are surfacing under Hugo 0.153+. Let's fix these before upgrading:

- [x] #8830
- [x] #8833
- [x] #8834
- [x] #8844
- [x] #8845
- [x] #8848
- [x] #8849

### Tasks

- [x] #8850 - also fixed shortcodes referencing the default language `Site`
- [x] #8932
- Will track separately: #8933

### Related

- https://github.com/google/docsy/issues/2431

chalin:
Reopening as I track more upgrade issues.

chalin:
All done.

PR #8844, opened 2026-01-11
[fr] Move zero-code/go index to its new position

chalin:
- Contributes to #8839
- Moves fr/docs/zero-code/go.md into go/_index.md so as to match the `en` page structure. If we don't do this, the `en` go/autodesk.md page ends up in the wrong place (immediately under zero-code) in the sidenav.
- No change in content. Page marked as "patched"

**Preview**: https://deploy-preview-8844--opentelemetry.netlify.app/fr/docs/zero-code/go/

### Screenshots when using Hugo 0.154.5+

| Before | After |
|--------|--------|
| image | image |

### DIFF
--- a/content/fr/docs/zero-code/go/_index.md
+++ b/content/fr/docs/zero-code/go/_index.md
@@ -2,8 +2,7 @@
 title: Instrumentation Zero-code Go
 linkTitle: Go
 weight: 16
-default_lang_commit: 3d179dbe1270b83aafff0d3b6aa3311afd482649
-drifted_from_default: file not found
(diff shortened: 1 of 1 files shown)


content/fr/docs/zero-code/go.md
```diff
@@ -2,8 +2,7 @@
 title: Instrumentation Zero-code Go
 linkTitle: Go
 weight: 16
-default_lang_commit: 3d179dbe1270b83aafff0d3b6aa3311afd482649
-drifted_from_default: file not found
```
chalin, line 6: @jaydeluca - this also fixes the "file not found", which you pointed out earlier (can't recall if it was this file or another).

PR #8912, opened 2026-01-15
[adjust-pages.pl] Add upper bound version support to patch helper

vitorvasc:
- Allows disabling patches via version constraints without removing code.
- Add optional 4th parameter `$maxVers` to `applyPatchOrPrintMsgIf()` function
- Patch skips when submodule version > maxVers (inclusive upper bound)
- Backward compatible with existing 3-parameter calls


tiffany76 (approved):
RSLGTM.

chalin (commented):
I'll take a look ASAIC.

chalin:
@vitorvasc - btw, I haven't forgotten about this PR. I've just been feeling that we can do better. Here are some ideas.

1. I think that we should implement a function that parses semver IDs (X.Y.Z-pre-release-tag) and git describe-tag IDs (of the form X.Y.Z-dd-gHASH), and does an version-ID-parts based comparison, rather than a string/lexical comparison. The comparison function should accept partial IDs too, such as X.Y.Z-dd. (Actually, maybe we only need to support git describe-tag IDs for now.)

2. We should try to move towards a (close to) data-driven approach. That is, where we'd have an array of patch info (spec, minver, maxver, anonymous subroutine to hold the code that will apply the patch).

WDYT?

vitorvasc:
@chalin, both ideas make sense and I agree with them! Just pushed a new commit with both implementations.

PS: I'm not familiar with Perl, so I had my fellow Claude help me out with that. Let me know if you see any issues or opportunities to improve the file in general!

chalin (commented):
Nice! LGTM. We can later read the patch info from a file.

Is the current patch data up-to-date?

chalin (approved):
Actually, this is good as is. I'll followup with suggestions or a PR later.

chalin:
Merging b/c I want to add a patch! :)

PR #9021, opened 2026-01-26
docs(contrib): Rework side note about build errors and npm install

chalin:
- Follow up to #9004

### DIFF
--- a/content/en/docs/contributing/development.md
+++ b/content/en/docs/contributing/development.md
@@ -7,6 +7,7 @@ what-next: >
   website files. For details on how to submit changes, see [Submitting
   content](../pull-requests).
 weight: 60
+cSpell:ignore: TOCSS
 ---
 
 > [!WARNING] Supported build environments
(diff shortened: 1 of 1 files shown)


content/en/docs/contributing/development.md
```diff
@@ -92,10 +93,26 @@ npm run build
...
+> ```
+>
+> Or:
+>
+> ```log
+> ERROR failed to load modules: module "github.com/FortAwesome/Font-Awesome" not found
+> ```
+>
+> This is usually because you didn't complete all the steps in the
+> [local setup](#local-setup). In particular, (re)run this command:
```
tiffany76, line 111: Could it also be that they set up their local environment a long time ago and haven't run `npm install` in a while? (Ask me how I know...)

chalin: That isn't usually the case, at least not for those errors. But we're covering all bases here by stating "(re)run". Is that good enough?

It can be that a new version of Node (and hence new versions of packages) are needed, but again, not with those errors.

tiffany76: Sounds good to me!

Issue #8899, opened 2026-01-14
Homepage revamp followup


chalin:
Followup items for #8652:

- [X] https://github.com/open-telemetry/opentelemetry.io/pull/8652/changes#r2690209968 - "rather than hardcoding color tweaks, we'll want to make style changes in a manner that is more Docsy-upgrade friendly."  https://github.com/open-telemetry/opentelemetry.io/pull/9033
- [X] Hide navbar search element (keeping only the in-page serach) https://github.com/open-telemetry/opentelemetry.io/pull/9022
- [X] Avoid use of `markdownify`. See https://github.com/open-telemetry/opentelemetry.io/pull/8652#discussion_r2690747121 https://github.com/open-telemetry/opentelemetry.io/pull/9026
- [x] Review @julianocosta89 feedback on graphics together with @leandrocaracciolo, see [comment 1](https://github.com/open-telemetry/opentelemetry.io/pull/8652#issuecomment-3749477386) and [comment 2](https://github.com/open-telemetry/opentelemetry.io/pull/8652#issuecomment-3749681527)
- [x] Consider adding back a hero image, see https://github.com/open-telemetry/opentelemetry.io/pull/8652#issuecomment-3753568818. @svrnm @chalin
- [x] Banner width needs fixing (again) as noted in https://github.com/open-telemetry/opentelemetry.io/pull/8652#pullrequestreview-3677980593 - #8953
- [x] Fix homepage `.cncf` link styling. For context, see https://github.com/open-telemetry/opentelemetry.io/pull/8652#discussion_r2704488084
- [x] Supported languages quantity++ https://github.com/open-telemetry/opentelemetry.io/pull/9024
- [x] SCSS cleanup and refactoring
  - [x] #9033
  - [x]  #9042
  - [x] #9045 
  - [x] #9048 
  - [x] #9053
  - [x] #9081
  - [ ] ...
- [x] #9071
- [x] #9075

Related:

- #5155

theletterf:
@chalin About Markdownify, I've tried assess the situation with the help of AI and it's tricky:

[...]

PR #9022, opened 2026-01-27
Add search bar exception for homepage

theletterf:
- Contributes to https://github.com/open-telemetry/opentelemetry.io/issues/8899

This adds an exception to the navbar code so that it doesn't render in the nav when the homepage loads.


theletterf:
@chalin I preferred a CSS trick rather than a full layout override.

PR #9023, opened 2026-01-27
Move kotlin position in telemetry support table

fractalwrench:
- [X] I have read and followed the [Contributing](https://opentelemetry.io/docs/contributing/) docs, especially the "**First-time contributing?**" section.
- [ ] This PR has content that I did not fully write myself.
  - [ ] I used AI and I have read and followed the [Generative AI Contribution Policy](https://github.com/open-telemetry/community/blob/main/policies/genai.md).
- [X] I have the experience and knowledge necessary to understand, review, and validate all content in this PR.[^I-know-my-stuff]

[^I-know-my-stuff]:
    Yes, I can answer maintainer questions about the content of this PR, without using AI.

---

#8983 had a review comment asking for Kotlin to be in alphabetical order in the telemetry-support-table file. It got merged before I could commit a fix so I'm opening a separate PR to address the comment.

PR #9027, opened 2026-01-27
Update language docs weights while preserving order

chalin:
- Followup to #9023
- Adjusts the language section weights to that:
  - SDK Config at the top (10)
  - Other at the bottom (999)
  - The rest in alphabetical order, but ...
  - .NET treated as if named “DotNet”
  - We have weight-space between languages

| Section | Before | After |
|---------|--------|-------|
| SDK Configuration | 1 | 10 |
| C++ | 11 | 110 |
| .NET | 12 | 120 |
| Erlang/Elixir | 14 | 130 |
| Go | 16 | 140 |
| Java | 18 | 150 |
| JavaScript | 20 | 160 |
| Kotlin | 20 | 170 |
| PHP | 21 | 180 |
| Python | 22 | 190 |
| Ruby | 24 | 200 |
| Rust | 26 | 210 |
| Swift | 28 | 220 |
| Other | 99 | 999 |

### DIFF
--- a/content/en/docs/languages/cpp/_index.md
+++ b/content/en/docs/languages/cpp/_index.md
@@ -1,6 +1,6 @@
 ---
 title: C++
-weight: 11
+weight: 110
 description: >
   <img width="35" class="img-initial otel-icon"
   src="/img/logos/32x32/C++_SDK.svg" alt="C++"> A language-specific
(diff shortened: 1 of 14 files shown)


vitorvasc (approved):
🙌🏼

PR #9043, opened 2026-01-27
Fix Go getting-started: use HandleFunc for route handlers

a-ramses:
mux.Handle() expects a value implementing http.Handler (a type with ServeHTTP method), but rolldice is a plain function. Using HandleFunc() correctly adapts the function signature func(http.ResponseWriter, *http.Request) to satisfy the http.Handler interface.

### DIFF
--- a/content/en/docs/languages/go/getting-started.md
+++ b/content/en/docs/languages/go/getting-started.md
@@ -343,8 +343,8 @@ func newHTTPHandler() http.Handler {
 	mux := http.NewServeMux()
 
 	// Register handlers.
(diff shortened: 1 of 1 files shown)


theletterf:
@a-ramses Could you sign the CLA? Thanks!

dmathieu:
We keep these code examples in sync from the examples in go-contrib. And there is already a PR in there updating this code.

pellared:
@a-ramses, I submitted https://github.com/open-telemetry/opentelemetry.io/pull/9052

PR #9123, opened 2026-02-08
docs: update deprecated Prometheus 1.8 links in all languages

mihir-dixit2k27:
Description
I noticed that the Kubernetes Collector `components.md` documentation was linking to the outdated Prometheus 1.8 configuration guide (released in 2017).

This PR updates those links to point to the `latest` canonical version. This ensures users are directed to the current `scrape_config` options and improves SEO by removing redirects.

I applied this fix to the English documentation as well as the Chinese, Japanese, and Spanish translations to maintain consistency.

Link Verification
- **Old:** `https://prometheus.io/docs/prometheus/1.8/...` (Deprecated)
- **New:** `https://prometheus.io/docs/prometheus/latest/...` (Verified 200 OK)

### DIFF
--- a/static/refcache.json
+++ b/static/refcache.json
@@ -19251,6 +19251,10 @@
     "StatusCode": 200,
     "LastSeen": "2026-02-03T09:54:58.38730773Z"
   },
+  "https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config": {
+    "StatusCode": 200,
+    "LastSeen": "2026-02-13T07:43:58.041386222Z"
+  },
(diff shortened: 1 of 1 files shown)


theletterf:
@mihir-dixit2k27 Thanks. To avoid localization has mismatch, you should only edit the EN version and then follow the localization update process for the other locales. +CC @vitorvasc

mihir-dixit2k27:
@theletterf  I have reverted the changes to the localized files (es, ja, zh) as requested.

The checks seem to be stuck on "Waiting for status" (likely due to the earlier GitHub actions outage), but the code itself is ready for review whenever they clear up!

Issue #8578, opened 2025-12-04
Update blog submission guidelines around links into repositories


svrnm:
follow up to https://github.com/open-telemetry/opentelemetry.io/pull/7513:

We should add a section to the blog submission guidelines that links to github repos should not point to `main` (or an equivalent default branch) but to a commit or release that is up-to-date point in time. This way we avoid links breaking in the future when we no longer update the blog posts.

AbhinavKumar-Jha:
Hi, I'd like to work on this issue.

chalin:
@vitorvasc @svrnm et all: I'm wondering if we should apply the same policy to the registry entries? WDYT?

vitorvasc:
@chalin - I don't think that's a bad idea at all, but I wonder how we could keep the entries updated whenever the destination repo changes. Taking this commit (https://github.com/open-telemetry/opentelemetry.io/pull/9278/commits/56809dc96054c62d10de5080f376261f6df9994d) from the refcache update as an example, we'd be always pointing to an older commit where only the `javalin-5.0` existed, meaning `javalin-7.0` wouldn't be a known entry at that point.

On the other hand, I can see how allowing entries from `main` branch would create a constant burden for maintainers. 

I don't have a strong opinion either way, I can see the pros and cons of both approaches.

chalin:
Closing as the originally scoped ask has been covered. If we can revisit later regarding registry entries if the need arises.

PR #9162, opened 2026-02-13
chore(blog/2026): replace GitHub URL branch name with commit hash

chalin:
- Contributes #8578 by enforcing the guidance
- Replaces GitHub URL branch name with the last commit hash of the named resrouce. Tags `?from_branch=BRANCH_NAME` as a query parameter.

/cc @svrnm @vitorvasc

### DIFF
--- a/content/en/blog/2026/2025-year-in-review/index.md
+++ b/content/en/blog/2026/2025-year-in-review/index.md
@@ -205,7 +205,7 @@ You can also join us:
 Let's make 2026 another amazing year for [opentelemetry.io](/)!
 
 [redesign]:
-  https://github.com/open-telemetry/opentelemetry.io/blob/main/projects/landing-page-redesign.md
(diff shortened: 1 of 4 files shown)


content/en/blog/2026/dapr-workflow-observability/index.md
```diff
@@ -93,7 +93,7 @@ On the application side, the
 [durabletask-java SDK](https://github.com/dapr/durabletask-java/pull/46) was
 updated to read this context and restore it before running activity code. A
 version of this can be seen in the
-[exploratory branch](https://github.com/kaspernissen/pizza/tree/agentic-workflows)
+[exploratory branch](https://github.com/kaspernissen/pizza/tree/b09aaf5?from_branch=agentic-workflows)
```
chalin: One might question whether we should pin to a commit when the root of a branch is being referred to, but I'd say that we should (for now). WDYT @svrnm @vitorvasc et all?

chalin: Actually, changed my mind. Reverting this.

Issue #9167, opened 2026-02-13
Render-link hook rework/consolidation, and ref-link cleanup


chalin:
- [x] Simplify our path processing logic for cases that the URL is a path that starts with `/`. Use https://github.com/gohugoio/hugo/blob/master/tpl/tplimpl/embedded/templates/_markup/render-link.html as (re-)inspiration - #9168
- [x] Drop "backwards compatibility" code introduced in 9168 - #9169
- [x] Http checks:
  - [x] Clean up http processing (see TODO in file) - #9168
  - [x] Move http check to an mdl rule - #9171
  - [x] #9220, with followup PRs improving the mdl rules, see below.
- [x] #9216
- [x] #9218
- [x] Factor out spec-specific link processing - #9221
- [x] Keeping render-link code DRY: #9229
- [x] #9240

Ref link cleanup:

- [x] #9243 
- [x] #9245 
- [x] #9248 
- [x] #9249 
- [x] #9250 
- [x] #9252

svrnm:
closing this as all checklist items are completed! 🎉

PR #9171, opened 2026-02-13
chore(render-hook): move localhost link check to mdl rule

### LINKED ISSUE #6409: Use canonical link to connectrpc.com instead of connect.build
- [ ] Update semconv
- [ ] Update OTel Go
- [ ] Update refcache - drop entries using old URL, esp. the `http` reference `http://connect.build` (it should be `https`).

chalin:
- Contributes to #9167
- Fixes #6409 by wrapping up all tasks
- Moves non-localhost `http` URL checking into an mdl rule, out of the render-link hook
- Shortened the disable query parameter to `http_ok`


tiffany76 (approved):
RSLGTM.

Issue #8793, opened 2026-01-09
[otelcol docs] Add new content to "Deploy" section


tiffany76:
Phase 2

Add new content to the "Deploy the Collector" section as follows:

- Deploy the Collector
  - New: Choose a deployment pattern
  - New: Combined agent-to-gateway pattern

PR #9173, opened 2026-02-13
[otelcol] Rewrite the agent-to-gateway deployment page

tiffany76:

[^I-know-my-stuff]:

This PR shuffles the Deploy section a bit, but mostly it rewrites the agent-to-gateway deployment pattern page. This pattern is popular with users and the current documentation lacked thorough explanations and examples.

**EDIT**: I asked an AI agent to analyze the #otel-collector Slack threads to find out what we really needed to add to this page. I tried vetting everything by cross-checking across different agents. And I spent a good chunk of time editing. Hopefully there aren't too many mistakes. 🤞 

This is part of the cleanup from Phase 1 and contributes to #8793.

**PREVIEW**: https://deploy-preview-9173--opentelemetry.netlify.app/docs/collector/deploy/other/agent-to-gateway/

cc: @jade-guiton-dd and @mx-psi


content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
```diff
@@ -0,0 +1,646 @@
...
+  optional load balancing.
+- Gateway collectors receive data from agents, perform centralized processing,
+  such as filtering and sampling, and then export the data to backends.
+- Applications communicate with local agents using `localhost`, agents
+  communicate with gateways over the internal cluster network, and gateways
+  securely communicate with external backends using TLS.
+
+![gateway](otel-gateway-arch.svg)
+
+```mermaid
```
tiffany76: For the reviewer(s): I tried creating a Mermaid diagram to replace the current architecture svg for two reasons:

- Mermaid is compatible with the website's dark mode, but the svg isn't.
- I find the current svg hard to read in its rendered form and having to open it in a new tab to see it causes friction.

I've left the svg file in place for now, so we can compare them in the [preview](https://deploy-preview-9173--opentelemetry.netlify.app/docs/collector/deploy/other/agent-to-gateway/). Please let me know your thoughts!

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
```diff
@@ -0,0 +1,646 @@
...
+  manage credentials.
+
+## Example agent-to-gateway architecture
+
+The following diagram shows an architecture for a combined agent-to-gateway
+deployment:
+
+- Agent collectors run on each host in a DaemonSet pattern and collect telemetry
+  from services running on the host as well as the host's own telemetry, with
+  optional load balancing.
```
jade-guiton-dd: I think some kind of load balancing is a requirement of the gateway deployment pattern, otherwise you could only have a single Collector in your gateway. But maybe what's optional is Collector-level load balancing, as opposed to network-level load balancing?

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
```diff
@@ -0,0 +1,646 @@
...
+- Gateway collectors receive data from agents, perform centralized processing,
+  such as filtering and sampling, and then export the data to backends.
+- Applications communicate with local agents using `localhost`, agents
```
jade-guiton-dd: A lot of the examples are Kubernetes-centric, in which `localhost` only works within a single pod, so this may be misleading. Maybe "the internal host network" would be clearer?

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
```diff
@@ -0,0 +1,646 @@
...
+  memory_limiter:
+    check_interval: 1s
+    limit_mib: 512
+    spike_limit_mib: 128
+
+  # Use smaller batches on agents
+  batch:
+    send_batch_size: 1024
+    send_batch_max_size: 2048 # safety limit
+    timeout: 1s
```
jade-guiton-dd: We're trying to move away from recommending the batch processor in favor of the batching settings in exporters. The equivalent of this config would be adding the following under the `sending_queue` section of the exporter:

```yaml
batch:
  sizer: items
  min_size: 1024
  max_size: 2048
  flush_timeout: 1s
```

Although, unless we have very good reasons to recommend these specific values, maybe it would be best to just stick with the defaults?

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
```diff
@@ -0,0 +1,646 @@
...
+
+exporters:
+  # Send to gateway
+  otlp:
+    endpoint: otel-gateway:4317
+    tls:
+      insecure: false
```
jade-guiton-dd: Nit: TLS being enabled is already the default, so `insecure: false` isn't necessary. I think it's the same thing for `retry_on_failure.enabled` and `sending_queue.enabled`: as long as the section is set it will be enabled by default.

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
```diff
@@ -0,0 +1,646 @@
...
+      ca_file: /etc/ssl/certs/ca.crt
```
jade-guiton-dd: I think this is specific to gateways which provide a server certificate, ie. with `cert_file` and `key_file` set in the OTLP receiver. I'm not sure how common that setup is, so maybe it's best to leave this TLS section out?

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
```diff
@@ -0,0 +1,646 @@
...
+    sending_queue:
+      enabled: true
+      num_consumers: 2
+      queue_size: 5000
+
+service:
+  pipelines:
+    traces:
+      receivers: [otlp]
+      processors: [resourcedetection, memory_limiter, batch]
```
jade-guiton-dd: `memory_limiter` should be as early in the pipeline as possible.

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
```diff
@@ -0,0 +1,646 @@
...
+    num_traces: 100000 # requires more memory
+    policies:
+      # Always sample traces with errors
+      - name: errors-policy
+        type: status_code
+        status_code: { status_codes: [ERROR] }
+      # Sample 10% of other traces
+      - name: probabilistic-policy
+        type: probabilistic
+        probabilistic: { sampling_percentage: 10 }
```
jade-guiton-dd, line 261: It's confusing to include the `tail_sampling` processor in a gateway config right after an agent config which, as stated further below, _won't_ work with tail sampling (ie. network-level load balancing instead of using the load balancing exporter).

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
```diff
@@ -0,0 +1,646 @@
...
+based on their role.
+
+### Essential processors for production
+
+Both agents and gateways should include:
+
+- **Memory limiter processor**: This processor prevents out-of-memory issues by
+  applying backpressure when memory usage is high. Configure this as the first
+  processor in your pipeline. Agents typically need smaller limits (256-512 MB),
+  while gateways require more memory (1-4 GB or higher) for batching and
```
jade-guiton-dd: I feel like these numbers kind of come out of nowhere. The memory limiter configuration will need to be adjusted based on how much memory the Collector containers have available, what the workload is, and what level of reliability is expected when the gateway is temporarily unavailable, which seems pretty user-specific.

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
```diff
@@ -0,0 +1,646 @@
...
+  > [!CAUTION]
+  >
+  > The tail-sampling processor can make accurate decisions only if all spans
+  > for a trace arrive at the same Collector instance. While the load balancing
+  > exporter supports routing by trace ID, running tail sampling across multiple
+  > gateway instances is an advanced setup and has practical caveats, such as
+  > re-splitting of routing when backends change and cache/decision consistency.
+  > Test carefully and prefer a single well-resourced tail-sampling gateway
+  > unless you have a robust sticky-routing strategy.
+
```
jade-guiton-dd, line 402: Nit: Not sure how many users currently use the cumulative-to-delta processor (at Datadog we're probably going to start recommending its use more widely), but maybe we should call out that it's another component that requires data-aware load balancing, so all points of a given metric series reach the same gateway Collector.

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
```diff
@@ -0,0 +1,646 @@
...
+Agents need to reliably send telemetry data to gateways. Configure the
+communication protocol, endpoints, and security settings appropriately for your
+environment.
+
+### Protocol selection
+
+Use the OTLP protocol for communication between agents and gateways. OTLP
+provides the best performance and compatibility across the OpenTelemetry
+ecosystem. Configure the OTLP exporter in your agents to send data to the OTLP
```
jade-guiton-dd: The compatibility claim is fine since it's the canonical protocol for OTel data, but the performance claim is kind of [citation needed]

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
```diff
@@ -0,0 +1,646 @@
...
+
+- **Without tail-based sampling**: Use any load balancer or Kubernetes service
+  with round-robin distribution. All gateway instances operate independently.
+
+  > [!NOTE]
+  >
+  > When scaling gateway instances that export metrics, ensure your deployment
+  > follows the single-writer principle to avoid multiple Collectors writing the
+  > same time series concurrently. Some backends require consistent time series
+  > ownership. See the
```
jade-guiton-dd: I'm not sure this is really a backend issue, it's more of a spec requirement

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
```diff
@@ -0,0 +1,646 @@
...
+  to route spans by trace ID. The load balancing exporter ensures all spans for
+  a trace go to the same gateway instance, which is required for tail-based
+  sampling decisions.
+
+For automatic scaling in Kubernetes, use
+[Horizontal Pod Autoscaling (HPA)](https://kubernetes.io/docs/concepts/workloads/autoscaling/horizontal-pod-autoscale/)
+based on CPU or memory metrics. Configure the HPA to scale gateways based on
+your workload patterns.
+
+## Resource planning
```
jade-guiton-dd: What's the difference with the "scaling" section? Feels a bit redundant

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
```diff
@@ -0,0 +1,646 @@
...
+usage, isolation, and management complexity.
+
+### Agent deployment options
+
+- **DaemonSet (Kubernetes)**: Recommended for most Kubernetes deployments. Runs
+  one agent per node, automatically scaling with cluster size.
+
+- **Sidecar**: Use only when you need isolated resource limits per application
+  or application-specific configuration. Sidecars increase resource overhead and
+  management complexity. Prefer DaemonSets when possible.
```
jade-guiton-dd: Including side cars as an instance of the Agent deployment pattern conflicts with the idea that the agent-to-gateway allows you do host-based data collection.

I think it's probably reasonable to include Sidecars as Agents? But you would want to specify that the section about host-based data collection doesn't really apply. Maybe there is some pod-based data collection you could do?

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
```diff
@@ -0,0 +1,646 @@
...
+
+### Handling variable workloads
+
+For workloads with significant variability (seasonal traffic, batch jobs):
+
+- Configure memory limiter processor with appropriate limits and spike
+  thresholds.
+- Use queue settings in exporters to handle temporary traffic spikes.
+- For gateways, configure HPA with appropriate scaling policies.
+- Monitor queue sizes and retry counts to identify capacity issues.
```
jade-guiton-dd: I feel like this is also redundant.

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
```diff
@@ -0,0 +1,646 @@
...
+
+Monitor these key metrics from your Collectors:
+
+- `otelcol_processor_refused_spans`: Indicates memory pressure or processing
+  issues
+- `otelcol_exporter_queue_size`: Watch for growing queues indicating export
+  problems
+- `otelcol_exporter_send_failed_spans`: Indicates backend connectivity or
+  capacity issues
+- Memory and CPU usage from the Kubernetes metrics API or Prometheus
```
jade-guiton-dd: I feel like everything here besides the part about tail sampling is not specific to the Agent to Gateway pattern and should be in the "Collector > Troubleshooting" page. (And the warning about tail sampling has already been covered a few times on this page.)

(Additionally, you'd probably want to list internal metrics for logs and metrics as well, and call out complexities of the `loadbalancing` exporter's internal metrics... But as I said that probably shouldn't be on this page.)

jade-guiton-dd (commented):
Sorry in advance for the very long review. I think that despite what AI tends to think, "less is more" is probably the right philosophy here, especially given the maintenance upkeep of keeping things up-to-date later. I think the new page is a bit overconfident about some of the recommended numbers and claims, and I think some parts are best left for other pages. There are some technical claims about specific components I'm not qualified to evaluate, so I think feedback from other Collector maintainers would be helpful.

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
```diff
@@ -0,0 +1,646 @@
...
+        end
+    end
+
+    subgraph "External Network"
+        Backend["Observability<br/>backend"]
+    end
+
+    App1 -->|"OTLP<br/>(localhost)"| Agent1
+    App2 -->|"OTLP<br/>(localhost)"| Agent2
+    AppN -->|"OTLP<br/>(localhost)"| AgentN
```
mx-psi: If we go with @jade-guiton-dd's suggestion above I would change `localhost` here as well maybe to this?
```suggestion
    App1 -->|"OTLP<br/>(local)"| Agent1
    App2 -->|"OTLP<br/>(local)"| Agent2
    AppN -->|"OTLP<br/>(local)"| AgentN
```

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
```diff
@@ -0,0 +1,646 @@
...
+
+- Separation of concerns: Avoid placing complex configuration and processing
+  logic on every machine or in every node. Agent configurations stay small and
+  focused, while central processors handle the heavier collection tasks.
+- Scalable cost control: Make better sampling and batching decisions in gateways
+  that can receive telemetry from multiple agents. Gateways can see the full
+  picture, including complete traces, and can be independently scaled.
+- Security and stability: Send telemetry over local networks from agents to
+  gateways. Gateways become a stable egress point that can handle retries and
+  manage credentials.
```
mx-psi: These do not have bold titles, other lists like this do, is that intentional?

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
```diff
@@ -0,0 +1,646 @@
...
+
+```yaml
+receivers:
+  # Receive telemetry from applications
+  otlp:
+    protocols:
+      grpc:
+        endpoint: 0.0.0.0:4317
+      http:
+        endpoint: 0.0.0.0:4318
```
mx-psi: We have a warning in some other docs pages like the one here about the `0.0.0.0` address https://opentelemetry.io/docs/collector/configuration/#basics. Should we have it here?

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
```diff
@@ -0,0 +1,646 @@
...
+
+  # Collect host metrics
+  hostmetrics:
+    collection_interval: 30s
```
mx-psi: If we want to go for a more minimal configuration we can just remove this
```suggestion
```

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
```diff
@@ -0,0 +1,646 @@
...
+    sending_queue:
+      enabled: true
+      num_consumers: 4
+      queue_size: 20000
+
+  # Optional: Export metrics to Prometheus
+  prometheusremotewrite:
+    endpoint: https://prometheus:9090/api/v1/write
+    tls:
+      insecure: false
```
mx-psi: I think we can remove this is we want to keep the example simple
```suggestion
```

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
mx-psi: I tend to agree and they may get outdated quickly

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
```diff
@@ -0,0 +1,646 @@
...
+  processor in your pipeline. Agents typically need smaller limits (256-512 MB),
+  while gateways require more memory (1-4 GB or higher) for batching and
+  sampling operations.
+
+- **Batch processor**: This processor improves efficiency by batching telemetry
+  data before export. Configure agents with smaller batch sizes and shorter
+  timeouts (for example, `send_batch_size: 1024`, `timeout: 1s`) to minimize
+  latency and memory usage. Configure gateways with larger batch sizes and
+  longer timeouts (for example, `send_batch_size: 10000`, `timeout: 10s`) for
+  better throughput and backend efficiency.
```
mx-psi: Let's talk about batching generally since we don't wnat more people to use the batch processor

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
```diff
@@ -0,0 +1,646 @@
...
+
+### Sampling considerations
+
+- **Probabilistic sampling**: You can configure this on either agents or
+  gateways. When using probabilistic sampling across multiple collectors, ensure
+  they use the same hash seed for consistent sampling decisions.
```
mx-psi: I think we can make this shorter
```suggestion
- **Probabilistic sampling**: Ensure all collectors use the 
  same hash seed for consistent sampling decisions.
```

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
```diff
@@ -0,0 +1,646 @@
...
+
+- **Tail-based sampling**: Configure tail-based sampling on gateways only
+  because the processor must see all spans for a trace to make sampling
+  decisions. When using tail-based sampling with multiple gateway instances, you
+  need to ensure all spans for a given trace ID are routed to the same gateway
+  instance. Use the
```
mx-psi: I feel like we get the point across without using that sentence?
```suggestion
  decisions. Use the
```

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
```diff
@@ -0,0 +1,646 @@
...
+
+In Kubernetes environments, use service names for endpoint configuration. For
+example, if your gateway service is named `otel-gateway`, configure your agent
+exporter with `endpoint: otel-gateway:4317`.
+
+### Retries
+
+In production, configure exporter queue and retry settings (for example,
+`retry_on_failure` or `sending_queue` settings) on agents and gateways to handle
+temporary outages between agents and gateways or between gateways and backends.
```
mx-psi:
```suggestion
Adjust exporter queue and retry settings (for example,
`retry_on_failure` or `sending_queue` settings) on agents and gateways to handle
temporary outages between agents and gateways or between gateways and backends.
```

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
```diff
@@ -0,0 +1,646 @@
...
+    GC2 -->|OTLP| B1
+    GC3 -->|OTLP| B1
+```
+
+## Processor configuration in agents and gateways
+
+When deploying an agent-to-gateway pattern, configure processors differently
+based on their role.
+
+### Essential processors for production
```
mx-psi:
```suggestion
### Recommended processors
```

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
```diff
@@ -0,0 +1,646 @@
...
+  ([mTLS](https://pkg.go.dev/go.opentelemetry.io/collector/config/configtls)) to
+  authenticate both agents and gateways.
+- Consider token-based authentication using the
+  [`bearertokenauthextension`](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/bearertokenauthextension)
+  when mTLS is not practical.
+
+In Kubernetes environments, automate certificate issuance and rotation (for
+example, with [cert-manager](https://cert-manager.io/docs/)) and mount
+certificates into Collector pods as Secrets. Certificate reload behavior can
+vary by component, so plan rolling restarts if needed when certificates rotate.
```
mx-psi: Simplification
```suggestion
certificates into Collector pods as Secrets. Plan rolling restarts
when certificates rotate.
```

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
```diff
@@ -0,0 +1,646 @@
...
+As your telemetry volume grows, you need to scale your Collectors appropriately.
+Agents and gateways have different scaling characteristics and requirements.
+
+### Agents
+
+Agents typically don't require horizontal scaling because they run on each host
+(DaemonSet pattern in Kubernetes). Instead, scale agents vertically by adjusting
+resource limits:
+
+- Monitor CPU and memory usage through Collector internal metrics.
```
mx-psi: We can add a link to https://opentelemetry.io/docs/collector/internal-telemetry/

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
```diff
@@ -0,0 +1,646 @@
...
+- Increase memory limits if you see memory pressure or memory limiter
+  backpressure.
+- Increase CPU limits if you experience CPU throttling.
```
mx-psi: If we want to make it shorter I would keep this out, it seems a bit too obvious?

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
mx-psi: +1

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
```diff
@@ -0,0 +1,646 @@
...
+
+For workloads with significant variability (seasonal traffic, batch jobs):
+
+- Configure memory limiter processor with appropriate limits and spike
+  thresholds.
+- Use queue settings in exporters to handle temporary traffic spikes.
+- For gateways, configure HPA with appropriate scaling policies.
+- Monitor queue sizes and retry counts to identify capacity issues.
+
+## Troubleshooting
```
mx-psi: Is this intended to replace https://opentelemetry.io/docs/collector/troubleshooting/ ?

mx-psi (commented):
Left a bunch of comments. I am worried that we will end up with a "Best practices for Kubernetes deployments", IMO (and I am not part of the Comms SIG so feel free to disagree) we should tell users here only about the specifics of using the Collector in Kubernetes and link to Kubernetes documentation on how to do more generic stuff

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
tiffany76: Good catch! Not intentional, just the result of different drafting sessions. I'll make them consistent!

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
tiffany76: > otherwise you could only have a single Collector in your gateway

Good point. 😆 

I'll remove the "optional" from here and the diagram.

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
tiffany76: I've removed the batch processor from the example and added the `sending_queue::batch` section. 

As for the values, I think the point was to indicate agents should have smaller batches than gateways. But I can see how specific values are problematic. I've left them in for now but added a TODO so I won't forget to revisit.

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
tiffany76: Excellent point. I moved the agent configuration with tail sampling up a section.

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
tiffany76: Given @mx-psi's concern about duplicating Kubernetes docs, I'm going to just remove the `## Deployment considerations` section.

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
tiffany76: I've pulled out the Troubleshooting content and made a note to incorporate it into the Troubleshooting page when we get there.

tiffany76 (commented):
Thank you, @jade-guiton-dd and @mx-psi, for your thorough reviews. 🙏 

Well, I think it's fair to say that although it wasn't a complete failure, this generative AI experiment did not succeed. I waited to see how this review would go before working on other rewrites. I have absolutely no wish to create a giant review burden for you and other Collector folks, so I'll be changing how I refactor and rewrite the rest of the pages. 

I think maybe having the agent do analysis (e.g. Slack) and produce an outline that we (Collector SMEs and I) can edit and collaborate on is a better approach. WDYT?

As for this particular PR, I made all of your suggested changes and am about to push them up. There are just a couple outstanding questions:

1. Should we keep the mermaid diagram and ditch the svg?
2. Should we remove the specific values in the example configs, especially with respect to batching? As I noted in my followup comment, I think the idea is to show that agents should have smaller batches than gateways. If you still think it's worth removing the custom settings so the defaults apply, let me know.

Also, I failed to spell this out, but I do intend to break up this page a little as we go along. We will eventually have a "choose a deployment pattern" page in this section, where I will move the "when to use" and "when not to use" sections. I'm keeping them here for now since that page doesn't exist yet. I removed the Troubleshooting section and will add it to the existing Troubleshooting page as soon as we get there.

Many thanks, again, for your help!

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
tiffany76: I added a section just below this. Take a look and let me know what you think!

mx-psi:
> I think maybe having the agent do analysis (e.g. Slack) and produce an outline that we (Collector SMEs and I) can edit and collaborate on is a better approach. WDYT?

I think that would make more sense, yes :)

> Should we keep the mermaid diagram and ditch the svg?

Yes, +1 from my side to do that

> Should we remove the specific values in the example configs, especially with respect to batching? As I noted in my followup comment, I think the idea is to show that agents should have smaller batches than gateways. If you still think it's worth removing the custom settings so the defaults apply, let me know.

I would rather remove then, at least for a first version

jade-guiton-dd:
Agree with all the above. Mermaid is certainly easier to modify than SVG, even if it's not as flexible.

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
```diff
@@ -0,0 +1,543 @@
...
+
+The following examples show typical configurations for agents and gateways in an
+agent-to-gateway deployment.
+
+> [!WARNING]
+>
+> While it is generally preferable to bind endpoints to `localhost` when all
+> clients are local, our example configurations use the “unspecified” address
+> `0.0.0.0` as a convenience. The Collector currently defaults to `0.0.0.0`, but
+> the default will be changed to `localhost` in the near future. For details
```
mx-psi: This is outdated, both here and in other PRs since https://opentelemetry.io/blog/2024/hardening-the-collector-one/:
```suggestion
> `0.0.0.0` as a convenience. The Collector defaults to `localhost`. For details
```

(we can also change this on another PR)

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
jade-guiton-dd: Maybe it would make sense to add a warning up front about how this section is for cases where no data-aware load balancing is needed (ie. no tail sampling, no cumulative-to-delta conversion)

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
```diff
@@ -172,47 +182,84 @@ processors:
...
+      grpc:
+        endpoint: 0.0.0.0:4317
+
+processors:
+  memory_limiter:
+    check_interval: 1s
+    limit_mib: 512
+  batch:
+    send_batch_size: 1024
+    timeout: 1s
```
jade-guiton-dd: Probably best to use the loadbalancing exporter's batch settings instead of the batch exporter here as well, something like:
```yaml
    sending_queue:
      batch:
        sizer: items
        min_size: 1024
        flush_timeout: 1s
```

(Under `loadbalancing`, not `loadbalancing.protocol.otlp`; both are possible, but I would recommend the former)

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
```diff
@@ -447,6 +439,16 @@ Both agents and gateways should include:
...
 
+### Other processing considerations
+
+- **Cumulative-to-delta calculations**: Cumulative-to-delta metric processing
+  requires data-aware load balancing because the calculation is only accurate if
+  all points of a given metric series reach the same gateway Collector. Take
+  care when using the
+  [`cumulativetodelta` processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/cumulativetodeltaprocessor)
+  in an agent-to-gateway deployment. Each data source should send data to a
+  single Collector.
```
jade-guiton-dd: "Each data source" is a bit ambiguous, it might sound like each application instance can only send to a single gateway Collector. "Each metric stream should only be sent to a single Collector" may be more accurate.

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
```diff
@@ -172,47 +182,84 @@ processors:
...
+    send_batch_size: 1024
+    timeout: 1s
+
+exporters:
+  # Load balance by trace ID
+  loadbalancing:
+    protocol:
+      otlp:
+        tls:
+          insecure: false
```
jade-guiton-dd: `insecure: false` is the default, so no need to set the `protocol` section here I think

jade-guiton-dd (commented):
Looks good (although there are TODOs to resolve)

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
tiffany76: Updated this PR in 9a7d5be and raised a [follow-up issue](https://github.com/open-telemetry/opentelemetry.io/issues/9404) to fix the others.

tiffany76:
Welcome back to the PR-that-never-ends! 😆 

@mx-psi @jade-guiton-dd I made all of your suggested changes, including removing the specific values or redundant defaults from the config examples. I also reordered the page to focus more on load balancing and a little less on tail sampling.

All the TODOs have been resolved and removed.

Hopefully we're close.

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
```diff
@@ -0,0 +1,513 @@
...
+
+```yaml
+receivers:
+  # Receive from agents
+  otlp:
+    protocols:
+      grpc:
+        endpoint: 0.0.0.0:4317
+      http:
+        endpoint: 0.0.0.0:4318
```
mx-psi: Not technically needed for your agent configuration since it uses the `otlp` exporter, not the `otlphttp` one

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
```diff
@@ -0,0 +1,513 @@
...
+### Retries
+
+Configure exporter queue and retry settings (for example, `retry_on_failure` or
+`sending_queue` settings) on agents and gateways to handle temporary outages
+between agents and gateways or between gateways and backends. Gateways often
+need larger queues and retry policies to handle backend outages. Also consider
+setting a `max_size` for batches to avoid transient backend rejections due to
+oversized payloads.
+
+### Security
```
mx-psi: Maybe better for a generic security page like https://deploy-preview-9173--opentelemetry.netlify.app/docs/security/config-best-practices/ ?

mx-psi (commented):
Looking good! Just two comments, otherwise looks good from my side

theletterf (approved):

mx-psi (approved):

tiffany76:
@jade-guiton-dd Any last thoughts before I hit merge?

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
```diff
@@ -0,0 +1,489 @@
...
+
+## When to use this pattern
+
+The agent-to-gateway pattern adds operational complexity compared to simpler
+deployment options. Use this pattern when you need one or more of the following
+capabilities:
+
+- **Host-specific data collection**: You need to collect metrics, logs, or
+  traces that are only available on the host where your applications run, such
+  as host metrics, system logs, or resource detection. For example, receivers
```
jade-guiton-dd: Not sure why "Host-specific data collection" is listed as a reason to use the "Agent to Gateway" pattern. It sounds like a reason to use Agent Collectors, but you don't necessarily need to add a Gateway to it.

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
```diff
@@ -0,0 +1,489 @@
...
+      batch:
+        sizer: items
+        flush_timeout: 1s
+
+service:
+  pipelines:
+    traces:
+      receivers: [otlp]
+      processors: [memory_limiter, resourcedetection]
+      exporters: [otlp]
```
jade-guiton-dd, line 181: @mx-psi do you know if the new alias system supports declaring an exporter as `otlp_grpc` but referencing it in a pipeline as `otlp`?

jade-guiton-dd (approved):
Looks pretty good to me, just two minor concerns

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
mx-psi: I don't think that pattern is something we want to encourage even if supported. I think we should use `otlp` at the top and make a switch to `otlp_grpc` once more time has passed since the first release that supports that naming scheme

tiffany76:
Sorry for all the pings, folks! Git got the best of me. Since this is a complete mess, I'm closing now.

tiffany76:
Some fairy godmother decided to work some git magic and now this PR is...back to where it should be? Tentatively going to try again.

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
tiffany76: Removed it and will relocate it to the Agent page.

content/en/docs/collector/deploy/other/agent-to-gateway/_index.md
tiffany76: Changed both occurrences to `otlp` instead of `otlp_grpc`.

PR #9201, opened 2026-02-18
feat: add daemonset resiliency docs

### LINKED ISSUE #5932: [docs/collector] Disaster recovery recommendations for Daemonset deployments
**Desired feature or idea**: We would like to document recommendations and best practices related to Daemonset deployments with the Collector, since these are prone to have catastrophic failures with traffic increases. We don't necessarily want to add documentation ourselves, but possibly link to 'best practices' documents, specially for things that are not Collector-specific.

**Additional context**: Part of the discussion in open-telemetry/opentelemetry-collector/issues/11499, see https://docs.google.com/document/d/1BO7hMg0K7zFQ0219z61yRj-2JMMTW_HfoX3Vt6kXW8E/edit?disco=AAABblcXbKI

jpkrohling:
Documents the failures modes when running the collector as a daemonset. This doc gap was spotted during the technical review of OTel in the context of the project's graduation. 

Fixes #5932

Signed-off-by: Juraci Paixão Kröhling <juraci@kroehling.de>


jade-guiton-dd:
Would it make sense to have this as a section under the "Agent deployment pattern" page? Since it seems to me like Daemonsets are the most common way to implement it

jpkrohling:
PR updated

content/en/docs/collector/deploy/agent.md
```diff
@@ -123,6 +123,294 @@ Cons:
...
+
+```yaml
+env:
+  - name: GOMEMLIMIT
+    value: '400MiB' # Match or slightly below the memory_limiter limit
+```
+
+> [!TIP] Monitor `otelcol_processor_refused_spans` (and the metrics/logs
+> equivalents). A sustained increase indicates the Collector is under pressure
+> and you should either scale resources or reduce incoming traffic.
```
tiffany76:
```suggestion
> [!TIP]
>
> Monitor `otelcol_processor_refused_spans` (and the metrics/logs
> equivalents). A sustained increase indicates the Collector is under pressure
> and you should either scale resources or reduce incoming traffic.
```

content/en/docs/collector/deploy/agent.md
```diff
@@ -123,6 +123,294 @@ Cons:
...
+      queue_size: 5000
+    retry_on_failure:
+      max_elapsed_time: 10m
+
+service:
+  extensions: [file_storage]
+  pipelines:
+    traces:
+      receivers: [otlp]
+      processors: [memory_limiter, batch]
```
tiffany76: I think we're trying to limit examples with the batch processor in favor of batching in the exporter.

content/en/docs/collector/deploy/agent.md
```diff
@@ -123,6 +123,294 @@ Cons:
...
+
+Mount a persistent volume or use the node's local storage for the WAL directory.
+When using local storage (such as `emptyDir` or `hostPath`), be aware that data
+does not survive node failures.
+
+> [!WARNING] If a bug in the Collector causes data in the persistent queue to
+> trigger a crash on replay, the Collector will enter a crash loop. Monitor for
+> `CrashLoopBackOff` and have a plan to clear the WAL directory if needed. See
+> [opentelemetry-collector#12095](https://github.com/open-telemetry/opentelemetry-collector/issues/12095)
+> for more details.
```
tiffany76:
```suggestion
> [!WARNING]
>
> If a bug in the Collector causes data in the persistent queue to
> trigger a crash on replay, the Collector will enter a crash loop. Monitor for
> `CrashLoopBackOff` and have a plan to clear the WAL directory if needed. See
> [opentelemetry-collector#12095](https://github.com/open-telemetry/opentelemetry-collector/issues/12095)
> for more details.
```

content/en/docs/collector/deploy/agent.md
```diff
@@ -123,6 +123,294 @@ Cons:
...
+```
+
+Setting `maxUnavailable: 0` with `maxSurge: 1` ensures the new pod starts before
+the old one terminates. This requires Kubernetes 1.22+ and sufficient node
+resources to run two Collector pods simultaneously during the update.
+
+> [!NOTE] `maxSurge` for DaemonSets is available since Kubernetes 1.22 as a
+> stable feature. If you're on an older version, `maxUnavailable: 1` (the
+> default) is unavoidable, and you should rely on persistent queues to minimize
+> data loss during updates.
```
tiffany76:
```suggestion
> [!NOTE]
>
> `maxSurge` for DaemonSets is available since Kubernetes 1.22 as a
> stable feature. If you're on an older version, `maxUnavailable: 1` (the
> default) is unavoidable, and you should rely on persistent queues to minimize
> data loss during updates.
```

content/en/docs/collector/deploy/agent.md
```diff
@@ -123,6 +123,294 @@ Cons:
...
+
+- **Isolation is critical**: Use the
+  [sidecar pattern](/docs/collector/scaling/#scaling-stateless-collectors-and-using-load-balancers)
+  when you need per-application isolation, so that one application's telemetry
+  spike cannot affect another.
+- **Processing is heavy**: Offload processors like `batch`, `tail_sampling`, or
+  `transform` to a [Gateway tier](/docs/collector/deploy/gateway/) and keep the
+  DaemonSet Collector lightweight (receive + forward only).
+- **High pod-to-node ratio**: When you have many small pods per node, a
+  DaemonSet works well. When you have few large pods, a sidecar avoids the blast
```
tiffany76:
```suggestion
  DaemonSet works well. When you have a few large pods, a sidecar avoids the blast
```

tiffany76 (approved):
LGTM with a couple small changes!

The admonitions need a blank line between the heading and content to display correctly.

tiffany76:
@open-telemetry/collector-approvers, PTAL. Thanks!

jpkrohling:
Just wanted to mention that I'm mostly afk this week but will work on the feedback next week. If you prefer, feel free to add a commit to the PR with the suggestions, they all seem reasonable to me.

tiffany76:
Thanks, @jpkrohling! I applied the simple copy edits. That just leaves the question of how you want to address batching. No rush from my perspective.

content/en/docs/collector/deploy/agent.md
jpkrohling: I agree! I need to get used to this change :-)

tiffany76:
@open-telemetry/collector-approvers, PTAL! Thanks.

content/en/docs/collector/deploy/agent.md
```diff
@@ -123,6 +123,301 @@ Cons:
...
+Kubernetes terminates it with an OOMKill. The pod restarts, but any data
+buffered in memory is lost. Meanwhile, telemetry from applications on that node
+accumulates. When the restarted Collector begins processing the backlog, it
+might immediately exceed its memory limit again, creating a crash loop.
+
+The `memory_limiter` processor can help prevent this, but it is not bulletproof:
+incoming data must be deserialized and converted into the Collector's internal
+data model before the memory limiter can reject it. A sudden spike that arrives
+between memory checks can push memory usage past the limit before the processor
+has a chance to act.
```
jade-guiton-dd: Sorry, I know I was the one who said to move things to the Agent page, that was before I actually read through the new text.

Honestly, I don't think most of this is specific to Agent deployments, it's just advice about dealing with Collectors under high load, which can in principle happen in Sidecars, and can definitely happen in Gateways as well.

My non-blocking recommendation would be to:
- move all of the generic advice ("OOMKill cascades", "Limited back-pressure in async pipelines", "Resource management", and honestly everything from "Resilience configuration" to "Monitoring and early warning") back to the "Resiliency" page
- only keep the Agent-specific advice ("Node-scoped blast radius", "Noisy neighbors", maybe "Rolling update gaps") on this page
- add a section warning generically about "the load is likely higher than a sidecar, so please look at this section (link) of the resiliency page about high load".

content/en/docs/collector/deploy/agent.md
```diff
@@ -123,6 +123,301 @@ Cons:
...
+an external load balancer to handle rate limiting when needed.
+
+#### Limited back-pressure in async pipelines
+
+When the Collector is overwhelmed, it responds to clients with errors (gRPC
+Unavailable or HTTP 503). However, in the typical configuration with an
+exporting queue enabled, back-pressure signals from the export destination do
+not propagate back to the sending applications. This means applications continue
+sending telemetry at full rate even when the Collector cannot keep up with
+exporting, which accelerates queue fill-up and data loss.
```
jade-guiton-dd: Is this true? If you're using the exporterhelper queue, then the only errors that can fail to be propagated up the pipeline (if you don't use `wait_for_result`) are exporting errors, whereas errors when the Collector is overwhelmed usually come from the queue being full, or from an upstream memorylimiter processor.

content/en/docs/collector/deploy/agent.md
```diff
@@ -123,6 +123,301 @@ Cons:
...
+env:
+  - name: GOMEMLIMIT
+    value: '400MiB' # Match or slightly below the memory_limiter limit
+```
+
+> [!TIP]
+>
+> Monitor `otelcol_processor_refused_spans` (and the metrics/logs equivalents).
+> A sustained increase indicates the Collector is under pressure and you should
+> either scale resources or reduce incoming traffic.
```
jade-guiton-dd: Could be good to link to the "Internal telemetry" page with the list of these metrics and how to monitor them.

content/en/docs/collector/deploy/agent.md
```diff
@@ -123,6 +123,301 @@ Cons:
...
+  remaining budget
+- Monitor memory consumption under both normal and degraded conditions and
+  adjust
+
+### Resilience configuration
+
+#### Enable persistent queues
+
+By default, the sending queue buffers data in memory. If the Collector pod
+restarts, all queued data is lost. Enable
```
jade-guiton-dd: I'm pretty sure the sending queue will try to flush its data during a graceful shutdown. So the bigger data loss risk is a full Collector crash like an OOM.

content/en/docs/collector/deploy/agent.md
```diff
@@ -123,6 +123,301 @@ Cons:
...
+| `otelcol_exporter_queue_size`           | Current queue depth; rising trend indicates backend slowdown |
+| `otelcol_exporter_queue_capacity`       | Maximum queue size; compare with queue_size for headroom     |
+| `otelcol_exporter_send_failed_spans`    | Permanent export failures; backend may be unhealthy          |
+| `otelcol_exporter_enqueue_failed_spans` | Queue is full; data is being dropped                         |
+| Container memory usage                  | Approaching the limit means OOMKill risk                     |
+
+Set alerts on:
+
+- `otelcol_processor_refused_spans` > 0 sustained for more than 5 minutes
+- `otelcol_exporter_queue_size` > 70% of `otelcol_exporter_queue_capacity`
```
jade-guiton-dd: Would probably be good to link to the "Internal telemetry" page here as well, I think this section is slightly redundant with it.

jpkrohling:
@jade-guiton-dd , do you feel like you could take over this PR? Quite a few things here are from the new sending queue batcher, of which I still need to catch up on.

ManniMansoor:
pls check repo

ManniMansoor:
guess my AI agent ist working properly - it seems to be good code. Sorry for inconvience or misunderstanding caused
Kind regards
Mansoor

jade-guiton-dd:
@jpkrohling I can try to help with my two comments related to `sending_queue::batcher` (although I'm no expert on it myself), but I don't think I have the bandwidth right now to fully take over the PR, sorry...

Regarding the comment about shutdown behavior, it doesn't seem to be very well documented, but I think the relevant code is [here](https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/internal/queue/async_queue.go) (where the exporter waits until all queue consumers have shut down during Collector shutdown), and [here](https://github.com/open-telemetry/opentelemetry-collector/blob/2f20ee608cbb4243ae8f60f66b283a243164b02e/exporter/exporterhelper/internal/queue/memory_queue.go#L140-L159) (where the consumer checks for remaining elements before checking if the queue has been stopped, so it keeps going until the queue is empty). This matches our [general coding guidelines](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/coding-guidelines.md#graceful-shutdown), which ask that _"if components contain any temporary data they need to process and export it out of the Collector before shutdown is completed"_.

Regardless, there is still a risk of losing data when using the in-memory queue in the case of a crash, or if graceful shutdown times out, so I would just reword:

> By default, the sending queue buffers data in memory. If the Collector pod restarts, all queued data is lost.

to:

> By default, the sending queue buffers data in memory. If the Collector pod crashes or times out during a restart, all queued data is lost.

Similarly, regarding this part:

> However, in the typical configuration with an exporting queue enabled, back-pressure signals from the export destination do not propagate back to the sending applications. This means applications continue sending telemetry at full rate even when the Collector cannot keep up with exporting.

I think you could stay accurate by rewording to:

> However, in the typical configuration with an exporting queue enabled, back-pressure signals from the export destination do not propagate back to the sending applications. This means applications continue sending telemetry at full rate even when the destination cannot keep up with the received data.

> Note that if the Collector itself is the bottleneck, the exporter's sending queue will fill up, causing errors that should propagate back to the sender, except when using asynchronous processors in the pipeline such as the `batch` processor.

In any case, these are somewhat pedantic nitpicks; I am more concerned about being fair to Daemonset deployments and keeping the page focused by moving generic recommendations about dealing with high load that aren't specific to Daemonsets to the appropriate page.

tiffany76:
I'll take over on this one and update the PR based on @jade-guiton-dd's input.

Thanks for getting things started, @jpkrohling!

tiffany76:
@jade-guiton-dd, when you get a chance, could you take another look? I think I addressed all the comments.

Thank you!

content/en/docs/collector/deploy/agent.md
```diff
@@ -123,6 +123,71 @@ Cons:
...
+  spike cannot affect another.
+- **Processing is heavy**: Offload processors like `tail_sampling` or
+  `transform` to a [Gateway tier](/docs/collector/deploy/gateway/) and keep the
+  DaemonSet Collector lightweight (receive + forward only).
+- **High pod-to-node ratio**: When you have many small pods per node, a
+  DaemonSet works well. When you have a few large pods, a sidecar avoids the
+  blast radius problem.
+- **gRPC load balancing**: DaemonSet Collectors behind a Kubernetes Service
+  don't distribute gRPC connections evenly. Use a service mesh or sidecar
+  pattern for balanced gRPC distribution.
```
jade-guiton-dd, line 186: I'm not sure I understand this. Is this saying that "Typically, DaemonSet Collectors would be accessed through a Service with a Local internal traffic policy, meaning that pods communicate with the Collector on the same node, which means no distribution or load balancing happens"?

content/en/docs/collector/resiliency.md
```diff
@@ -27,21 +28,27 @@ sending queue.
...
     is dropped to prevent the Collector from running out of memory.
   - Retry Timeout: If the endpoint remains unavailable for longer than the
     configured maximum retry duration (default 5 minutes), the Collector will
     stop retrying for the oldest data in the queue and drop it.
+- Sizing the queue: The default queue size (1000 batches) might be too small for
+  high-throughput Collectors or too large for memory-constrained environments.
+  Tune `queue_size` based on:
+  - Expected telemetry volume
+  - Acceptable backend downtime (how long you want to buffer)
+  - Available memory or disk for the queue
```
jade-guiton-dd, line 51: Sounds good; but I would maybe also add the suggestion to look into adjusting [the sizer](https://github.com/open-telemetry/opentelemetry-collector/tree/main/exporter/exporterhelper#sending-queue) (sizing based on items or bytes instead of batches).

content/en/docs/collector/resiliency.md
```diff
@@ -181,6 +200,207 @@ backend, you can introduce a dedicated message queue like Kafka.
...
+    spec:
+      containers:
+        - name: collector
+          resources:
+            requests:
+              memory: 256Mi
+              cpu: 200m
+            limits:
+              memory: 512Mi
+              cpu: 500m
```
jade-guiton-dd, line 263: I would recommend including the `GOMEMLIMIT` environment variable in this example instead of leaving it until the next section, since it's useful and quite important even without the memory limiter processor.

You could also maybe note that the official Collector Helm chart [does this automatically](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/c738eda8d0bfa36513acdc9601165a308d9f506b/charts/opentelemetry-collector/templates/_pod.tpl#L93-L169). (Although they don't apply the "80%" rule mentioned below and in the [memorylimiter processor's "Best practices" section](https://github.com/open-telemetry/opentelemetry-collector/tree/main/processor/memorylimiterprocessor#best-practices)...)

content/en/docs/collector/resiliency.md
```diff
@@ -181,6 +200,207 @@ backend, you can introduce a dedicated message queue like Kafka.
...
+```yaml
+processors:
+  memory_limiter:
+    check_interval: 1s
+    limit_mib: 400 # Below the container memory limit (512Mi)
+    spike_limit_mib: 80
+```
+
+Set `limit_mib` to roughly 80% of your container memory limit, leaving room for
+memory that the Go runtime and other components allocate outside the pipeline.
```
jade-guiton-dd, line 286: Alternatively, we could suggest using `limit_percentage: 80` instead of hardcoding a calculated value in MiBs. (And the default value of 20% for `spike_limit_percentage` should work well enough here.)

content/en/docs/collector/resiliency.md
```diff
@@ -181,6 +200,207 @@ backend, you can introduce a dedicated message queue like Kafka.
...
+
+Additionally, set the `GOMEMLIMIT` environment variable to match the available
+memory. This helps the Go garbage collector make better decisions about when to
+reclaim memory, reducing the chance of OOMKills:
+
+```yaml
+env:
+  - name: GOMEMLIMIT
+    value: '400MiB' # Match or slightly below the memory_limiter limit
```
jade-guiton-dd, line 295: I don't think it makes sense to go below or even match the memory limiter limit here. The Go GC should pretty good at keeping the usage below GOMEMLIMIT, and I think you would want the memory limiter to kick in way before the GC starts thrashing.

For reference, what I've personally done in a past large scale experiment that was very memory bound was set `GOMEMLIMIT` to 92.5% of the hard memory limit, and `limit_percentage` to 85%. I was using the memory limiter _extension_ and a custom receiver to reject things at intake, so when using the memory limiter processor, you may want to be even more conservative.

content/en/docs/collector/resiliency.md
```diff
@@ -181,6 +200,207 @@ backend, you can introduce a dedicated message queue like Kafka.
...
+deserialized data in flight, and the sending queue (including its batch buffer).
+
+As a starting point:
+
+- Reserve at least 20% of the memory limit for the Go runtime and overhead
+- Allocate memory for the `memory_limiter` limit (the safety boundary)
+- Size the sending queue so that its maximum memory footprint fits within the
+  remaining budget
+- Monitor memory consumption under both normal and degraded conditions and
+  adjust
```
jade-guiton-dd, line 322: Not sure if this initial calculation makes sense. If the memory limiter limit is already 80% of the hard limit, and we need to reserve 20% for overhead, that leaves nothing for the sending queue.

Another problem is that sizing the sending queue based on a number of available bytes is non-trivial. If sizing on requests/items, the memory use per request/item can vary quite a bit, and even when sizing on "bytes", that refers to serialized OTLP bytes, not memory allocations.

Based on my limited prior experience, I would say a better "starting point" would be:
- initially set a very large amount of memory and no limits
- monitor the memory usage + queue size under stable load and normal conditions
- monitor how _quickly_ they increase if you artifically break the exporter to simulate the backend being unavailable
- decide how long of an outage you want to tolerate before you start dropping
- pick a queue size limit which triggers around that time
- pick a memory limit which triggers a bit later (in case memory use keeps increasing despite queue rejections)
- add 20% above that to get your final memory allocation

jade-guiton-dd (approved):
Looks pretty good, though I still have a few comments. Regarding the last point, I would be interested in hearing from other people with more expertise managing large fleets of Collectors under high load.

maryliag:
@jpkrohling and @jade-guiton-dd do you still want to collaborate on this PR or can we close it? There hasn't been activity in months

jpkrohling:
I don't have this PR on sight anymore, I believe @tiffany76 and @jade-guiton-dd took it over.

chalin:
👋 Please update this PR's branch over the latest `main`: it predates our link-checker [switch to Lychee][#10911] and can hit CI failures unrelated to your changes. For the how — including conflicts and after-update steps — see [#10990][]. Thanks for your contribution!

[#10911]: https://github.com/open-telemetry/opentelemetry.io/pull/10911
[#10990]: https://github.com/open-telemetry/opentelemetry.io/issues/10990

jpkrohling:
folks, I'm closing this but feel free to open a new PR based on this one.

PR #9229, opened 2026-02-19
chore(render-link): refactor render-link hooks and use shared partial

chalin:
- Contributes to #9167
- Extracts common link resolution and rendering into `_partials/hooks/render-link.html`
- Slims global `render-link.html` to a thin wrapper calling the partial
- Moves spec-specific logic into section-scoped `docs/specs/_markup/render-link.html`
- Drops `safeURL`/`safeHTML` from link rendering. See NOTES below.
- Fixes malformed `localhost:8080` URL in C++ getting-started page
- Adds `bench-render-link.sh` benchmark script under `scripts/stats/`

No changes to the generated site pages other than those due to the link fix and the timestamp:

```console
$ (cd public && git diff -bw --ignore-blank-lines -I '<a href="(http://)?localhost' -- ':(exclude)*.xml') | grep ^diff
```

---

I was concerned that using a partial in the link render hook might impact performance. Here are the benchmark results:

| Metric | baseline (6c685f4) | current-final | Delta |
|--------|-------:|-------:|------:|
| **Global hook** calls | 53,053 | 53,053 | |
| **Global hook** total | 904.5ms | 1,262.6ms | +358.1ms (+39.6%) |
| **Global hook** per call | 17.0µs | 23.8µs | +6.8µs (+40.0%) |
| **Spec hook** calls | 10,740 | 10,740 | |
| **Spec hook** total | 227.9ms | 319.1ms | +91.2ms (+40.0%) |
| **Spec hook** per call | 21.2µs | 29.7µs | +8.5µs (+40.1%) |
| **Partial** calls | -- | 63,793 | |
| **Partial** total | -- | 951.6ms | -- |
| **Partial** per call | -- | 14.9µs | -- |
| **Build total** | 4,216ms | 4,162ms | **-54ms (-1.3%)** |

Hook-level metrics show ~40% increase (inclusive of partial call time), but the **_overall_ build is 1.7% faster**. This is an average on 5 runs. Each starting from a empty folder. The partial itself costs only 15.1µs/call.

**NOTES:**

- **`safeHTML`** on `.Text` was redundant -- the text comes from Hugo's Markdown renderer and is already trusted HTML.
- **`safeURL`** on `$href` bypassed Go's `html/template` sanitization, which replaces `href` values with unrecognized URL schemes with `#ZgotmplZ`. The only URL that triggered this was `localhost:8080` in the C++ getting-started page (where `localhost:` was parsed as a scheme). That source has been fixed to use `http://localhost:8080`. With valid URLs in content, `safeURL` is unnecessary -- and removing it lets Go's built-in protection catch any future malformed URLs.
- **Detection strategy**: see TODO in comment below (that I'll create an issue from).


chalin:
**TODO:** add a post-build check that scans generated HTML for `#ZgotmplZ` markers. This gives a clear CI-time signal if any content introduces a URL that Go considers unsafe, without silently suppressing the protection at render time. (And we might be able to add some other checks too.)

See:

- #9230

PR #9239, opened 2026-02-20
docs: document prometheus metric suffixes and fix typo

### LINKED ISSUE #9183: [Docs]: internal telemetry page metrics are out of sync (eg. `otelcol_process_uptime`)
### What needs to be changed?

[This page](https://github.com/open-telemetry/opentelemetry.io/blob/main/content/en/docs/collector/internal-telemetry.md) shows metrics like `otelcol_process_uptime` whereas in new versions of the collector (unsure of which version but `0.145.0` shows this behaviour) the default metric name is actually: `otelcol_process_uptime_seconds_total`

You can get it back to the original with this config (which was hard to find documented):

```
service:
  telemetry:
    metrics:
      level: detailed
      readers:
        - pull:
            exporter:
              prometheus:
                host: '0.0.0.0'
                port: 8888
                without_type_suffix: true
                without_units: true
```

### Name + path of the page

content/en/docs/collector/internal-metrics.md

### Additional context

_No response_

mihir-dixit2k27:
This PR updates the Collector internal telemetry documentation to clarify how the Prometheus exporter applies unit and type suffixes to metric names.

Specifically:

Fixes an incorrect example in the _total suffix section (without_type_suffix: false → true).
Adds a new subsection, _seconds and other unit suffixes, explaining that:
The Prometheus exporter appends unit suffixes (e.g., _seconds) to metrics that carry a unit.
The _total suffix is appended for counters.
As a result, otelcol_process_uptime is exported as otelcol_process_uptime_seconds_total by default.
Provides a configuration example using both without_type_suffix: true and without_units: true to restore the original short metric names.
Adds a NOTE callout in the Metric views section clarifying that otelcol_process_uptime appears as otelcol_process_uptime_seconds_total when using the default Prometheus exporter, with a link to the new subsection.

These changes improve accuracy and help prevent confusion when users compare OTel instrument names with exported Prometheus metric names.
Fixes #9183

### DIFF
--- a/.github/workflows/pr-approval-labels.yml
+++ b/.github/workflows/pr-approval-labels.yml
@@ -1,3 +1,5 @@
+# To learn more about this workflow, see:
+#  https://opentelemetry.io/site/build/ci-workflows/#pr-approval-labels
 name: PR approval labels
 
 on:
--- a/.github/workflows/pr-review-trigger.yml
+++ b/.github/workflows/pr-review-trigger.yml
@@ -1,3 +1,5 @@
+# To learn more about this workflow, see:
+# https://opentelemetry.io/site/build/ci-workflows/#pr-approval-labels
 name: PR review trigger
 
 on:
--- a/content/en/site/_index.md
+++ b/content/en/site/_index.md
@@ -27,8 +27,8 @@ Tentatively planned content organization:
   choices, theme related decisions, and other design-level artifacts.
 - **Implementation** — Code-level structure and conventions, Hugo/Docsy
   templates, SCSS/JS customizations, patches, and internal shims.
-- **Build** — Tooling, local development setup, CI/CD workflows, deployment
-  environments, and automation details.
+- [**Build**](./build/) — Tooling, local development setup, CI/CD workflows,
+  deployment environments, and automation details.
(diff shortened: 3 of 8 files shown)


content/en/docs/collector/internal-telemetry.md
```diff
@@ -160,6 +160,13 @@ You can further configure how metrics from the Collector are emitted by using
 configuration updates the metric named `otelcol_process_uptime` to emit a new
 name `process_uptime` and description:
 
+> [!NOTE]
```
codeboten, line 163: This section is a bit confusing. The default Prometheus exporter configured in the Collector (if no readers are configured) already `without_type_suffix` and `without_units` to true, for backwards compatibility.

The note may be less confusing if written as "When configuring the Prometheus exporter for internal metrics"

mihir-dixit2k27: Thanks for the clarification, @codeboten! That makes a lot of sense. I didn't realize the default configuration preserved backward compatibility. I've pushed an update to clarify that this suffix behavior only applies when manually configuring the exporter via readers. Let me know if this reads better!

codeboten: Thanks @mihir-dixit2k27! it looks great

PR #9244, opened 2026-02-21
Add new utility motel to the registry

andrewh:
## Summary

- Adds [motel](https://github.com/andrewh/motel) to the OpenTelemetry Registry as a utility
- motel is a synthetic OpenTelemetry signal generator that produces traces, metrics, and logs from YAML topology definitions

## Changes

- New file: `data/registry/tools-go-motel.yml`


data/registry/tools-go-motel.yml
```diff
@@ -0,0 +1,24 @@
+title: motel
```
vitorvasc, line 2:
```suggestion
# cSpell:ignore Hodgson
title: motel
```

PR #9245, opened 2026-02-21
chore(style-guide): mention no-shortcut ref links, add rule to enforce, fix non-links

chalin:
- Followup to #9220
- Contributes to #9167
- **Style guide**: Documents preference for collapsed reference links (`[text][]` instead of shortcut form `[text]`) and explains why
- **Rule**: Adds `markdownlint-rule-no-shortcut-ref-link` with `ignore_pattern` for footnote-style `[1]` and range `[0..1]` labels
  - The rule is disabled for now until this PR lands with select file fixes
- **Fixes** `[some text]` instances that are not links
- Adds ignore mdl directives for legitimate cases

### DIFF
--- a/.markdownlint.yaml
+++ b/.markdownlint.yaml
@@ -41,11 +41,12 @@ table-column-style: false # Causes problems in ja and zh pages
 # to provide configuration. Rules that need no config can be omitted.
 
 no-shortcut-ref-link:
+  false # TODO: re-enable when this PR lands with select file fixes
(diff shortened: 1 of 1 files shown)


chalin:
The changes for SIGs other than @open-telemetry/docs-approvers are editorial, so with @vitorvasc's approval, I'm going to merge. Feedback is still welcome from locales and other SIGs.

PR #9248, opened 2026-02-21
chore(cleanup): applies `no-shortcut-ref-link --fix` mdl rule to all en pages

chalin:
- Followup to #9245
- Applies `no-shortcut-ref-link` mdl rule with `--fix` argument to all en pages

> [!NOTE]
> The current rule checks and fixes only single word ref links.


chalin:
Editorial changes for non SIG-docs SIGs. Merging with @vitorvasc's approval. Feedback is still welcome.

Issue #5374, opened 2024-10-10
Update go example submodule from core to contrib


svrnm:
> Yes, the dice demo is indeed moved to the contrib repository.
https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/examples/dice

_Originally posted by @dmathieu in https://github.com/open-telemetry/opentelemetry.io/issues/5364#issuecomment-2404258288_

ranjaniradha:
I'm working on this task.

jack-berg:
FYI, I'm working on this because I want to add more go examples for #9263

ManniMansoor:
see this and that coming

chalin:
@jack-berg - has this been completed?

jack-berg:
Yes completed in #9462.

PR #9263, opened 2026-02-23
Prometheus client compatibility guide page

jack-berg:

[^I-know-my-stuff]:

---

Solves https://github.com/open-telemetry/opentelemetry-specification/issues/4875.

Proposal to add a new page at location "Docs -> Compatibility -> Prometheus Clients".

The point of this document is to act as a resource for users familiar with prometheus client APIs and concepts, and explain how those APIs / concepts map to opentelemetry.

Its a combination of prose describing conceptual differences, and code snippets showing how to achieve key API usage patterns in both prometheus and opentelemetry, and in a variety of languages.

PRs for referenced code excerpts:

- opentelemetry-java-examples: ttps://github.com/open-telemetry/opentelemetry-java-examples/pull/1032
- opentelemetry-go-contrib: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8715

I'm less confident in the go examples, but thought it was important to include to show the vision for the page. That is, I envision the page demonstrating these key API usage patterns in a variety of languages [which have popular prometheus client libraries](https://prometheus.io/docs/instrumenting/clientlibs/).

~~Also, the go examples are going to be dependent on resolving #5374, so that we can have a proper home for these snippets with compile validation.~~

### [See Page Preview](https://deploy-preview-9263--opentelemetry.netlify.app/docs/compatibility/prometheus/)

cc @open-telemetry/prometheus-interoperability


chalin (commented):
Nice work.

A comment about the content organization. Let me first make a proposal about a possible _logical_ organization and we can discuss how to implement it later -- if you like the proposed logical organization.

As a logical organization, what if we had:

- An index file with the content of the conceptual differences section
- Then subsection pages per language. In these pages the tabbed panes would be for OTel vs Prometheus

It feels like that would offer a good experience for readers who might want to focus on one language or at least one at a time. 

WDYT @jack-berg?

jack-berg:
That's definitely an option! 

One observation I made developing this page that I was surprised by: if I click the "Go" or "Java" tab of a particular tabpane, all the other tabpanes on the page update together. It was unintuitive at first, but I grew to like it because as you note, a typical use would want to consume all the java content, or all the go content. It would be unusual to want to consume a combination of both at the same time.

One benefit of keeping the content all on one page is that it forces a degree of uniformity in the set of examples that are discussed and how they are presented. I didn't was to build an exhaustive guide and found that focusing on the concepts / flows that all or most clients helped resist going down the rabbit hole of too much detail.

Of course, there are likely techniques that we could leverage to achieve uniformity which don't display everything all on one page. 

Ultimately I'm happy to defer the docs maintainers on info architecture questions like this.

content/en/docs/migration/prometheus.md
```diff
@@ -0,0 +1,1881 @@
...
+
+In Prometheus, metrics register to a registry — by default a global one. You can
+declare a metric anywhere in your code, and it becomes available for scraping
+once registered. The exporter (HTTP server or OTLP push) is attached to the
+registry as a separate, independent step.
+
+In OpenTelemetry, metrics flow through a `MeterProvider`, which you configure
+upfront with exporters and collection schedules. You obtain a `Meter` — scoped
+to your library or component — from the provider, and create instruments from
+that `Meter`. The practical consequence is that **you configure where metrics go
```
dashpole: > you configure where metrics go before you create them

Not sure about this...  At least in Go, most instrumentation libraries default to the global MeterProvider, which seems equivalent to defaulting to the global Registry. In both OTel and Prometheus, you can create a non-global MeterProvider or Registry, and pass it explicitly to instrumentation libraries that support an explicit one. So I think in either case you can define metrics before-or-after creating the MeterProvider/Registry

content/en/docs/migration/prometheus.md
```diff
@@ -0,0 +1,1881 @@
...
+configurable — for example, to preserve UTF-8 characters or suppress unit and
+type suffixes. See the
+[Prometheus exporter](/docs/specs/otel/metrics/sdk_exporters/prometheus/)
+configuration reference for details.
+
+### Stateful and callback instruments
+
+Both systems support two recording modes:
+
+- **Prometheus** calls these _stateful_ (`Counter`, `Gauge`) and _callback_
```
dashpole: Interesting.  I don't think the Go Prometheus client has stateful vs callback.  It just has the instruments (Gauge, Counter), and then "Const" versions of the instrument, which can be passed directly to Collect() as part of a "giant asynchronous callback" of sorts.

content/en/docs/compatibility/migration/prometheus.md
```diff
@@ -0,0 +1,1881 @@
...
+Both systems support two recording modes:
+
+- **Prometheus** calls these _stateful_ (`Counter`, `Gauge`) and _callback_
+  (`CounterWithCallback`, `GaugeWithCallback`). Stateful metrics maintain their
+  own value; callback metrics invoke a function at scrape time to get the
+  current value.
+- **OpenTelemetry** calls these _synchronous_ (counter, histogram, etc.) and
+  _asynchronous_ (observed via a registered callback). The semantics are the
+  same.
+
```
dashpole: This seems like a good place to cover UpDown counters vs gauges? Prometheus doesn't have updown counters, and things that OTel represents as UpDown counters (e.g. memory usage) are gauges in Prometheus.

content/en/docs/compatibility/migration/prometheus.md
```diff
@@ -0,0 +1,1881 @@
...
+Instrumentation code should depend only on the API. The SDK is configured once
+at application startup and wired to an API reference that gets passed to the
+rest of the codebase. This keeps instrumentation library code decoupled from any
+specific SDK version and makes it straightforward to swap in a no-op
+implementation for testing.
+
+### OTel: Instrumentation scope
+
+Prometheus metrics are global: every metric in a process shares the same flat
```
dashpole: Prometheus would say that the name is always namespaced.  E.g. Prometheus would do `acme_http_requests_total` vs OTel's `http.requests.total` in scope `go.acme.io/client`.

content/en/docs/migration/prometheus.md
```diff
@@ -0,0 +1,1881 @@
...
+
+Prometheus metrics are global: every metric in a process shares the same flat
+namespace, identified only by name and labels.
+
+OpenTelemetry scopes each group of instruments to a `Meter`, identified by a
+name and optional version (for example, `smart.home`). When exporting to
+Prometheus, the scope name and version are added as `otel_scope_name` and
+`otel_scope_version` labels on every metric point. These labels appear
+automatically and may be unfamiliar to users coming from Prometheus. The
+exporter also emits a separate `otel_scope_info` gauge (always 1) for each
```
dashpole: This isn't the case anymore since scope attributes became identifying.  We've removed this _info metrics.

content/en/docs/migration/prometheus.md
```diff
@@ -0,0 +1,1881 @@
...
+
+```go
+package main
+
+import "github.com/prometheus/client_golang/prometheus"
+
+func counterCallbackUsage(reg *prometheus.Registry) {
+	// The smart energy meter maintains its own cumulative joule total in firmware.
+	// Use CounterFunc to report that value at scrape time.
+	reg.MustRegister(prometheus.NewCounterFunc(
```
dashpole: I've never seen this used in the wild.  Doesn't mean it isn't... but not being able to provide lables means almost everyone implements a custom collector.

content/en/docs/compatibility/migration/prometheus.md
dashpole: Ah, I read more about it down below.  Maybe just mention it here, and link to the section below?

content/en/docs/migration/prometheus.md
```diff
@@ -0,0 +1,1881 @@
...
+package main
+
+import (
+	"context"
+
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/metric"
+)
+
+// Preallocate attributes when values are static to reduce per-call allocation.
```
dashpole: This doesn't actually reduce allocations.  The important thing to do is to preallocate the _option_ and the _option slice_ (which both can result in allocations from copying the input attributes, and from the slice allocation).  E.g. 

```golang
zoneUpstairsOpts   = []metric.RecordOption{metric.WithAttributes(attribute.String("zone", "upstairs"))}
```

content/en/docs/compatibility/migration/prometheus.md
```diff
@@ -0,0 +1,1881 @@
...
+Prometheus `Histogram` maps to the OpenTelemetry `Histogram` instrument.
+
+### Classic (explicit) histogram
+
+Both systems support classic histograms, where fixed bucket boundaries partition
+observations into discrete ranges.
+
+- **Bucket configuration**: Prometheus declares bucket boundaries on the
+  instrument itself at creation time. In OpenTelemetry, bucket boundaries are
+  set on the instrument as a hint that can be overridden or replaced by views
```
dashpole: From an end user perspective, I think this is overcomplicated.  In both Prom and OTel, you set the histogram buckets at instrument creation time.

Additionally OTel has views that let you configure a wide variety of things... histogram buckets, min/max, names, attributes, units, etc. Probably best to just cover views elsewhere and leave this section simple.

content/en/docs/migration/prometheus.md
```diff
@@ -0,0 +1,1881 @@
...
+
+{{% /tab %}} {{< /tabpane >}}
+
+### Native (base2 exponential) histogram
+
+Both systems support native (base2 exponential) histograms, which automatically
+adjust bucket boundaries to cover the observed range without requiring manual
+configuration.
+
+- **Format selection**: Prometheus selects the native histogram format at
```
dashpole: Hmmm... I would double-check this with a prom java maintainer.  Prometheus _should_ be collecting both the fixed-bucket and native histogram in-memory, and can provide both simultaneously when scraping in the Proto format.  This was designed to make migrations between fixed-bucket and native histograms easy.

content/en/docs/migration/prometheus.md
```diff
@@ -0,0 +1,1881 @@
...
+
+{{% /tab %}} {{< /tabpane >}}
+
+### Summary {#summary}
+
+Prometheus `Summary` computes quantiles client-side at scrape time and exposes
+them as labeled time series (for example, `{quantile="0.95"}`). OpenTelemetry
+has no direct equivalent.
+
+A histogram with no explicit bucket boundaries is a good stand-in for most
```
dashpole: I'm not super sure about this... I would probably recommend exponential histograms as the best alternative since it provides quantiles with bounded errors.

I think histograms without buckets are much more useful for delta than they are for cumulatives because min and max are actually useable and useful.

dashpole (commented):
Amazing work. This must have taken a long time to put together.

I know it is already long. Did you think about covering bridges?

cartermp:
Hmmm. TBH, I would prefer that we keep it as-is, file an issue for future refactor later, and let a copilot agent or something handle moving language-specific content into separate files ... if that's ultimately what we prefer.

content/en/docs/migration/prometheus.md
```diff
@@ -0,0 +1,1881 @@
...
+Prometheus requires label _names_ to be declared at metric creation time. Label
+_values_ are bound at record time, via `labelValues(...)`.
+
+OpenTelemetry has no upfront label declaration. Attribute keys and values are
+both provided together at the time of the measurement via `Attributes`.
+
+### Naming conventions
+
+Prometheus uses `snake_case` metric names. Counter names must end in `_total`;
+the client library appends this automatically if omitted.
```
ArthurSens: Hmmmm, at first glance, this looks a bit weird. I don't think the Prometheus client library adds this, at least not in Go. I putting on my TODO list to verify this later

content/en/docs/compatibility/migration/prometheus.md
```diff
@@ -0,0 +1,1881 @@
...
+Prometheus, the exporter translates names: dots become underscores, unit
+abbreviations expand to full words (for example, `s` → `seconds`), and counters
+receive a `_total` suffix. An OpenTelemetry counter named `hvac.on` with unit
+`s` is exported as `hvac_on_seconds_total`. See the
+[compatibility specification](/docs/specs/otel/compatibility/prometheus_and_openmetrics/)
+for the complete set of name translation rules. The translation strategy is
+configurable — for example, to preserve UTF-8 characters or suppress unit and
+type suffixes. See the
+[Prometheus exporter](/docs/specs/otel/metrics/sdk_exporters/prometheus/)
+configuration reference for details.
```
ArthurSens: Well, it's in the spec but as we also realized... almost no SDKs have implement this part of the spec completely 🤷

content/en/docs/migration/prometheus.md
ArthurSens: Isn't this a callback?
https://github.com/prometheus/client_golang/blob/4b86739a4e19a9d33b78ab94fc9d0c54bfc6ebf6/prometheus/gauge.go#L286-L311

The name certainly doesn't align with `GaugeWithCallback`, if that's what Java uses 🤔

content/en/docs/migration/prometheus.md
ArthurSens: Yep, just to clarify... all scope attributes become label as `otel_scope_[attr name]=[attr value]`

content/en/docs/migration/prometheus.md
ArthurSens: CounterFunc and GaugeFunc is used quite a few times inside the Prometheus codebase :)

ArthurSens: Oh, maybe you mean metric creation inside a callback? Then that's also something I've never seen before 🤔

ArthurSens: These are good callback examples:

https://github.com/prometheus/prometheus/blob/1751685dd4f6430757ba3078a96cffeffcb2bb47/cmd/prometheus/main.go#L1557-L1573

It's used mostly to access data at scrape time.

ArthurSens (commented):
I did a first pass, could review everything in detail yet!

content/en/docs/migration/prometheus.md
dashpole: I think this behavior is only for OpenMetrics

content/en/docs/migration/prometheus.md
jack-berg: > It just has the instruments (Gauge, Counter), and then "Const" versions of the instrument, which can be passed directly to Collect() as part of a "giant asynchronous callback" of sorts.

What about prometheus.NewGauageFunc, prometheus.NewCounterFunc? Is this not the equivalent of a callback?

If callback and stateful isn't the right vocabulary, is there a more idiomatic way to talk about this?

content/en/docs/migration/prometheus.md
jack-berg: Addressed in the latest commit, indicating that counter names end in `_total` by convention without stating that the libraries enforce this

content/en/docs/compatibility/migration/prometheus.md
jack-berg: I think its fine -- too complex to have to consider the implementation status of features when writing, on top of everything else

content/en/docs/migration/prometheus.md
jack-berg: Sounds good - addressed in latest commit.

content/en/docs/migration/prometheus.md
jack-berg: You are correct. Addressed in latest commit

content/en/docs/compatibility/migration/prometheus.md
jack-berg: 👍 Addressed in latest commit

content/en/docs/migration/prometheus.md
jack-berg: 👍 Addressed in latest commit

content/en/docs/migration/prometheus.md
jack-berg: I've adjusted all the examples to record labels on the callback instruments, and use the custom collector approach. Since this is the more common case, it makes sense to emphasize it, even if its more verbose than `NewCounterFunc`.

content/en/docs/migration/prometheus.md
jack-berg: 👍 I got that wrong. Addressed in latest commit.

content/en/docs/compatibility/migration/prometheus.md
jack-berg: Addressed in latest commit

content/en/docs/compatibility/migration/prometheus.md
jack-berg: Addressed in latest commit

content/en/docs/migration/prometheus.md
jack-berg: I rephrased in the latest commit. Let me know what you think

content/en/docs/migration/prometheus.md
jack-berg: Addressed in latest commit

jack-berg:
Thanks for the reviews @dashpole and @ArthurSens. I've merged in the #9379 so that this page is now at `Compatibility -> Prometheus Clients`, instead of `Migration -> Prometheus Client`. I've also taken a stab at addressing all your feedback - pretty sure I got all of it! 😁 

> I know it is already long. Did you think about covering bridges?

I could add a section about bridges. In java, we have a dedicated doc section listing all of our shims: https://opentelemetry.io/docs/languages/java/instrumentation/#shims

Would be good if there was a more detailed external page we could link to for each language.

@dashpole what do you think about the go examples? To questions in particular:

- As mentioned in the description, I added them here to illustrate the vision of the page but not being a go developer, am not an expert. Do you think they are good enough for this initial pass?
- As mentioned in the description, the java examples are guaranteed to compile because I added them to `opentelemetry-java-examples`. The go examples are plain text and may have uncaught typos. Ideally I would add them as compiled examples in one of the go repos and embed via the excerpt mechanism, but that's blocked by #5374. Do you think I / we should try to resolve #5374 before landing this, or land this and later split out the examples as proper excerpts?

dashpole:
#5374 sounds trivial, but it hasn't been done in 2 years so maybe i'm missing something. Another option is to pull from the metrics api examples themselves if the examples are simple: https://pkg.go.dev/go.opentelemetry.io/otel/metric#pkg-examples.  We can expand those if it makes sense.

jack-berg:
Cool - I updated in #9462. After that merges, I'll open a PR against opentelemetry-go-contrib to propose adding the examples here so that they can be referenced using the excerpts functionality.

content/en/docs/compatibility/migration/prometheus.md
```diff
@@ -0,0 +1,1977 @@
...
+### OTel: Aggregation temporality
+
+Prometheus metrics are always cumulative. OpenTelemetry supports both cumulative
+and delta temporality, but the Prometheus exporter enforces cumulative for all
+instruments. For developers migrating from Prometheus, this is transparent — the
+behavior you already rely on is preserved.
+
+### OTel: Resource attributes
+
+Prometheus identifies scrape targets using `job` and `instance` labels, which
```
dashpole: `target_info` is actually a metric defined in the openmetrics 1.0 spec.  Prometheus applications are encouraged to use this metric to provide information about themselves at scrape time. It isn't a particularly well-adopted pattern in Prometheus, but it isn't something that was invented for OTel compatibility.  A prometheus user who is creating a target_info metric with labels would want to migrate to setting resource attributes.

content/en/docs/compatibility/migration/prometheus.md
```diff
@@ -0,0 +1,1977 @@
...
+
+	hvacOnTime.Add(ctx, 127.5, zoneUpstairsOpts...)
+	hvacOnTime.Add(ctx, 3600.0, zoneDownstairsOpts...)
+}
+```
+
+Key differences:
+
+- `Add(value)` → `Add(ctx, value, metric.WithAttributes(...))`. All instrument
+  calls require a `context.Context` as the first argument.
```
dashpole: The context argument reminds me that I would love for this to compare how to add exemplars in both.  But lets do that in a follow-up.

content/en/docs/compatibility/prometheus.md
```diff
@@ -0,0 +1,1977 @@
...
+}
+```
+
+OpenTelemetry
+
+In OpenTelemetry, the instrumentation code is identical to the classic histogram
+case. The base2 exponential format is configured separately, outside the
+instrumentation layer.
+
+The Go OTLP exporter does not support exporter-level aggregation selection
```
dashpole: I think it does?

https://github.com/open-telemetry/opentelemetry-go/blob/e9449e3b34f1cb7e8818f90f1955e9c4b4c3dd53/exporters/otlp/otlpmetric/otlpmetrichttp/config.go#L215

content/en/docs/compatibility/prometheus.md
```diff
@@ -0,0 +1,1977 @@
...
+	return sdkmetric.NewView(
+		sdkmetric.Instrument{Name: "device.command.duration"},
+		sdkmetric.Stream{Aggregation: sdkmetric.AggregationBase2ExponentialHistogram{}!},
+	)
+}
+```
+
+Key differences:
+
+- `NativeHistogramBucketFactor` must be set to a value greater than 1.0 to
```
dashpole: One thing i've noticed out in the wild is that since most users set this to 1.1, users usually end up with fewer buckets in native histograms (maybe 50 on average) compared to OTel which tends to be closer to the max number of buckets. If my math is right, this would be the equivalent of setting a MaxScale of 3 on exponential histograms (2^2^^-3 ~= 1.1).

In practice, the reason you might want to note this is that if migrating to OTel raises someone's bill due to the higher number of buckets, they can set MaxScale to keep their existing behavior.

content/en/docs/compatibility/prometheus.md
jack-berg: Oops. Updated to reflect this.

content/en/docs/compatibility/migration/prometheus.md
jack-berg: Updated to talk about this.

content/en/docs/compatibility/prometheus.md
jack-berg: Updated to add a comment saying that setting MaxScale=3 is roughly equivalent in density to the commonly used value of 1.1.

jack-berg:
Alright happy to address more comments, but just a point of process: The build is failing because the submodules checked in currently in this PR are different than the pinned hashes. Once we reach consensus on the content, we can merge the respective PRs to update opentelemetry-java-examples / opentelemetry-go-contrib. At this point, I'll update the pinned hashes to reflect the latest on main of those repos, and the build will pass.

dashpole (approved):

jaydeluca (approved):

jack-berg:
Hey @chalin / @cartermp, with two approvals on this from prometheus SMEs, and with approvals on the respective example PRs to go and java repos, I think we're waiting for a positive signal from you to continue.

If it looks good from your side, I'll ask that the go / java PRs be merged and update the hashes here, but don't want to make that request if the examples are going to churn.

MrAlias (approved):

jack-berg:
Friendly ping @open-telemetry/docs-maintainers 🙂

tiffany76 (approved):
LGTM from a copy edit perspective. Let us know when the build is passing!

chalin (changes requested):
This will need to be revisited given that #9638 removed the tools folder. Let me take a closer look.

@jack-berg - if I can resolve the link issues and adjust to the tools folder going away, are you ok if I just push a commit to this PR?

chalin (commented):
@jack-berg - which content-module update are essential to this PR? Go and Java I presume?

The build is failing which versions of those do you want exactly?

jack-berg:
> @jack-berg - if I can resolve the link issues and adjust to the tools folder going away, are you ok if I just push a commit to this PR?

Yes you can push to the branch. But links is failing because I checked in the changed to the submodules for demonstration purposes. To fix, I need to [merge the corresponding PRs to go / java](https://github.com/open-telemetry/opentelemetry.io/pull/9263#issuecomment-4121546979), and update the submodules to point to the latest refs. Now that @tiffany76 has approved, [I've requested the opentelemetry-go-contrib](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8716#issuecomment-4238826609) PR be merged, and have [merged the opentelemetry-java-examples PR](https://github.com/open-telemetry/opentelemetry-java-examples/pull/1032).

jack-berg:
@chalin dependent PRs in opentelemetry-go-contrib, opentelemetry-java-examples are merged, .gitmodules updated to point to latest refs, and cleaned up. 

I think this is ready for a final look.

chalin (approved):
LGTM. Thanks for this! 🙌🏻

PR #9286, opened 2026-02-26
Update Ruby instrumentation documentation

ltickett:
It took me 4 days to figure out how to instrument my Ruby on rails application to send trace and log data.

I do wonder if we should remove the exclusion of the logger instrumentation from https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/all/Gemfile#L31 maybe?

But if not, I added instructions for installing the logger instrumentation package and suggested using the all instrumentation package and not needing to manually setup tracing if so.

- [ ] I have the experience and knowledge necessary to understand, review, and validate all content in this PR.[^I-know-my-stuff]

[^I-know-my-stuff]:


ltickett:
@open-telemetry/docs-approvers could you take a look please?

ltickett:
> @open-telemetry/docs-approvers could you take a look please?

I guess you can't mention a group like this :)

@jaydeluca @maryliag could you take a look at this please? thanks

content/en/docs/languages/ruby/instrumentation.md
```diff
@@ -20,8 +20,29 @@ First, ensure you have the SDK package installed:
...
+Then install one or more instrumentation packages:
+
+```sh
+gem install opentelemetry-instrumentation-all
+```
+
+If you use the all instrumentation package, you do not need to follow the
+instructions below to enable [traces](#traces).
+However, logging is not instrumented by default, so you will need to install
+the package mentioned below to enable [logs](#logs).
```
chalin: Thanks @ltickett. I think that we might be able to improve on the flow here (I'll let y'all ensure the technical accuracy):

~~~suggestion
Then include configuration code that runs when your program initializes. Make
sure that `service.name` is set by configuring a service name (for example via
the `OTEL_SERVICE_NAME` environment variable).

To instrument many common libraries without enabling each instrumentation
individually, use `opentelemetry-instrumentation-all`:

```sh
gem install opentelemetry-instrumentation-all
```

Then configure the SDK to enable available instrumentations:

```ruby
require 'opentelemetry/sdk'

OpenTelemetry::SDK.configure do |c|
  c.use_all
end
```

> [!TIP]
>
> In a Ruby on Rails app, you can place this configuration in
> `config/initializers/opentelemetry.rb`.

This approach automatically enables supported library instrumentations (so you
typically don’t need to follow the per-instrumentation steps in the
[Traces](#traces) section below). You may still need to configure an exporter
and environment variables appropriate for your deployment.
~~~

content/en/docs/languages/ruby/instrumentation.md
```diff
@@ -351,6 +372,12 @@ The metrics API & SDK are currently under development.
 
 The logs API & SDK are currently under development.
 
+Ensure you have the logger instrumentation package installed:
+
+```sh
+gem install opentelemetry-instrumentation-logger
+```
+
```
chalin: Here's what co-pilot had to suggest as a rewrite. WDYT?

~~~suggestion
To capture application logs emitted via Ruby’s standard `Logger`, install the
logger instrumentation package:

```sh
gem install opentelemetry-instrumentation-logger
```

After installing, ensure you have configured an exporter/endpoint that supports
logs for example, via the exporter configuration referenced in the next section.
~~~

ltickett:
Thanks @chalin - I like your suggestions and have applied them 🙏

ltickett:
> Thanks @chalin - I like your suggestions and have applied them 🙏

GitHub really screwed up applying the suggestions- but I think I manually unpicked it :/

chalin (approved):
LGTM from the techdocs side of things. Ruby folks, awaiting your review.

cartermp:
So I am certainly in favor of adding this at the top, but we do have this information in the _getting started_ docs, and the general pattern in these docs is:

- _getting started_ docs show how to install instrumentation libraries, use an autoinstrumentation agent, etc.
- _instrumentation_ docs show how to create custom instrumentation in code

This PR would deviate from how the rest of the language do it.

**Now, here's why I think that's fine:**

1. Most Ruby devs aren't going to care that the Java, .NET, and Go docs all follow a similar pattern
2. A little duplication in the docs helps ensure that people who land on one page don't get unintentionally misled, such as anyone who thinks that instrumentation + <language> = custom instrumentation only, which is certainly not true

@svrnm @chalin I'll also raise this in the approvers channel.

ltickett:
> So I am certainly in favor of adding this at the top, but we do have this information in the _getting started_ docs, and the general pattern in these docs is:
> 
> * _getting started_ docs show how to install instrumentation libraries, use an autoinstrumentation agent, etc.
> * _instrumentation_ docs show how to create custom instrumentation in code
> 
> This PR would deviate from how the rest of the language do it.
> 
> **Now, here's why I think that's fine:**
> 
> 1. Most Ruby devs aren't going to care that the Java, .NET, and Go docs all follow a similar pattern
> 2. A little duplication in the docs helps ensure that people who land on one page don't get unintentionally misled, such as anyone who thinks that instrumentation +  = custom instrumentation only, which is certainly not true
> 
> @svrnm @chalin I'll also raise this in the approvers channel.

Thanks @cartermp - the key missing bit I think was the bit about logs. I could consider adding that to the getting started page instead or as well. Happy either way- just know I couldn't find that information anywhere :)

I'm working on metrics now, so will probably raise a follow-up PR adding docs around how to instrument them too :)

chalin:
Thanks for the context @cartermp. If we can keep things DRY (possibly using includes or cross-referenciing), that would be great, but I support your arguments to have the info accessible in more than one place.

svrnm:
> the key missing bit I think was the bit about logs. I could consider adding that to the getting started page instead or as well. Happy either way- just know I couldn't find that information anywhere :)

Yes, this is missing indeed, if you look for other languages like Java, etc, they provide details about logs, so you could use this as a point of reference on how to add this to the getting started and/or instrumentation docs.

For the question where and how we want to have this info, for some languages we have a info box at the top of the page calling out that there are automatic instrumentations and instrumentation libraries (https://opentelemetry.io/docs/languages/js/instrumentation/), although this would require the user to know what these terms mean. 

I think I would be fine to have this added for now.

svrnm:
@open-telemetry/ruby-approvers PTAL!

kaylareopelle (approved):
This is fantastic! Thank you, @ltickett!

kaylareopelle:
> It took me 4 days to figure out how to instrument my Ruby on rails application to send trace and log data.
>
> I do wonder if we should remove the exclusion of the logger instrumentation from https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/all/Gemfile#L31 maybe?
>
> But if not, I added instructions for installing the logger instrumentation package and suggested using the all instrumentation package and not needing to manually setup tracing if so.

First off, @ltickett, I'm really sorry that our documentation wasn't in a better place to help you get your application set up. I appreciate your efforts to improve the docs to help others. 

Right now, our thinking is to exclude logger instrumentation from instrumentation-all because instrumentation-all currently only emits traces. Our API/SDK for traces is stable, but our API/SDK for logs is still experimental. Even though the instrumentation packages are at a 0.x level, we want to keep experimental signals separate from stable signals. 

We are working on adding metrics instrumentation, so these goals may change when metrics are emitted.

If you have a different opinion, we'd be happy to hear it and discuss further!

ltickett:
Thanks @kaylareopelle - i've also just managed to get metrics working in my application.

I did read somewhere that's it's possible to get some automatic metric instrumentation? But I so far have only been able to add manual metrics.

Happy to document that in a follow-up too.

kaylareopelle:
Hi @ltickett, that's exciting too! We don't have automatic metrics instrumentation for any of the libraries yet. That will hopefully be available soon. 

I wondering if you were reading about the older [Metrics Reporter](https://github.com/open-telemetry/opentelemetry-ruby/blob/main/sdk/lib/opentelemetry/sdk/trace/export/metrics_reporter.rb) feature? That one isn't an implementation of the OTel Metrics spec, but instead a separate off-spec implementation to get metrics on how the OpenTelemetry SDK/Exporters/etc. are behaving. It was developed before OTel metrics were stabilized.

svrnm (approved):

svrnm:
I will merge this for now, but create a follow up issue for the larger docs to revisit how we can make this more streamlined

PR #9400, opened 2026-03-13
[ja] Update Collector components table to match English version

vitorvasc:
- [x] I have read and followed the [Contributing](https://opentelemetry.io/docs/contributing/) docs, especially the "**First-time contributing?**" section.
- [ ] This PR has content that I did not fully write myself.
  - [ ] I used AI and I have read and followed the [Generative AI Contribution Policy](https://github.com/open-telemetry/community/blob/main/policies/genai.md).
- [x] I have the experience and knowledge necessary to understand, review, and validate all content in this PR.[^I-know-my-stuff]

[^I-know-my-stuff]:
    Yes, I can answer maintainer questions about the content of this PR, without using AI.

---

As some of the Collector components have been deprecated over the past few releases, we've started seeing errors in our daily reference cache updates (e.g. [this](https://github.com/open-telemetry/opentelemetry.io/actions/runs/23045651210/job/66934198746?pr=9395) and [this](https://github.com/open-telemetry/opentelemetry.io/actions/runs/22997295247/job/66772928682?pr=9390) workflow run). 

To address this, we initially began manually patching the Japanese Collector component pages by removing them (as seen in [this PR](https://github.com/open-telemetry/opentelemetry.io/pull/9390/changes/80e1842122b426280e85ed7f1a2f22f8de3d2780)). However, following the changes introduced in #9315 , we are now linking deprecated components to their last known versions to avoid them suddently disappearing from the list. This PR synchronizes that same behavior for the Japanese content.

/cc @open-telemetry/docs-ja-approvers

PR #9402, opened 2026-03-13
Make collector component docs more translatable

### LINKED ISSUE #8644: Ensure localization of Collector component tables is preserved during regeneration
## Context

The [Collector components pages](https://opentelemetry.io/docs/collector/components/) were introduced in #8001 and are generated via automation, which scrapes Collector metadata and regenerates the tables in these pages after each Collector release.

As of #8464, we already have translations for those pages. The translated table headings are surrounded by the `<!-- BEGIN GENERATED: [...] -->` generation block. A subsequent regeneration could potentially overwrite or affect the localized content.

See the discussion in https://github.com/open-telemetry/opentelemetry.io/pull/8464/changes#r2613525033.

## Proposal

Ensure the generated content only outputs the table rows (data) within the generation markers so that future updates to that content won't affect the localized content.

- [ ] Move Collector components data into the [`./data`](https://github.com/open-telemetry/opentelemetry.io/tree/main/data) directory, aligning it with other files such as registry entries, community members, etc.
- [ ] ...

/cc @jaydeluca

jaydeluca:

[^I-know-my-stuff]:

---

Resolves #8644

This should simplify the way we handle translations for these pages, but I know that using the `i18n/*.yaml` files is not the current pattern. I wanted to propose the use of it for these pages, but can go back to the drawing board if there is a reason to avoid using them.

This changes the automation, which will now update the data yaml files instead of the pages directly and then use a shortcode to generate the tables

If this is accepted, I can update the [localization docs](https://opentelemetry.io/site/build/localization/) to include a step to translate these values in the `i18n` folder as well. 

I can followup and update the Japanese page to follow this new pattern as well (ref: #9400).

------

## Context

The current state of things prior to this PR is that we have code in scripts/collector-sync/src/documentation_sync/ that references the [registry](https://github.com/jaydeluca/opentelemetry-ecosystem-explorer/tree/main/ecosystem-registry/collector) in the ecosystem explorer, and when it detects changes (from a new release), it will automatically open a PR. It uses markers within each page to determine what sections to update (example automated PR: #9314).

The challenge with handling localization is that we only need to translate the table headers and footnotes for each language, but the current state of things is not setup well for that, for a few reasons:

- For each language, we would need to maintain an entire page of content, and then update that as well. This isn't a huge dealbreaker, but would require us to update the automation every time a new language is added.
- The table headings that need to be translated cause an issue because the markdown rendering breaks if we put the markers in a position that excludes the table headers, which makes it really difficult for us to have automation that excludes overwriting translated headers
- On every update, we also have to have a step in the automation that identifies any new cspell issues, and adds them to the frontmatter automatically

[This PR](https://github.com/open-telemetry/opentelemetry.io/pull/9400) is an example of the challenges with the current approach that this PR solves for.

With this new change, here's what happens instead:

- Instead of having the script update markers and the pages directly, we update a data file
- For each new language that we want to support, instead of copying the contents of the entire page and all the items, only the frontmatter and intros would need to be translated on each page, and then a handful of terms just need to be added to the appropriate `i18n/<language>.yaml` file. The shortcode handles the rest
- We avoid the cspell issues entirely

So this allows us to more easily translate the headers and footnotes, and avoids having to copy/paste the entire contents of the tables per language.

When new collector releases happen, the only updates that then happen in this repo would be changes to the `data/` files, and all languages would be updated automatically

### DIFF
--- a/.cspell.yml
+++ b/.cspell.yml
@@ -16,6 +16,7 @@ ignorePaths:
   - content/{ja,zh}
   - data/community/members.yaml
   - data/ecosystem/vendors.yaml
+  - data/collector
   - public/_redirects
   - static/refcache.json
 patterns:


theletterf:
@jaydeluca Thanks! I appreciate the concern. How would this make things simpler? Could you run me a bit through the process and what it would look like?

jaydeluca:
@theletterf I've updated the PR description with more context and describing what this solves for. Just let me know if it's still not clear and I can add more details. I intend on following up with adding [internal docs](https://opentelemetry.io/site/) after (i want to avoid making this PR any larger)

theletterf:
@jaydeluca Thanks for the explanation! This sounds very reasonable to me. 

@vitorvasc WDYT?

vitorvasc (approved):
@jaydeluca, thanks so much for taking care of that! 🙌🏼 

I like the idea of having the translations in the `i18n/*.yaml` files. Even though it's not the current pattern, I'm not sure if there's a specific reason behind that. @chalin might be the one who can provide us more context once he's back. 

For now, I'd say we can move forward! We can always revert or improve the approach later if we notice something that isn't quite right. 😄

PR #9443, opened 2026-03-18
Clarify the unification of event representation in the Span Events API

pellared:
Addresses https://github.com/open-telemetry/community/issues/3312#issuecomment-4080255128

> I don't understand the benefit of switching to the log API: I don't want information scattered across several systems/APIs, and I'm really happy to have a way to build rich events (spans regular data + span events attached to them with exact timestamps) that carry all the context I need.

This blog post provides more detailed context on the decision and the benefits of moving to the log-based approach.


pellared:
@open-telemetry/docs-maintainers, it will be nice if this little clarification could be merged soon :wink:

PR #9444, opened 2026-03-18
Fix links to Tracing API and Logs API in Deprecating Span Events API

pellared:
Current links are to some old versions of Tracing API and Logs API.

This is an important bugfix given https://github.com/open-telemetry/opentelemetry-specification/blob/v1.40.0/specification/logs/api.md says:

> The API is not intended to be called by application developers directly.

which is no longer the case (and would be against the content of the blog post!)

### DIFF
--- a/data/community/members.yaml
+++ b/data/community/members.yaml
@@ -915,7 +915,6 @@ maintainers:
     teams:
       - docs-ja-approvers
       - docs-ja-maintainers
-      - docs-triagers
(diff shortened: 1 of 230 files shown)


pellared:
@open-telemetry/docs-maintainers, it will be awesome if this is merged soon given the bad hyperlinks can bring a lot of confusion. I am sorry for inconvenience.

PR #9509, opened 2026-03-26
Remove using var metricProvider pattern in .NET doc

### LINKED ISSUE #4075: [.NET] remove the pattern of "using var meterProvider"
> It'd be good to get a follow up PR to remove the pattern of "using var meterProvider", and prefer an explicit dispose().. We have a lot of people copying this to helper functions, and accidently causing it to get disposed() when helper method exits!

https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/docs/trace/getting-started-console/Program.cs#L27-L29

Separate PR to not further delay this PR

_Originally posted by @cijothomas in https://github.com/open-telemetry/opentelemetry.io/pull/4016#discussion_r1504577727_

schielc:
This PR addresses issue #4075  by replacing the `using var` pattern for MeterProvider with explicit Dispose() calls including a comment. This was already partly addressed in #4310 (comment copied from there) but not adjusted everywhere.
 
Why was it removed? The `using var` pattern is being removed because it causes issues when developers copy-paste these patterns into helper methods without giving it much thought, leading to the providers being accidentally disposed when the helper method exits instead of when the entire application lifecycle completes.

There are still other cases for the traces provider but for now I wanted to keep it simple since #4075 did not mention them.

Fixes #4075.

### DIFF
--- a/data/registry/instrumentation-js-fastify-otel.yml
+++ b/data/registry/instrumentation-js-fastify-otel.yml
@@ -16,5 +16,5 @@ createdAt: 2025-01-27
 package:
   registry: npm
   name: '@fastify/otel'
-  version: 0.17.1
+  version: 0.18.0
 isFirstParty: true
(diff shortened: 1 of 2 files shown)


schielc:
Btw this is my first cncf pr and I got motivated to do something small from the otel maintainer talks at kubecon. You guys inspired me :+1:

martincostello (approved):

PR #9513, opened 2026-03-27
  Add Chinese translations for Collector components docs

wucm667:
## Summary

  - add Chinese translations for the Collector components landing page and child pages
  - add localized pages for receivers, processors, exporters, connectors, and extensions
  - preserve shared include usage and stable English anchors for translated headings

  ## Test plan

  - [x] Ran `npm run test-and-fix`


jaydeluca:
if you want to also translate the table headings and footnotes, create a `zh.yaml` file in the `i18n` directory, and translate [these strings](https://github.com/open-telemetry/opentelemetry.io/blob/main/i18n/en.yaml#L30-L46)

wucm667:
@jaydeluca Done! Added `i18n/zh.yaml` for the table translations.

i18n/zh-cn.yaml
```diff
@@ -0,0 +1,43 @@
+# Docsy overrides
```
jaydeluca, line 1: my apologies, it looks like this file actually needs to be named `zh-cn.yaml`

PR #9535, opened 2026-03-31
[Infra] Pin GitHub Actions by semver and npm dependencies by exact version

vitorvasc:

[^I-know-my-stuff]:

---

- Pin all action references to full semver tags (e.g., `actions/checkout@v6.0.2`) instead of major-only tags or commit SHAs
- Consolidate `actions/checkout` from mixed v5.0.1/v6 to v6.0.2
- Bump actions to latest patch/minor where applicable
- Pin dependencies to exact versions (remove `^` ranges) in:
  - `scripts/generate-community-data/package.json`
  - `scripts/registry-scanner/package.json`
  - `scripts/ai-content-detection/package.json`
- Add npm update schedules for `scripts/generate-community-data`, `scripts/registry-scanner`, and `scripts/ai-content-detection`


jaydeluca (approved):

.github/copilot-instructions.md
```diff
@@ -205,3 +205,22 @@ npm run fix:markdown     # Auto-fix markdown issues
...
+- **Never generate or guess SHAs** - always look up the correct SHA from the
+  action's repository
+- **Include version comments** alongside SHAs for readability (e.g., `# v4.1.0`)
+- **Format**: `uses: owner/action@<full-sha> # vX.Y.Z`
+
+Example:
+
+```yaml
+- uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5.0.1
+```
```
chalin: Docs like these should be under our site docs, with a cross-reference made from here (IMHO).

scripts/generate-community-data/package.json
```diff
@@ -1,8 +1,8 @@
 {
   "type": "module",
   "dependencies": {
-    "@octokit/rest": "^18.0.0",
-    "dotenv": "^10.0.0",
-    "js-yaml": "^4.1.0"
+    "@octokit/rest": "18.12.0",
+    "dotenv": "10.0.0",
+    "js-yaml": "4.1.1"
```
chalin, line 6: By pinning here, which tool will inform us that updates are needed?
(I check for the main package.json, but not in this scripts folder. Should this be done here too?)

chalin (changes requested):
See inline questions before merging.
Edit: also see https://github.com/open-telemetry/opentelemetry.io/pull/9535#discussion_r3018213904, which I forgot to add before submitting this review.

.github/workflows/ai-content-detection.yml
```diff
@@ -53,15 +53,15 @@ jobs:
           private-key: ${{ secrets.OTELBOT_PRIVATE_KEY }}
 
       - name: Checkout
-        uses: actions/checkout@v6
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
```
chalin: Why not pin to a specific version here, as is done for copilot below?

It is impossible to know at a glance if this hash actually matches / is for v6 (as stated in the comment).
Does any of the tooling enforce the correspondence between the hash and the comment?

chalin (commented):
See inline questions before merging.

.github/copilot-instructions.md
vitorvasc: I agree with that, but according to GitHub's docs on custom instructions, Copilot can't follow external links. The recommendation is to "copy the relevant content directly into your instruction file instead", so I'm assuming it also applies to cross-reference files. 😞 

https://docs.github.com/en/copilot/tutorials/use-custom-instructions#unsupported-instruction-types

.github/workflows/ai-content-detection.yml
vitorvasc: We went through the details on Slack, but just to close the loop here as well:

Dependabot should update both the hash and the comment together, but we don't have any tooling that explicitly enforces that they match. 

Since the full semver will make it much easier to validate at a glance, I'll update the PR to use that pattern!

scripts/generate-community-data/package.json
vitorvasc: The dependabot configs for this path were missing, I've just committed them: https://github.com/open-telemetry/opentelemetry.io/pull/9535/commits/9d1d59dd5877c5ab9c01073812b9f8a767b0516d

.github/copilot-instructions.md
vitorvasc: Just committed a different approach for that and moved this content to a separate instructions file: https://github.com/open-telemetry/opentelemetry.io/pull/9535/commits/3c9afb9407a0c2b6e14ad6dcca273e46496d21de

.github/workflows/ai-content-detection.yml
```diff
@@ -46,22 +46,22 @@ jobs:
       pull-requests: write
       issues: write
     steps:
-      - uses: actions/create-github-app-token@f8d387b68d61c58ab83c6c016672934102569859 # v3.0.0
+      - uses: actions/create-github-app-token@v3.0.0
```
thompson-tomo, line 49: This version was actually correct as per https://scorecard.dev/viewer/?uri=github.com/open-telemetry/opentelemetry.io which is published to the security dashboard https://github.com/open-telemetry/sig-security/blob/main/security-dashboard.md

As such Should we not be switching from dependabot to renovate so that we get the pinned sha and improve the security score?

.github/instructions/github-actions.instructions.md
```diff
@@ -0,0 +1,21 @@
...
+  releases page to ensure the tag is published
+- **Never generate or guess versions** — always look up the correct version from
+  the action's repository
+- **Format**: `uses: owner/action@vX.Y.Z`
+
+Example:
+
+```yaml
+- uses: actions/checkout@v6.0.2
+```
```
thompson-tomo, line 21: If we switched to renovate like the other otel repos we would have renovate automatically applying this logic

.github/workflows/ai-content-detection.yml
thompson-tomo: Actually renovate does enforce they match

chalin:
Thanks @thompson-tomo, it's worth a thought. @vitorvasc, shall we consider/propose renovate for this repo. I think that we can give it a try.

vitorvasc:
@chalin @thompson-tomo +1 on Renovate. I've just opened #9871 to track the migration proposal, so I'll close this PR in favor of that.

Issue #6230, opened 2025-02-06
Analytics for schema files


chalin:
_Originally posted by @lmolkova in [#3468](https://github.com/open-telemetry/opentelemetry.io/issues/3468#issuecomment-2629490496)_:

> let me bring it up on the semconv call tomorrow. 
> 
> /cc @open-telemetry/specs-semconv-maintainers 
> 
> A side question - @chalin do we have some stats on successful hits to schemas? E.g. I'd love to see a distribution of versions being requested,  user-agents, etc? 

I've started looking into this.

chalin:
@lmolkova @trask - FYI, [/schemas/v1.13.0](https://opentelemetry.io/schemas/v1.13.0) is a 404 with 4.5K hits in the past month, so it seems that a lot of folks think that there's a "v" as part of the schema file name, which there is not as you know.

Interestingly, https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/22852 contains a reference to 

```yaml
"schemaUrl": "https://opentelemetry.io/schemas/v1.13.0",
```

Maybe that reference should be corrected so that folks don't get the wrong impression? Or worse, is that referring to a tool that is generating the invalid path? If so, the tool should be fixed if it hasn't already!

/cc @open-telemetry/specs-semconv-approvers

lmolkova:
found it appearing in some old otel-dotnet versions https://github.com/open-telemetry/opentelemetry-dotnet/blob/303ab7a25a81d7804ac7414134366e61a9501114/src/OpenTelemetry.SemanticConventions/Resource/ResourceSemanticConventions.cs#L15, but it no longer there.

I think we can't fix it without time machine.

svrnm:
was this a one off request or is there a permanent need to track those?

chalin:
@lmolkova et all - Here is an example of raw data (from Netlify) for the past 30 days:

<img width="325" height="1094" alt="Image" src="https://github.com/user-attachments/assets/d1786ce2-0d2f-4412-8143-5fdc98208b4d" />

Is this helpful?

I'm working on a broad solution to make such data accessible via GA, but wanted to run this by you first. We'll be able to create a report for, say, the top X schemas accessed in the past Y days.

/cc @vitorvasc 

Btw, here's a histogram for the data above:

<img width="803" height="301" alt="Image" src="https://github.com/user-attachments/assets/baaf9ab8-739e-4521-96c8-a03aee207cde" />

chalin:
[...]

PR #9555, opened 2026-04-02
Add GA4 asset-fetch Edge Function (phase 1) and plan

chalin:
- Contributes to
  - #6230
  - #9559
- Adds a Netlify asset-fetch Edge Function as a means of capturing asset-fetch data under GA so that we can later share it.
- For details, see the plan file.
- This PR implements phase 1.
- Codex (GPT-5.4) was my copilot in the development of this solution. Reviews by Claude (Opus) and Gemini.
- Note that we won't be able to test this until it lands in production.
- I'll be adding to our site docs (if) after this lands to explain the design and implementation of analytics etc

/cc @vitorvasc

### DIFF
--- a/projects/2026/asset-fetch-analytics.plan.md
+++ b/projects/2026/asset-fetch-analytics.plan.md
@@ -498,6 +498,12 @@ Do not send:
 
 If coarse origin analysis is useful, send only a normalized `referrer_host`.
 
+Phase 1 intentionally does not forward the original request `User-Agent` header
+to GA4. The primary reporting goal is event counts and top-accessed asset paths,
+not browser or device attribution, and Netlify Observability is the better
+source for request-level traffic classification such as browsers, crawlers, and
+AI agents.
+
 ## Cardinality controls
 
 ### Known GA4 risk


chalin (commented):
Copilot review (obtained separately):

---

## Review of `schema-analytics.ts` (Phase 1)

The implementation is clean, well-structured, and aligns with the plan. A few items below, split by urgency.

### Must address

**GA4 Measurement Protocol silently accepts invalid payloads**

The `sendGa4Event` function checks `response.ok` (line 166) to detect failures. However, the GA4 `/mp/collect` endpoint **always returns 2xx**, even when payloads are malformed or contain invalid parameters. This means the error-handling branch will essentially never fire, and silent misconfiguration (wrong measurement ID, bad parameter names, etc.) won't surface in logs.

This matters especially for Phase 1 since post-deploy validation is the only testing path. Suggestion: add a comment documenting this quirk, and during initial validation, temporarily point at the [GA4 debug endpoint](https://developers.google.com/analytics/devguides/collection/protocol/ga4/validating-events) (`/debug/mp/collect`) which returns detailed error responses.

### Follow-up

**Misleading `content_type` for redirect responses**

Phase 1 intentionally tracks 3xx redirects (e.g., `/schemas/latest`). For those responses, `response.headers.get('content-type')` will typically be `null`, causing `buildPayload` (line 40) to fall back to `'application/yaml'`. That value gets sent to GA4 as the `content_type` parameter, which misrepresents what actually happened. Consider sending `'redirect'` or letting `normalizeContentType` return `'none'` (which it already does for `null`) by removing the `?? 'application/yaml'` fallback.

**`[[headers]]` content-type rule for schemas may be shadowed**

The existing `netlify.toml` header rule setting `content-type: application/yaml` for `/schemas/:version` (line 23–25) is now effectively redundant since the edge function's `ensureSchemaContentType` owns that header for all 2xx responses on the same route. This isn't broken, but a comment in `netlify.toml` noting the relationship would help future maintainers understand why both exist.

chalin:
Also reviewed with Gemini, and got a thumbs up for this phase 1 work.

chalin:
The new custom event (`asset_fetch`) is being triggered, as can be seen from the Realtime panel:

image

Issue #9560, opened 2026-04-03
Address top 404s for 26Q2


chalin:
Description TBC

## Tasks

- [ ] ...

PR #9563, opened 2026-04-04
Publish favicons at the root and drop Docsy icons

chalin:
- Contributes to #9560
- Applies a patch until https://github.com/google/docsy/issues/2595 is fixed. The patch is to:
  - Delete the remaining Docsy icons for more exotic icon formats (which OTel doesn't override).
  - Add a redirect rule for requests to `/favicons/**`

---

**Preview**: https://deploy-preview-9563--opentelemetry.netlify.app/

Icons and the site manifest links are being served (no 404s):

> image


theletterf (approved):
Thanks!

Issue #9521, opened 2026-03-30
Declarative Configuration Schema Documentation


jaydeluca:
Creating a tracking issue for the introduction of 2 new pages related to documenting Declarative Configuration. This will involve several components:

- A script to transform https://github.com/open-telemetry/opentelemetry-configuration/blob/main/opentelemetry_configuration.json into a compatible shape for consumption by the frontend
- A new page for the declarative configuration schema that lays out the types, including properties and code snippets
- A new page under the declarative configuration page for "language implementation status"

Creating this issue to track the work as I will submit it several phases to reduce the size of the PRs.

## Data sources

- The types page will use the sync'd registry in order to render the types
- The language implementation status will use https://github.com/open-telemetry/opentelemetry-configuration/blob/main/language-support-status.md

Related / WIP:

- https://github.com/open-telemetry/opentelemetry.io/pull/8723 (initial sketch PR, superseded by 9383)
- https://github.com/open-telemetry/opentelemetry.io/pull/9383 (this is the most current)

## Work to do

- [x] javascript for converting json schema to format used in accordion #9654 
- [x] create schema json during build time and available as static resource #9842 
- [ ] Create the Types page #9850
- [ ] Language implementation status page
- [ ] Code snippets for types page
- [ ] LLM Markdown support

chalin:
@jaydeluca - recording notes from our 1-1 yesterday so I/we can keep a trace. Context/scope: this issue and #9265. While understanding that the current work is incremental, we agreed to these "objectives":

- Explore alternative ways (including Hugo's page generation from data approach) to build-time rendering of the page(s) that #9265 currently is supporting dynamic/client-side rendering
- Think Markdown first: any new page should have a clean md rendering, ideally one from which the HTML version is produced, but that isn't a strict requirement.
- Where the page(s) should live. IIRC, we had discussed https://opentelemetry.io/docs/specs/otel/configuration/ as a possible home for these non-normative pages.

jaydeluca:
comments from @jack-berg:

-----

There's a few missing features from [schema-docs.md](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/schema-docs.md) which I still find useful, but not necessary to start. Let's take [OtlpHttpExporter](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/schema-docs.md#otlphttpexporter-) as an example:

- There's a language support status table that shows the support of each property in each language at a glance.
[...]

PR #9448, opened 2026-03-18
Add declarative configuration docs for Spring Boot starter

zeitlinger:
## Summary

- New `declarative-configuration.md` page for the Spring Boot starter covering getting started, mapping rules, differences from agent DC, env var overrides, and duration format
- New `programmatic-configuration.md` page (extracted from `sdk-configuration.md`) with a warning that `AutoConfigurationCustomizerProvider` doesn't work with DC
- DC flavor tabs on `out-of-the-box-instrumentation`, `sdk-configuration`, `additional-instrumentations`, and `other-spring-autoconfig` pages
- DC note on `annotations` page for enable/disable mapping
- Updated agent DC page: removed stale "Spring Boot starter limitations" section, added cross-link
- Mentioned DC as a configuration option in `_index.md`

## Follow-up

Automatic converter: https://github.com/open-telemetry/opentelemetry.io/pull/9456

## Test plan

- [ ] Hugo builds without errors (`npm run serve`)
- [ ] New page renders at `/docs/zero-code/java/spring-boot-starter/declarative-configuration/`
- [ ] Tabs render correctly on sibling pages
- [ ] Agent DC page no longer mentions Spring Boot limitation
- [ ] All internal links resolve

Deploy preview:

- https://deploy-preview-9448--opentelemetry.netlify.app/docs/zero-code/java/spring-boot-starter/
- https://deploy-preview-9448--opentelemetry.netlify.app/docs/zero-code/java/spring-boot-starter/declarative-configuration/

PR #9582, opened 2026-04-06
Add script for synchronizing Configuration schema data

jaydeluca:

[^I-know-my-stuff]:

---

This is the first part of #9521 

This script pulls the schema from the [configuration repo](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/opentelemetry_configuration.json), and then converts it into a simplified structure for us to use in a new page for displaying the information. 

The page itself will come in a followup PR, but you can see the latest [draft preview here](https://deploy-preview-9383--opentelemetry.netlify.app/docs/specs/declarative-configuration/) to get an idea of how this will be used

cc @jack-berg

### DIFF
--- a/data/opentelemetry/configurationTypes.json
+++ b/data/opentelemetry/configurationTypes.json
@@ -1,0 +1,2747 @@
+{
+  "types": [
+    {
+      "id": "aggregation",
+      "name": "Aggregation",
+      "isExperimental": false,
+      "hasNoProperties": false,
+      "properties": [
+        {
+          "name": "default",
+          "type": "DefaultAggregation",
+          "default": "If omitted, default behavior applies.",
+          "constraints": "",
+          "description": "Configures the stream to use the instrument kind to select an aggregation and advisory parameters to influence aggregation configuration parameters. See https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#default-aggregation for details. If omitted, ignore."
+        },
+        {
+          "name": "drop",
+          "type": "DropAggregation",
+          "default": "If omitted, default behavior applies.",
+          "constraints": "",
+          "description": "Configures the stream to ignore/drop all instrument measurements. See https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#drop-aggregation for details. If omitted, ignore."
+        },
+        {
+          "name": "explicit_bucket_histogram",
+          "type": "ExplicitBucketHistogramAggregation",
+          "default": "If omitted, default behavior applies.",
+          "constraints": "",
+          "description": "Configures the stream to collect data for the histogram metric point using a set of explicit boundary values for histogram bucketing. See https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#explicit-bucket-histogram-aggregation for details If omitted, ignore."
+        },
+        {
+          "name": "base2_exponential_bucket_histogram",
+          "type": "Base2ExponentialBucketHistogramAggregation",
+          "default": "If omitted, default behavior applies.",
+          "constraints": "",
+          "description": "Configures the stream to collect data for the exponential histogram metric point, which uses a base-2 exponential formula to determine bucket boundaries and an integer scale parameter to control resolution. See https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#base2-exponential-bucket-histogram-aggregation for details. If omitted, ignore."
+        },
+        {
+          "name": "last_value",
+          "type": "LastValueAggregation",
+          "default": "If omitted, default behavior applies.",
+          "constraints": "",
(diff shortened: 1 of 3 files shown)


jack-berg:
FYI, @open-telemetry/configuration-approvers. @jaydeluca is working on incorporating declarative config into the docs site. Basically, he took the idea of [generated schema docs](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/schema-docs.md) and was able to make the UX delightful with some simple javascript / interactivity:

- Filtering 
- Collapsible sections
- Improved display by tighter control over the table column widths

Please check out the following preview pages to see the vision:

- [Type documentation](https://deploy-preview-9383--opentelemetry.netlify.app/docs/specs/declarative-configuration/)
- [Language support status](https://deploy-preview-9383--opentelemetry.netlify.app/docs/specs/declarative-configuration/language-status/)

Personally, I think this is fantastic. Once its fully realized, I say we can drop doc generation in `opentelemetry-configuration` and instead link directly to this.

data/opentelemetry/configurationTypes.json
jack-berg, line 1: Given that this isn't dramatically different that the format or content of https://github.com/open-telemetry/opentelemetry-configuration/blob/main/opentelemetry_configuration.json, I wonder if we can just add a git submodule and generate directly from the JSON schema.

jaydeluca: @zeitlinger is actually already pulling over the configuration repo as a submodule in https://github.com/open-telemetry/opentelemetry.io/pull/9448/files#diff-fe7afb5c9c916e521401d3fcfb4277d5071798c3baf83baf11d6071742823584 so we could do that. We'll just need to come up with a way to keep it up to date, which shouldn't be hard

jaydeluca: and just to clarify, we'll still generate this `configurationTypes.json` representation of the data, we just won't [fetch it](https://github.com/open-telemetry/opentelemetry.io/pull/9582/files) any more

jack-berg: > We'll just need to come up with a way to keep it up to date, which shouldn't be hard

Should probably only update it when new releases are published, yeah? Else we risk confusing readers by advertising types / properties that have no chance of being implemented yet.

> and just to clarify, we'll still generate this configurationTypes.json representation of the data, we just won't [fetch it](https://github.com/open-telemetry/opentelemetry.io/pull/9582/files) any more

Does it need to be checked into git, or at that point is it just an ephemeral intermediate representation to assist in generation?

jaydeluca: > Should probably only update it when new releases are published, yeah?

yea, i just mean we'll need to create some automation to keep it up to date like is currently done with the [opentelemetry-specification](https://github.com/open-telemetry/opentelemetry.io/pull/9341) one. Or we can just rely on remembering to manually bump it.

> Does it need to be checked into git, or at that point is it just an ephemeral intermediate representation to assist in generation?

`configurationTypes.json` is needed as the "transformed" representation of the raw schema, but we don't need the raw schema in git. The benefit of having this other format is to simplify the logic within the hugo shortcode template. I had originally tried working off the raw schema alone and it added a lot of complexity so that's why I opted for massaging it into a simpler shape

jack-berg: > configurationTypes.json is needed as the "transformed" representation of the raw schema, but we don't need the raw schema in git. 

If `opentelemetry-configuration` is available as a submodule, then the raw schema is already available in git. But I actually don't see that: https://github.com/open-telemetry/opentelemetry.io/tree/main/content-modules

jaydeluca: the [pr](https://github.com/open-telemetry/opentelemetry.io/pull/9448) hasn't been merged yet

jack-berg: Ah got it. 

>> Does it need to be checked into git, or at that point is it just an ephemeral intermediate representation to assist in generation?

> configurationTypes.json is needed as the "transformed" representation of the raw schema, but we don't need the raw schema in git. 

Does `configurationTypes.json` need to be checked into git? With the raw schema available all the time as a submodule, I would think that whatever tooling is needed to generate the docs can depend on an intermediate step to generate `configurationTypes.json` from the raw schema. Then generate from `configurationTypes.json`. I.e. no need for it to be checked into git because its can be cheaply created on the fly as needed.

jaydeluca: ah I see what you are saying. I'm going to put this PR into draft and test that out

jaydeluca:
> Does configurationTypes.json need to be checked into git? With the raw schema available all the time as a submodule, I would think that whatever tooling is needed to generate the docs can depend on an intermediate step to generate configurationTypes.json from the raw schema. Then generate from configurationTypes.json. I.e. no need for it to be checked into git because its can be cheaply created on the fly as needed.

Thanks for this call out @jack-berg, I've updated my implementation to use this approach, so i'll be submitting new PRs starting today that don't require this piece

Issue #9449, opened 2026-03-18
Provide agent-friendly content & support


vitorvasc:
## Summary

Make opentelemetry.io content discoverable and consumable by AI agents and LLMs by adding the following capabilities:

- **`llms.txt`** at the site root - a machine-readable index listing docs and recent blog posts, following the [llms.txt proposal](https://llmstxt.org)
- **`text/markdown`** output format, where every page gets an `index.md` alternative with rendered content
- ...

Inspired by [Make Hugo Site Agent-Friendly](https://dacharycarey.com/2026/03/01/make-hugo-site-agent-friendly/)

### Tasks

- [x] #9447 
- [x] Prep:
  - [x] #9581
  - [x] #9586
- [ ] Add support for Markdown output
  - [x] #9597
  - [x] #9606
  - [x] Add support for "View Markdown" - #9607
- [ ] Add support for Markdown content negotiation
  - [x] #9603
  - [x] #9633
- [ ] Improve the Markdown output (to avoid unnecessary HTML elements) for the following pages:
  - [ ] Homepages
  - [ ] Blog section pages, since they currently contain a `<script>` element
  - [x] #9659
  - [ ] ...
- [ ] Add support for `llms.txt`
- [ ] Other
  - [ ] #3537
  - [ ] #9747

### Examples

- https://www.elastic.co/docs/get-started.md
- ...

### Related

- #9662

PR #9586, opened 2026-04-06
Consolidate & clean up output fmt config

chalin:
- Cleanup and prep for #9449
- Consolidates `outputs` config into the main `hugo.yaml` file so that it is easier for localization teams to ensure they support the right outputs by default
- That is, this removes `outputs` front matter from locale pages
- With more uniform `outputs`, this PR enables RSS for all locale homepages

```console
$ (cd public && git diff -bw --ignore-blank-lines -I 'link rel="alternate|dateModified' -- ':(exclude)*.xml') | grep ^diff | head
```

### DIFF
--- a/config/_default/hugo.yaml
+++ b/config/_default/hugo.yaml
@@ -120,7 +120,7 @@ cascade:
       [HTML, RSS]
     target:
       kind: section
-      path: '/blog{,2026,2025}'
+      path: '/blog'
 
 params:
   copyright:


config/_default/hugo.yaml
```diff
@@ -106,6 +106,22 @@ cascade:
...
+      kind: home
+      sites: { matrix: { languages: [en] } }
+
+  # Enable RSS for blog section
+  - outputs:
+      # This must match the equivalent of [*section-outputs, RSS]
+      [HTML, RSS]
+    target:
+      kind: section
+      path: '/blog{,2026,2025}'
```
vitorvasc: I assume next year we'll need to manually add 2027 to this line, right?

chalin: Oops. Thanks for catching that. I had meant to remove it. We don't need the year specifier. Fixed.

Issue #4427, opened 2024-05-06
[IA] Zero-code as top-level section


chalin:
As [discussed during the 2024-05-06 Comms meeting](https://docs.google.com/document/d/1wW0jLldwXN8Nptq2xmgETGbGn9eWP8fitvD5njM-xZY/edit#bookmark=id.psd7kvp6ejwz), we're moving ahead with the proposal discussed via #3809. This meta issue will be used to track the migration work (and possibly some new additions). We'll be starting with Java.

Languages whose "Automatic" section will be moved to the new Zero-code top-level section:

- [x] .NET
  - [x] #4532 
  - [x] #4570
- [x] JavaScript
  - [x] #4527
  - [x] #4571
- [x] Java - #4428
- [x] PHP - #4716
- [x] Python - #4717
- [x] (Go) -- does not exist as of today

/cc @svrnm @zeitlinger

svrnm:
this is completed

Issue #3783, opened 2024-01-15
[Documentation] Mov "automatic instrumentation" out of "Language APIs & SDKs"


svrnm:
@open-telemetry/docs-approvers As a follow up to #3761 I'd like to suggest a significant change to the "Automatic Instrumentation" pages as well:

Move all  "Automatic" pages that provide "true"[^1] zero-code instrumentation into a top level section that is separate from "Language APIs & SDKs"

# Why?

As of today we heavily confuse end-users by not consistently using the term "automatic", it often means "automated approaches to extracting portable telemetry data with zero source code modification" (see [OTEP-1](https://github.com/open-telemetry/oteps/blob/main/text/0001-telemetry-without-manual-instrumentation.md), but it is also often used for throwing a bundle of instrumentation libraries into your SDK initialization and seeing a lot of telemetry dropping out of your application. Both are fair use cases for the word _automatic_, but as @martinjt pointed out in #3228 this is "_is making people [...] a little confused about the "right" approach._", so getting rid of _manual_ (as in #3761) was step 1 towards that goal, and getting rid of _automatic_ is step 2.

The obvious solution would be renaming the "Automatic" category under "Language APIs & SDKs > `Language`", but spending some time thinking about it, this still does not solve all the problems:

* We still mix the use of Zero-Code instrumentation and code-based instrumentation throughout the API&SDK documentation, e.g. we do a getting started with zero-code but then the rest of the docs is code-based, or we have exporter pages that say "here is how you do it with code, but by the way if you come from the zero-code getting started, here's how you do it with environment variables". By separating the 2 topics we can untangle a lot of that.
* Having the Automatic/Zero-Code so close by in the content tree, gives people that are exploring what they can do with the APIs & SDKs the feeling that this is something they should read about, and that this is something they need to do, because "_If I'm using Manual Instrumentation I'm doing it wrong_". By separating the 2 topics we make it explicit that Code-based and Zero-Code is for different audiences, depending on their **needs**
* Also, technically the zero code instrumentations are not part of the APIs&SDKs. They are not covered by the spec, and for most languages they live in a dedicated repository <language>-instrumentation. They are tools built on top of the SDKs, similar to the K8s operator which is built "on top" of the collector. By separating the 2 topics we make this much clearer to end-users, that this is something "different".
[...]

Issue #3228, opened 2023-09-06
Terminology Suggestion: Change "Manual Instrumentation" to "Code based Instrumentation" in docs


martinjt:
This is more of a wider suggestion based on a lot of conversations I've been having recently about the ways people instrument their applications.

The general feeling I've been hearing is that "If I'm using Manual Instrumentation I'm doing it wrong". For context, this is from engineering conferences primarily focused on JS/TS and .NET, I understand that languages like Java may put more focus on AOP style techniques that may make Manual actually be more of an anti-pattern.

Automatic instrumentation, using an external application to generate/setup tracing/metrics/logs is not a bad thing, I'm not at all saying that, I'm also not saying that Manual is the one true way. I'm saying that the terminology is making people I speak to a little confused about the "right" approach.

I talk about 3 different avenues to instrumentation with Otel.

* **Auto/Agent based** - Useful when you don't have access to the code (i.e. an infrastructure or some SRE's who aren't coders too), or when it's cost/time prohibitive to change each application, and you're willing to have over-instrumentation (not an exhaustive list).
* **Instrumentation Libraries** - Coded into your application code, getting opinionated Spans/Attributes from the authors of the libraries you use like Database calls and making/receiving HTTP requests.
* **Manual** - Adding your own context to existing spans and creating new spans around interesting areas of your code that would be useful to understand in production when debugging.

In our docs, we have only "Auto" and "Manual", meaning that engineers feel that to get anything using "Manual" they have to write all the code for all the spans.

I'm not sure of the "right" answer here, I just know that it's an issue, and I'm hearing it more and more (likely a good sign as people are starting to talk about OTel more in my communities).

One thought I had would be use the term "Code based" instead of "Manual". Which I think signifies the idea that you can get what Auto does, but configure it in code rather than via an external application. The other was that we have a standard "Getting Started" page for each language that jumps off into Auto and Code-based detailing reasons to choose each with the specific language idioms in mind (i.e. Java is more of an agent based/AOP style user base so maybe it's more prevalent there to use Auto).

I'd love some thoughts, especially from people outside of the .NET community. If you are putting in opinions though, please let us know which language, and also role like SRE/Application Engineer, you're coming from.

pellared:
1. How you "bootstrap the SDK" - it can be either automatic or manual.
2. How you "instrument libraries" (apply instrumentation libraries) - it can be either automatic or manual
3. How you "instrument your code" - I think this is always manual.

[...]

PR #9593, opened 2026-04-07
distinguish manual and auto instrumentation

y1yang0:
This patch unifies terminology, replacing "language & sdk/API" and "zero-code instrumentation" with "manual instrumentation" and "auto instrumentation" respectively.


theletterf (changes requested):
Thanks!

Please split the PR into two separate contributions, one for the terminology and another for the CN localization. Focus this PR on the EN changes — changes in other localized docs must happen in separate PRs.

y1yang0:
> Thanks!
> 
> Please split the PR into two separate contributions, one for the terminology and another for the CN localization. Focus this PR on the EN changes — changes in other localized docs must happen in separate PRs.

I didn't push the CNlocalization part to this branch, so this patch only contains the changes for terminology distinction. Sorry for the confusion. I've updated the PR description.

katzchang:
Hi, thanks for working on this!
                                                                                                                        
Before the lang:ja SIG can approve the Japanese localization changes in this PR, we'd like to understand the broader　context of this terminology change.                                                                                   
   
As you may know, there was a deliberate effort in 2024 to move away from the term "automatic instrumentation" to  "zero-code instrumentation," precisely because "automatic" was considered ambiguous (see #3228, #3783, #4427).　That migration involved significant cross-language work to establish "zero-code" as the standard term in the docs.　This PR appears to reverse that decision by replacing "zero-code" back with "auto instrumentation."

Before we can approve the lang:ja changes, we need clarity on:
                                                                                                                        
  1. Is there a community-level discussion or decision (e.g., an OTEP, SIG meeting note, or issue) that has agreed to　revert the terminology?
  2. If so, could you link to it?

Without a clear consensus on this terminology change at the project level, we're not in a position to approve the　localization portion independently — the Japanese translation should follow whatever the agreed-upon English terminology is.

y1yang0:
@katzchang Thanks for sharing the links and context, I wasn't aware of the history behind this terminology change.

My goal was to add documentation for `go-compile-time-instrumentation`. While drafting it, I noticed that the distinction between `zero-code instrumentation` and `Language APIs & SDKs` seemed inconsistent and a bit unintuitive (from my perspective), so I updated those references first as a prerequisite for the upcoming docs patch.

I’ll bring this up in the community for broader discussion. I won't open a new PR until a clear consensus is reached.

PR #9671, opened 2026-04-17
chore(ux): re-enable search in homepage navbar

chalin:
- ... because that's where some users are looking for it.
- Reverting from #9081

/cc @theletterf


tiffany76 (approved):
RSLGTM.

PR #9727, opened 2026-04-22
[adjust-pages.pl] clarify patch field semantics, and other changes

chalin:
- Adopt inclusive `minVers` / exclusive `maxVers` semantics, comparing via `semverCmp` instead of a prefix match
- Default `maxVers` to `minVers` with its patch segment incremented (for example, `1.55.0` → `1.55.1`), preserving the original prefix-match behavior
- Make `file` optional, defaulting to the module's spec/docs tree (`^tmp/otel/specification/`, `^tmp/otlp/docs/`, `^tmp/semconv/docs/`)
- Warn and skip when a patch's `module` is unknown and no `file` is provided
- Rename the `context` value `frontmatter` to `front matter` and drop the now-unneeded `cSpell:ignore` entry
- Clarify apply/skip/obsolete log messages to reference the triggering bound (min vs. max)
- Rewrite content-module-patches.md field descriptions to match and convert the reference tables to bulleted lists
- No changes to the generated site pages other than the updated site doc page (and sitemaps and build timestamps)

### DIFF
--- a/content/en/site/build/content-module-patches.md
+++ b/content/en/site/build/content-module-patches.md
@@ -3,7 +3,6 @@ title: Content module patches
 description: >-
   Creating and managing temporary patches for content modules between releases.
 weight: 15
-cSpell:ignore: frontmatter
 ---
 
 Spec pages published on this site (OTel specification, OTLP, semantic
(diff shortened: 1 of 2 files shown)


theletterf (approved):
Thought it was pl for Polish, then remembered that Perl still exists and is healthy :_)

chalin:
> Thought it was pl for Polish, then remembered that Perl still exists and is healthy :_)

😆 

Yeah, Perl is great for this kind of application, but I am considering rewriting the script in JS/TS.

PR #9712, opened 2026-04-21
Docs: Add index page for compatibility section

Wineshuga:
- [x] I have read and followed the [Contributing](https://opentelemetry.io/docs/contributing/) docs, especially the "**First-time contributing?**" section.
- [x] This PR has content that I did not fully write myself.
  - [x] I used AI and I have read and followed the [Generative AI Contribution Policy](https://github.com/open-telemetry/community/blob/main/policies/genai.md).
- [x] I have the experience and knowledge necessary to understand, review, and validate all content in this PR.[^I-know-my-stuff]

[^I-know-my-stuff]:
    Yes, I can answer maintainer questions about the content of this PR without using AI.

---
### Description

Related to #9650 

This PR introduces a new Prometheus section under docs/compatibility/. It adds an `_index.md` file to serve as the overview for Prometheus-related interoperability topics.

It also reorganizes existing content by moving the Prometheus Clients documentation into this directory so that all Prometheus-related compatibility content is grouped in one place.

PR #9746, opened 2026-04-25
Clean up compatibility docs metadata and rename Prometheus clients page

chalin:
- Followup to #9712, etc
- Renames the Prometheus compatibility guide file to a leaner slug that avoids the prometheus stutter
- Normalizes front matter metadata in compatibility and migration pages.
- Removes obsolete local cSpell ignores and adds shared dictionary coverage.
- Removes unused `tools/content/compatibility` path alias.
- Removes no-op alias

/cc @tiffany76 @Wineshuga

### DIFF
--- a/content/en/docs/compatibility/prometheus/client-libraries.md
+++ b/content/en/docs/compatibility/prometheus/client-libraries.md
@@ -1,10 +1,7 @@
 ---
-title: Prometheus Client Libraries vs. OpenTelemetry
-linkTitle: Prometheus Clients
-aliases: [/docs/compatibility/prometheus/]
-weight: 5
-# prettier-ignore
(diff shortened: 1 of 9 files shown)


content/en/docs/compatibility/prometheus/client-libraries.md
```diff
@@ -1,10 +1,7 @@
 ---
-title: Prometheus Client Libraries vs. OpenTelemetry
-linkTitle: Prometheus Clients
-aliases: [/docs/compatibility/prometheus/]
```
chalin, line 4: FYI: this alias is a no-op because that page exists on our site already.

content/en/docs/compatibility/prometheus/client-libraries.md
```diff
@@ -1,10 +1,7 @@
...
-weight: 5
-# prettier-ignore
-cSpell:ignore: AggregationBase2ExponentialHistogram base2ExponentialBucketHistogram bedroomTemperatureCelsius buildAndStart buildWithCallback classicUpperBounds connectedDeviceCount CounterValue defaultAggregation deviceCommandDuration devicesConnected errcheck gaugeBuilder GaugeFunc GaugeValue GaugeWithCallback hvac hvacOnTime initLabelValues InstrumentKindHistogram InstrumentSelector InstrumentType labelValues livingRoomTemperatureCelsius LongUpDownCounter MustNewConstMetric NativeHistogramBucketFactor nativeOnly nolint OtelHistogramAsSummary otlpmetrichttp PrometheusHistogramNative PrometheusRegistry PrometheusSummary promhttp sdkmetric setDefaultAggregationSelector setExplicitBucketBoundariesAdvice thermostatSetpoint totalEnergyJoules upDownCounterBuilder
```
chalin, line 7: cSpell is smart about spell-checking compound var names. You don't need to add every variable here, only var name parts that it doesn't recognize.

tools/content/compatibility
chalin, line 1: We don't use the tools directory anymore, so don't any any links under here from now on. (I'll add a check for this soon.)

tiffany76 (approved):

PR #9638, opened 2026-04-13
Migrate code-excerpts tooling to npm package version

chalin:
- Fixes #1635
- **Replaces** the Dart-based code excerpt tooling with new Node [code-excerpter][] tool, rewritten in TypeScript, as a **drop-in** replacement
- Illustrates **drop-in** replacement by hooking up the tool into the Test PR job (details below)
- Adjusts `content/en/docs/languages/java/getting-started.md`, as an example of how usage will be simplified by using this new version of the tool
  - @jack-berg: I've not updated any other Java pages with `<?code-excerpt...?>` directives b/c I know you're working on them. If you _would_ like me to clean up those files, let me know.
- Drops the old `tools/`, which was in support of the Dart tooling
- npm scripts:
  - Adds `check:code-excerpts`
  - Adds `fix:code-excerpts`
  - Keeps a deprecated `code-excerpts` alias that prints a **red** deprecation notice
- Includes `check:code-excerpts` in `test:compound-tests` so that it gets run in the Test PR check
- Refreshes`content/en/site/build/npm-scripts.md` to document the new commands and the deprecated alias.
- Adjusts `content/en/docs/languages/java/getting-started.md` snippet markup so excerpt-related Prettier/Markdownlint comment markers stay clean.

/cc @vitorvasc 

[code-excerpter]: https://www.npmjs.com/package/code-excerpter

PR #9751, opened 2026-04-25
Extend filename check to flag obsolete paths (e.g., tools/)

chalin:
- Extends `check:filenames` to also flag obsolete top-level paths in addition to underscore-named files, so leftover artifacts like `tools/` (made obsolete by #9638) don't linger in working copies after pulling
- Updates `fix:filenames` to remove flagged obsolete paths alongside renaming files to kebab-case, giving contributors a single command to bring their checkout back into compliance
- Documents the new behavior in the PR-checks contributor guide, including an "Obsolete files and folders" list that links each entry to the PR/issue providing context, and adds a note warning that `fix:filenames` may delete files
- Refreshes the npm-scripts reference page so `check:filenames` / `fix:filenames` descriptions match the broader scope and link back to the PR-checks doc
- Tightens the `_filename-error` message to mention "files or directories" since the check is no longer filename-only


maryliag (approved):

chalin:
TY @maryliag!

PR #9831, opened 2026-05-05
[ja] add translation of content/en/docs/zero-code/obi/configure/export-data.md

ymotongpoo:
## Summary

Translates `content/en/docs/zero-code/obi/configure/export-data.md` into Japanese as `content/ja/docs/zero-code/obi/configure/export-data.md`.

## Checks

- [x] Followed the [localization guide](https://opentelemetry.io/docs/contributing/localization/)
- [x] `npm run check:i18n -- -n` reports no missing `default_lang_commit`
- [x] `npm run check:links` passes
- [x] `npm run fix:format:diff` (Prettier on changed files) ran cleanly


ymotongpoo:
@chalin `<!-- markdownlint-disable-next-line no-emphasis-as-heading -->` doesn't work when the target line is enclosed with `**`. However, in Japanese, we don't use colon at the end of the line (i.e. the lines in PR should be the correct format in Japanese). Could you help us pass the CI without changing the current Japanese expressions?

vitorvasc:
> @chalin `<!-- markdownlint-disable-next-line no-emphasis-as-heading -->` doesn't work when the target line is enclosed with `**`. However, in Japanese, we don't use colon at the end of the line (i.e. the lines in PR should be the correct format in Japanese). Could you help us pass the CI without changing the current Japanese expressions?

Hi @ymotongpoo! 

The `disable-next-line` directive doesn't work here because it applies to the next literal line, which is the blank line, rather than the bold text. This makes it a no-op.

There are two options that should actually work:

**Option A: File-level disable**

Add this right after the frontmatter:

```markdown
<!-- markdownlint-disable no-emphasis-as-heading -->
```

Since this is a file-level instruction, you won't need to add a disable comment before every individual line. A single instruction at the top of the file should do the trick.

**Option B: Locale-wide config (long-term)**

Create a new file at `content/ja/.markdownlint.yaml`:

```yaml
extends: ../../.markdownlint.yaml
no-emphasis-as-heading: false
```

This mirrors how `content/en/.markdownlint.yaml` handles English-specific rules, and should cover future Japanese localizations that run into the same issue.

aabmass:
👋 👋

ymotongpoo:
@vitorvasc the file wide notation worked! Thanks for the advice!

katzchang (approved):
全体的によさそうです。

#9881 で変更がありますが、追従は続くPRで対応する方針でいいと思います。

vitorvasc (approved):

PR #9900, opened 2026-05-13
[CI/Infra] Migrate from Dependabot to Renovate

### LINKED ISSUE #9871: [Infra] Migrate from Dependabot to Renovate for dependency updates
### Desired feature or idea

Replace the current Dependabot setup ([`.github/dependabot.yml`](https://github.com/open-telemetry/opentelemetry.io/blob/main/.github/dependabot.yml)) with [Renovate](https://docs.renovatebot.com), bringing this repository in line with other OpenTelemetry repositores that have already adopted it.

### Additional context

- https://github.com/open-telemetry/opentelemetry.io/pull/9535#issuecomment-4360504645)
- Adopting Renovate was also discussed previously in the approvers Slack channel.
- #9535 is being closed in favor of this issue, Renovate should handle the manual SHA pinning automatically.

vitorvasc:
Context:

- Fixes #9871
- Supersedes #9535.

Replaces Dependabot with Renovate for npm and GitHub Actions updates.

- Adds `.github/renovate.json5`
- Weekly schedule (`before 9am on monday`) for parity with the previous Dependabot cadence
- `minimumReleaseAge: 7 days` so just-released versions are not picked up until they have had a week to settle
- `internalChecksFilter: strict` so PRs only open once all internal checks (including the release-age gate) pass
- Groups non-major `devDependencies` updates into a single weekly PR; runtime deps stay individual for safer review
- Keeps `hugo-extended` ignored, since it is intentionally pinned and updated manually via `npm run update:hugo`
- Applies the existing `dependencies` label to all Renovate PRs
- Removes `.github/dependabot.yml`

/cc @chalin


thompson-tomo:
Fyi @vitorvasc you will need to raise a pr in the community repo after merging this PR. See https://github.com/open-telemetry/community/issues/3156 as an example.

.github/renovate.json5
```diff
@@ -0,0 +1,24 @@
+{
+  $schema: 'https://docs.renovatebot.com/renovate-schema.json',
+  extends: [
+    'config:recommended',
```
thompson-tomo:
```suggestion
    'config:best-practices',
```

vitorvasc:
@thompson-tomo, good to know, thank you!

.github/renovate.json5
```diff
@@ -0,0 +1,24 @@
...
+  ],
+  minimumReleaseAge: '7 days',
+  internalChecksFilter: 'strict',
+  labels: ['dependencies'],
+  schedule: ['before 9am on monday'],
+  packageRules: [
+    {
+      // hugo-extended is intentionally pinned; updated manually via `npm run update:hugo`.
+      matchPackageNames: ['hugo-extended'],
+      enabled: false,
```
thompson-tomo, line 16: As a thought what about the below, that way you can see the dependency is there and if there is an update but it is not raised unless you tell it to.

```suggestion
      matchDepTypes: ['devDependencies'], 
      dependencyDashboardApproval: true,
      groupName: "hugo",
```

I have setup very similar in the ruby repo's where runtime updates are gated away until we are ready.

Note the second match is to stop the next rule overriding it.

.github/renovate.json5
```diff
@@ -0,0 +1,24 @@
...
+    'docker:pinDigests',
+    'helpers:pinGitHubActionDigestsToSemver',
+  ],
+  minimumReleaseAge: '7 days',
+  internalChecksFilter: 'strict',
+  labels: ['dependencies'],
+  schedule: ['before 9am on monday'],
+  packageRules: [
+    {
+      // hugo-extended is intentionally pinned; updated manually via `npm run update:hugo`.
```
chalin, line 14: 👍 (that is, let's continue to exclude Hugo)

.github/renovate.json5
chalin: I prefer excluding any checks for `hugo-extended`

chalin (approved):
(RS)LGTM, esp. keeping `hugo-extended` out of the loop.

Issue #6193, opened 2025-02-04
[CI] Move away from the use of Gulp


chalin:
Roll our own scripts indepentent of Gulp, even for (file) stream processing. 

Here are the tasks to roll over:

```console
$ npx gulp --tasks
Tasks for ~/tmp/opentelemetry.io/gulpfile.js
├── default            Display usage instructions
├── lint-md            Check markdownlint rules. For details, use --info.
├── prune              Prune --num <n> entries from static/refcache.json file. For details, use --info.
└── validate-registry
```

WDYT @svrnm et all?

svrnm:
Yeah. 👍 when I wrote my task I only plugged it into gulp because it was there already. I would not mind replacing it

PR #9901, opened 2026-05-13
chore(deps): drop through2 in favor of node:stream.Transform

vitorvasc:
## Description

Dependabot opened #9879 and #9935 to bump `through2` from 4.0.2 to 5.0.x. This PR replaces `through2` with Node's built-in `node:stream.Transform`. 

Supersedes and closes #9935. 

- Replaces `through2.obj(...)` calls in `gulp-src/validate-registry.js` with a small local `objTransform(fn)` helper that returns `new Transform({ objectMode: true, transform: fn })`
- Removes `through2` from `devDependencies` in `package.json`

## Verification

- `npx gulp validate-registry` runs clean: `Processed N files, 0 had issues.`
- `npx gulp validate-registry --debug` still logs `Processing file: ...` for every entry, exercising the `logFiles` transform.
- `git grep through2` returns no source references after the change.

### DIFF
--- a/gulp-src/validate-registry.js
+++ b/gulp-src/validate-registry.js
@@ -1,6 +1,6 @@
 const gulp = require('gulp');
+const { Transform } = require('node:stream');
 const { taskArgs } = require('./_util');
-const through2 = require('through2');
 const yaml = require('js-yaml');
 const Ajv = require('ajv');
 const addFormats = require('ajv-formats');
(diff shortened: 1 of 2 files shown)


chalin (approved):
(RS)LGTM. We should really be getting rid of Gulp tasks (#6193) but it's ok to do this cleanup now.

Issue #10144, opened 2026-06-03
Auto-update version adjustments are needed


chalin:
### Tasks

- [x] #10117 
- [x] #10145
- [x] Ensure that at most `en` pages are updated - #10147

### Followup

- #1016

chalin:
Done

PR #10147, opened 2026-06-03
ci(version-in-file.sh): don't update non-en files

chalin:
- Contributes to #10144
- @vitorvasc do you agree? The (compensating) proposal is to disable link checking for drifted pages.


vitorvasc (approved):
Yes, let's go that way! 🙂

PR #10152, opened 2026-06-03
ci: Add reusable workflow failure reporting, and more

### LINKED ISSUE #5762: Auto-update of versions: failures should be reported.
This issue offers some analysis of the failures that lead to a fix being proposed in the following PR:

- #5758

After some investigation, here is what I've found:

(1) On `Nov 13, 2024, 1:17 AM EST`, the `Auto-update versions #15240` job failed: https://github.com/open-telemetry/opentelemetry.io/actions/runs/11811741891/job/32905820377.

It failed, as @zeitlinger reports in #5758, because two runs of the subscript are attempting to create a PR for a single repo. The first subscript is successful, but the second fails because "a PR already exists" for the given repo version update.

I can't recall if maintainers get any notice about such failures? If not, maybe the script(s) should be altered so create an issue? Or we should have a process to monitor the GH actions and report any failures? Maybe @zeitlinger's proposal to have only one run per repo is the simplest solution.

(2) The next scheduled run at `Nov 13, 2024, 2:16 AM EST` of the script `Auto-update versions #15241` passed because it saw that a PR had already been created and assumed that all was wel.

I think that we have an opportunity to improve things here. If there are changes pending, and a PR exists but has been merged, a new PR should be created. This would allow us to catch partial updates.

---

There is a separate problem that #5758 has surfaced, because it's build is currently failing, see https://github.com/open-telemetry/opentelemetry.io/actions/runs/12278556636/job/34260597540?pr=5758. For details, see:

- #5761

chalin:
- Fixes #5762
- Addresses silent failures in scheduled auto-update and integration-branch workflows by opening or updating a tracking GitHub issue when a run fails
- Adds `reusable-report-failure.yml` as a reusable workflow callers invoke with `if: failure()`; usage details live in the workflow file header
- Adds `scripts/gh/report-failure/` (CLI, testable orchestration, unit tests) to create or comment on deduplicated `CI/infra` issues
- Wires a `report-failure` job into the auto-update and spec/semconv integration-branch workflows
- Renames `scripts/docsy-alerts-to-md/test.js` to `convert.test.mjs` so the Perl converter sanity tests run under `test:local-tools`
- Simplifies `test:local-tools` to `node --test "scripts/**/*.test.mjs"`
- Documents the failure-report workflow in `content/en/site/build/ci-workflows.md` and updates `npm-scripts.md`
- Codeveloped with Copilot and Cursor

### DIFF
--- a/.github/workflows/auto-update-community-members.yml
+++ b/.github/workflows/auto-update-community-members.yml
@@ -10,7 +10,7 @@ permissions:
   contents: read
 
 jobs:
-  auto-update-versions:
+  auto-update:
     name: Auto-update community members page
     runs-on: ubuntu-latest
     # Remove the if statement below when testing against a fork
@@ -64,3 +64,11 @@ jobs:
           title: Update community members
           body: |
             This pull request contains automated updates to files by the GitHub Action.
+
+  report-failure:
+    needs: auto-update
+    if: failure() && github.repository == 'open-telemetry/opentelemetry.io'
+    uses: ./.github/workflows/reusable-report-failure.yml
+    permissions:
+      issues: write
+      contents: read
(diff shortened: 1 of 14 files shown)


tiffany76 (approved):
RSLGTM.

Issue #10165, opened 2026-06-04
Ensure that drifted pages are not link-checked


chalin:
### Reasoning

Drift pages often contain outdated links to GitHub and other resources. This causes headaches for our refcache maintenance, and just doesn't scale with the increasing number of locales. Besides, we post a banner at the top of drifted pages warning that links might not work.

### Tasks

- [x] Prep: #10167
- [x] #10230
- [x] Take 2: #10260

### Context

Followup to:

- #10144

Related:

- https://github.com/open-telemetry/opentelemetry.io/issues?q=is%3Aissue%20state%3Aopen%20drift

chalin:
Done for now.

chalin:
Reopening so that we can solve this differently. Committing a modified config results in merge conflicts too often.

chalin:
Fixed by:

- https://github.com/open-telemetry/opentelemetry.io/pull/10260

PR #10230, opened 2026-06-05
ci(feat): config + script to auto-ignore link checking for drifted localized pages

chalin:
- Contributes to #10165

- **Skip link checks on drifted localized pages**
  - Pages with `drifted_from_default: true` are excluded because their links are often stale relative to English and produce expected, non-actionable failures.
  - Uses htmltest `IgnoreDirs` (not `data-proofer-ignore`) so drifted pages are still parsed as anchor targets; inbound `#fragment` links from in-sync pages keep being validated.

- **Auto-generate htmltest config from front matter**
  - Extends `scripts/htmltest-config.pl` to derive end-anchored `IgnoreDirs` regexes from drift status; the trailing `$` isolates a drifted section page without suppressing its non-drifted children.
  - Skips `_includes/` and other fragment files, which render to no output page.
  - Regenerates `.htmltest.yml` under a labeled, auto-generated block via `fix:htmltest-config`.

- **Remove manual workarounds**
  - Drops the `htmltest:` override on `zh/docs/contributing/_index.md`.
  - Drops a per-link `?link-check=no` workaround on the drifted `zh` supported-libraries page.

- **CI guard for stale config**
  - Adds a `check-links` step that regenerates the htmltest config and fails on diff, so `.htmltest.yml` stays in sync when drift status changes.

- **Docs**
  - Notes that `drifted_from_default` drives both the outdated banner and link-check skipping in the Drift status section.
  - Clarifies that link-fix guidance applies to in-sync pages, since drifted pages are now skipped by the link checker.

- Co-developed with Copilot/Opus.


tiffany76 (approved):
RSLGTM.

PR #10371, opened 2026-06-12
fix(ci): restore missing:sig-approval for localization PRs

vitorvasc:
## Context

PR #10295 removed the `## Localization Teams` block from `.github/component-owners.yml` when it moved locale ownership to CODEOWNERS. The SIG-approval logic in `pr-approval-labels.sh` reads `component-owners.yml` to decide which teams must approve, so locale PRs were no longer labeled `missing:sig-approval`. 

This issue was flagged via the Slack #otel-docs-localization channel: https://cloud-native.slack.com/archives/C076RUAGP37/p1781230209175099

## Changes

- Adds `get_locale_teams_for_files` to `pr-approval-labels.sh`, read locale approver teams from `CODEOWNERS` so locale PRs are labeled with `missing:sig-approval` as they did before #10295


.github/scripts/pr-approval-labels.sh
```diff
@@ -193,7 +194,71 @@ get_sig_teams_for_files() {
...
 # ---------------------------------------------------------------------------
-# Check if the PR has potentially a publish date in the frontmatter of 
+# Parse CODEOWNERS for locale-approval teams owning the given PR files.
+# Since #10295, CODEOWNERS is the single source of truth for locale ownership.
+#
+#
+# We read only directory rules under content/ (pattern starts with /content/,
+# ends with /, no glob characters). That reproduces the prefix matching the
+# locale block in component-owners.yml used before #10295, without
+# reimplementing CODEOWNERS glob and last-match-wins resolution here.
```
chalin: This introduces a hidden coupling: the labeler now silently depends on the locale rules in CODEOWNERS staying plain `/content/<loc>/` directory rules. If those rules are ever reworked into glob form (e.g. `/content/ja/**`), this parser silently stops matching and the regression this PR fixes returns unnoticed.

Suggestion (can be a follow-up): add a short comment in the localization section of `.github/CODEOWNERS` (and in its embedded generator snippet) noting that `pr-approval-labels.sh` parses the `/content/<loc>/` rules and that they must remain plain directory rules.

.github/scripts/pr-approval-labels.sh
```diff
@@ -193,7 +194,71 @@ get_sig_teams_for_files() {
 }
 
 # ---------------------------------------------------------------------------
-# Check if the PR has potentially a publish date in the frontmatter of 
+# Parse CODEOWNERS for locale-approval teams owning the given PR files.
+# Since #10295, CODEOWNERS is the single source of truth for locale ownership.
+#
+#
```
chalin: Nit: doubled empty comment line.

.github/scripts/pr-approval-labels.sh
```diff
@@ -193,7 +194,71 @@ get_sig_teams_for_files() {
...
+    line="${line//$'\r'/}"      # tolerate CRLF
+    [[ -z "${line//[[:space:]]/}" ]] && continue
+
+    local pattern owners
+    pattern="${line%%[[:space:]]*}"   # first token is the path pattern
+    owners="${line#"${pattern}"}"     # remainder is the owners list
+
+    # Only directory rules under content/, with no glob characters.
+    [[ "${pattern}" == /content/*/ ]] || continue
+    [[ "${pattern}" == *"*"* ]] && continue
```
chalin, line 266: Known gap, fine to leave as-is since it preserves exact pre-#10295 parity: PRs touching only `.cspell/<loc>-*.txt` or `prh/<loc>.yml` require locale-team review per branch protection, but are skipped here (glob / non-directory patterns), so they won't get `missing:sig-approval`. If we ever want to close that gap, bash `[[ "$file" == ${pattern#/} ]]` glob matching would handle `ja-*.txt` natively.

chalin (approved):
Thanks for tracking this down — the diagnosis is correct, and the fix is sound.

**Verified:** #10295 removed the `content/<loc>:` entries from `component-owners.yml`, and `get_sig_teams_for_files` is the only input to `missing:sig-approval`, so locale PRs were labeled `ready-to-be-merged` on docs approval alone. Branch protection still gated merges (CODEOWNERS makes the locale team the sole required reviewer), so nothing unsafe merged — but the label produced false positives in the ready-to-merge backlog.

**Tested locally** against the real CODEOWNERS file:
- Correct teams returned for `ja`/`zh`/`pt`/`fr` content files; multi-locale PRs return the union.
- `content/en/...`, near-miss `content/ja-something/`, `.cspell/ja-*.txt`, and `prh/ja.yml` correctly excluded.
- CRLF, tabs, and inline comments handled; the empty-union path in `main()` is safe under `set -euo pipefail`.
- Behavior is exact parity with pre-#10295 (the removed block covered only `content/<loc>`).

Design-wise, parsing CODEOWNERS is the right call vs. restoring the duplicate block in `component-owners.yml` — it keeps the single source of truth that #10295 established.

A few non-blocking inline comments; the cross-reference suggestion in CODEOWNERS is the one I'd most like to see (here or as a follow-up).

chalin (approved):
Re-reviewed the update — all three previous comments are addressed, and the revision goes beyond what was asked:

1. **Hidden coupling** ✅ — cross-reference comments added in both the CODEOWNERS locale section and its embedded generator snippet, *plus* a runtime `::warning::` when a `/content/<loc>/` locale rule degenerates into a glob. That's a better guard than a comment alone.
2. **Doubled comment line** ✅ — fixed.
3. **`.cspell`/`prh` gap** ✅ — now explicitly documented in the function header as intentional pre-#10295 parity.

Also verified the new refinements locally against the PR-head script and CODEOWNERS:
- Single-segment restriction correctly skips nested dirs (`content/en/site/...` no longer considered) while `ja`/`zh`/`pt` still resolve.
- The glob-degeneration warning fires for `/content/ja/**` + locale team, and stays quiet for `/content/en/site/**` and `/content/**/.markdownlint.yaml` (docs-maintainers) — no spurious warnings against the real CODEOWNERS.
- `dedupe_words` union handles mixed newline/space input and the all-empty case safely under `set -euo pipefail`.
- Missing-CODEOWNERS case now warns instead of silently returning.
- ShellCheck CI passes.

LGTM, thanks @vitorvasc!

PR #10440, opened 2026-06-16
[i18n][ko] Setup Korean language support

vitorvasc:
- Part of #9577 
- Adds support for Korean `(ko)` localization
  - Update Hugo configs to include `ko` language definition
  - Adds the `ko-words` cSpell dictionary
  - Maps the `lang:ko` label to `content/ko/**` and adds the `ko` entry to `data/locale-teams.yaml`
  - Regenerates the locale section of `.github/CODEOWNERS` from the registry with `npm run fix:codeowners`
  - Adds `content/ko/_index.md` as an English placeholder homepage, with `default_lang_commit` pinned to the current English source

PR #6905, opened 2025-05-16
[bn] Infra setup and config for bn

chalin:
- Contributes to #6765
- This PR contains the infra prep files from #6775, which seems to be stuck

/cc @debjotyms @sajibAdhi

PR #10481, opened 2026-06-19
feat: add setup-new-localization skill

vitorvasc:
- Adds `.claude/skills/setup-new-localization/SKILL.md` that wires a new locale end-to-end:
  - Hugo language block, content mounts, cSpell custom list with an upstream-dict lookup against the `cspell-dicts` repo, the `lang:<lang>` labeler config and GitHub label, `data/locale-teams.yaml` with CODEOWNERS regeneration, and the `projects/localization.md` entries
  - Takes a kickoff issue or a bare ISO 639-1 code as input
  - Documents the two maintainer/org-admin actions it can't do itself: creating the `lang:<lang>` label with the country's flag color, and the `docs-<lang>-*` org teams in `open-telemetry/admin` (example: admin#716)
  - Flags the pre-#10295 gotcha so contributors don't re-add locale owners to `component-owners.yml`

/cc @chalin 

---

[^I-know-my-stuff]:

> Co-authored with Claude Opus 4.8.

### DIFF
--- a/.claude/skills/setup-new-localization/SKILL.md
+++ b/.claude/skills/setup-new-localization/SKILL.md
@@ -6,7 +6,7 @@ description: >-
   onboarding a localization team.
 argument-hint: '<kickoff-issue | lang-code>'
 allowed-tools: Bash Read Edit Write Grep Glob
-cSpell:ignore: endonym gitkeep nowrap unstaffed
+cSpell:ignore: endonym unstaffed
 ---
 
 # Set up a new localization
@@ -38,9 +38,10 @@ Either way, derive these; don't ask for them as separate arguments:
 
 > [!NOTE]
 >
-> All list entries below stay in **alphabetical order by `<lang>`** (English
-> `en` stays first as the default). Match the placement of an existing neighbor
-> (e.g. insert `ko` between `ja` and `pl`).
+> Most lists below are ordered alphabetically by `<lang>` (English `en` stays
+> first). Some aren't — `projects/localization.md`'s "Current language teams"
+> section is ordered by English name. At each insertion point, match the
+> ordering of the existing entries rather than assuming one global rule.
 
 ## Steps
 
@@ -97,7 +98,7 @@ Create it empty — git won't track an empty folder, so the `.gitkeep` is the
 tracked file:
 
 ```bash
-touch content/<lang>/.gitkeep
+mkdir -p content/<lang> && touch content/<lang>/.gitkeep
 ```
 
 **Do NOT copy the English homepage into the setup PR.** The translated homepage
@@ -126,12 +127,9 @@ published package names are inconsistent (`dict-es-es` with a hyphen vs
 `dict-pl_pl` with an underscore), but the [dictionary folders][cspell-dicts]
 follow one rule: `<lang>` (e.g. `bn`) or `<lang>_<REGION>` (e.g. `es_ES`,
 `pl_PL`, `pt_BR`, `uk_UA`). A matching folder means a dict exists; no folder (no
-`ko`, `ja`, `zh`) means there isn't one.
-
-```bash
-gh api repos/streetsidesoftware/cspell-dicts/contents/dictionaries \
-  --jq '.[].name' | grep -i '^<lang>'
-```
+`ko`, `ja`, `zh`) means there isn't one. Read the [dictionary
+folders][cspell-dicts] directly — a folder-name match is the source of truth,
+not a guessed npm name.
 
 The **dictionary id** is the folder name lowercased with `_` → `-` (`pl_PL` →
 `pl-pl`). For the exact **package name** and version, read them from npm
@@ -156,20 +154,22 @@ leave the `package.json` devDependencies unchanged.
 touch .cspell/<lang>-words.txt
 ```
 
-### f. `package.json` — Prettier nowrap group (ask a maintainer)
+### f. Prettier prose-wrapping — default: leave it alone
 
-Some locales break under Prettier's default prose wrapping; the
-`_check:format:nowrap` script runs those with `--prose-wrap preserve` instead.
-It currently lists `content/ja content/ko content/uk content/zh` (CJK plus
-Ukrainian), so there's no clean a-priori rule. **Confirm with a maintainer**
-whether `<lang>` belongs in this group. If it does, add `content/<lang>`
-(alphabetical) to the script:
+**Default: no Prettier changes for a new locale.** Per `localization.md`,
+prose-wrap exceptions exist only for languages Prettier mishandles; a locale
+opts in later, when its team hits the problem — not at setup time.
 
-```json
-"_check:format:nowrap": "npm run __check:format:nowrap -- content/ja content/<lang> content/uk content/zh"
-```
+If the team does opt in, it's **two coupled edits**, and one without the other
+is a silent no-op:
 
-Then run `npm run check:format` to confirm the locale's content passes.
+- a `/content/<lang>` line in `.prettierignore` (exempts the dir from the
+  default `proseWrap: always` pass), **and**
+- `content/<lang>` in the `_check:format:nowrap` script in `package.json`
+  (re-checks it with `--prose-wrap preserve`).
+
+Read `.prettierignore` and that script for the current members rather than
+trusting a snapshot here.
 
 ### g. `lang:<lang>` label — labeler config **and** the GitHub label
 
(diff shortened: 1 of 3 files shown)


.claude/skills/setup-new-localization/SKILL.md
```diff
@@ -0,0 +1,252 @@
...
+
+# Set up a new localization
+
+Wire a new language into the site end-to-end. Start from one of:
+
+- **A kickoff issue (preferred)** — its number or URL, e.g.
+  [#9577][kickoff-example]. Read it with
+  `gh issue view <n> -R open-telemetry/opentelemetry.io` and pull out:
+  - ISO 639-1 code → `<lang>`
+  - Language name → `<native-label>`
```
chalin: We want to know both:

- A `<lang>` (required)
- An optional `<locale>`

E.g. `pl` and `pl-PL`, `zh` and `zh-CN`, both of which appear in the Hugo config.

chalin: Ah, I see that you mention locale below. But the issue might specify it so it should be mentioned here too.

chalin (commented):
Great! A few initial drive-by comments:

- Just double checking: labels aren't configured under Terraform, right? (If that is the case, then the guidance in the skill is ok).
- See inline comment.

chalin:
Btw, probably as a separate skill, when reviewing the new locale's homepage, we need to check for structural parity with the `en` homepage, and equivalence in the translation.

.claude/skills/setup-new-localization/SKILL.md
vitorvasc: Updated the guidance to include `<locale>` as an optional value, in case the issue already provides it.

Also added a couple steps to match the last-minute changes applied for the `ko` locale, such as creating the `.gitkeep` placeholder and confirming whether the locale should be added to the Prettier nowrap group. 

On the Terraform question: confirmed, labels aren't managed there.

chalin (changes requested):
> [!NOTE]
> This review was iteratively co-developed by @chalin and Copilot (Fable).

**Method**: each step was cross-checked against the real locale-setup diffs (ko #10440, pl #9389, bn #6905), and every embedded command was executed against the current repo and this PR's head. All CI checks are green. The recent head revisions (dictionary-folder rule, inline label colors, locale-decision procedure) already fixed several earlier concerns — nice.

This skill captures genuinely hard-won knowledge (details at the end), but a few items need fixing before merge, plus one structural theme worth adopting while you're at it.

<details>
<summary><b>Required changes</b></summary>

**1. Add the skill to the skills catalog.** Every skill to date is listed in `content/en/site/skills/_index.md` (entry + link definition); this PR should add its own entry in the same PR, as prior skill PRs did.

**2. Step (c): the `touch` fails.** `content/<lang>/` doesn't exist yet, and `touch` doesn't create parent directories:

```bash
mkdir -p content/<lang> && touch content/<lang>/.gitkeep
```

**3. Step (d): the lookup command contradicts its own (correct) prose.** The folder rule is exactly right — but the pinned command undermines it: `grep -i '^ko'` matches `kotlin`, and `^ja` matches `java` (verified against the live cspell-dicts listing), while the adjacent text says "no folder (no `ko`, `ja`, `zh`) means there isn't one". An agent that trusts the command output over the prose wires up a bogus dictionary. Either anchor the match (`grep -iE '^<lang>(_|$)'`) or — better, per the structural note below — drop the pinned command and let the folder rule + listing be the source of truth.

**4. Step (f): incomplete mechanism, stale snapshot, inverted default.** Three intertwined problems:

- The nowrap regime is **two coupled edits**: a `.prettierignore` entry (which exempts the dir from the default `proseWrap: always` pass) *plus* the `_check:format:nowrap` script entry (which re-checks it with `--prose-wrap preserve`). A script-only entry is a silent no-op — which is exactly what happened to `content/ko` in #10440, the latent bug this snapshot inherited. (Repo-side fix: #10770 drops the inert entry.)
- "It currently lists `content/ja content/ko content/uk content/zh`" goes stale the moment #10770 merges. Point at the sources of truth (`package.json` + `.prettierignore`) instead of quoting today's state.
- The default for a new locale is **no Prettier special-casing at all** — per `localization.md`, ignore rules are only for when the language "isn't well supported by Prettier". (Korean is well supported: Prettier wraps Hangul at existing spaces only; the Korean-specific bugs prettier/prettier#5028 and prettier/prettier#6516 were fixed long ago, and the open prettier/prettier#14936 covers Han/Kana only.) Locales opt in later when their team wants it, as uk did in #7445 — it's not a setup-time maintainer question. Also: `npm run check:format` at setup time runs over a dir containing only `.gitkeep`, so it verifies nothing about the locale.

Suggested shape for (f): *Default: no Prettier changes for a new locale. If the locale team later finds Prettier mishandles their language (criterion in `localization.md`), add the coupled pair — a `/content/<lang>` line in `.prettierignore` **and** `content/<lang>` in the `_check:format:nowrap` script; one without the other is a no-op.*

</details>

<details>
<summary><b>Structural suggestion: spend the detail budget on the non-derivable</b></summary>

The pattern behind items 3 and 4: the skill sometimes pins exact commands, values, or snapshots of current state for things an executing agent can derive at execution time — and each pinned detail is something that can be subtly wrong or go stale silently. A litmus test that would both trim and harden the skill: **could a competent agent derive this at execution time from the repo?**

- **Derivable** → state the *intent* and the *source of truth*, not the command. (The dictionary folder rule is the right shape; the grep command bolted onto it is the wrong one.)
- **Non-derivable institutional knowledge** → spell it out fully. This is where the skill already shines: `.gitkeep`-not-homepage, the post-#10295 `component-owners.yml` trap, the admin-repo team structure, unstaffed-locale fallback semantics.

Related: consider a short preamble naming the authoritative artifacts (live `lang:*` labels, `hugo.yaml`, `locale-teams.yaml`, the ko setup diff #10440) so future edits stay anchored to sources rather than snapshots.

</details>

<details>
<summary><b>Smaller suggestions</b></summary>

- **Step (a)**: practice note — recent setups set `locale:` explicitly even for arguably single-variant languages (`ko-KR`, `pl-PL`, `ro-RO`, `uk-UA`). The "omit" branch is defensible but diverges from what maintainers actually did; consider "follow the kickoff issue, else recent practice".
- **Step (d)**: a second decision the step doesn't mention — `.cspell.yml` `ignorePaths` already exempts `content/{ja,zh}` from spell-checking entirely. A locale whose script cSpell can't handle belongs there, rather than (or in addition to) getting a words file. (Hangul spell-checks fine, so ko rightly isn't listed.)
- **Step (j)**: the top-of-file "alphabetical by `<lang>`" note doesn't hold for the "Current language teams" section, which is ordered by **English name** (Chinese second) — say "match the neighboring entries" per insertion point. This list also drifts easily (#10770 fixes a ja/ko ordering slip and a missing pl), so a self-check — "verify the list against `hugo.yaml` `languages`" — would pay for itself.
- **Slack channel** (step (j), item 4): for a brand-new locale the channel may not exist yet — say where the URL comes from (kickoff issue, or created by maintainers) or mark it optional-initially.
- **Validation**: add `npm run build` (catches mount/config typos nothing else exercises), and consider anchoring the checklist to the canonical worked example: "your diff should match the shape of #10440".
- **Front matter**: `cSpell:ignore` can be trimmed to `endonym unstaffed` — `gitkeep` and `nowrap` are already covered (verified by running cSpell on this head with the line stripped).

</details>

<details>
<summary><b>Beyond this PR (no action needed here)</b></summary>

- `localization.md` still carries the pre-#10295 maintainer checklist that this skill correctly contradicts (e.g., `component-owners.yml`); we should reconcile the docs in a follow-up so skill and docs don't fight.
- Found while testing: `check:spelling`'s `projects/**/*.md` glob is unquoted in `package.json`, so top-level `projects/*.md` (including `localization.md`) are never spell-checked in CI. Separate fix candidate.

</details>

### What's especially good

- The **expected-paths list matches the real ko setup (#10440) exactly** — I diffed them.
- The `.gitkeep` trick and do-not-bundle-the-homepage guidance encode the trickiest discoveries from the #10440 review, precisely.
- The post-#10295 `component-owners.yml` gotcha and never-hand-edit-CODEOWNERS rule are exactly the traps a future agent would hit without this skill.
- The dictionary-folder rule (folders as source of truth; dict-id vs package-name distinction) is the right shape for this class of problem.
- Label flag-colors check out against the live labels (ko=0047A0, pl=DC143C, pt=009739, uk=ffdd00), as does the admin-repo example (open-telemetry/admin#716).
- Rightly omits link-check shard wiring — not needed at setup time.

chalin (approved):
Thanks for the updates. Main suggestions adopted, let's land this and iterate.

PR #10485, opened 2026-06-19
chore(specs): drop obsolete semconv patch

chalin:
- Disables now obsolete semconv patch. Kept it as an example in comments
- Generated site is the same modulo the build timestamp

### DIFF
--- a/scripts/content-modules/adjust-pages.pl
+++ b/scripts/content-modules/adjust-pages.pl
@@ -59,25 +59,25 @@ my %versFromSubmod = %versions; # Actual version of submodules. Updated by getVe
 # =================================================================================
 
 my @patches = (
-  {
-    # RocketMQ docs renumbered the consumer group page from 07 to 08:
-    # https://rocketmq.apache.org/docs/domainModel/08consumergroup/
-    #
-    # Upstreamed fix:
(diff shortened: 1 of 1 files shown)


scripts/content-modules/adjust-pages.pl
```diff
@@ -59,25 +59,25 @@
...
-    minVers => '1.41.0',
-    maxVers => '1.42.0',
-    file    => qr|^tmp/semconv/docs/messaging/rocketmq\.md$|,
-    apply   => sub {
-      s{
-        (https://rocketmq\.apache\.org/docs/domainModel/)07(consumergroup)
-      }{${1}08$2/}gx;
-    },
-  },
+  # {
```
maryliag, line 62: can't we just remove this? and if we need again we can still look at history here and get the code back?

chalin: Sure. Just a matter of preference: I like to keep an example or two in the file. It is easier than search history when comes time to create a new entry.

maryliag: ah okay, we can leave it then

maryliag (approved):

PR #10606, opened 2026-06-29
feat: Update the ruby instrumentation metadata

### LINKED ISSUE #9920: Enrich otel registry entries using package registry
Currently we have a sync process which creates registry entries for otel packages. In addition we also have a script which keeps the version of these packages up to date.

Looking at the otel ruby entries they are light on details ie no docs url.

It would be great if when we detected a new version we fetched additional metadata for that version and added/updated the entry accordingly.

For instance https://rubygems.org/api/v1/gems/opentelemetry-api.json would give us doc url, description and can ensure licences are accurate.

Reason this would be useful is that I want to redirect the ruby gh-pages to the registry and for that it would be useful to have direct links to the docs like the current gh pages.

thompson-tomo:

[^I-know-my-stuff]:

---

This introduces metadata enrichment of the ruby gems published in the regisry. All metadata is sourced from rubygems.org which is the same source as the versioning. In the process a registry scan has been performed for ruby to bring entried up to date.

A handful of entries have been deprecated to match the deprecation notices on the repos.

Closes: https://github.com/open-telemetry/opentelemetry.io/issues/9920

See approval in linked issue.


thompson-tomo:
Cc @open-telemetry/ruby-maintainers

svrnm (approved):
oh wow, that looks really geat and it is indeed only a small change 👍 

Asking @open-telemetry/ruby-approvers to take a look, also @open-telemetry/ecosystem-explorer-approvers FYI!

thompson-tomo:
Thanks, it didn't take much at all. :)

For reference you can compare the old entries (https://opentelemetry.io/ecosystem/registry/?language=ruby) to the new ones (https://deploy-preview-10606--opentelemetry.netlify.app/ecosystem/registry/?language=ruby). The new have the additional links in particular the docs links which are updated with each version and is what the gh pages currently provide.

jaydeluca (approved):
nice, lgtm! just needs passing CI which im guessing are unrelated

thompson-tomo:
@jaydeluca i have fixed most of the ci checks. The only one is the link checker which i don't know how to resolve.

thompson-tomo:
Thanks @chalin it looks like it is having an issue with the url's i was worried about.

chalin (changes requested):
> Thanks @chalin it looks like it is having an issue with the url's i was worried about.

@thompson-tomo - there remain six 404s because

- Core Ruby GitHub Pages docs lag behind gem releases
- Contrib GitHub Pages site doesn’t exist at all

The source of the problem is the Ruby Gem metadata, ie docs links, are invalid.
@jaydeluca @svrnm et all: how do we want to resolve this?
We shouldn't be using invalid links in our registry entries IMHO.

thompson-tomo:
I agree the contrib is wrong but for the core packages the 404 page is strangely ok as the 404 by design returns a list of links. My understanding is that was the design back then.

The options i see are:
- Manually Remove urls for contrib packages & allow 404 on core
- Manually Remove urls for all packages and await for next release to add them back
- Manually set urls for all packages as per other packages
- Await releases of the gems & trigger an update of versions.

It does raise the question if we are scraping info from public registries which the developer has published to, what verification of that info should be done? What happens if the link is valid but 2 years later it breaks what happens then? Maintainers need to fix it? Maintainers inform submitter that it is broken & it should to fixed? Note urls is a required property so it might not be possible to just remove it.

jaydeluca:
if this is something that is going to happen every time a new package is released, then we probably need a better way to handle it so that the automation doesn't try to add anything that doesn't exist in the gem registry yet

thompson-tomo:
The packages exist in the upstream (gem) registry, the issue is that the metadata from the upstream registry which the developer published to was in 2 cases wrong. This wrong data is in the form of broken links to additional docs. For instance if a user visits the upstream registry ie https://rubygems.org/gems/opentelemetry-instrumentation-gruf and click documentation they get the broken link. In the other cases it is a poor legacy design which misuses the 404 page to provide an index. Both scenarios which will hopefully be fixed in the next publish of the offending packages to the upstream registry.

My question was more general as it applies regardless of how the data was populated into otel registry. Note the sync process here can help mitigate risk of the issue as if the upstream registry is updated it flows to otel without manual effort.

kaylareopelle:
@thompson-tomo When we release new versions of gruf and grpc that include https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/2424, will this issue be resolved?

kaylareopelle (approved):
Overall, this looks great! Thank you for your work @thompson-tomo. 

I don't want the Ruby approval to be a blocker here, so I'll approve for now, leaning on @chalin's requested changes to shore up the CI issue.

thompson-tomo:
@kaylareopelle correct once the 6 gems (2 contrib & 4 core) are released the ci will be resolved by the sync without workarounds.

chalin:
> if this is something that is going to happen every time a new package is released, then we probably need a better way to handle it so that the automation doesn't try to add anything that doesn't exist in the gem registry yet

As @jaydeluca points out, before we move forward with the improvement in entry updates, we need an automated solution, not a manual one, that addresses the issue of invalid metadata. Will give this some more thought and report here. Suggestions are welcome.

thompson-tomo:
@chalin links are now fixed as @kaylareopelle has performed a release of the packages.

In relation to the checking of the metadata, if i used curl to check the url before setting it would that be sufficient?

.github/scripts/update-registry-versions.sh
```diff
@@ -73,6 +73,36 @@ for yaml_file in ${FILES}; do
...
+                    ${UPDATE_YAML} ".urls.repo = \"$url\"" "$yaml_file"
+                fi
+                url="$(jq -r '.homepage_uri' <<< "$gem_json")"
+                if [ -n "$url" ] && [ "$url" != "null" ]; then
+                    ${UPDATE_YAML} ".urls.website = \"$url\"" "$yaml_file"
+                fi
+                ;;
+            *)
+                echo "Registry not supported for metadata update."
+                ;;
```
chalin, line 119: This will likely make the output very noisy. If so, drop it.

.github/scripts/update-registry-versions.sh
```diff
@@ -73,6 +73,36 @@ for yaml_file in ${FILES}; do
...
+                fi
+                url="$(jq -r '.source_code_uri' <<< "$gem_json")"
+                if [ -n "$url" ] && [ "$url" != "null" ]; then
+                    ${UPDATE_YAML} ".urls.repo = \"$url\"" "$yaml_file"
+                fi
+                url="$(jq -r '.homepage_uri' <<< "$gem_json")"
+                if [ -n "$url" ] && [ "$url" != "null" ]; then
+                    ${UPDATE_YAML} ".urls.website = \"$url\"" "$yaml_file"
+                fi
+                ;;
```
chalin: My suggestion is to factor this out into an `update_gems_metadata()` helper. The code can then fetch all URLs to ensure they are valid, and only update the entry if they are all valid.

.github/scripts/update-registry-versions.sh
```diff
@@ -73,6 +73,36 @@ for yaml_file in ${FILES}; do
...
     }
 
+    update_metadata() {
+        package_name=$1
+        registry=$2
+        yaml_file=$3
+
+        case $registry in
+            gems)
+                gem_json="$(curl -s "https://rubygems.org/api/v1/gems/${package_name}.json")"
```
chalin: Validate the result of the curl call here.

chalin (commented):
> In relation to the checking of the metadata, if i used curl to check the url before setting it would that be sufficient?

Sure. See inline comments for details.

.github/scripts/update-registry-versions.sh
thompson-tomo: Thanks but We actually don't want that but rather to just ignore urls which don't produce 2xx status codes. See updated script.

thompson-tomo:
Thanks @chalin i have made the changes similar to your suggestion but instead am only skip setting the invalid urls rather than skipping the update of all the metadata as the rest is still useful.

chalin:
Hi @thompson-tomo — while we wrap up the review, could you resolve the merge conflicts with `main`?

Most of the conflict is likely `.lycheecache`: it's fine to take `main`'s version wholesale (it'll get regenerated by CI), then re-merge. The registry YAMLs, if any conflict, should be straightforward.

Thanks!

thompson-tomo:
No worries @chalin conflicts resolved

chalin:
> No worries @chalin conflicts resolved

Thx. Link cache needs updating, and to be committed.

chalin (commented):
Thanks for the updates: most of my earlier asks are now addressed. One remaining ask, plus optional hardening notes.

**Remaining ask (blocking):**

- Validate the gem-metadata fetch before using it. `gem_json="$(curl -s …)"` silently yields an empty/error body on failure, which then feeds the yq writes. Suggestion: `gem_json="$(curl -sf "https://rubygems.org/api/v1/gems/${package_name}.json")" || return`

**Optional (fine as a follow-up issue; won't block on these):**

- Quote yq string writes via env to avoid expression breakage on special chars: `VALUE="$description" yq -i '.description = strenv(VALUE)' "$file"`
- `.licenses | join(", ")` errors when `licenses` is null; guard with `// []`.

If you don't address these now, create an issue to track these changes, and link to it here.

thompson-tomo:
All addressed @chalin

chalin (approved):
LGTM

PR #10623, opened 2026-06-30
docs: clarify Ruby in_span exception handling

### LINKED ISSUE #2022: Explain `in_span` more for ruby docs
It will set the status and record an exception for you, so this is worth mentioning and explaining in the status section as well.

Jaegeonyoon:
Clarifies how Ruby's `in_span` handles exception recording and span status when an exception escapes the block.

This PR adds a short note explaining that `in_span` records the exception on the span by default, sets the span status to `Error`, and re-raises the exception. It also clarifies how to disable automatic exception recording and adds a short example for exceptions rescued inside the block.

Fixes #2022.

[^I-know-my-stuff]:


content/en/docs/languages/ruby/instrumentation.md
```diff
@@ -117,6 +117,14 @@ def do_work
 end
 ```
 
+If an exception escapes the `in_span` block, the tracer records the exception on
+the span by default, sets the span status to `Error`, and re-raises the
+exception. You can disable automatic exception recording by passing
+`record_exception: false`.
```
kaylareopelle: `record_exception` also happens to be the name of a configuration option for our Trilogy instrumentation. When I first read this, I thought that's what the change referred to. 

What do you think about this adjustment?
```suggestion
exception. You can disable automatic exception recording by passing
`record_exception: false` as an argument to the `in_span` method.
```

content/en/docs/languages/ruby/instrumentation.md
```diff
@@ -117,6 +117,14 @@ def do_work
...
+
+If you rescue an exception inside the block and don't re-raise it, set the span
+status and record the exception manually when appropriate.
```
kaylareopelle: Do you think users might be helped by an example code snippet for this recommendation?

kaylareopelle (commented):
Hi @Jaegeonyoon! Thanks for opening this PR! 😄

content/en/docs/languages/ruby/instrumentation.md
Jaegeonyoon: Thank you, it is a good point. I updated it to say that this is an `in_span` argument.

content/en/docs/languages/ruby/instrumentation.md
Jaegeonyoon: Thanks again. Yes, added a short example for the rescue-exception case.

kaylareopelle (approved):
Thanks for the updates! 🎉

vitorvasc (approved):
Thanks!

PR #10741, opened 2026-07-09
Consolidate spec/semconv integration workflows into specs-integration.yml

chalin:
- Follow-up to
  - #10738 and
  - #10739
- Consolidates the two spec integration workflows into a single [specs-integration.yml](https://github.com/chalin/opentelemetry.io/blob/chalin-m24-spec-workflow-matrix-2026-0709/.github/workflows/specs-integration.yml) that runs one matrix job per upstream repo (`otel`, `semconv`), eliminating the near-duplicate YAML.
- **Workflow**:
  - Keeps step bodies unchanged, except for the semconv-only Hugo-mounts refresh, now a shell conditional keyed on `$SPEC`. It stays inside the Update-submodule step so that the upstream-controlled `$commit_desc` remains step-local rather than being exported through `$GITHUB_ENV` (relevant to the shell-hardening follow-up).
  - Makes run blocks template-expansion free: matrix values arrive via job-level `env:`, keeping zizmor's template-injection check happy, and the `${{ env.* }}` occurrences inherited from the old files become plain shell vars.
  - Sets `fail-fast: false`; the `report-failure` job waits on both matrix jobs, firing if either fails.
- **Scripts**:
  - Rephrases the pick-branch tracking-issue lead to name the matrix job within the consolidated workflow.
  - Cleans up `version-in-file.sh` submodule switching: drops the two dead spec arms (automation stopped calling it for the spec repos in #10738, and a manual invocation would recreate the duplicate release PR that this workflow owns), then derives the switch from whether `.gitmodules` is among the repo's file targets — the `all-versions.sh` argument list becomes the single source of truth.
- **Docs and links**:
  - Rewrites the EN `ci-workflows.md` section (matrix-job table) and the EN `sig-practices.md` trigger step for the single workflow.
  - JA/RO pages: retargets link-definition URLs only, leaving labels and prose to the locale-sync process, and marks the pages `# patched` — without this, the old workflow URLs would 404 post-merge and break link checking.
  - Replaces the old-workflow refcache entries with `206`-seeded entries for the new workflow URLs, which resolve only post-merge.
- **Preview(s)**:
  - [site/build/ci-workflows/](https://deploy-preview-10741--opentelemetry.netlify.app/site/build/ci-workflows/)
  - [docs/contributing/sig-practices/](https://deploy-preview-10741--opentelemetry.netlify.app/docs/contributing/sig-practices/)

---

Validation:

- First successful run: https://github.com/open-telemetry/opentelemetry.io/actions/runs/29054675063


vitorvasc (approved):
🤩

didiViking (approved):

PR #10765, opened 2026-07-10
chore: Compare refcache JSON canonically

zeitlinger:
## What changed

- Normalize `static/refcache.json` and each shard refcache with `jq --sort-keys` before comparing them in the refcache CI check.
- Add local-tool tests that cover order-only differences and real value differences.

## Why

The current check uses a raw file diff, so a PR merge commit can fail when Git orders JSON object keys differently from the shard-generated refcache even though the cached URL data is equivalent.

## Validation

- `node --test scripts/check-refcache.test.mjs`
- `bash -n .github/scripts/check-refcache.sh`

### DIFF
--- a/.github/scripts/check-refcache.sh
+++ b/.github/scripts/check-refcache.sh
@@ -32,6 +32,17 @@ else
   echo "Compare shard refcache files against '$REFERENCE_FILE':"
 fi
 
+if ! command -v jq >/dev/null 2>&1; then
+  echo "ERROR: jq is required to compare refcache JSON files"
+  exit 1
+fi
+
+TMP_DIR=$(mktemp -d)
+trap 'rm -rf "$TMP_DIR"' EXIT
+
+REFERENCE_NORMALIZED="$TMP_DIR/reference.json"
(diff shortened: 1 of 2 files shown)


svrnm:
@chalin @vitorvasc PTAL

vitorvasc (approved):
LGTM! Leaving the final approval and merge to @chalin.

chalin (changes requested):
- Thx @zeitlinger -- is this a proactive change, or can you share a link to a PR job where there was a problem?
- Btw, the refcache is going a way soon 🤷🏼‍♂️ (the format, not the cache) ... but that doesn't mean that this can't land in the meantime.
- Setting the review to "request changes" so I can to a full and context-aware review before this lands.
- Thanks for the ping @vitorvasc

zeitlinger:
it was a real PR - but if you'll change it then this can be closed

chalin:
Thanks @zeitlinger - I'd really like to see the false-positive PR job failure if you can track it down because we still might be using shards with the new format. If you can't easily find it, that's ok too.

PR #10770, opened 2026-07-10
Drop inert ko Prettier-nowrap entry; fix supported-languages list

chalin:
- Drops `content/ko` from the `_check:format:nowrap` script — the entry (added with the ko locale setup, #10440) was inert: the nowrap regime requires a matching `.prettierignore` entry to exempt the directory from the default `proseWrap: always` pass, and `content/ko` never had one. Korean pages are therefore already prose-wrapped, and the extra nowrap pass over `content/ko` checked nothing.
  - Leaving ko under the default `proseWrap: always` is safe: Prettier wraps Hangul at existing spaces only (verified empirically against Prettier 3.9.4, with round-trip check); the Korean-specific wrap bugs were fixed long ago (prettier/prettier#5028, prettier/prettier#6516), and the open CJK no-wrap issue (prettier/prettier#14936) scopes Han and Kana only.
  - If the ko localization team prefers unwrapped prose, the coupled pair (`.prettierignore` entry + nowrap-script entry) can be added, as uk did in #7445.
- Fixes the supported-languages list in `projects/localization.md`:
  - Restores lang-code ordering (moves ko after ja).
  - Adds Polish (pl) — live since March (#9389) but never listed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "__find:md:not-old-blog": "sh -c 'find \"$@\" -name \"*.md\" -not -path \"content/en/blog/2019/*\" -not -path \"content/en/blog/202[0-4]/*\" -print0' --",
     "_build": "npm run _hugo -- --logLevel info -e dev --buildDrafts --buildFuture --baseURL \"${DEPLOY_PRIME_URL:-http://localhost}\"",
     "_check:format:any": "npm run __check:format -- --check --ignore-path ''",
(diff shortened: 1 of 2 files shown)


chalin:
@vitorvasc - btw, I realize that we might be in this situation because I _thought_ that Prettier might have issues with Korean (and so made comments to this effect on the original PR), but it doesn't. This PR fixes that.

vitorvasc (approved):

PR #10844, opened 2026-07-15
Update community submodule to include self-observability engineering value (#3544)

cijothomas:
The `content-modules/community` submodule was pinned to `e9411ee` (2025-05-02), so the [community mission page](https://opentelemetry.io/community/mission/#engineering-values) does not reflect [open-telemetry/community#3544](https://github.com/open-telemetry/community/pull/3544), which added self-observability as a core engineering value.

This bumps the submodule (and the `community-pin` in `.gitmodules`) to the current community `main` tip `96374af`, which includes that change.

Only two community files are mounted into the site (`mission-vision-values.md` and `roadmap.md`). Since the previous pin, the only change to either is #3544 itself, so the net rendered change is exactly the new self-observability engineering value:

```diff
-_resiliency_, and _performance_.
+_resiliency_, _performance_, and _self-observability_.
```

plus a new `### We value _self-observability_` section. `roadmap.md` is unchanged.


vitorvasc (approved):
Thanks!

Issue #9219, opened 2026-02-19
Website Localizations: Steps forward to make teams more independent of docs-maintainers


svrnm:
This is a follow up to yesterdays SIG Comms call, where we discussed how we can help the language teams to be more independent, i.e. work towards them being able to work on their issues and PRs without docs-maintainers support.

As an important upfront note: this is aiming at mature localizations, with a healthy community and a healthy record of continuously keeping their documents updated, our goal is to make it possible for those groups to have maintainers that can self-organize, so this issue is aimed towards current approvers and maintainers of those groups.

The challenge that we are facing right now is that we can not provide a simple way of having merge permissions isolated to a localization folder, as github only provides an "all or nothing" permission. This means we need to work around this by one of the following solutions:

* Separate repository
* Merge Bot
* Do nothing and wait for GitHub to fix this
* Other solutions?

From @open-telemetry/docs-maintainers perspective the bot solution would be the cleanest, but we are currently not able to address this ourselves, so we would need help from localization approvers&maintainers to take a look into this and come up with a solution.

chalin:
Related:

- #7331

Andygol:
Related:
- https://github.com/open-telemetry/opentelemetry.io/discussions/7921

didiViking:
I'm not yet an approver for the entire docs repo, I'm only approver for the Romanian localization. I would say that I'm inclined for the simplest solution, that doesn't involve a lot of overhead for the maintainers. If the merge-bot is the cleanest, I vote for this! 
I'm assuming that the merge bot will check if all the localization requirements are met and proceeds to merge the PR. 
@svrnm Maybe adding a short description of what the merge bot will do, will help clarify the situation for the other approvers/maintainers.

chalin:
I'm currently exploring solutions. I'll track progress in this comment. 

As we move towards granting locale teams more autonomy, we'll count on locale maintainers to enforce site conventions with a particular attention given to [Localization guidelines](https://opentelemetry.io/docs/contributing/localization/). We'll be setting up as many PR checks as we can to catch obvious violations.

Note: task list moved to opening comment.

PR #10930, opened 2026-07-21
l10n: restrict multi-locale PRs to keeping build and checks green

chalin:
- Contributes to #9219
- Tightens the locale-span policy: a PR that changes localized page **content** may span locales only when required to keep the site build and its checks green; any other content change — including additions to drifted pages — is a semantic change that belongs to the locale's team. Content-neutral maintenance is exempt.
- Gives the `# patched` marker its own section, "Patching localized pages": mechanical build and check fixes only, whether or not the fix spans locales; minimal-edit discipline and marker lifecycle.
- Restructures the cross-locale link-fix how-to as a case list by target fate: moved targets are patched, deleted targets have their drift status derived with `fix:i18n:status`, and aliases are clarified as serving site visitors only (page moves — fragments can't be redirected).
- Aligns the review-pull-request skill, the localization instructions, and the `ensure-default-lang-commit-patched` script header with the new policy.
- **Preview(s)**:
  - https://deploy-preview-10930--opentelemetry.netlify.app/docs/contributing/localization/#prs-should-not-span-locales
  - https://deploy-preview-10930--opentelemetry.netlify.app/docs/contributing/localization/#patched

### DIFF
--- a/.claude/skills/review-pull-request/SKILL.md
+++ b/.claude/skills/review-pull-request/SKILL.md
@@ -82,10 +82,9 @@ validates and the local fix command. Caveats:
 
 - Submodules: non-maintainer PRs should not touch them; a maintainer fixes
   before merge — [`sig-practices.md#general`][general].
-- Locale span: semantic changes are per-locale; editorial cross-locale edits are
-  OK and append `# patched` to `default_lang_commit` —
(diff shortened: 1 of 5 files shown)


svrnm (approved):
LGTM!

PR #10946, opened 2026-07-22
Zoom Migration Announcement

svrnm:
- Add an announcement that we are migrating our meetings to LFX Project Control Center
- **Preview**: https://deploy-preview-10946--opentelemetry.netlify.app/

cc @open-telemetry/governance-committee 

- [x] I have read and followed the [Contributing](https://opentelemetry.io/docs/contributing/) docs, especially the "**First-time contributing?**" section.
- [x] This PR has content that I did not fully write myself.
  - [x] I used AI and I have read and followed the [Generative AI Contribution Policy](https://github.com/open-telemetry/community/blob/main/policies/genai.md).
- [x] I have the experience and knowledge necessary to understand, review, and validate all content in this PR.[^I-know-my-stuff]

[^I-know-my-stuff]:
    Yes, I can answer maintainer questions about the content of this PR, without using AI.

PR #10911, opened 2026-07-20
Switch link checking from htmltest to Lychee

chalin:
- Fixes #10912.
- **Switches link checking from htmltest to Lychee** (mirrors the recent [google/docsy#2665](https://github.com/google/docsy/pull/2665) switch), dropping the htmltest + refcache.json toolchain:
  - `check:links` now runs Lychee through the [link-cache](https://github.com/chalin/link-cache) helpers; `fix:refcache` and friends keep their names — and their 5 workflow consumers plus `/fix:refcache` keep working — but now mean "check links, updating the committed `.lycheecache`".
  - Replaces the perl → `.htmltest.yml` → translate config chain with a lychee-native generator (`scripts/lychee/config/index.mjs`, tested): `lychee.toml` = `lychee.base.toml` + `exclude_path` entries computed from `drifted_from_default` and the new `link_check_exclude_path` front-matter key (renamed from `htmltest:`; 2 pages).
  - Commits `.lycheecache` (seeded by translating `refcache.json` and normalizing with a full run), so checks only fetch new or expired URLs. Lychee never caches failures, so broken links are retried on every run.
- **CI**:
  - `check-links.yml`: the 3 htmltest shards and the `REFCACHE updates?` job collapse into a single blocking `CHECK LINKS` job over the shared build artifact: lychee sanity tests, authenticated full check, and a `git diff --exit-code -- .lycheecache` staleness guard.
  - Adds the `.github/actions/install-lychee` composite action as the single version-pin home (v0.24.2) for the six consuming workflows.
  - `refcache-refresh.yml` keeps its name, bot branch, and cadence; it now prunes the N oldest cache entries (N is a workflow input) and re-checks with Lychee. The Chrome/4xx double-check steps are gone (failures are never cached).
- **Drops htmltest and refcache artifacts** (~20 files): configs, bridge translators, refcache maintenance scripts, and `static/refcache.json`; `.lycheecache` inherits its no-owner role (CODEOWNERS, component-owners, locale-auto-merge).
- **Docs and agent skills** describe the Lychee-only reality: `site/build/link-checking` rewritten; related build/contributing pages updated; `refresh-refcache-pr-fix`, `resolve-refcache-conflicts`, and `review-pull-request` skills reworked for the new cache semantics.
[...]

PR #10968, opened 2026-07-22
/fix: guard against pre-Lychee-switch branches; KubeCon Japan banner tweak

chalin:
Two small follow-ups, one commit each:

- **`/fix` stale-branch guard** — PR branches cut before the Lychee link-checker switch (#10911) still carry the old refcache wiring, and the fix pipeline runs the *head's* scripts, so `/fix` updates the deleted `static/refcache.json` instead of `.lycheecache` and `CHECK LINKS` stays red (seen on #10946, which took two `/fix` rounds plus a merge-main to recover).
  - Adds a guard step in `generate-patch` (the workflow runs from `main`, so it covers all stale heads): when the checked-out head still has `static/refcache.json`, skip the fix scripts.
  - The report job relays the reason via a new generic `--not-run-reason` path in the patch-report composer; the /fix-specific reason text lives with the directive parser, alongside its hint. Sample outcome comment: "⚠️ `fix:refcache` was not run: this PR branch predates the Lychee link-checker switch (#10911), so its fix scripts would update the old link cache. Update the branch from `main`, then re-run `/fix`."
  - The guard is marked removable (TODO) once pre-switch branches age out.
- **KubeCon Japan announcement** — drops the forced `<br>` (`d-none d-md-inline`) so the banner is a single line on wide displays and wraps naturally on narrow ones, like the Zoom-migration announcement. (The same pattern exists in the already-expired announcement files; left untouched.)

- **Preview**: https://deploy-preview-10968--opentelemetry.netlify.app/ (banner on the home page)

### DIFF
--- a/.github/workflows/pr-actions.yml
+++ b/.github/workflows/pr-actions.yml
@@ -96,6 +96,7 @@ jobs:
       action_exit_status: ${{ steps.patch.outputs.action_exit_status }}
       patch_name: pr-fix-${{ github.run_id }}
       patch_skipped: ${{ steps.patch.outputs.patch_skipped }}
+      stale_branch: ${{ steps.stale-branch-guard.outputs.stale_branch }}
 
     steps:
       - uses: step-security/harden-runner@bf7454d06d71f1098171f2acdf0cd4708d7b5920 # v2.20.0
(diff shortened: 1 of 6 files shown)


chalin:
Closing in favor of separate PRs for independent review (the review effort differs significantly): #10970 (`/fix` stale-branch guard) and #10971 (KubeCon Japan banner).

Issue #10990, opened 2026-07-23
Old PRs need a branch update following the link-checker switch


chalin:
#10911 switched the site's link checking from htmltest to Lychee and deleted `static/refcache.json`. PR branches created before then still carry the old wiring, which can cause CI failures unrelated to the PR's changes.

## What to do

> [!NOTE]
>
> All affected PRs have received a comment, including details for PRs that need different handling (such as a rebase). See your PR's comment for the exact steps to follow.

In general, update your PR branch over the latest `main` as follows.

- If your PR page offers an **Update branch** button, click it. Done.
- Otherwise, merge from the command line — with [`upstream` pointing to this repo](https://opentelemetry.io/docs/contributing/pull-requests/#clone-and-set-upstream):

  ```sh
  git fetch upstream
  git merge upstream/main
  ```

  If git reports `CONFLICT (modify/delete): static/refcache.json`, accept the deletion and finish the merge:

  ```sh
  git rm static/refcache.json
  git commit --no-edit
  git push
  ```

  For any other conflicts, see [Merge conflicts and rebasing](https://opentelemetry.io/docs/contributing/pull-requests/#merge-conflicts-and-rebasing) — GitHub's web editor can't resolve this kind of conflict, so use the command line.

- Stuck? Comment on your PR and a maintainer can help.

If `CHECK LINKS` fails after you update, see [`BUILD` and `CHECK LINKS`](https://opentelemetry.io/docs/contributing/pr-checks/#build-and-check-links).

The `FILENAME check` now flags PRs that edit or restore `static/refcache.json` — the usual sign of a stale branch (#10993). Questions? Ask here.

chalin:
Closing as resolved: the post-switch update wave is behind us, and the `FILENAME check` (#10993) now flags stale branches automatically. This issue intentionally stays as the permanent reference — the check's error message and the [PR checks docs](https://opentelemetry.io/docs/contributing/pr-checks/) link here, and the update instructions above remain current.

Landing here from a failed check? Follow the steps above; if you're stuck, comment on your PR and a maintainer will help.

PR #10991, opened 2026-07-23
Add STALE BRANCH check for pre-Lychee-switch PR branches

chalin:
- Contributes to #10990
- Extends the Files workflow with a `STALE BRANCH check` job that fails when the PR head still has `static/refcache.json` (deleted by the Lychee switch, #10911), giving authors a named failure with a pointer to the update-or-rebase instructions in #10990 — instead of confusing link-check errors.
- Inspects the PR head rather than the merge preview, since a clean merge resolves main's deletion and would mask the staleness; skipped in merge groups, which have no PR head.
- Keeps the `/fix` stale-branch guard (#10970): this check *informs* authors, while the guard *prevents* the bot from committing to the resurrected cache when `/fix` is run on a stale branch — which a red check makes more likely, not less. Both self-retire once pre-switch branches age out (TODO(2026-08)).


chalin:
Check signal proven live on this PR, both ways:

- **Red** — head carrying `static/refcache.json` (ffe3302): [STALE BRANCH check ❌](https://github.com/open-telemetry/opentelemetry.io/actions/runs/30021333246/job/89254506814), with annotation: _"This branch predates the Lychee link-checker switch (#10911) — it still has static/refcache.json. Update your branch over main, accepting main's deletion of that file. For step-by-step instructions, see #10990."_
- **Green** — file removed again (d8e1186): [STALE BRANCH check ✅](https://github.com/open-telemetry/opentelemetry.io/actions/runs/30021484833/job/89255039958)

The two `Red-proof` commits are that demonstration pair — they cancel out tree-wise and disappear at squash-merge.

chalin:
Closing in favor of a leaner approach that extends the existing FILENAME check (which already runs on both the PR merge preview and the merge queue) with a data-driven obsolete-paths list and per-file guidance messages — no new CI job needed.

Keeping this PR closed-but-in-history as a possible alternative: a head-checkout job like this one is the only variant that can flag stale-but-cleanly-merging branches proactively, should that signal ever prove necessary.

PR #11009, opened 2026-07-24
Add whatbroke to the registry

arthi-arumugam-git:
Adds whatbroke to the registry as a utilities entry. It is a CLI I built that diffs an AI agent's behavior between two runs, and since v0.3.0 it imports OTLP JSON span exports that follow the GenAI semantic conventions. MIT licensed, TypeScript, published on npm as whatbroke-cli.


data/registry/tools-js-whatbroke.yml
```diff
@@ -0,0 +1,27 @@
+# cSpell:ignore whatbroke arthi arumugam
```
vitorvasc:
```suggestion
# cSpell:ignore whatbroke arthi arumugam genai
```

vitorvasc (commented):
Could you also run the `npm run fix:link-cache` script locally and commit the results?

Thanks!

arthi-arumugam-git:
Thanks for the `genai` suggestion, that was the SPELLING failure exactly: the word list is `caseSensitive: true`, so the lowercase tag never matched `GenAI`. Applied in b264c4e. Thanks also for merging main into the branch.

On the link cache, I ran `/fix:link-cache` twice rather than locally, and it could not converge. The cause was the `docs` URL in my entry, `https://github.com/arthi-arumugam-git/whatbroke#readme`. In the CHECK LINKS job it resolved 200 at 11:09:28 UTC and the entry was written into the generated cache. In the `/fix:link-cache` run 15 minutes later, lychee reported `Cannot find fragment` for that same URL, exited 2, and so never wrote it. That is why the verify job kept seeing a stale cache while the fix run had nothing to apply. A GitHub repository landing page does not reliably render a `readme` anchor, so that fragment is a coin flip, and running it locally as you suggested would have hit the same error.

Fixed by dropping the fragment and pointing `docs` at the README directly:

```yaml
docs: https://github.com/arthi-arumugam-git/whatbroke/blob/main/README.md
```

That URL returns 200 and has nothing to resolve. I have also merged main again, since the generated cache had picked up one unrelated entry while the branch sat behind.

Happy to use a different `docs` value if you would rather.

vitorvasc (approved):

PR #10993, opened 2026-07-23
FILENAME check: data-driven checker with per-path guidance

chalin:
- Contributes to #10990; supersedes #10991 (closed). The `FILENAME check` already runs on the merge preview and merge queue — no new CI job needed.
- Replaces the four shell one-liners behind `check:filenames` / `fix:filenames` with a tested, data-driven script (`scripts/validate/filenames/`) that enforces both rules — kebab-case names and no obsolete paths:
  - Each obsolete-path entry carries its own guidance message, emitted as a per-file error annotation.
  - `--fix` validates the full rename plan before deleting or renaming anything, refusing destination collisions and renaming deepest-first.
- Docs: `static/refcache.json` added to the "Obsolete files and folders" list with stale-branch guidance.

Red/green proven live — [proof comment](https://github.com/open-telemetry/opentelemetry.io/pull/10993#issuecomment-5060917444). **Preview**: [PR checks: FILENAME check](https://deploy-preview-10993--opentelemetry.netlify.app/docs/contributing/pr-checks/#filename-check)

PR #11011, opened 2026-07-24
Revert "Guard against pre-Lychee-switch PR branches using `/fix:*` (#10970)"

chalin:
- Reverts #10970 (`git revert 2728c8fbf`): removes the `/fix` stale-branch guard, its patch-report plumbing, and tests.
- The guard never fired: 220 pr-actions runs and 6 real `/fix` executions since merge, zero declines.
- Coverage is already in place:
  - All affected PRs got recovery instructions in the #10990 sweep; the cohort is dormant and shrinking (91 → 51 open).
  - FILENAME (#10993) guards the re-add class, not this one: stale branches fail as merge conflicts, where checks don't run.
- Why now: the revert is still clean, and upcoming pr-actions work touches these files.
- Residual risk (`/fix` on a stale branch): recovery is documented in #10990; if it ever bites, the right shape is a branch-update action, not a guard.

### DIFF
--- a/.github/workflows/pr-actions.yml
+++ b/.github/workflows/pr-actions.yml
@@ -96,7 +96,6 @@ jobs:
       action_exit_status: ${{ steps.patch.outputs.action_exit_status }}
       patch_name: pr-fix-${{ github.run_id }}
       patch_skipped: ${{ steps.patch.outputs.patch_skipped }}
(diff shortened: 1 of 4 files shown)


jaydeluca (approved):

chalin:
From Copilot:

> Confirmed: all four files are byte-for-byte identical to their state immediately before #10970 (matching blob SHAs). This is an exact revert with no unrelated changes.

PR #11076, opened 2026-07-28
[fr] Update Java agent pages to match English

0xdorian-sm:
Part of #8777. Second lot after #11075 (Python).

Resyncs eight French pages under `zero-code/java` with their English sources and clears their `drifted_from_default` flags.

### Content changes, per upstream

- Bumps the cascaded versions to instrumentation 2.30.0 / otel 1.64.0 and adds the new `contrib` version.
- `api`: `GlobalOpenTelemetry.get()` becomes `getOrNoop()`.
- `getting-started`: translates the new **Declarative configuration** section.
- `server-config`: translates the new Tomcat installation paths (package-managed, download, Windows service) and the new **Enable predefined JMX metrics** section.
- `disable`: refreshes the instrumentation table, which gained **21 entries** (Activej, Avaje Jex, Alibaba Druid, ElasticJob, Iceberg, Elasticsearch API client, Failsafe, Helidon, Java HTTP Server, JFinal, NATS, OpenAI, OpenSearch, Payara, Powerjob, Spring Boot Actuator Autoconfigure, Spring Cloud AWS, and others), bumps the page weight to 12, and extends the cSpell list accordingly.
- Converts the remaining `alert` shortcodes on these pages to GitHub-style callouts.

English heading anchors are preserved, per the French localization convention.

### Scope

The Java lot is split so it stays reviewable. Still to come: `agent/extensions.md` (large enough to deserve its own PR) and the `spring-boot-starter/` pages.

### Note

I still cannot run `npm run fix:format` locally (no Node on this machine), so please flag any Prettier complaints from CI and I will push a fixup right away. The `disable` table rows were regenerated from the English source at identical column widths, so they should be stable.

/cc @dmathieu

PR #11075, opened 2026-07-28
[fr] Update zero-code/python pages to match English

0xdorian-sm:
Part of #8777.

Resyncs the six French `zero-code/python` pages with their English sources and clears their `drifted_from_default` flags.

### Content changes, per upstream

- **Logging config**: replaces `OTEL_PYTHON_LOGGING_AUTO_INSTRUMENTATION_ENABLED` with `OTEL_PYTHON_LOG_AUTO_INSTRUMENTATION` and `OTEL_PYTHON_LOG_CODE_ATTRIBUTES`, keeping the pre-1.40.0 note.
- Adds `OTEL_PYTHON_AUTO_INSTRUMENTATION_EXPERIMENTAL_GEVENT_PATCH` and the new *gevent applications* section on the operator page.
- Translates the new **Pre-fork server issues** section, including the workarounds table and the Gunicorn/UvicornWorker, programmatic auto-instrumentation, Prometheus and single-worker options.
- Translates the new **Sanitization of captured headers** section and extends the list of supported HTTP instrumentations.
- Converts the two `alert` shortcodes to GitHub-style callouts.
- Adds the `collector_vers` cascade to `content/fr/docs/zero-code/python/_index.md` so `logs-example` can resolve the `param` shortcode, and pins the collector image accordingly.

### Drive-by fix

`logs-example` told readers to save the collector config as `otel-collector-config.yaml` but then mounted `otel-collector-config.yml`, so the documented `docker run` command could not work as written.

English heading anchors are preserved, per the French localization convention.

### Note

I could not run `npm run fix:format` locally (no Node on this machine), so please flag any Prettier complaints from CI and I will push a fixup right away.

/cc @dmathieu

PR #9529, opened 2026-03-30
Java agent extensions - Writing extensions

jaydeluca:
- [x] I have read and followed the [Contributing](https://opentelemetry.io/docs/contributing/) docs, especially the "**First-time contributing?**" section.
- [x] This PR has content that I did not fully write myself.
  - [x] I used AI and I have read and followed the [Generative AI Contribution Policy](https://github.com/open-telemetry/community/blob/main/policies/genai.md).
- [x] I have the experience and knowledge necessary to understand, review, and validate all content in this PR.[^I-know-my-stuff]

[^I-know-my-stuff]:
    Yes, I can answer maintainer questions about the content of this PR, without using AI.

---

Deployment Preview: https://deploy-preview-9529--opentelemetry.netlify.app/docs/zero-code/java/agent/extensions/

Following #9329 

This is part of breaking down #8264 into smaller pieces for easier review

There will be ~2 more parts following this one (writing custom instrumentation, muzzle and other)

PR #9329, opened 2026-03-03
Java Agent Extensions Rewrite - Part 1

jaydeluca:
- [x] I have read and followed the [Contributing](https://opentelemetry.io/docs/contributing/) docs, especially the "**First-time contributing?**" section.
- [x] This PR has content that I did not fully write myself.
  - [x] I used AI and I have read and followed the [Generative AI Contribution Policy](https://github.com/open-telemetry/community/blob/main/policies/genai.md).
- [x] I have the experience and knowledge necessary to understand, review, and validate all content in this PR.[^I-know-my-stuff]

[^I-know-my-stuff]:
    Yes, I can answer maintainer questions about the content of this PR, without using AI.

---

Related to #8264 which is in a good place but is unreasonable large, so I decided to break it into smaller pieces for easier review.

I plan on breaking it into three PRs:

- Quick start and how to load and use extensions (this PR)
- Writing Extensions and custom instrumentation
- Other (muzzle, troubleshooting, misc)

The code snippets are synced from the [examples repo](https://github.com/open-telemetry/opentelemetry-java-examples/tree/main/doc-snippets/extensions-minimal)

Deploy Preview: https://deploy-preview-9329--opentelemetry.netlify.app/docs/zero-code/java/agent/extensions/

PR #11077, opened 2026-07-28
[fr] Translate the rewritten Java agent extensions guide

0xdorian-sm:
Part of #8777. Third lot, after #11075 (Python) and #11076 (Java agent).

Upstream rewrote and greatly expanded this page in #9329, #9263 and #9529 — it went from 56 to 815 lines. The French page was still the old 56-line version, so this is a **translation of the new guide**, not a resync.

### What is translated

- The new introduction and the **Quick start**.
- **Using extensions**: runtime loading (single, multiple, directory, mixed paths), how extension loading works, and embedding extensions in the agent including the `extendedAgent` Gradle task.
- **Writing extensions**: dependency scopes (`compileOnly` vs `implementation`), extension points overview, configuration access and naming conventions, and `@AutoService`.
- The **extension point reference** for `AutoConfigurationCustomizerProvider`, `InstrumenterCustomizerProvider`, `ConfigurablePropagatorProvider`, `ConfigurableSamplerProvider` and `ResourceProvider`.

### Care taken with code blocks

The eight blocks under a `<?code-excerpt?>` directive are left **byte-identical** to the English source, so they keep matching their example projects and the excerpt check stays green. I verified this programmatically after writing the page. Comments are translated only in blocks that are *not* excerpt-verified, where they are part of the explanation rather than of a checked sample.

English heading anchors are preserved on all 32 headings, per the French localization convention, and both tables are column-aligned.

### Note

I still cannot run `npm run fix:format` locally (no Node on this machine), so please flag any Prettier complaints from CI and I will push a fixup right away.

/cc @dmathieu

### DIFF
--- a/content/fr/docs/zero-code/java/agent/extensions.md
+++ b/content/fr/docs/zero-code/java/agent/extensions.md
@@ -5,56 +5,822 @@ description:
   Les extensions ajoutent des capacités à l'agent sans avoir à créer une
   distribution séparée.
 weight: 300
-default_lang_commit: 3d179dbe1270b83aafff0d3b6aa3311afd482649
-drifted_from_default: true
+default_lang_commit: cb8364effee3fd3f2dc33c15da7c47bde0432122
+cSpell:ignore: Customizer Dotel myextension
 ---
+
+<!-- markdownlint-disable blanks-around-fences -->
+<?code-excerpt path-base="examples/java/extensions-minimal"?>
 
 ## Introduction {#introduction}
 
-Les extensions sont conçues pour surcharger ou personnaliser l'instrumentation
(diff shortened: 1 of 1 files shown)


chalin:
Thanks for your interest in OpenTelemetry. Per our [first-time contributor policy](https://github.com/open-telemetry/opentelemetry.io/issues/11243), we are closing first-timer issues and PRs that don't comply with it; in particular, keep to a **single** open contribution at a time. Also note that the issues these PRs reference are CI/infra tasks reserved for maintainers, not `good first issue` items. This closure is about process, not the merit of the change; see the policy for how to proceed, and note that we may reopen this later if capacity allows.

PR #11094, opened 2026-07-29
docs(demo): update collector exporter names for v0.145.0

mwimpelberg28:
Addresses item 4 of #10962.

The demo's Collector configuration renamed the OTLP exporters (`otlp` → `otlp_grpc/jaeger`, `otlphttp/prometheus` → `otlp_http/prometheus`) and moved span metrics to the `span_metrics` connector in open-telemetry/opentelemetry-demo#2942. The docs still referenced the old names.

### Changes

**`docker-deployment.md`**

- `spanmetrics` → `span_metrics` in the bring-your-own-backend pipeline snippet and the accompanying note. Reworded the note to call it a connector rather than an exporter, and to say why it must stay in the array if you override the `traces` pipeline — the `metrics` pipeline consumes it as a receiver.
- Replaced the "merges the configuration from two files" description with the current four-file layering: `otelcol-config.yml` → `otelcol-config-full.yml` → `otelcol-config-observability.yml` → `otelcol-config-extras.yml`.
- Dropped the now-unused `spanmetrics` from `cSpell:ignore`.

**`sample-configurations/tail-sampling-service-criticality.md`**

- `exporters: [otlp, debug, spanmetrics]` → `[otlp_grpc/jaeger, debug, span_metrics]`.

### On `kubernetes-deployment.md`

The tracking issue lists the same rename for `kubernetes-deployment.md`, but that page documents the Helm chart rather than the compose stack, and the released chart still uses the old names — `charts/opentelemetry-demo` v0.40.10 (appVersion 2.2.0) has `connectors: spanmetrics: {}` and `exporters: [otlp/jaeger, debug, spanmetrics]`. Renaming there now would break the copy-paste snippet for anyone on the current chart. The rename is queued in open-telemetry/opentelemetry-helm-charts#2326 ("[demo] Release 3.0"); the docs should follow once that ships, so I've left that box unchecked.

### Note on the layering wording

`compose.observability.yaml` mounts `otelcol-config-full.yml` as well, so `make start-minimal` loads all four config files even though it skips the full *compose* layer. The prose reflects that rather than implying minimal mode drops a config file.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #10962, opened 2026-07-22
[demo] Update demo docs to match opentelemetry-demo main (removed/added services, k6, collector renames)


mwimpelberg28:
## Summary

The OpenTelemetry Demo documentation under `content/en/docs/demo/` has drifted from the
current [`opentelemetry-demo`](https://github.com/open-telemetry/opentelemetry-demo) `main`
branch. This is a meta/tracking issue enumerating each needed update, grouped by theme.
File paths are relative to `content/en/docs/demo/`. Linked PRs are in `opentelemetry-demo`.

Happy to split any group into its own issue/PR — this is a checklist to coordinate the work.

### 1. Remove deleted services (`product-reviews`, `llm`) — — open-telemetry/opentelemetry.io#10963

`product-reviews` and `llm` were removed from the demo
(open-telemetry/opentelemetry-demo#3568, open-telemetry/opentelemetry-demo#3587,
open-telemetry/opentelemetry-demo#3599) but the docs still reference them.

- [x] Delete `services/product-reviews.md` (entire page is for the removed service)
- [x] `architecture.md` — remove `llm` and `product-reviews` nodes and all their edges from the `graph TD` service diagram
- [x] `_index.md` — remove Product Reviews from the Python row of the Language Feature Reference table, and from the Service Documentation list
- [x] `services/_index.md` — remove the `product-reviews` table row
- [x] `feature-flags/_index.md` — remove the `llmInaccurateResponse` and `llmRateLimitError` flags (mock LLM service is gone)
- [x] `telemetry-features/metric-coverage.md`, `trace-coverage.md`, `log-coverage.md` — remove the Product Reviews rows
- [x] `telemetry-features/manual-span-attributes.md` — remove the "Product Reviews" section and `app.product_reviews.*` attributes
- [x] `sample-configurations/tail-sampling-service-criticality.md` — remove `product-reviews` from the medium-criticality tier

### 2. Add docs for new services — see #10574

New services added but undocumented: Agent/MCP/Chatbot
(open-telemetry/opentelemetry-demo#3455), OpAMP server
(open-telemetry/opentelemetry-demo#3566), telemetry-docs
(open-telemetry/opentelemetry-demo#2794).

- [x] Create `services/agent.md` — Python 3.14, FastAPI + Uvicorn, LangChain/LangGraph ReAct agent, `POST /prompt`, port 8010, instrumented with the Traceloop SDK + OTLP; built-in tools vs MCP tool mode
[...]

PR #11095, opened 2026-07-29
docs(demo): fix outdated Docker Compose commands

mwimpelberg28:
Addresses item 5 of #10962.

> [!NOTE]
> **Stacked on #11094** (item 4), because both edit `docker-deployment.md` and would otherwise conflict on the `cSpell` front matter. Only the second commit, `docs(demo): fix outdated Docker Compose commands`, is new here — the diff will shrink to just that once #11094 merges. Happy to rebase instead if you'd rather review them independently.

The demo moved to layered `compose.*.yaml` files, and several of the files and Make targets this page referenced no longer exist. Every command below was checked against `opentelemetry-demo` `main`.

### Changes

**Start commands**

- The Docker tab for the full demo was a bare `docker compose up`, which loads only core `compose.yaml` — the web store with no Kafka and no observability backends, so there's nowhere to view telemetry. It now passes the same four files `make start` uses, with a note explaining the difference.
- Added a short list of what each of the four layers contributes: `compose.yaml`, `compose.full.yaml`, `compose.observability.yaml`, `compose.extras.yaml`.

**Minimal mode**

- `docker compose -f docker-compose.minimal.yml ...` → `-f compose.yaml -f compose.observability.yaml -f compose.extras.yaml`, matching `make start-minimal`.
- Dropped `flagd-ui` from the exclusion list — it's in core `compose.yaml` now, so minimal mode includes it. The excluded set is `accounting`, `fraud-detection`, `kafka`.

**Testing step**

The "API observability-driven testing" step was entirely dead: `make run-tracetesting`, `docker-compose-tests.yml`, and the `traceBasedTests` service have all been removed, and Tracetest now appears in the demo only as a vendor-fork link in the README. Replaced with the two current entry points:

- `make run-frontend-tests` — Cypress, via `compose.tests.yaml`'s `frontendTests` service, run against an already-running demo.
- `make run-telemetry-tests` — asserts each service emits the expected traces, metrics, and logs. Noted that this target starts and stops the demo itself, so it wants the demo stopped first.

**New stacks documented**

- `make start-agentic` / `compose.agent.yaml`, with the Chatbot UI at `/chatbot/`. Noted that `USE_VCR=True` is the default, so the agent replays recorded LLM responses and no API key is needed to try it.
- `make start-profiling` / `compose.profiling.yaml`, with the Firepit UI at `/profiles/`.

**Verify-URL list**

Removed the Tracetest UI entry (`localhost:11633`, no longer exists) and added `/telemetry/` for the Weaver-generated telemetry docs, plus `/chatbot/` and `/profiles/` marked as mode-specific. All routes confirmed against `src/frontend-proxy/envoy.tmpl.yaml`.

### Adjacent drift, not included

`content/en/docs/demo/tests.md` has the same problem — it still documents `make run-tests`, `integrationTests`, `traceBasedTests`, and Tracetest, none of which exist. It isn't enumerated in #10962, so I left it out rather than widen this PR. Happy to fix it here or in a follow-up, whichever you prefer.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


content/en/docs/demo/docker-deployment.md
julianocosta89: could you rename this to the new name?

```suggestion
    otlp_http/example:
```

content/en/docs/demo/docker-deployment.md
```diff
@@ -154,15 +236,15 @@ with an editor.
   service:
     pipelines:
       traces:
-        exporters: [spanmetrics, otlphttp/example]
+        exporters: [span_metrics, otlphttp/example]
```
julianocosta89:
```suggestion
        exporters: [span_metrics, otlp_http/example]
```

julianocosta89 (approved):
one minor nit, but LGTM

maryliag (approved):

PR #11168, opened 2026-08-03
Propagate exit codes from log:* scripts; single-build double-check in fix-loop skill

chalin:
- Contributes to the [#11133](https://github.com/open-telemetry/opentelemetry.io/pull/11133) double-check arc; carries the skill fix-loop step deliberately moved off the policy-scoped [#11163](https://github.com/open-telemetry/opentelemetry.io/pull/11163) (the missing step bit a live run on [#11166](https://github.com/open-telemetry/opentelemetry.io/pull/11166))
- **`log:*` exit codes**: `tee` masked the underlying script's failure, so `npm run log:check:links` reported success on failing links; `bash -o pipefail` wrappers fix both scripts (repo scripts already require a POSIX shell — `${PRUNE_N:-800}`, `${BUILD_KIND:-lean}` — so no portability regression)
- **CI adoption**: refcache-refresh's link-check step collapses onto `log:check:links`, converging CI and local runs on one entry point with the same log path (`tmp/check-links-log.txt`) the double-check step reads
- **Skill single-build flow**: fix-loop step 1 now uses `log:check:links`, so the restored double-check step (new step 4) probes the captured log without a second build
- **Red-first proof**: with `check:links` stubbed to fail, the old form exited 0 (false-green) and the fixed wrapper exited non-zero; stub then reverted


chalin:
Closing, for a take 2.

PR #11226, opened 2026-08-06
Move to api & sdk types

thompson-tomo:

[^I-know-my-stuff]:

---

This switches a batch of registry entries which were previously using core to now use api or sdk. In case of otel additional entries were added when possible.


ymotongpoo (approved):
LGTM for JA part

chalin (changes requested):
Website maintainers are freezing the curated ecosystem lists, which affects whether this change will land. For details see:

- #11377

svrnm:
@thompson-tomo as you might know we froze the registry, so we can not accept this PR right now, or is there any reason outside adding new items and renaming them to have this updated?

chalin:
All, let's consider the proposal described in the following issue instead:

- #11666

thompson-tomo:
@svrnm no other reason and note the only reason why i spent the time going through the existing entries is because you mentioned ww should update them.

svrnm:
> @svrnm no other reason and note the only reason why i spent the time going through the existing entries is because you mentioned ww should update them.

thanks, let's keep this open while we decide how to move forward with the registry

Issue #7510, opened 2025-08-08
[CI] `check:i18n` adds incorrect commit hashes on `default_lang_commit`


vitorvasc:
### Description

This issue raises two items to be investigated for more details:

1. In #7450, we found that when using the `check:i18n` command to add the `default_lang_commit` field in the front matter of localized files, the command adds an incorrect hash instead of the actual hash for the given file. 
2. In both #7431 and #7450, when files had an invalid commit hash, the merge only failed after the PRs were added to the merge queue. We should validate why the checks aren't failing in the PR workflow instead.

### Next steps

- [ ] Verify how the `check:i18n` determines the commit hash and update it to get the most recent commit hash for the given file.
- [ ] Ensure that CI catches invalid or mismatched hashes during PR checks, not only when PR is added to the merge queue.

PR #5891, opened 2025-01-07
🇺🇦 Ukrainian localization of OpenTelemetry documentation

Andygol:
I'm thrilled to share the news that I've translated the OpenTelemetry.io website into Ukrainian — https://andygol-otel.netlify.app/uk/! If this is something you're interested in, you're more than welcome to merge it to the main repo.

| image |  image |
| --- | --- |

PR #11260, opened 2026-08-08
Use the SHA from the latest EN commit for the `default_lang_commit` parameter

Andygol:
This PR adds additional functionality that allows you to use not only the HEAD or a specific commit number, but also to retrieve the latest SHA of a commit from the English source to populate the `default_lang_commit` field in localized files.

[^I-know-my-stuff]:

### DIFF
--- a/.github/localization.instructions.md
+++ b/.github/localization.instructions.md
@@ -51,7 +51,7 @@ npm run check:i18n -- diff content/zh  # Show detailed diff for drift
 **Update Localization Tracking:**
 
 ```bash
-npm run check:i18n -- commit HEAD --new content/zh # Add default_lang_commit to new pages
+npm run check:i18n -- --from-src-latest --new content/zh # Add default_lang_commit to new pages
 npm run check:i18n -- commit <HASH> content/zh     # Update existing pages commit hash (also syncs drift status)
 npm run fix:i18n:status -- <PATHS>                 # Sync drift status; in PRs, pass only the pages a failing check reports (tree-wide runs are Housekeeping's)
 ```
@@ -130,7 +130,8 @@ default_lang_commit: <commit-hash-of-english-page>
 1. Review changes: `npm run check:i18n -- diff content/xx/path/to/page`
 2. Update your localized page to match new English content
 3. Update `default_lang_commit` to latest hash (this also clears the page's
-   drift status): `npm run check:i18n -- commit HEAD content/xx/path/to/page`
+   drift status): `npm run check:i18n -- commit HEAD content/xx/path/to/page` or
+   `npm run check:i18n -- --from-src-latest content/xx/path/to/page`
 
 **Patching (build fixes only):**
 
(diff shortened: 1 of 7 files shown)


chalin (commented):
Thanks @Andygol.

**TL;DR**: A file-scoped pin is exactly what #7510 calls for, but the core needs rework before it's safe, and the natural companion (CI-side pin validation, #7510's second half) is beyond this PR's scope. I'd rather pick up the full work in a maintainer PR. Details are next.

1. **Anchor the file-scoped SHA to upstream `main`.**
   - `git log -1 -- FILE` in a contributor clone returns fork-local SHAs whenever the branch (or a diverged fork `main`) touched the EN file (the exact failure mode of #5891).
   - `writePinsFromEnLatest` also bypasses the `git branch --contains … main` guard that `writePins` applies, so the one existing safety net is lost.
   - Fix: resolve against main (`git log -1 main -- FILE`, with a help-text tip to fetch upstream first), or at minimum run the resolved SHA through the same on-main guard.
2. **Don't silently repin `error`-status pages.**
   - An unresolvable pin is a signal to investigate; overwriting it in the default (`drifted`) list kind hides bad pins.
   - Fix: keep `error` out of the default selection, or gate it behind `--all`.
3. **`--from-src-latest` shouldn't mutate the parsed noun/write mode.**
   - Forcing `noun = 'commit', write = true` turns what reads like a flag into an implicit write command.
   - Fix: make it an argument form of the `commit` noun (for example, `commit --from-src-latest PATHS`) so read vs write stays explicit.
4. **Split out the EN → "source-language" rename, or drop it.**
   - The rename (comments, variable names, the `enCounterpartOf` alias, the compat option keys) is roughly 2/3 of this diff, and it's purely cosmetic while the default language is `en`; the compat shims add surface area of their own.
   - Fix: a separate PR, or drop it. Either leaves a small, reviewable behavioral change here.
5. **Docs shouldn't call the two forms "equivalent"** (`localization.md`, the `drift-pr-guard.mjs` message).
   - `commit HEAD` pins main's HEAD (repo-scoped, guarded); `--from-src-latest` pins a per-file SHA (currently unguarded). Different SHAs, different guarantees.
   - Fix: once (1) lands, describe `--from-src-latest` as the _preferred_ form and say why.

With (1)-(3) addressed, this closes the first half of #7510. The second half (CI catching bad pins before the merge queue) remains open; a checker-side "pin is an ancestor of main" validation would cover it, and belongs in the follow-up.

Andygol:
@chalin, thanks for the review. I've made the changes based on your recommendations. I hope these changes will be helpful to all localization teams.

chalin:
Thanks @Andygol, but in the end, this doesn't work with our i18n roadmap. I'll followup with roadme PRs as soon as I can.

chalin:
FYI @Andygol: the policy that surfaced in this PR's close-out is now in the contributing docs, see https://opentelemetry.io/docs/contributing/issues/#ci-infra. This isn't a judgment of your submission here; it's us recognizing that this class of issue can't be meaningfully worked on without full maintainer context.

Issue #11243, opened 2026-08-07
Contribution policy for first-time / new contributors


chalin:
If you have not yet had a non-trivial PR merged in this repository, you are a **first-time contributor** here. Welcome! 

### Choose _one_

As a **first-time or new contributor**, you can do **one** of the following:

- **Open a single issue** describing a problem or proposal, and wait for maintainer feedback before doing anything else. Keep to one open issue at a time.
- **Submit a single PR** for _one_ of:
  - An open issue labeled [`good first issue`](https://github.com/open-telemetry/opentelemetry.io/labels/good%20first%20issue):
    - If there are no such issues, check back later.
    - **Link to the issue** from your PR description.
      ⚠️ **NOTE**: issues are **not** claimed, reserved, or [assigned](https://opentelemetry.io/docs/contributing/#first-time-contributing); the first adequate PR wins.
  - A [registry](https://opentelemetry.io/ecosystem/registry/adding/) or [vendor](https://opentelemetry.io/ecosystem/vendors/#how-to-add) self-addition, following the linked instructions. As with all first-timer PRs: one open PR at a time.
  - A _trivial editorial fix_ (typo, broken link) without an issue; such PRs are accepted at maintainer discretion.

> [!IMPORTANT]
> You remain a first-time contributor until a non-trivial PR of yours is merged: a merged trivial fix or registry/vendor self-addition doesn't change your status.

### Answer the PR template questions

Every PR opens with our [pull request template](https://github.com/open-telemetry/opentelemetry.io/blob/main/.github/PULL_REQUEST_TEMPLATE.md). First-time contributors **must** complete its checklist truthfully; PRs with an unaddressed or dishonestly completed checklist will be closed. As a reminder, the checklist asks you to confirm:

> - [ ] I have read and followed the [Contributing](https://opentelemetry.io/docs/contributing/) docs, especially the "**First-time contributing?**" section.
> - [ ] This PR has content that I did not fully write myself.
>   - [ ] I used AI and I have read and followed the [Generative AI Contribution Policy](https://github.com/open-telemetry/community/blob/main/policies/genai.md).
> - [ ] I have the experience and knowledge necessary to understand, review, and validate all content in this PR. (Yes, I can answer maintainer questions about the content of this PR, without using AI.)

> [!CAUTION]
[...]

PR #11340, opened 2026-08-15
docs(languages): specify OTEL_EXPORTER_OTLP_HEADERS format

### LINKED ISSUE #11093: [Page feedback]: Clarify the format of `OTEL_EXPORTER_OTLP_HEADERS`
### URL

https://opentelemetry.io/docs/languages/sdk-configuration/otlp-exporter/#otel_exporter_otlp_headers

### Description

The header format is unspecified.

There is an example, `api-key=key,other-config-value=value`.
However, this does not provide a definitive specification.

- What is the syntax of keys?
  Can they consist of any non-`=` characters or is there some kind of encoding?
- What is the syntax of values?
  Can they consist of any non-`,` characters or is there some kind of encoding?

Grafana Cloud provides this header using some kind of scheme where the value is partially urlencoded.
However, the use of `,` to separate key-value pairs implies that this value is not an URI query.

santhiprakash:
Fixes #11093

## What

The [OTLP Exporter Configuration](https://opentelemetry.io/docs/languages/sdk-configuration/otlp-exporter/) page showed an example (`api-key=key,other-config-value=value`) but never defined the syntax for keys and values — which characters are allowed, how commas and equals are handled, and whether quoting/escaping is possible.

## Why

Users configuring exporters cannot tell from this page whether they need to escape commas in values or how to include special characters. The protocol specification already answers this: headers must match the [W3C Baggage](https://www.w3.org/TR/baggage/#header-content) format, and semi-colon delimited metadata is not supported.

## How

Added a paragraph to the Header configuration section stating the W3C Baggage format, the `key1=value1,key2=value2` shape, the semi-colon metadata exclusion, and a link to the authoritative [Specifying headers via environment variables](/docs/specs/otel/protocol/exporter/#specifying-headers-via-environment-variables) section of the protocol spec.

English source only. The pt and ja locale mirrors of this page have the same gap, but I'm leaving those for the localization teams to sync.

## Verification

- All three linked URLs return 200 (checked with `curl`).
- Content-only change, no code.

### DIFF
--- a/content/en/docs/languages/sdk-configuration/otlp-exporter.md
+++ b/content/en/docs/languages/sdk-configuration/otlp-exporter.md
@@ -117,8 +117,8 @@ Headers are represented in a format matching to the
 [W3C Baggage](https://www.w3.org/TR/baggage/#header-content), for example
 `key1=value1,key2=value2`. Semi-colon delimited
 [metadata](https://www.w3.org/TR/baggage/#property) is not supported. For more
-details, see [Specifying headers via environment
-variables](/docs/specs/otel/protocol/exporter/#specifying-headers-via-environment-variables)
+details, see
+[Specifying headers via environment variables](/docs/specs/otel/protocol/exporter/#specifying-headers-via-environment-variables)
 in the protocol specification.
 
 ### `OTEL_EXPORTER_OTLP_HEADERS`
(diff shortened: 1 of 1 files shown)


chalin:
**Note from the maintainers**: per our [first-time contributor policy](https://github.com/open-telemetry/opentelemetry.io/issues/11243), contributors without a non-trivial merged PR in this repository should work on **one contribution at a time**, working through review feedback before opening more. You currently have 4 open PRs. We are keeping #11363, which has a complete PR checklist, and closing the others.

Once #11363 is resolved, you are welcome to resubmit this change -- one PR at a time, with the PR-description checklist completed.

PR #11370, opened 2026-08-18
Add blog post: Zero-code trace-log correlation with OBI

### LINKED ISSUE #11230: blog: Zero-code trace-log correlation with OBI
### Blog Post Title

Zero-code trace-log correlation with OBI

### Blog Post Outline

OBI can enrich application logs with trace_id/span_id at the kernel level, with
no code changes. The post explains what the feature does, how it works internally, and shows it running
end to end.

Outline:
1. Introduction
2. What you get — before/after examples
3. How it works — log-write interception via kprobes on tty_write/pipe_write, and
   how OBI keeps thread -> trace context accurate across Go goroutines, the
   Node.js event loop, Java thread pools, Ruby (Puma) and Python asyncio
4. Configuration — enabling it per service, field names, plain-text options
5. Demo with screenshots
6. Limitations
7. Getting started + links

### Technologies Used

- OpenTelemetry eBPF Instrumentation (OBI)
- Jaeger
- OpenTelemetry Collector
- Go (demo services, standard library only)

### Related Special Interest Groups (SIGs)

- SIG eBPF Instrumentation

### Sponsoring SIG

SIG eBPF Instrumentation

### Sponsor Name

@MrAlias 
@grcevski

### Additional Information

_No response_

mmat11:

[^I-know-my-stuff]:

---

Adds a blog post introducing OBI's zero-code trace-log correlation.

The post covers:

- What the enriched output looks like for JSON, NDJSON, and plain-text logs
- How the correlation works internally, including how OBI tracks context across goroutine migrations (Go), event-loop callbacks (Node.js), thread-pool handoffs (Java), Puma's reactor/worker split (Ruby), and asyncio task switches (Python)
- Configuration examples and a two-service demo (uninstrumented Go services + Jaeger) showing bidirectional log-trace correlation
- Current limitations (synchronous writes only, 8 KiB cap, `CAP_SYS_ADMIN`, Java virtual threads pending)

Sponsored by @MrAlias and @grcevski 

Closes https://github.com/open-telemetry/opentelemetry.io/issues/11230


content/en/blog/2026/obi-trace-log-correlation/index.md
```diff
@@ -0,0 +1,260 @@
...
+missing fields, and appends the enriched line to the same pty or pipe the
+application was writing to. From the container runtime's point of view, the
+application simply logged a line with trace context in it. Writes are fanned
+out to parallel workers sharded by output file, so ordering is preserved per
+log stream. The placeholder lines are trivially dropped in your log shipper
+with a single filter.
+
+The whole flow, end to end:
+
+```mermaid
```
skl, line 105: This is failing to render ([link](https://github.com/mmat11/opentelemetry.io/blob/21be5a0639900464cf58d66386f3c98a951bdb86/content/en/blog/2026/obi-trace-log-correlation/index.md#intercepting-the-write))

Image

svrnm:
did @MrAlias and @grcevski agree to sponsor this?

content/en/blog/2026/obi-trace-log-correlation/index.md
mmat11: fixed

MrAlias:
Yes. It looks like I missed this notification. I'll be sure to review.

content/en/blog/2026/obi-trace-log-correlation/index.md
```diff
@@ -0,0 +1,262 @@
...
+You get paged. A trace shows a request failing in one of your services, and you
+know the answer is in the logs — but which log lines belong to _that_ request?
+If the service never adopted structured logging with trace context, the honest
+answer is: you grep by timestamp and hope.
+
+Injecting trace context into logs traditionally means touching every service:
+adding an SDK, configuring the logger to emit `trace_id` and `span_id`, and
+redeploying.
+[OpenTelemetry eBPF Instrumentation (OBI)](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation)
+now does this from the kernel instead. It intercepts your application's log
```
MrAlias, line 21: The opening promise needs the boundaries that matter to someone evaluating the feature. OBI does not enrich arbitrary application writes: this applies to supported stdout/stderr streams, and identifiers are added only when OBI has active trace context for the write. Enabling it also requires an OBI configuration rollout and a placeholder filter in the log pipeline, so `no redeploys` should be scoped to `no application rebuild or redeploy`. Could we state those boundaries in the introduction and carry them into the adoption checklist proposed below?

content/en/blog/2026/obi-trace-log-correlation/index.md
```diff
@@ -0,0 +1,262 @@
...
+plain text — free-form lines get a `key=value` annotation:
+
+```text
+payment authorized trace_id=4bf92f3577b34da6a3ce929d0e0e4736 span_id=00f067aa0ba902b7
+```
+
+If your logger already emits one of the configured fields, OBI preserves it and
+only fills in what's missing.
+
+## How it works
```
MrAlias, line 62: This heading is where the post changes audiences. The opening speaks to someone responding to an incident, but the next hundred lines speak to an OBI or eBPF contributor: map layouts, `pid_tgid`, kernel hooks, buffer rewriting, runtime-private symbols, and scheduler behavior. That material explains how the implementation was built, but it pushes the feature's value, applicability, and rollout guidance behind details that the likely user of this feature does not need.

I realize #11230 was approved with a detailed `How it works` section. This is an intentional request to revisit that editorial decision after seeing the complete draft, not a claim that the post failed to follow the proposal. The execution shows the tradeoff more clearly than the outline could: preserving the promised internals makes this read as developer documentation and weakens the announcement for the people most likely to adopt the feature. The OTel blog serves both end users and developers, but this post still needs a primary reader; implementation contributors can follow durable developer documentation without making every blog reader work through the implementation.

I don't think the right fix is to shorten this section while preserving the existing structure. The post should be rebuilt around one primary reader: an SRE, platform engineer, or observability practitioner who runs or is evaluating OBI and needs better incident response for services that cannot easily be modified. Application owners with legacy or uninstrumented services are a useful secondary audience. OBI/eBPF contributors are not the primary audience for this announcement; the repository's developer documentation is the better home for the implementation design.

The editorial thesis should be something like: **OBI can add trace and span context to existing application logs, allowing responders to move between a failed trace and the relevant logs without adding application instrumentation.** That also suggests a more outcome-driven title, such as `Find the logs behind a failed trace with OBI`. The post then needs to help that reader answer five questions: Does this solve my problem? What will I see? Does it work in my environment? What do I need to change? What risks should I account for before enabling it?

I suggest rebuilding the post around this arc:

1. **The incident-response problem and the promise.** Keep the current opening scenario, then announce the capability and distinguish no application changes from the required OBI and log-pipeline rollout.
2. **What changes for the user.** Keep one before/after example and explain the resulting workflow: search by trace ID, link logs to traces, and move in either direction during an incident.
3. **Is this a fit for my environment?** Add a short compatibility checklist covering the OBI and Linux versions, supported log destinations, the need for active trace context, runtime qualifications, and the configuration-v1/v2 scoping difference.
4. **Enable it.** Show the smallest correct configuration and the required placeholder filter. This should be an adoption step, not another explanation of the mechanism.
5. **See it work.** Make the existing two-service demo and screenshots the centerpiece. Start with the failed request, show the correlated logs, and finish with the operational outcome.
6. **Before enabling it in production.** Summarize user-visible limitations, failure behavior, and a safe rollout/rollback approach. Keep the consequences here; link to technical documentation for mechanisms.
7. **Try it.** Link directly to documentation that matches the released behavior, the release, and a runnable example.

I would remove this entire implementation section, including the diagram and runtime-hook table. At most, the post needs one conceptual sentence: `OBI associates active trace context with supported writes to application stdout and stderr, then adds the correlation fields before the container logging pipeline receives them.` Readers who want the map, probe, or runtime-hook design can follow a link to OBI developer documentation.

The tone should stay in the direct operator voice established by `You get paged`: `your service`, `your logs`, `your OBI deployment`, and concrete outcomes. Avoid private symbol names and absolute claims such as `no redeploys`; explain what the reader gains, what they must operate, and where the feature does not apply. A successful revision lets a reader decide within a couple of minutes whether to try the feature and how to do so safely.

content/en/blog/2026/obi-trace-log-correlation/index.md
```diff
@@ -0,0 +1,262 @@
...
+_running_, a uprobe on the runtime's status-change function checks whether that
+goroutine has an in-flight server request, client call, or database operation,
+and refreshes the map entry for the OS thread it just landed on. When the
+handler returns, the entry is deleted. This way a log write resolves to the
+right span even after the goroutine has migrated to a different OS thread.
+
+## Turning it on
+
+The enricher is opt-in per service. Select services under `ebpf.log_enricher`
+(or `extensions.obi.correlation.log_trace_annotation` in version 2 of the
```
MrAlias, line 180: This per-service statement does not apply to Config v2 in the v0.12.2 release. The v0.12.2 converter [explicitly rejects a nonempty `correlation.log_trace_annotation.filter`](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/v0.12.2/internal/config/convert/import.go#L195-L213), while both its full and partial conversion paths [map `enabled: true` to a wildcard `*` service selector](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/v0.12.2/internal/config/convert/import.go#L2031-L2054). Config v2 can therefore enable annotation for all workloads already selected for OBI instrumentation, but it cannot select a separate subset specifically for log annotation. Could we keep the scoped example version-1-only and describe the version-2 behavior accurately, or omit the version-2 reference until filtering is supported?

content/en/blog/2026/obi-trace-log-correlation/index.md
```diff
@@ -0,0 +1,262 @@
...
+JSON with no trace fields; OBI injected matching context — the same `trace_id`
+in both services, each with its own `span_id`:
+
+![Enriched logs from both services carrying the same trace ID](logs-and-trace.png)
+
+Searching Jaeger for the `trace_id` from either log line lands on exactly the
+trace shown above — correlation works in both directions, log to trace and trace
+to log.
+
+## Limitations and future work
```
MrAlias, line 238: This section should become `Before enabling in production` and describe observable behavior rather than implementation mechanics. The current guidance also needs several corrections before it can support an adoption decision:

- Node.js stdout is asynchronous when backed by a POSIX pipe, as it normally is in a Linux container. Under backpressure, a write can occur after the request context has changed, so Node should not be included in an unconditional statement that synchronous request-thread logging is the default.
- A write over 8 KiB does not pass through intact and merely lose enrichment. The captured prefix is suppressed and re-emitted with context while the remainder reaches the stream separately and without enrichment, splitting one logical record; `writev()` can split earlier at a segment boundary.
- Ordinary `write()` enrichment requires Linux 6.0 or later. On older kernels, the `writev()` path may still work, but the common `write()` path is not enriched.
- If OBI's finite enrichment queue fills, bytes that were suppressed before queue submission cannot be re-emitted, so logs can be lost under sustained saturation. This needs to be disclosed together with rollout and monitoring guidance.

Could we replace the current list with a compact operator-facing checklist covering supported destinations and active-context requirements, runtime caveats, kernel and capability requirements, record-size behavior, saturation behavior, and a canary/rollback recommendation? The details of why those limits exist can live in the linked technical documentation.

MrAlias (commented):
The feature is worth announcing, but seeing the complete draft makes clear that the accepted outline needs an editorial reset. The post currently splits between an end-user incident-response story and OBI implementation documentation; I think it should be rebuilt as an operator-focused adoption post, with the user-facing behavior and production boundaries corrected before publication.

content/en/blog/2026/obi-trace-log-correlation/index.md
```diff
@@ -0,0 +1,262 @@
...
+  The carrier thread ID doesn't identify the request, and a wrong attribution is
+  worse than a missing one. Platform-thread enrichment is unaffected.
+- **Writes are capped at 8 KiB** per `write()`/`writev()` call; larger writes
+  pass through un-enriched.
+- **It needs `CAP_SYS_ADMIN`** and a kernel that is not in lockdown mode,
+  because rewriting the user-space buffer uses `bpf_probe_write_user`.
+
+## Try it
+
+Trace-log correlation ships in OBI today. Point it at one service, add the
```
grcevski, line 253: Should we explicitly state differences in how this behaves if OBI instruments the application and if the application is instrumented with an SDK. I think it's important to mention that the log enrichment will work with an application that doesn't use eBPF trace instrumentation, but native SDK trace instrumentation. The only difference is that we will be reporting the originating span_id and the outgoing span_id, so span_ids may not match, but you can still find your transaction in the logs by the trace_id. So in a way this feature completes SDK instrumentation where OTLP logs are not enabled or supported.

grcevski (approved):
LGTM! One minor comment for suggestion.

grcevski:
> did @MrAlias and @grcevski agree to sponsor this?

Yes I agreed to sponsor!

mmat11:
after discussing it in the SIG, we agreed to make it more operator-focused -- converting as draft for now

content/en/blog/2026/obi-trace-log-correlation/index.md
```diff
@@ -0,0 +1,214 @@
...
+  by trace ID.
+- **Configuration version.** With version 1 configuration you select exactly
+  which services get log enrichment. With version 2 configuration (as of OBI
+  v0.13.0), enabling `correlation.log_trace_annotation` applies to all workloads
+  OBI instruments — per-service selection for log enrichment is not supported
+  there yet.
+
+## Enable it
+
+The enricher is opt-in. With version 1 configuration, select the services under
```
MrAlias, line 124: Non-blocking: My earlier feedback allowed keeping the scoped Config v1 example if Config v2 was described accurately. Seeing the revised adoption flow, I think the post would be even stronger if standalone Config v2 were the primary path. The checklist and enablement section currently lead with the older format and devote two examples to it, while the Config v2 boundaries are split across later prose.

If you want to make that shift here, could we lead with the Config v2 example, state that annotation is standalone-only and applies to every workload selected by `capture`, and align the canary and rollback guidance with that deployment-wide scope? Config v2 has no independent per-workload log-annotation selector: canarying one service requires a deployment whose `capture` rules isolate it, while rollback otherwise disables `correlation.log_trace_annotation` for every workload that deployment captures. The older format could then be reduced to a short compatibility note linking to the detailed configuration documentation.

MrAlias (approved):
This looks amazing. I'm even more excited about this feature 🚀 

The config v2-first comment is non-blocking and editorial.

Great work.

svrnm (approved):
LGTM, happy to merge & publish this whenever you're ready

content/en/blog/2026/obi-trace-log-correlation/index.md
svrnm: @mmat11 PTAL, if you want to address this or not

mmat11: while addressing this and updating the examples to use configv2, I found https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/issues/3384 which is a blocker

PR #11403, opened 2026-08-22
Clarify .NET TracerProvider lifetime

### LINKED ISSUE #2048: Document use of `Sdk.CreateTracerProvider` (i.e., not-web scenarios) for .NET
As per discussion here: https://github.com/open-telemetry/opentelemetry.io/pull/2017#discussion_r1024238516

People have been haphazardly copy/pasting initialization code into `void`-returning methods that immediately disposed of the `TracerProvider` due to using `uisng var`. However, explicitly calling `Dispose()` somewhere else in an example file instead of a `using var` can also lead to problematic behavior if people copy/paste code without also doing the dispose call.

One example can be to have tracing initialization happen as a method call like so:

```csharp
TracerProvider ConfigureTracing()
{
    return Sdk.CreateTracerProviderBuilder(). //...
}

//...

using var tracerProvider = ConfigureTracing
```

So that if people copy/paste the contents of `ConfigureTracing`, they still need to deal with the fact that it is a disposable.

LarryHu0217:
- [x] I have read and followed the [Contributing](https://opentelemetry.io/docs/contributing/) docs, especially the **First-time contributing?** section.

[^I-know-my-stuff]:

---

Clarify the lifetime requirements for a manually created .NET `TracerProvider`.
The new callout explains that console and other non-DI applications must keep the
provider alive while they produce telemetry, avoid disposing it inside a
`void` helper method, and dispose it during application shutdown.

Validation:

- Prettier
- markdownlint
- CSpell
- textlint
- Full Hugo site build

Fixes #2048

### DIFF
--- a/content/en/docs/languages/dotnet/instrumentation.md
+++ b/content/en/docs/languages/dotnet/instrumentation.md
@@ -307,11 +307,8 @@ loggerFactory.Dispose();
 > [!IMPORTANT]
 >
 > Keep the `TracerProvider` alive for as long as the application creates
-> telemetry. If you move its creation into a helper method, return the
-> `TracerProvider` to the caller. Don't create it with a local `using`
-> declaration in a `void` method, because the provider is disposed when the
-> method returns. Dispose the provider during application shutdown so that it
-> can flush any remaining telemetry. For more information, see
+> telemetry, and explicitly dispose it during application shutdown to flush any
+> remaining telemetry. For more information, see
 > [TracerProvider management](/docs/languages/dotnet/traces/best-practices/#tracerprovider-management).
 
 For debugging and local development purposes, the example exports telemetry to
--- a/content/en/docs/languages/dotnet/instrumentation.md
+++ b/content/en/docs/languages/dotnet/instrumentation.md
@@ -304,6 +304,16 @@ meterProvider.Dispose();
 loggerFactory.Dispose();
 ```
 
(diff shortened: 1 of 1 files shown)


content/en/docs/languages/dotnet/instrumentation.md
```diff
@@ -304,6 +304,16 @@ meterProvider.Dispose();
 loggerFactory.Dispose();
 ```
 
+> [!IMPORTANT]
+>
+> Keep the `TracerProvider` alive for as long as the application creates
+> telemetry. If you move its creation into a helper method, return the
+> `TracerProvider` to the caller. Don't create it with a local `using`
+> declaration in a `void` method, because the provider is disposed when the
+> method returns. Dispose the provider during application shutdown so that it
```
martincostello: I'm not sure we need to explain how C# `using` scopes with to the reader - we can assume they have a working knowledge of C#.

I think just calling out that it needs to be kept alive while telemetry is collected and then explicitly disposed to flush is sufficient.

LarryHu0217: Updated in 1fd5f0a to keep the note focused on the provider lifetime and explicit disposal at shutdown.

martincostello (approved):

maryliag (approved):

cijothomas (approved):

Issue #11377, opened 2026-08-19
Ecosystem lists are frozen, retirement planned


chalin:
We're putting a hold on **lists under [ecosystem][]** while we work out each list's future.

Curating these lists consumes a disproportionate share of maintainer time relative to the value they add:

- Registry and vendor self-additions are a steady stream of first-time-contributor PRs (~290 registry-labeled PRs in the last 12 months; see also #11243).
- The registry carries significant infrastructure in this repo: 1160+ data files, schema validation, dedicated workflows and tooling.
- Listing unvetted entries carries risk (CVEs, malware, non-compliant implementations) and the maintainer team lacks the bandwidth and tooling to validate them.

Direction per list, with details to follow in dedicated issues:

- **[Registry][]**: retire; the community ecosystem explorer becomes the source of truth.
- **[Vendors][] & [distributions][]**: retire, or reframe around vendors actively investing in and contributing to OpenTelemetry.
- **[Integrations][] & [adopters][]**: continue in some form, or retire; decision to come (compare Kubernetes' evolution of its adopter list into [case studies][]).

Input welcome, especially from SIGs and users that depend on the registry data.

[ecosystem]: https://opentelemetry.io/ecosystem/
[adopters]: https://opentelemetry.io/ecosystem/adopters/
[case studies]: https://kubernetes.io/case-studies/
[distributions]: https://opentelemetry.io/ecosystem/distributions/
[integrations]: https://opentelemetry.io/ecosystem/integrations/
[registry]: https://opentelemetry.io/ecosystem/registry/
[vendors]: https://opentelemetry.io/ecosystem/vendors/

jaydeluca:
I would also add security posturing and lack of validation as part of the justification. We don't have the bandwidth or tooling to validate every entry into the registry does not contain CVEs, malware, or some non-compliant implementation, and that carries risk in us listing them.

chalin:
Done, thx @jaydeluca.

kamphaus:
Hopefully https://github.com/open-telemetry/semantic-conventions-conformance can add an ecosystem list with qualitative criteria.

chalin:
Some usage data to ground the per-list decisions, from the site's public [analytics report](https://lookerstudio.google.com/s/tSTKxK1ECeU) (Jul 22 - Aug 20, 2026; 741,921 site page views):

| Page | Views | % of site views |
|---|---:|---:|
| `/ecosystem/registry/` | 10,199 | 1.37% |
[...]

PR #11606, opened 2026-09-06
Rename qryn vendor entry to Gigapipe

rita7lopes:
## What this changes

The vendor entry named `qryn` is renamed to **Gigapipe**, its URL repointed to the current documentation, and the entry moved into alphabetical order.

qryn was renamed to Gigapipe. The project is the same one — same maintainers, same repository lineage — and the name change is reflected everywhere else, including this repo's own registry entry, which already reads [`collector-exporter-gigapipe.yml`](https://github.com/open-telemetry/opentelemetry.io/blob/main/data/registry/collector-exporter-gigapipe.yml) and describes it as "Gigapipe (formerly qryn)". The vendors list is the last place still carrying the old name.

## Details

- **Name** — `qryn` → `Gigapipe`.
- **URL** — `https://qryn.metrico.in/#/support?id=tempo-api&link-check=no` → `https://gigapipe.com/docs/api`. The old URL still resolves, so this is not a broken-link fix; it points at the superseded docs site and carries a `link-check=no` escape that is no longer needed.
- **Contact** — updated to a current maintainer address; the previous contact is no longer the right first point of contact.
- **Position** — moved from between Parseable and Sentry (its old alphabetical slot under "q") to between Elastic and Google Cloud Platform, so the file stays sorted under the new name.

`oss: true`, `commercial: true` and `nativeOTLP: true` are unchanged and still accurate: the project is AGPL-3.0 with a commercial hosted offering, and ingests OTLP over both HTTP and gRPC.

## Verification

- `data/ecosystem/vendors.yaml` parses, with the vendor count unchanged at 105.
- The new URL returns HTTP 200.
- `data/ecosystem/vendors.yaml` is listed in `ignorePaths` in `.cspell.yml`, so no dictionary entry is needed for the new name.

## Disclosure

I work at Gigapipe. Submitting this as a correction to an existing entry for a project that has since been renamed, not as a new listing.

### DIFF
--- a/.lycheecache
+++ b/.lycheecache
@@ -840,6 +840,7 @@ https://geminicli.com/docs/cli/telemetry/#opentelemetry-integration,200,17879167
 https://geode.apache.org/,200,1787981986
 https://getcomposer.org/,200,1787981969
 https://getcomposer.org/doc/01-basic-usage.md#autoloading,200,1788327798
+https://gigapipe.com/opentelemetry,200,1788975322
 https://gist.github.com/braydonk/b2381da98dc3c4fd5ac064045d556634,200,1787916982
 https://gist.github.com/fardjad/ea3c38d566c845e0b353237d3959e365,200,1787916982
 https://gist.github.com/zeitlinger/09585b1ab57c454f87e6dcb9a6f50a5c,200,1787916712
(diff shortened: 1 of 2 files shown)


svrnm:
The vendor page is currently frozen, so we are not accepting any PRS to it right now, for more details please take a look here: https://github.com/open-telemetry/opentelemetry.io/issues/11377

Issue #6899, opened 2025-05-15
Go language docs: Add documentation for how to test instrumentation


dashpole:
**What needs to be changed?**

We document how to instrument a library, but not what to use to test instrumentation.  This seems important enough to have some basic documentation for.

**What is the name + path of the page that needs changed?**

https://opentelemetry.io/docs/languages/go/

**Additional context**: anything else you can think about that would be helpful.

Context: https://cloud-native.slack.com/archives/C01NPAXACKT/p1747330549116979?thread_ts=1747174039.804719&cid=C01NPAXACKT

cc @open-telemetry/go-approvers

svrnm:
Added `help wanted`, might be something someone can pick up and help with.

pellared:
For testing trace instrumentation this can be helpful:
- https://pkg.go.dev/go.opentelemetry.io/otel/sdk/trace/tracetest (we should add a package example)

For testing metrics instrumentation this can be helpful:
- https://pkg.go.dev/go.opentelemetry.io/otel/sdk/metric/metrictest (we should add a package example)

For testing logs instrumentation this can be helpful:
- https://pkg.go.dev/go.opentelemetry.io/otel/log/logtest (there already is a package example)

svrnm:
@open-telemetry/go-approvers double checking on that, I assume this is still a valid request?

lavishpal:
Can i work on this issue?

lavishpal:
I'm thinking of creating a new `testing.md` page under `content/en/docs/languages/go/` covering how to test trace, metric, and log instrumentation. 
WDYT @svrnm , @pellared

pellared:
@lavishpal, works for me. You could also include guidance on how to test additional components, such as custom resource detectors, trace/metric/log processors. It would be helpful to keep one PR per aspect being tested to make reviews clearer and more focused.

lavishpal:
Will raise the PR shortly. 😊

PR #11675, opened 2026-09-11
docs(go): add trace instrumentation testing guide

lavishpal:
## Description
this PR adds the new `testing.md` page which have the steps and instructions on how to test trace instrumentation using [trace.SpanRecorder](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/trace/tracetest#SpanRecorder)

this PR fixes the subpart(tracer) of the issue #6899. Will raise the follow-up PR for the metrics and logs

[^I-know-my-stuff]:

---

### DIFF
--- a/content/en/docs/languages/go/testing.md
+++ b/content/en/docs/languages/go/testing.md
@@ -15,7 +15,7 @@ Use the `tracetest` package to capture spans in memory during tests.
 
 Ensure you have the right packages installed:
 
-```bash
+```sh
 go get go.opentelemetry.io/otel/sdk/trace \
   go.opentelemetry.io/otel/sdk/trace/tracetest
 ```
--- a/content/en/docs/languages/go/testing.md
+++ b/content/en/docs/languages/go/testing.md
@@ -1,0 +1,64 @@
+---
+title: Testing Instrumentation
+weight: 35
+description: How to test OpenTelemetry Go instrumentation
+---
+
+When instrumenting a library or application, it is important to verify
+that your instrumentation works as expected without needing a real backend
+such as Jaeger or Prometheus. OpenTelemetry Go provides testing helper
+packages for this purpose.
+
+## Traces
+
+Use the `tracetest` package to capture spans in memory during tests.
+
+Ensure you have the right packages installed:
+
+```bash
+go get go.opentelemetry.io/otel/sdk/trace \
+  go.opentelemetry.io/otel/sdk/trace/tracetest
+```
+
+The `SpanRecorder` acts as a `SpanProcessor` that records all started
+and ended spans in memory. You can then assert on those spans in your
+test.
+
+```go
+package yourpackage_test
+
+import (
+    "context"
+    "testing"
+
+    sdktrace "go.opentelemetry.io/otel/sdk/trace"
+    "go.opentelemetry.io/otel/sdk/trace/tracetest"
+)
+
+func TestSpanRecorder(t *testing.T) {
+    ctx := context.Background()
+
+    // Set up an in-memory span recorder and tracer provider.
+    sr := tracetest.NewSpanRecorder()
+    tp := sdktrace.NewTracerProvider(
+        sdktrace.WithSpanProcessor(sr),
+    )
+    defer tp.Shutdown(ctx)
+
+    tracer := tp.Tracer("example/simple")
+
+    // Start and end a span.
+    _, span := tracer.Start(ctx, "test-span")
+    span.End()
+
+    // Assert on the recorded spans.
+    spans := sr.Ended()
+    if len(spans) != 1 {
+        t.Fatalf("expected 1 span, got %d", len(spans))
+    }
+    if spans[0].Name() != "test-span" {
+        t.Errorf("expected span name 'test-span', got %s",
+            spans[0].Name())
+    }
+}
+```


lavishpal:
Can you PTAL?
cc: @pellared , @svrnm

maryliag:
Hi @lavishpal , thank you for your interest in contributing. 
I can see you're a first time contributor here, and that your PR has no issue associated with. We request you create an issue first so we can confirm if that is something we want to do, the format, etc, specially because you're creating a page the vast majority of languages don't have (currently only Erlang has a similar page), so we want to make sure not only the work is agreed on, but that you also have a sponsor from the related SIG that will be able to review for you.

Also, if you notice on the top right, there are already teams tagged to review (docs approvers and go approvers), so there is no need to tag anyone directly here, specially right after opening a PR. Give approvers/maintainers some time to get to this.

If someone from the go approvers team can review, then we can leave this open, if this gets no traction I'll request you to follow the process of opening the issue/getting a sponsor first :)

lavishpal:
thanks for the clarification @maryliag , actually i was working on #6899 where in [comment](https://github.com/open-telemetry/opentelemetry.io/issues/6899#issuecomment-4210330066) i proposed a idea of creating `testing.md` page for the instrumentation testing


> Also, if you notice on the top right, there are already teams tagged to review (docs approvers and go approvers),

also misunderstood the docs approvers and go approvers teams as labels rather than the teams already assigned for review.

content/en/docs/languages/go/testing.md
```diff
@@ -0,0 +1,64 @@
...
+---
+
+When instrumenting a library or application, it is important to verify
+that your instrumentation works as expected without needing a real backend
+such as Jaeger or Prometheus. OpenTelemetry Go provides testing helper
+packages for this purpose.
+
+## Traces
+
+Use the `tracetest` package to capture spans in memory during tests.
```
pellared, line 14: Can we refer to https://pkg.go.dev/go.opentelemetry.io/otel/sdk/trace/tracetest and add [a GoDoc example](https://go.dev/blog/examples) instead of the code snippets here? We can then do the same for other signals. Note that e.g. we have an example already in `metricdatatest`: https://pkg.go.dev/go.opentelemetry.io/otel/sdk/metric/metricdata/metricdatatest#example-AssertEqual. For logs there are two packages https://pkg.go.dev/go.opentelemetry.io/otel/log/logtest (the most useful, for testing instrumentation), https://pkg.go.dev/go.opentelemetry.io/otel/sdk/log/logtest (for testing custom processors).

content/en/docs/languages/go/testing.md
```diff
@@ -0,0 +1,64 @@
...
+    sr := tracetest.NewSpanRecorder()
+    tp := sdktrace.NewTracerProvider(
+        sdktrace.WithSpanProcessor(sr),
+    )
+    defer tp.Shutdown(ctx)
+
+    tracer := tp.Tracer("example/simple")
+
+    // Start and end a span.
+    _, span := tracer.Start(ctx, "test-span")
```
MrAlias, line 51: This span is created directly by the test, so the example verifies `SpanRecorder` rather than any instrumentation in `yourpackage`. It will still pass if that instrumentation emits no span or records the wrong data. Could the canonical example invoke a small instrumented function with this provider and assert on that function's span?

MrAlias (commented):
The guide still needs to connect the recorder setup to the instrumentation under test, and the spelling/formatting failures on the current head also need to be corrected.

content/en/docs/languages/go/testing.md
lavishpal: checking that if i got this correct. should i remove the code snippet from this page and instead add a GoDoc example in the tracetest package or just add the reference link here to already present example in [example_test.go](https://github.com/open-telemetry/opentelemetry-go/blob/sdk/metric/v1.46.0/sdk/trace/tracetest/example_test.go)? And follow the same approach for metrics and logs?

cijothomas: Can users not use InMemoryExporter? Or is tracetest a wrapper around  it?

lavishpal: Yes, users can use InMemoryExporter, but it serves a slightly different purpose from SpanRecorder and tracetest is not a wrapper around InMemoryExporter.

cijothomas: I am curious what is the difference between them. InMemoryExporter is a spec requirement and should exist for all signals. Wondering why we cannot use that? 
(Sorry I haven't explore OTel go enough, but this was based on my observation from other languages)

content/en/docs/languages/go/testing.md
```diff
@@ -0,0 +1,64 @@
...
+
+func TestSpanRecorder(t *testing.T) {
+    ctx := context.Background()
+
+    // Set up an in-memory span recorder and tracer provider.
+    sr := tracetest.NewSpanRecorder()
+    tp := sdktrace.NewTracerProvider(
+        sdktrace.WithSpanProcessor(sr),
+    )
+    defer tp.Shutdown(ctx)
```
dashpole: `tp.Shutdown(ctx)` returns an error that gets silently ignored here. In tests, consider checking the error or using `t.Cleanup` so shutdown failures surface in test results:
```suggestion
    t.Cleanup(func() {
        if err := tp.Shutdown(ctx); err != nil {
            t.Error(err)
        }
    })
```

content/en/docs/languages/go/testing.md
lavishpal: SpanRecorder implements SpanProcessor interface which captures span as soon as they end, before export and InMemoryExporter implementts the SpanExporter which captures the spans only after they go through the export pipeline. 
In context of this PR InMemoryExporter operates at the exporter level which means it need wiring up an export pipeline(WithBatcher or WithSyncer) and imo adding complexity for the verifying instrumentation output is not required

PR #11677, opened 2026-09-11
Document how to set trace state and trace flags in Python

### LINKED ISSUE #11470: [Page feedback]: Please add some words about defining trace_state and trace_flags
### URL

https://opentelemetry.io/docs/languages/python/cookbook/

### Description

I've spent a day or so trying to use Python to pass in a trace_state and trace_flags.  I want to pass w3.tracestate, so it can be propagated  along with my data.

Please can you give an example of how to do this.

I've tried
```

INVALID_SPAN_ID = 0x0000000000000000
INVALID_TRACE_ID = 0x00000000000000000000000000000000
DEFAULT_TRACE_OPTIONS = TraceFlags.get_default()

tracestate = trace.TraceState([("vendor_key","vendor_value")    ])
cc = {"trace_id":INVALID_TRACE_ID,
             "span_id":INVALID_SPAN_ID,
             "is_remote":True,
             "trace_flags":DEFAULT_TRACE_OPTIONS,
             "trace_state":tracestate}
con =  Context(cc)

with tracer.start_as_current_span("myPythonProgram",  context=con):

but it complains about invalid root.

Looking at the code, it seems I can do it using a sampler ...  but this feels I'm heading in the wrong direction.

Thank you

```

ocelotl:
Fixes #11470

- [ ] I have read and followed the [Contributing](https://opentelemetry.io/docs/contributing/) docs, especially the "**First-time contributing?**" section.
- [ ] I have the experience and knowledge necessary to understand, review, and validate all content in this PR.[^I-know-my-stuff]

[^I-know-my-stuff]:

---

The reporter of #11470 spent about a day trying to set a `trace_state` and
`trace_flags` on a Python span and could not find any guidance. The Python docs
never mention `tracestate` outside of empty console output, so there was nothing
to find.

The underlying source of confusion is that `trace_state` and `trace_flags` are
not parameters of `start_span` or `start_as_current_span`. They are results of
starting a span. The SDK takes the trace state of a new span from the
`SamplingResult` the sampler returns, and it derives the `sampled` bit from the
sampling decision. So there are only two ways in: inherit from a valid parent
span context, or have a sampler produce it.

Two details make the failure hard to diagnose, because neither raises an error:

- A `Context` is an opaque key-value mapping whose keys are generated at run
  time, not a record with `trace_id` and `span_id` fields. Building one from a
  plain dictionary of trace fields stores values that nothing reads, so the span
  starts as a root span with an empty trace state.
- A span context built from all-zero IDs is not valid, so the SDK discards it
  and starts a root span, dropping the trace state along with it. These are
  exactly the values named `INVALID_TRACE_ID` and `INVALID_SPAN_ID`.

## Changes

Add a **Setting trace state and trace flags** section to the Python cookbook,
covering the three cases a reader can hit, plus trace flags:

- inheriting a trace state from a valid parent span context
- adding an entry to an incoming trace state, since `TraceState` is immutable
  and `add`, `update` and `delete` return a new instance
- seeding a trace state onto a root span through a sampler, which is the only
  way in when there is no parent to inherit from
- why the `sampled` bit cannot be set on a span directly, and how `ParentBased`
  makes a remote parent's decision carry over

Also fix the receiving-service example in the Python propagation page. It built
its carrier as `{'traceparent': headers['Traceparent']}`, which silently
discarded any incoming `tracestate` before extraction, and extracted baggage
from a second one-key carrier for the same reason. Propagators look up lowercase
header names, which is why the original code could not pass
`dict(request.headers)` straight through, so the keys are now normalized to
lowercase and the whole carrier is handed to both propagators.

## Validation

Each Python example was extracted from the rendered markdown and run as a
standalone script. All of them execute as written and print the output their
comments claim. `check:format`, `check:spelling`, `check:markdown` and
`check:text` pass on both changed files.

### DIFF
--- a/content/en/docs/languages/python/cookbook.md
+++ b/content/en/docs/languages/python/cookbook.md
@@ -126,6 +126,174 @@ finally:
     context.detach(token)
 ```
 
+## Setting trace state and trace flags
+
+The [trace state](https://www.w3.org/TR/trace-context/#tracestate-header) of a
+span, which carries the vendor-specific `tracestate` header of
+[W3C Trace Context](https://www.w3.org/TR/trace-context/), and its
+[trace flags](https://www.w3.org/TR/trace-context/#trace-flags) are not
+parameters of `start_span` or `start_as_current_span`. They are results of
+starting a span, not inputs to it, so there are only two ways to set them:
+inherit them from a valid parent span context, or have a sampler produce them.
+
+A common mistake is to build a `Context` out of a plain dictionary of trace
+fields. That does not work, because a `Context` is an opaque key-value mapping
+whose keys are generated at run time rather than a record with `trace_id` and
+`span_id` fields. Writing those names into it stores values that nothing reads,
+and the span is started as a root span with an empty trace state.
+
+### Inheriting trace state from a parent
+
+To start a span under a parent that carries a trace state, wrap the parent span
+context in a `NonRecordingSpan` and set it in a `Context`, as in
+[Manually setting span context](#manually-setting-span-context):
+
+```python
+from opentelemetry import trace
+from opentelemetry.context import Context
+from opentelemetry.trace import NonRecordingSpan, SpanContext, TraceFlags, TraceState
+from opentelemetry.sdk.trace import TracerProvider
+
+trace.set_tracer_provider(TracerProvider())
+tracer: trace.Tracer = trace.get_tracer("my.tracer")
+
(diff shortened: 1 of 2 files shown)


emdneto (approved):
LGTM

didiViking (changes requested):
@ocelotl Could you please tick the corresponding options from the issue's description regarding the contribution? Thank you!

